PR #1195, opened 2021-07-13
updates to documentation for alpha

punit-codaio:


jonathan-codaio (commented):
Should we keep the alpha doc more private than this? This repo is now open-sourced, I don't think we want to invite the public to read through the doc.

punit-codaio:
Agreed. We're only going to direct-share the alpha documentation doc with those who sign up via this form -- so the link will lead to nowhere for everyone else. Wdyt? 

https://coda.io/d/Coda-Packs-Alpha-Ts-Cs_diO2tbNTL4G/Coda-Packs-Alpha-Developer-Registration_suk_C?utm_source=slack#_luMFF

jonathan-codaio (approved):
Ok, that's fine if the doc is private.

Just be aware that even these PRs are public to the world so we just want to be thoughtful about when we're ready to mention/link to things.

PR #1196, opened 2021-07-14
create separate ivm context for each sync formula execution

huayang-codaio:
https://staging.coda.io/d/Packs-IA-go-packs_d36WK4zgrYx/Bugs-Feature-requests_suyN0#Open-Bugs_tu6FU/r26

This PR fixes the bug and further narrows the gap between the lambda execution and the sdk execution. Instead of having IVM thunk to iterate though all the pages, now it's the sdk iterating the pages and thunk only executes a single request. 

Now if a pack author uses a global variable to store page number, it's going to fail in the CLI execution (with VM).

Testing:

```
➜  packs-sdk git:(hg-sync-cli) node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake.ts Students Smith  --vm
[
  { Name: 'Alice' },
  { Name: 'Bob' },
  { Name: 'Chris' },
  { Name: 'Diana' }
]
➜  packs-sdk git:(hg-sync-cli) node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake.ts Students Smith      
[
  { Name: 'Alice' },
  { Name: 'Bob' },
  { Name: 'Chris' },
  { Name: 'Diana' }
]

```

### DIFF
--- a/testing/execution.ts
+++ b/testing/execution.ts
@@ -1,6 +1,5 @@
 import type {Credentials} from './auth_types';
 import type {ExecuteOptions} from './execution_helper';
-import type {ExecuteSyncOptions} from './execution_helper';
 import type {ExecutionContext} from '../api_types';
 import type {MetadataContext} from '../api';
 import type {MetadataFormula} from '../api';
@@ -8,7 +7,8 @@ import type {PackVersionDefinition} from '../types';
 import type {ParamDefs} from '../api_types';
 import type {ParamValues} from '../api_types';
 import type {SyncExecutionContext} from '../api_types';
-import { compilePackBundle } from './compile';
+import type {SyncFormulaResult} from '../api';
+import {compilePackBundle} from './compile';
 import {getPackAuth} from '../cli/helpers';
 import * as helper from './execution_helper';
 import * as ivmHelper from './ivm_helper';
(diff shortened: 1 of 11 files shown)


testing/execution.ts
```diff
@@ -87,10 +88,31 @@ export async function executeFormulaOrSyncFromCLI({
...
-    const result = vm
-      ? await executeFormulaOrSyncWithRawParamsInVM({formulaName, params, manifestPath, executionContext})
-      : await executeFormulaOrSyncWithRawParams({formulaName, params, manifest, executionContext});
-    print(result);
+    const syncFormula = helper.tryFindSyncFormula(manifest, formulaName);
+    if (syncFormula) {
+      const result = [];
+      let iterations = 1;
+      do {
+        if (iterations > MaxSyncIterations) {
```
jonathan-codaio, line 96: I meant to allow you to override this with a CLI flag, we should do that at some point.

jonathan-codaio (approved):
Smart solution to this.

How fast is the VM execution when using `coda execute` vs not using it? And are error stacks basically as clear as when not using the VM now? I wonder if we should invert the default to use the VM.

testing/execution.ts
huayang-codaio: yeah, I found it was not currently used so just remove it for simplicity. we can always add this if needed.

huayang-codaio:
> How fast is the VM execution when using coda execute vs not using it? And are error stacks basically as clear as when not using the VM now? I wonder if we should invert the default to use the VM.

Yes I do wonder if we should do so. Testing without VM env isn't solid locally. I need to do more testing to confirm if it's the time. I'll follow up.

patrick-codaio (approved):

PR #1198, opened 2021-07-14
Add Makefile command to publish to npm with proper tags.

jonathan-codaio:
PTAL @huayang-codaio @patrick-codaio @coda/packs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,4 +102,4 @@ validate-no-changes: compile
 
 .PHONY: npm-publish
 npm-publish:
-  npm publish --tag alpha --access public
+	npm publish --tag alpha --tag latest --access public
--- a/Makefile
+++ b/Makefile
@@ -99,3 +99,7 @@ validate-no-changes: compile
 		echo "dist directory is not clean. run 'make build'"; \
 		exit 1; \
 	fi
+
+.PHONY: npm-publish
+npm-publish:
+  npm publish --tag alpha --access public


Makefile
```diff
@@ -99,3 +99,7 @@ validate-no-changes: compile
 		echo "dist directory is not clean. run 'make build'"; \
 		exit 1; \
 	fi
+
+.PHONY: npm-publish
+npm-publish:
+  npm publish --tag alpha --access public
```
patrick-codaio, line 105: just to make sure i understand the intended workflow here:
- 1x/week we `make npm-publish` to update the alpha tag, which our alpha users will use
- `coda`, `packs`, & `pack-examples` will all depend on the alpha tag too
- To test SDK changes before the weekly release, we must explicitly set the commit hash in our dev repo but not check it in.

is that right?

cc @coda/packs

jonathan-codaio: The main thing is that we don't want alpha users to experience breaking changes more than once a week. How we decide to achieve that is somewhat flexible. I suspect that most of the time what we'll do is make changes in a backwards-compatible way, and still update the `coda` side with the new SDK hash so that we don't have painful SDK bumps that include lots of changes, but we can remove backwards compatibility after a week if we need to.

I think we can also publish to npm whenever we want, but we just wouldn't set the `latest` tag to a version that includes breaking changes more than once a week.

`packs-examples` will depend on the published npm package at whatever version we want to pin it too--maybe we could pin it to `latest`.

`coda` and `packs` can continue to use commit hashes for now until we decide what to do. Most likely `packs` will switch to using the published package version at `latest` when we start publishing those packs using the v2 flow. `coda` may use a pre-release tag of some sort or continue to use commit hashes.

huayang-codaio: ack. 

+1 for manual packs-sdk release and using hashes for packs & coda repo for now. frequency can be decided with # of changes.

Makefile
```diff
@@ -99,3 +99,7 @@ validate-no-changes: compile
 		echo "dist directory is not clean. run 'make build'"; \
 		exit 1; \
 	fi
+
+.PHONY: npm-publish
+npm-publish:
+	npm publish --tag alpha --tag latest --access public
```
huayang-codaio, line 105: learned.

huayang-codaio (approved):

Makefile
huayang-codaio: btw do you need any credentials?

Makefile
patrick-codaio: I think it would be great to pin `packs-examples` to latest and have circleci tell us if we break it.

If we're going to be flexible in our general process, can we clarify the release process for a breaking change then? Would we `make npm-publish` at the same time that we deploy to the packs adhoc? And when we're aware there's a breaking chance coming, the team just gets really careful about deploying the adhoc?

huayang-codaio: imo in the alpha phase breaking changes are tolerated. after a public release, version numbers will indicate breaking changes. we can decide testing / announcement accordingly.

jonathan-codaio: We agree we can make breaking changes during the alpha, but preferably not more than once a week, so that stuff isn't breaking on you randomly all the time.

We will probably set the "latest" tag on npm at the same time we deploy the adhoc when we're ready to "apply" a breaking change. If we want, we can still publish npm versions before that if people want to be early-adopters, we just don't have to set the "latest" tag on them. Users will get the "latest"-tagged version when they run `npm install`, and will only get a different version if they specify it manually.

Makefile
jonathan-codaio: Yeah, you need credentials. Would need to create an account at npmjs.org and get added to our organization.

PR #1205, opened 2021-07-16
Add sdkVersion to PackVersionMetadata

alan-codaio:
For the Packs migration, we want to include the SDK version that the Pack was built with, and drop that into the `PackVersionMetadata`. So we need to include this as a field so that the Packs compiler can insert that field, and that the Coda migration workflow can read from it. 

Question: `make build` changed a bunch of `docs/` files, including creating several untracked files. Should I include those in this PR?

### DIFF
--- a/compiled_types.ts
+++ b/compiled_types.ts
@@ -54,6 +54,7 @@ export type PackVersionMetadata = Omit<
   formats: PackFormatMetadata[];
   syncTables: PackSyncTable[];
   defaultAuthentication?: AuthenticationMetadata;
+  sdkVersion?: string;
 };
 
 /** Stripped-down version of `PackDefinition` that doesn't contain formula definitions. */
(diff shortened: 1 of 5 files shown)


jonathan-codaio (commented):
Was actually thinking real quick and dirty, just tack it on in the `packs` repo without even changing the SDK. Will be confusing to have this in two places in the SDK (it's already on PackUpload).

So could just do `(metadata as any).sdkVersion = whatever` in the `packs` compiler.

The untracked files issue should go away if you rebase, there were some files that didn't get checked in that I fixed the other day.

alan-codaio:
> Was actually thinking real quick and dirty, just tack it on in the `packs` repo without even changing the SDK. Will be confusing to have this in two places in the SDK (it's already on PackUpload).
> 
> So could just do `(metadata as any).sdkVersion = whatever` in the `packs` compiler.
> 
> The untracked files issue should go away if you rebase, there were some files that didn't get checked in that I fixed the other day.

Ok, I can close this then. My worry is that we'd have to do the same thing in the workflow, i.e. grab the version by doing `metadata as any...`

PR #1206, opened 2021-07-16
Return executeSyncFormulaFromPackDef to its original behavior of running the sync in a loop to preserve backwards compatibility.

jonathan-codaio:
I'm trying to bump the SDK in the packs repo and the change in functionality of executeSyncFormulaFromPackDef is breaking a bunch of tests there, that rely on the fact that it runs a sync to completion. It will also break packs-examples, and possibly customers who bothered to write unittests (don't know if there are any).

It is fair that it's probably more useful to test a single iteration and be able to inspect the returned continuation, and sidestep the fact that we don't run the loops in isolation in the context of a unittest or integration test yet. But I'd rather not have this be a breaking change and have to fix all those tests and tell users to do the same, we can try to have this use the VM too when it's ready.

PTAL @huayang-codaio @alan-codaio @coda/packs

### DIFF
--- a/development.ts
+++ b/development.ts
@@ -7,6 +7,7 @@ export type {ExecuteOptions} from './testing/execution';
 export {executeFormulaFromPackDef} from './testing/execution';
 export {executeMetadataFormula} from './testing/execution';
 export {executeSyncFormulaFromPackDef} from './testing/execution';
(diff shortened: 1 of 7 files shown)


alan-codaio (approved):

huayang-codaio (approved):
Makes sense. Thanks for the fix.

PR #1214, opened 2021-07-21
add generated examples

lucas-codaio:
# Summary 
* Add generated examples config
* Consolidate compilation of snippets and examples
* Add ColumnFormat as an example `example`


documentation/documentation_config.ts
```diff
@@ -140,3 +142,11 @@ export const Snippets: Snippet[] = [
     codeFile: './snippets/dynamic_sync_table.ts',
   },
 ];
+
+export const Examples: Example[] = [
+  {
+    contentFile: './examples/column-format/column_format.md',
+    codeFiles: ['./examples/column-format/no_matchers.ts'],
```
jonathan-codaio, line 149: This is indeed what I had in mind. It just occurred to me, as we start using this, if we decide that would be useful to have content and code interspersed, we could generalize this to have a single ordered list of blocks to display, which each block would be content or code:

```
blocks: [
  {type: BlockType.Content, file: '...'],
  {type: BlockType.Code, file: '...'],
  {type: BlockType.Content, file: '...'],
  {type: BlockType.Code, file: '...'],

],
```

No need to change anything now, just something to keep in mind if we need more flexibility in what we present in the side panel.

documentation/examples/column-format/no_matchers.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const pack = coda.newPack();
+
+// BEGIN
+
+// Replace all <text> with your own text
+// a custom column type that you apply to any column in any Coda table.
+pack.addColumnFormat({
+  name: '<format name>',
+  formulaNamespace: '<formulaNamespace of Pack>', // this will be deprecated
```
jonathan-codaio, line 11: Ah yeah, we should (separately) probably change `addColumnFormat()` so that it doesn't require a namespace, and populates it with a dummy default one if you don't provide one, so that users don't have to think about this during the deprecation period.

documentation/examples/column-format/no_matchers.ts
```diff
@@ -0,0 +1,16 @@
...
+  formulaName: '<the name of a formula to execute using the value in the column>',
```
jonathan-codaio, line 12: Nit, 'the name of a formula in this pack' is probably clearer.

documentation/examples/column-format/no_matchers.ts
```diff
@@ -0,0 +1,16 @@
...
+  instructions: '<instruction text that will be visible by Users',
```
jonathan-codaio, line 13: Nit, add closing brace.

documentation/examples/column-format/no_matchers.ts
```diff
@@ -0,0 +1,16 @@
...
+  // regular expressions used to detect inputs of a certain format type.
```
jonathan-codaio, line 14: Nit, proper capitalization for all comments visible to users. Maybe,

"Optional, regular expressions that detect urls that are handleable by this column format. When a user pastes a url that matches any of these, the UI will hint to the user that they can apply this format to their column."

jonathan-codaio (approved):
Nice

PR #1215, opened 2021-07-22
copy most ivm runtime files from coda

huayang-codaio:
I am about to merge the IVM runtime files in coda repo and the packs sdk repo. There are a lot of methods that are duplicated over these two repos and are slightly different in implementation. 

This PR is really doing nothing but copying over the files from coda repo into the packs-sdk repo. I am sending this out first so that it's clear what I changed in the following PRs. 

A few things changes during the copying:
- I updated `StatusCodeError` in the packs-sdk repo. I just realized this thing exists. Maybe we should switch to it everywhere and get rid of the dependency from request-promise `StatusCodeError`. 
- I didn't copy over marshalError/unmarshalError since I plan to redo it together with other native types (e.g. Buffer, Date, etc).

Most part of this PR is no-op except the StatusCodeError change. I searched the packs repo and confirmed that it's ok.


huayang-codaio:
Oh the documentation changes are probably my local cache issue. Let me take a look.

api.ts
```diff
@@ -61,11 +62,25 @@ export class UserVisibleError extends Error {
   }
 }
 
+interface StatusCodeErrorResponse {
+  body?: any;
+  headers?: {[key: string]: string | string[] | undefined};
+}
+
+/**
+ * StatusCodeError is a simple version of StatusCodeError in request-promise to keep backwards compatibility.
```
jonathan-codaio, line 71: This class is exported in index.ts so it'll show up in public docs. Is this comment intended for Codans or users? If the latter, what does it mean to a user to keep backwards compatibility? I don't think they have context for what this is about.

huayang-codaio:
still there are a lot of documentation files changed in a legit way. kind of making PRs larger than they should be.

runtime/bootstrap/index.ts
```diff
@@ -0,0 +1,141 @@
...
+import type {ParamDefs} from '../../api_types';
+import type {ParamValues} from '../../api_types';
+import type {SyncFormulaResult} from '../../api';
+
+/**
+ * Setup an isolate context with sufficient globals needed to execute a pack.
+ *
+ * Notes:
+ * 1. JSON.parse/stringify are built into v8, so we don't need to inject those.
+ * 2. It is critically important that we do not leak isolated-vm object instances (e.g., Reference, ExternalCopy,
```
jonathan-codaio, line 16: Probably worth getting thoughts from @chris-codaio too, we obviously aren't doing security-through-obscurity but maybe we want to give some thought to the nature of what we describe here as we start bringing pieces of the runtime into the open-source repo.

jonathan-codaio (commented):
This seems reasonable, I'm curious to see the next stage of this though before we merge. Do you maybe want to cut a branch of this and send another PR based on this branch that shows some of the next phase of this?

jonathan-codaio:
> still there are a lot of documentation files changed in a legit way. kind of making PRs larger than they should be.

Yeah, because the docs link to line numbers, not sure if there's much we can do about that. They're auto-collapsed though and easy to sanity-check, so I'm not too bothered by it yet.

api.ts
huayang-codaio: ah good call. let me just remove the comment.

runtime/bootstrap/index.ts
huayang-codaio: ok let me give Chris a heads up.

huayang-codaio:
> This seems reasonable, I'm curious to see the next stage of this though before we merge. Do you maybe want to cut a branch of this and send another PR based on this branch that shows some of the next phase of this?

Sounds good. I'll keep this PR open while using it as a base branch to merge related changes.

runtime/bootstrap/index.ts
chris-codaio: Ack

huayang-codaio: haha, I was thinking about tomorrow but apparently you are already anywhere. :P

huayang-codaio:
oh I'd close this. Just realized that I don't need this to be open to merge into this branch.

PR #1218, opened 2021-07-22
add custom marshaling helpers

huayang-codaio:
Note: @jonathan-codaio I made this PR standalone to be merged into main because I feel like it will be useful anyway.

**Context**

We have to do value marshaling/unmarshaling on a few interfaces: 
- in/out IVM context
- grpc on certain unstructured data (e.g. fetcher response)

In these situations, we need to marshal/unmarshal on both sides to allow data transportation. At this moment, we use JSON.stringify + JSON.parse in most places plus a few special handling to Errors. 

This is not good enough. We should
- have a global marshaler/unmarshaler that applies to all these data transformation.
- and it should take care of most of the common data types that need to be marshaled. 

Particularly, this marshaling is needed in the packs-sdk for Pack local development with IVM. So this helper needs to live in packs-sdk or its dependency. Not the main coda repo.

**High level idea**

I searched open source solutions a bit but didn't find something that's close to what we need. So I decided to implement it. 

Types to be supported:
- selected [native javascript types](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects). You may notice that even simple types like Buffer, Date, NaN, regex can't be marshaled with JSON. 
- Specifically, selected Errors are supported. This include all the javascript system errors plus a whitelist of Coda errors. 

What's not supported:
- checkout `test/marshaling_test.ts` for a few test cases that I skipped.
- custom classes are not supported. 
- functions, native arrays, classes, promises are not supported but will not lead to error. In most cases they will become `{}`.

The implementation is straightforward: we pass a custom JSON replacer / reviver that handles specific types using custom injected keys (e.g. `__is_marshaled_buffer__`). 

**Testing**

Unittest.

**Caution**

~Some data types (i.e. Date) has its own toJSON method but returns a string that can't be used to identify types. I had to override the toJSON method to return an object but this breaks any code that relies on Date marshaled to a string in JSON. So I added a flag that makes this override only takes place during marshaling. This design kind of relying on marshaling to not run in parallel of any other code.~


runtime/common/marshaling/config.ts
```diff
@@ -0,0 +1,17 @@
+// This file is trying to create a global flag that toJSON overrides know to apply or not.
+// It's no-op for some objects but will change the toJSON format for Date.
+// The flag is not designed to be used by marshalers running in parallel. It's probably okay
+// if marshalers don't yield explicitly. In most cases JS runs with no parallelism.
+let _isMarshaling: boolean = false;
```
jonathan-codaio, line 5: Not sure I quite follow the need for the global flag and overriding prototype.toJSON. Doesn't passing a custom serializer to JSON.stringify() alone provide the necessary context and functionality to do custom marshaling?

huayang-codaio: The JS JSON.stringify lib works in a weird way. 
1. it first checks if the targeting object has a toJSON method. If there is, the library will call the toJSON function first.
2. then it calls the replacer to parse the intermediate result. 
3. lastly it serializes the result in the typical JSON way. 

For example, since Buffer.prototype.toJSON is defined, the replacer in step 2 will never receive an object that's instance of Buffer. The object will be serialized to an object before the replacer receives it. 

--- 

A few more things. 
- The protocol of toJSON is not symmetric in general since JSON.parse has nothing like fromJSON.
- there's no context passed all the way down to toJSON() so that we can't tell if toJSON() is called by the marshaler, or the user code. that's why we need a global flag here.

huayang-codaio: Other libraries. like jackson-js, allows passing down a stringify context, however that context is not passed to toJSON function.

huayang-codaio: I can add these to the code doc if that makes it more clear.

jonathan-codaio: Ah got it.

Can we instead just have the marshaler recursively traverse objects and convert Errors/Dates/Buffers to your new object representation, and then just call normal JSON.stringify() once at the end?

In theory what you have is probably safe, but it seems a little scary to be replacing the native toJSON() impls any time this code gets imported.

huayang-codaio: ok let me do that. it might be okay with a simplified version that only do the replacing for recognizable structures. at least it should work for our own use case.

runtime/common/marshaling/constants.ts
```diff
@@ -0,0 +1,11 @@
+export enum MarshalingInjectedKeys {
+  IsError = '__is_marshaled_error__',
+  ErrorClassName = '__error_class_name__',
+  ErrorClassType = '__error_class_type__',
+
+  IsBuffer = '__is_marshaled_buffer__',
```
jonathan-codaio, line 6: Nit, any reason to do separate keys for each of these rather than have a single key that's something like `__marshaled_value_type__` with values like `'number', 'buffer', 'date'`, etc? Seems like it scales a bit better to new marshalable types.

runtime/common/marshaling/marshal_numbers.ts
```diff
@@ -0,0 +1,18 @@
+import {MarshalingInjectedKeys} from './constants';
+
+export function marshalNumber(val: any): object | undefined {
+  // Most numbers don't need to be marshaled. The only special cases are NaN and Infinity.
```
jonathan-codaio, line 4: Ah good catch. TIL `JSON.stringify(NaN) === 'null'`, another JS head-scratcher

runtime/common/marshaling/marshal_errors.ts
```diff
@@ -0,0 +1,103 @@
...
+  StatusCodeError as any,
+];
+
+function getErrorClassType(err: Error): ErrorClassType {
+  if (recognizableSystemErrorClasses.some(cls => cls === err.constructor)) {
+    return ErrorClassType.System;
+  }
+
+  if (recognizableCodaErrorClasses.some(cls => cls === err.constructor)) {
+    return ErrorClassType.Coda;
```
jonathan-codaio, line 31: Any particular reason the Coda errors need to be handled differently than the system errors?

jonathan-codaio (commented):
Cool, looking good

runtime/common/marshaling/marshal_errors.ts
huayang-codaio: oh, theoretically, we can create a Coda error class called URIErrorr. 

actually I first separated them because there's one system error called InternalError. but then I realized that it was non-standard. 

let me simplify this a bit.

runtime/common/marshaling/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    throw new Error('marshaling value is too deep or containing circular strcture');
+  }
+
+  if (val === undefined || val === null) {
+    return val;
+  }
+
+  const serializedValue = serialize(val);
+
+  if (Array.isArray(serializedValue)) {
```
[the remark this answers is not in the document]
huayang-codaio: you are right.

runtime/common/marshaling/constants.ts
huayang-codaio: good idea.

runtime/common/marshaling/marshal_errors.ts
huayang-codaio: actually. let's leave this as is. I feel like it's more clear to have a separate namespace for coda errors.

jonathan-codaio (approved):

PR #1220, opened 2021-07-23
Moved READMEs to published doc

punit-codaio:


jonathan-codaio (commented):
Can you try running `make docs`? Not sure if your environment is set up properly for this. The README gets bundled into what shows up at https://coda.github.io/packs-sdk/, so rebuilding will update things there too.

If it doesn't work for you I can just run it after you merge this.

punit-codaio:
oh hmm i don't think i have my dev environment setup enough such that `make docs` just errors out when i try to run. do you mind doing that after i merge?

jonathan-codaio (approved):
Yeah go for it.

punit-codaio:
super thanks jonathan

PR #1222, opened 2021-07-26
use formula specification for internal implementation

huayang-codaio:
This is some preparation work to narrow the gap between packs sdk and coda runtime. **No breaking changes.**

Two changes in this PR
- use `FormulaSpecification` instead of formula name for internal implementation. SDK APIs are not changed so no breaking change. 
- get rid of formula namespaces in the packs-sdk repo tests. (should I update the types as well? which is a breaking change)

You may find changes in `docs/index.html` that doesn't seem to be related to this PR but is generated by make build. No idea.


docs/index.html
```diff
@@ -59,527 +59,8 @@ <h1>@codahq/packs-sdk</h1>
 				<a href="#coda-packs-sdk" id="coda-packs-sdk" style="color: inherit; text-decoration: none;">
 					<h1>Coda Packs SDK</h1>
 				</a>
-				<p><em>This is the documentation for the SDK to build Packs either via our Online Pack Editor or the Command Line Interface (CLI) tool. Click <a href="https://coda.io/@coda/make-a-pack-private-alpha">here</a> to learn how to use the Online Pack Editor and Pack making more broadly, and <a href="README_CLI.md">here</a> to learn how to use the CLI.</em></p>
```
huayang-codaio, line 62: doesn't seem related to this PR.

runtime/types.ts
```diff
@@ -0,0 +1,63 @@
+export enum FormulaType {
```
huayang-codaio, line 1: copied from coda/.

huayang-codaio (commented):
~actually wait a second. let me check again if there's any breaking change.~

huayang-codaio (commented):
`executeFormulaOrSyncWithVM` is a breaking change. it's exported in development. let me fix that.

huayang-codaio:
also seems like we should separate exported code and internal code. it's not easy even to figure out if I had a breaking change in a PR at the moment.

jonathan-codaio (commented):
This looks reasonable.

Just out of paranoia, do you mind trying to use the SDK commit version in the `packs` repo and making sure all the unittests compile and run? Just checking backwards compatibility. I know you mentioned the public API isn't changing, so should be fine, but just making sure we weren't relying on any weird behavior for existing tests.

huayang-codaio:
sure let me do that.

huayang-codaio:
@jonathan-codaio I think the packs repo build is good. the bundle build changes are caused by https://github.com/coda/packs-sdk/pull/1216.

jonathan-codaio (approved):

PR #1224, opened 2021-07-26
Fix TS type defs for authentication definitions.

jonathan-codaio:
Some of the usual crap with a union of Omits being different than an Omit of a union.

PTAL @huayang-codaio @abhishek-codaio @coda/packs


abhishek-codaio (approved):

huayang-codaio (commented):
cool. thanks!

PR #1225, opened 2021-07-26
add a note for oauth providers only taking comma-separated scopes

huayang-codaio:
Caught today debugging the strava pack. 

This is not a fix but just adding a note before we want to decide if this is ever going to be fixed.

### DIFF
--- a/docs/enums/FeatureSet.html
+++ b/docs/enums/FeatureSet.html
@@ -87,7 +87,7 @@
 					<div class="tsd-signature tsd-kind-icon">Basic<span class="tsd-signature-symbol">:</span> <span class="tsd-signature-symbol"> = &quot;Basic&quot;</span></div>
 					<aside class="tsd-sources">
 						<ul>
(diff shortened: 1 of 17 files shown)


jonathan-codaio (commented):
This makes sense, but given that we track OAuth scopes as structured data in the DB now, this would break that tracking if we advised people to do this. So I'm not sure we want to endorse this workaround more broadly, behind internal use on the adhoc.

Would it be possible to fix this by adding an optional delimeter setting on the oauth def in the manifest? (Not saying it's a priority to do this)

FYI @patrick-codaio

huayang-codaio:
we definitely can if it's necessary to do so. there might be a few places to check & fix. 

so far I only saw one oauth provider not accepting the standard scopes protocol.

patrick-codaio:
Thanks for the heads up.

huayang-codaio:
let me close this one and add it to the bug list I guess.

PR #1238, opened 2021-07-27
use a default output directory build/ to write bundle files

huayang-codaio:
set a default output directory for CLI build / upload to avoid putting bundle files in the working directory. 

note: the bundle files can be put in a temporary dir outside the working directory because of the source map tools don't work with that. so we pick a common build/ directory for the output. 

tested in the packs repo.

success:

```
$. ./node_modules/.bin/coda upload packs/discourse/main/manifest.ts --codaApiEndpoint=https://packs.adhoc.coda.io
[info/2021-07-27T23:05:55.095Z]: Building Pack bundle...
[info/2021-07-27T23:05:55.096Z]: Existing directory ./_upload_build detected. Probably left over from previous build. Removing it...
[info/2021-07-27T23:06:03.664Z]: Validating Pack metadata...
[info/2021-07-27T23:06:03.674Z]: Registering new Pack version...
[info/2021-07-27T23:06:04.267Z]: Uploading Pack...
[info/2021-07-27T23:06:04.739Z]: Validating upload...
[info/2021-07-27T23:06:09.071Z]: 

Cleaning up...
[info/2021-07-27T23:06:09.073Z]: Intermediate files are moved to /var/folders/n1/7qfgvcqn04j0py98bvnsnd500000gp/T/coda-packs-2cad2218-1114-4b2b-82c8-333bc18f869frqre2u
[info/2021-07-27T23:06:09.073Z]: Done!

$ ls /var/folders/n1/7qfgvcqn04j0py98bvnsnd500000gp/T/coda-packs-2cad2218-1114-4b2b-82c8-333bc18f869frqre2u
browserify-bundle.js     browserify-bundle.js.map bundle.js
```

failure:

```
$ ./node_modules/.bin/coda upload packs/coda/debug/manifest.ts --codaApiEndpoint=https://packs.adhoc.coda.io
[info/2021-07-27T23:06:54.901Z]: Building Pack bundle...
[info/2021-07-27T23:06:57.836Z]: Validating Pack metadata...
[info/2021-07-27T23:06:57.842Z]: Registering new Pack version...
[info/2021-07-27T23:06:58.060Z]: 

Cleaning up...
[info/2021-07-27T23:06:58.062Z]: Intermediate files are moved to /var/folders/n1/7qfgvcqn04j0py98bvnsnd500000gp/T/coda-packs-9904a071-7284-424d-8ab4-be5a932f9cc2GTpYhq
Error while registering pack version: {
  statusCode: 400,
  statusMessage: 'Bad Request',
  message: 'New version 1.0.0 for Pack 2100000219 must be greater than the last registered version 4.0.6'
}

$ ls /var/folders/n1/7qfgvcqn04j0py98bvnsnd500000gp/T/coda-packs-9904a071-7284-424d-8ab4-be5a932f9cc2GTpYhq
browserify-bundle.js     browserify-bundle.js.map bundle.js                bundle.js.map            node-bundle.js           node-bundle.js.map
```


.gitignore
```diff
@@ -7,5 +7,4 @@ artifacts
 dist/test
 yarn-error.log
 local-docs
-bundle.js
-bundle.js.map
+./build
```
jonathan-codaio, line 10: As mentioned before, I think what we want to do is move the files from here to a temp dir after the compilation is completed, and not leave any artifacts in the working directory.

Having this git-ignored here is hiding the fact in our testing that when external developers compile/upload, it will leave git-visible artifacts in their project directory. They would need to know to git-ignore them manually.

If it's necessary to keep the build artifacts around after the CLI command has completed, we should move them to a temp dir and print the location, otherwise we should just delete them.

huayang-codaio: hmm. I thought we agreed on to use a predefined output directory for the bundle outputs. 

well I think it's necessary to keep these build artifacts around so that the developer has a record of what was uploaded by the CLI. IMO in a typical workflow it's not moved into a temporary dir because that means it will get lost soon.

huayang-codaio: oh I misunderstood your comment on the ticket but I disagree that it's a bad thing to leave temp build files in the working directory. I do feel like it's a common flow to allow ppl inspect / keep track of what's done. and that's why they need .gitignore.

jonathan-codaio: It's not that it's a problem to have artifacts in the working directory, it's just that I don't think the default behavior of the CLI should be to leave artifacts that need to be manually git-ignored by every single pack author. The reason I became aware that it was happening was because an alpha user already complained about it.

huayang-codaio: alright. I hold my opinion but let's follow the user ask in this case.

huayang-codaio: updated & tested.

jonathan-codaio (approved):
Thanks!

PR #1245, opened 2021-07-29
build isolated binary that runs on amazon linux 2

huayang-codaio:
per slack discussion, we want to build an isolated-vm binary that's compatible with lambda instance type (i.e. amazon linux 2). 

the 3 files in `runtime/isolated-vm/` are basically all we need for lambda functions to use isolated vm. 

@chris-codaio after thinking about this again, it seems better to build this binary file in the primary coda repo with `make build-packs-runtime`. what do you think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,12 +9,6 @@ YARN_CACHE_DIR=~/.yarncache
 ISOLATED_VM_VERSION_COMMAND="require('./node_modules/isolated-vm/package.json').version"
 ISOLATED_VM_VERSION=$(shell node -p -e $(ISOLATED_VM_VERSION_COMMAND))
 
-iv:
-	if [ `node -p -e "require('./runtime/isolated-vm/package.json').version"` == $(ISOLATED_VM_VERSION) ]; \
-		then echo 'here'; \
-		else echo 'there'; \
-	fi
-
 # Aliases
 bs: bootstrap
 
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,15 @@ ROOTDIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 # CircleCI yarn cache directory may also need to be updated in sync with this
 YARN_CACHE_DIR=~/.yarncache
 
+ISOLATED_VM_VERSION_COMMAND="require('./node_modules/isolated-vm/package.json').version"
+ISOLATED_VM_VERSION=$(shell node -p -e $(ISOLATED_VM_VERSION_COMMAND))
+
+iv:
+	if [ `node -p -e "require('./runtime/isolated-vm/package.json').version"` == $(ISOLATED_VM_VERSION) ]; \
+		then echo 'here'; \
+		else echo 'there'; \
+	fi
+
 # Aliases
 bs: bootstrap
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -6,6 +6,15 @@ ROOTDIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 # CircleCI yarn cache directory may also need to be updated in sync with this
 YARN_CACHE_DIR=~/.yarncache
 
+ISOLATED_VM_VERSION_COMMAND="require('./node_modules/isolated-vm/package.json').version"
+ISOLATED_VM_VERSION=$(shell node -p -e $(ISOLATED_VM_VERSION_COMMAND))
+
+iv:
+	if [ `node -p -e "require('./runtime/isolated-vm/package.json').version"` == $(ISOLATED_VM_VERSION) ]; \
+		then echo 'here'; \
```
jonathan-codaio, line 14: Think this is test code that you can remove.

Makefile
```diff
@@ -36,6 +45,27 @@ lint:
 lint-fix:
 	find . -name "*.ts" | grep -v /dist/ | grep -v /node_modules/ | grep -v .d.ts | xargs ${ROOTDIR}/node_modules/.bin/eslint --fix
 
+.PHONY: do-compile-isolated-vm
+do-compile-isolated-vm:
+	rm -rf build-isolated-vm
+
+	mkdir build-isolated-vm && \
```
jonathan-codaio, line 52: Should we make this a subdirectory of our existing `build` directory?

Makefile
```diff
@@ -36,6 +45,27 @@ lint:
...
+	cp build-isolated-vm/node_modules/isolated-vm/isolated-vm.js runtime/isolated-vm/
+	cp build-isolated-vm/node_modules/isolated-vm/out/isolated_vm.node runtime/isolated-vm/out/
+
+	rm -rf build-isolated-vm
+
+.PHONY: compile-isolated-vm
+compile-isolated-vm:
+	if [ `node -p -e "require('./runtime/isolated-vm/package.json').version"` != $(ISOLATED_VM_VERSION) ]; \
+		then $(MAKE) do-compile-isolated-vm; \
+		else echo "isolated-vm version matches, skipping."; \
```
jonathan-codaio, line 66: 👍 Smart

Makefile
```diff
@@ -6,6 +6,15 @@ ROOTDIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 # CircleCI yarn cache directory may also need to be updated in sync with this
 YARN_CACHE_DIR=~/.yarncache
 
+ISOLATED_VM_VERSION_COMMAND="require('./node_modules/isolated-vm/package.json').version"
```
jonathan-codaio, line 9: Little unclear on how the versions will change. This seems to be checking the version that's installed in the main project, which can get updated by dependabot. The separate package.json in this file is at a frozen version. Isn't the frozen version all that matters? Or we're ok letting dependabot update the version for us, and it triggers us building a new version for lambda, which will get used the next time we bump the SDK in `coda`?

Makefile
jonathan-codaio: Oops, forgot this wasn't in the `coda` repo, ignore :)

Makefile
huayang-codaio: oh package.json doesn't specify a fixed version. it uses something like ^4.3.2 and allows a patch version difference from yarn install. the idea is basically to use the version from yarn.lock which is more precise. 

instead of parsing yarn.lock, `require('./node_modules/isolated-vm/package.json').version` gives us the same version number.

huayang-codaio:
I am actually going to make another PR to do this in the coda repo and see if that works better.

huayang-codaio:
ok. think it's a better idea to move this to coda since packs-sdk doesn't really need the lambda binary otherwise.

huayang-codaio:
ok maybe we should actually do this. reopening.

jonathan-codaio (approved):
Ok let's see how this goes.

Makefile
```diff
@@ -6,6 +6,9 @@ ROOTDIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
...
+ISOLATED_VM_VERSION=$(shell node -p -e $(ISOLATED_VM_VERSION_COMMAND))
```
chris-codaio, line 10: I don't think node is installed in many CircleCI images.  We typically rely on python commands for our build scripts.

huayang-codaio: yeah I checked that we are using `cimg/node:14.13` for this repo's CI docker image. in fact most commands in this Makefile will fail if node is not available.

PR #1247, opened 2021-07-29
merge coda & packs-sdk ivm runtime

huayang-codaio:
This PR is a parent of a few PRs. The main ones are:
- copying over most IVM files from coda runtime (no PR, just copying files over).
- use thunk from coda runtime bundle instead of the local execution_helper bundle. (https://github.com/coda/packs-sdk/pull/1226)
- make the coda runtime interface general to be used by both packs-sdk and lambda. Get this working on an adhoc using thunk.js from packs-sdk. (https://github.com/coda/packs-sdk/pull/1241)

After all, this PR should not impact coda primary repo before we do something like https://github.com/coda/coda/compare/hg-common-packs-runtime?expand=1

It should also cause no impact to CLI users since no API interface is changed (except StatusCodeError, but I suppose nobody is using it).


huayang-codaio:
@jonathan-codaio can you take a look at this one? no new changes. most you have reviewed already.

jonathan-codaio (commented):
Yep sorry, just behind on stuff.

Remind me, does this currently still only invoke the VM when the --vm flag is used, and the normal invocation path `coda execute` is unchanged?

I patched this into the `packs-examples` repo and there are some TS errors, I suspect there will be similar errors in the `packs` repo. I think the errors are ok, I think what happened is that the inferred return type of `executeFormulaFromPackDef` changed, and is now `PackFormulaResult`, which seems correct, but maybe it used to be `any` or something? So many of the callers of that are now getting errors because PackFormulaResult is too generic for their usage.

I think it's probably correct that it returns `PackFormulaResult` and maybe we should just declare that it returns that instead of letting it infer a type that can flap over time? Then the usage in `packs-examples` and `packs` needs a fast followup to do proper type assertions/casts.

huayang-codaio:
let me take a look into the TS errors. 

re `--vm` flag. This PR is not supposed to introduce any user visible functional changes. I take it a parallel effort to make vm as default option.

jonathan-codaio:
> let me take a look into the TS errors.
> 
> re `--vm` flag. This PR is not supposed to introduce any user visible functional changes. I take it a parallel effort to make vm as default option.

Yep cool, just wanted to confirm the intent.

huayang-codaio:
I have https://github.com/coda/packs/pull/2582 and https://github.com/coda/packs-examples/pull/40

jonathan-codaio (approved):

api.ts
```diff
@@ -61,11 +62,25 @@ export class UserVisibleError extends Error {
...
+/**
+ * StatusCodeError is a simple version of StatusCodeError in request-promise to keep backwards compatibility.
+ */
 export class StatusCodeError extends Error {
   statusCode: number;
-  constructor(statusCode: number) {
-    super(`statusCode: ${statusCode}`);
+  body: any;
+  options: FetchRequest;
+  response: StatusCodeErrorResponse;
```
jonathan-codaio, line 77: Was just seeing the usage in the `packs` change--are all these fields intentionally required?

huayang-codaio: yeah, basically because statuscodeerror (which usually returns from request-promise) is treated as a user visible error, and we need extra information for logging / display purpose. 

the fact with packs repo is that we sometimes try to create a fake status code error. they display similarly to a real network error.

PR #1265, opened 2021-08-03
add @coda/infosec to runtime/ reviews

huayang-codaio:
I am not sure if this works. Following the github instruction this should be sufficient but I am not sure if it actually works after trying to change some random files in the runtime repo. 

I wonder if this only works after being committed?


CODEOWNERS
```diff
@@ -1 +1,2 @@
-* @coda/ecosystem
\ No newline at end of file
+*                 @coda/ecosystem
```
jonathan-codaio, line 1: Feel free to change this to coda/packs as well.

jonathan-codaio (approved):

huayang-codaio:
ok indeed it only works after being committed.

https://github.com/coda/packs-sdk/pull/1266

PR #1269, opened 2021-08-04
update autocomplete snippets and examples schema

lucas-codaio:
# Summary
* Update the schemas for autocomplete snippets and examples
* Modify the compiler to handle the schema changes


documentation/documentation_compiler.ts
```diff
@@ -32,32 +33,62 @@ function compileSnippets() {
...
+  const fileName = baseName.split('.')[0];
+  const snippetNameArray = fileName.split('_');
+  return snippetNameArray
+    .map(segment => {
+      return `${segment.charAt(0).toUpperCase()}${segment.slice(1)}`;
+    })
+    .join(' ');
+}
+
+function isValidReferencePath(referencePath: string): boolean {
```
lucas-codaio, line 81: Can only handle reference paths that include a `fragmentIdentifier` (i.e. `.hmtl#fragmentIdentifier`). 

It's fine for the current use case of examples for builder function, but this can't validate paths to `types`

documentation/documentation_compiler.ts
```diff
@@ -32,32 +33,62 @@ function compileSnippets() {
...
   return fs.readFileSync(path.join(DocumentationRoot, file), 'utf8').trim();
 }
 
-function compileCodeFiles(example: Example) {
-  const code = example.codeFiles.map(codeFile => {
-    return getCodeFile(codeFile);
+function compileExampleSnippets(example: Example): CompiledExampleSnippet[] {
+  return example.exampleSnippets.map(exampleSnippet => {
+    return {
+      name: formatSnippetName(path.basename(exampleSnippet.codeFile)),
```
jonathan-codaio, line 64: I would think the name should just be hand-written in the snippet definition.

documentation/documentation_compiler.ts
```diff
@@ -32,32 +33,62 @@ function compileSnippets() {
...
+    };
   });
-  return code;
 }
 
-function getCodeFile(file: string) {
-  const data = fs.readFileSync(path.join(DocumentationRoot, file), 'utf8');
-  const codeStart = data.indexOf(codeBegin) + codeBegin.length;
-  return data.substring(codeStart).trim();
+function formatSnippetName(baseName: string): string {
```
jonathan-codaio, line 71: There's a `pascalcase` utility in this repo already, but as above, I think we should just create a name explicitly for each example.

documentation/types.ts
```diff
@@ -1,35 +1,37 @@
...
+}
+
+export interface CompiledExampleSnippet {
+  name: string;
+  content: string;
+  code: string;
 }
 
 export interface Example {
+  tokens: string[];
```
jonathan-codaio, line 26: Should still be called `triggerTokens` I think? Their use is for triggering the example.

Should also call the autocomplete triggers `triggerTokens` for consistency.

documentation/documentation_config.ts
```diff
@@ -1,180 +1,182 @@
...
+      },
+    ],
   },
   {
+    tokens: ['addFormula'],
     contentFile: './examples/formula/formula.md',
-    codeFiles: ['./examples/formula/basic_formula.ts'],
-    categories: [Category.Formula],
-    triggerTokens: ['addFormula'],
+    referencePath: 'classes/PackDefinitionBuilder.html#addFormula',
```
jonathan-codaio, line 163: Nit, this one seems to be missing the leading slash.

documentation/types.ts
```diff
@@ -1,35 +1,37 @@
...
+  referencePath: string;
```
jonathan-codaio, line 27: Nit, can we call this `sdkReferencePath` for clarity.

documentation/types.ts
```diff
@@ -1,35 +1,37 @@
...
   contentFile: string;
-  codeFiles: string[];
-  categories: Category[];
-  triggerTokens: string[];
+  exampleSnippets: ExampleSnippet[];
 }
 
 export interface CompiledExample {
+  tokens: string[];
+  referencePath: string;
```
jonathan-codaio, line 34: Ditto, triggerTokens and sdkReferencePath.

documentation/documentation_compiler.ts
lucas-codaio: I thought it made sense for the example snippet file name to also be the name of the example.

jonathan-codaio: Yeah, they will likely track pretty closely in most cases, but it's a bit unintuitive that your filename will become a prettified name, and it eliminates the ability to do basic things like put non-alphanumeric characters in the title. Say I wanted a title that said "Dynamic Sync Tables (Advanced)", that would be impossible. Generally I prefer that easy things are easy to accomplish, so I'm hesitant for a system where somebody asks for a parenthesis in a title and we have to tell them they're out of luck.

lucas-codaio: Gotcha, def more flexible to have a separate name property. I'll make that change

documentation/types.ts
```diff
@@ -1,35 +1,38 @@
...
+  triggerTokens: string[];
   content: string;
   codeFile: string;
 }
 
-export interface CompiledSnippet {
-  name: string;
-  triggerWords: string[];
+export interface CompiledAutocompleteSnippet {
+  // The first item in triggerWords is used as the autocomplete label in the packs IDE
```
jonathan-codaio, line 8: Nit, triggerTokens in comment

jonathan-codaio (approved):
🎉

PR #1271, opened 2021-08-05
a few changes for debugging v1 packs

huayang-codaio:
sending this out since it's getting bigger than I expected. basically a few necessary changes to help debugging v1 packs. 

- support a minify flag for CLI buid for debugging purpose. (easier to identify issues).
- add a timer method in the fake pack for debugging. (to debug setTimeout)
- add an extra level of esbuild for shim injection. (for setTimeout)
- translate error stack from loadIntoIVM (this happens)
- add shim for timing methods (i.e. setTimeout)
- don't build another bundle for exec-in-vm

I am adding `timers` as an option instead of the default because I still need to test esbuild for browserifying capability.

### DIFF
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -34,6 +34,11 @@ if (require.main === module) {
           string: true,
           desc: 'For a dynamic sync table with a variable source location, specify the URL to test here.',
         },
+        timers: {
+          boolean: true,
+          default: false,
+          desc: 'Set to true to enable timer methods in the build. i.e. setTimeout',
+        },
       },
     })
     .command({
@@ -82,6 +87,15 @@ if (require.main === module) {
           string: true,
           alias: 'o',
           default: undefined,
+        },
+        minify: {
+          boolean: true,
+          default: true,
+        },
+        timers: {
+          boolean: true,
+          default: false,
+          desc: 'Set to true to enable timer methods in the build. i.e. setTimeout',
         },
       },
       handler: handleBuild,
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -39,6 +39,11 @@ if (require.main === module) {
           default: false,
           desc: 'Set to true to enable timer methods in the build. i.e. setTimeout',
         },
+        browserifyWithEsbuild: {
+          boolean: true,
+          default: false,
+          desc: 'Set to true to let esbuild do browserify',
+        },
       },
     })
     .command({
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -39,11 +39,6 @@ if (require.main === module) {
           default: false,
           desc: 'Set to true to enable timer methods in the build. i.e. setTimeout',
         },
-        browserifyWithEsbuild: {
-          boolean: true,
-          default: false,
-          desc: 'Set to true to let esbuild do browserify',
-        },
       },
     })
     .command({
--- a/testing/injections/timers_disabled_shim.js
+++ b/testing/injections/timers_disabled_shim.js
@@ -1,0 +1,13 @@
+// ts file compiled js file somehow end up with cjs format but esbuild needs esm format for shims.
+// so for now we just have this file in js.
+
+const ErrorMessage = 'Please use --timers option to enable timers. Be aware that timing methods ' + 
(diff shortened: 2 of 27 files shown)


testing/injections/timers_shim.js
```diff
@@ -0,0 +1,14 @@
+// ts file compiled js file somehow end up with cjs format but esbuild needs esm format for shims.
```
patrick-codaio, line 1: isn't that controllable in tsconfig?

testing/injections/timers_shim.js
```diff
@@ -0,0 +1,14 @@
...
+// so for now we just have this file in js.
+
+// per https://github.com/laverdet/isolated-vm/issues/136, passing the real setTimeout to 
+// an ivm context isn't safe. so we create a workaround to use a Promise to yield execution
+// of the current thread. 
+export function setTimeout(callback, _) { new Promise(_ => callback()) }
+
+// can't actually set internal in ivm. this would only be executed once. maybe we should
```
patrick-codaio, line 9: typo: internal => interval

patrick-codaio (commented):
I defer approval to folks with more experience in this code, but LGTM. Thanks for wrestling all these error flows Huayang!!

testing/injections/timers_shim.js
huayang-codaio: it's controllable, but not on a single-file granularity (-> I didn't figure this out). I am not sure if it's time to move everything to esm because of this.

cli/coda.ts
```diff
@@ -34,6 +34,16 @@ if (require.main === module) {
           string: true,
           desc: 'For a dynamic sync table with a variable source location, specify the URL to test here.',
         },
+        timers: {
+          boolean: true,
+          default: false,
+          desc: 'Set to true to enable timer methods in the build. i.e. setTimeout',
+        },
+        browserifyWithEsbuild: {
```
huayang-codaio, line 42: unfortunately I confirmed that esbuild is not worth the trust to browserify node modules. there are tons of issues to resolve "browser" dependencies that esbuild isn't correctly handling. for example https://github.com/evanw/esbuild/issues/772. 

I also tested with our packs repo. the local "browser" field mapping doesn't work well with multiple entries that depend on each other.

testing/injections/timers_disabled_shim.js
```diff
@@ -0,0 +1,13 @@
+// ts file compiled js file somehow end up with cjs format but esbuild needs esm format for shims.
+// so for now we just have this file in js.
+
+const ErrorMessage = 'Please use --timers option to enable timers. Be aware that timing methods ' +
```
jonathan-codaio, line 4: Nit, I would be pretty specific here, something like:

Please use the --timers flag to enable shimmed timing primitives. Native node timing primitives like setTimeout or setInternal aren't supported in the pack execution sandbox environment. However, if you are using a library that relies upon them, you can use the --timers flag to build your pack with shimmed implementations that approximate the native behavior. Because of this, be aware that packs that use timing primitives may not work reliably.

testing/injections/timers_disabled_shim.js
```diff
@@ -0,0 +1,13 @@
+// ts file compiled js file somehow end up with cjs format but esbuild needs esm format for shims.
```
jonathan-codaio, line 1: This files are both in the `testing` directory--just so I understand, these shims happen at compile time, and not at runtime in creation of the isolate? So if I used setTimeout() from the web editor, I would just get an error that it's not defined?

(Totally separate, we should move the compiler stuff into its own directory instead of `testing` at some point.)

huayang-codaio: yes, it's only at the compile time. 

re web editor: the current web editor compile flow is missing a few things from the CLI, e.g. shims or other browserify stuff. I've already seen issues with that and have a ticket to track. 

re: testing/compile. totally. let me do that later. I am currently in a big mess of changes to debug the drive pack.

jonathan-codaio: Yep re moving the files, just flagging, definitely not urgent.

jonathan-codaio (approved):
Ok LGTM mode the error text.

PR #1272, opened 2021-08-05
fix fetcher response on binary response.

huayang-codaio:
finally figured this out after a whole day of debugging ... turns out the fix is pretty simple.

context: when you run a pack that fetches binary response with the packs-sdk, it gets a different response from what it gets in the v1 world ...

so first `isBinaryResponse` needs to be passed through. this is easy. 

the tricky part is that request-promise-native returns a Buffer-ish object in that case. it looks identical to a node Buffer but Buffer.isBuffer returns false on that object! this leads to mysterious behavior in the system esp when it goes through marshal / unmarshal or anywhere uses isBuffer check. 

The solution is to rebuild the object into a native Buffer.


spencer-codaio (commented):
curious why does v2 get to skip the JSON stringify/parse step?

huayang-codaio:
in v1 you can only execute a pack with calc server. in v2 you can execute it with CLI or in a unittest, where the response isn't passed between services so that JSON stringify/parse isn't applied.

spencer-codaio:
@huayang-codaio ah ok so you can pass whatever format you want / expect whatever format inside a pack since a user could pass in whatever they want from CLI?

huayang-codaio:
oh, I guess I mislead you on this. In the v1 world, the JSON serialization happens on the network boundary when data is passed between services. 

In some v2 supported executions (e.g. in the CLI, unittests, etc), the data fetching and the pack execution happens in the same JS process where we don't apply JSON serialization. This has nothing to do with processing user CLI inputs, just that don't do it within a single process.

jonathan-codaio (approved):
Oof, that's subtle.

huayang-codaio:
ah after testing I realized that this only fixes the non-vm execution. the vm execution still creates a new Buffer object. 

specifically, the unmarshalBuffer method registered in one `eval` will give the pack in another `eval` a buffer-ish objet but Buffer.isBuffer still returns false. 

to fix that I think we need to get rid of the context fetcher from the pack code (which is registered in another eval) and use some helper method from the packs-sdk instead. this is too complicated..

huayang-codaio:
let me revert the buffer reconstruction change for now. it needs a better solution.

chris-codaio:
For added fun, `request-promise-native` is deprecated and we should be migrating to `node-fetch`.  Mind queuing that change up as a followup?

huayang-codaio:
> For added fun, request-promise-native is deprecated and we should be migrating to node-fetch. Mind queuing that change up as a followup?

I do believe we want to do that. Probably after the v1 packs migration though to keep the interface consistent during the v2 migration.

PR #1273, opened 2021-08-06
allow use a custom scope separator

huayang-codaio:
just send this out since a few scenarios we've seen recently. maybe it makes sense to support it now. 

two notes:
- this PR isn't sufficient. it needs a follow up in coda repo. 
- piggy back to 0.4.6 since it's not published yet. 

CLI tested with zoom pack:

```
./node_modules/.bin/coda auth packs/zoom/user/manifest.ts
```


huayang-codaio:
regarding `one option is to do the scope join before passing it into ClientOAuth2. not sure if we should do so.` 

instead of currently doing

```
  const oauth2Client = new ClientOAuth2({
    clientId,
    clientSecret,
    authorizationUri: authorizationUrl,
    accessTokenUri: tokenUrl,
    scopes: requestedScopes,
    redirectUri: makeRedirectUrl(port),
    query: additionalParams,
  });

```

we can do

```
  const oauth2Client = new ClientOAuth2({
    clientId,
    clientSecret,
    authorizationUri: authorizationUrl,
    accessTokenUri: tokenUrl,
    scopes: [requestedScopes.join(separator)],
    redirectUri: makeRedirectUrl(port),
    query: additionalParams,
  });

```

types.ts
```diff
@@ -155,6 +155,9 @@ export interface OAuth2Authentication extends BaseAuthentication {
 
   // Some OAuth providers require passing the token as a URL param.
   tokenQueryParam?: string;
+
+  // The charactor used to separate multiple OAuth scopes. Default is a whitespace.
+  scopeSeparator?: string;
```
jonathan-codaio, line 160: Total nit, can we call it `scopeDelimiter` to be precise?

types.ts
```diff
@@ -155,6 +155,9 @@ export interface OAuth2Authentication extends BaseAuthentication {
 
   // Some OAuth providers require passing the token as a URL param.
   tokenQueryParam?: string;
+
+  // The charactor used to separate multiple OAuth scopes. Default is a whitespace.
```
jonathan-codaio, line 159: Nit, would say `Default is a single space, per the OAuth standard.` Would use a `/** */` style comment while you're here so it makes it into TypeDoc (we'll eventually go clean up the rest of them).

jonathan-codaio (approved):
Thanks!

Yeah, if your approach for client-oauth2 works, go for it, would just leave a comment describing why we do it.

huayang-codaio:
The client-oauth2 hack works for zoom pack.

chris-codaio (commented):
From a security POV, I'm a little worried that you could hide other scopes in the "separator" here that we wouldn't show on a packs page or somewhere an IT admin can get at it.  Can we enforce that this is a single character and one of a small set of known safe separators?  Like `[' ', ':', ',', ';']`?

huayang-codaio:
> From a security POV, I'm a little worried that you could hide other scopes in the "separator" here that we wouldn't show on a packs page or somewhere an IT admin can get at it. Can we enforce that this is a single character and one of a small set of known safe separators? Like [' ', ':', ',', ';']?

Thanks for catching this! Updated.

PR #1274, opened 2021-08-08
browserify thunk bundle

huayang-codaio:
In the IVM context today, code executed comes from 3 sources:
1) the pack bundle. 
2) the thunk bundle.
3) the hard-coded code passed into context.eval or context.evalClosure. 

Ideally, 3) should be as simple as possible and usually should just be calling the thunk code instead. 

As this moment, pack bundle is browserified by thunk bundle is not. This causes problems when thunk bundle tries to access nodejs resources. The first thing I found is that our Fetcher may return a Buffer which is a node object. The marshaling method in the thunk bundle isn't working properly without knowing what a Buffer is. Beyond this case, other nodejs dependency may be introduced in the future so I think we should just browserify the thunk bundle. 

This PR leverages the existing compile helper to browserify & minify the thunk bundle. It also generates the source map of the thunk that we can leverage later.


chris-codaio:
What new code does this inject into the isolate and make available to the packs code in a new way (around the isolate boundary)?  I'm worried about the expansion of attackable surface area.

huayang-codaio:
> What new code does this inject into the isolate and make available to the packs code in a new way (around the isolate boundary)? I'm worried about the expansion of attackable surface area.

@chris-codaio I briefly looked into the new thunk_bundle. browserify doesn't introduce new packages for the bundle. The only change at this moment is to inject a new Buffer class and replace all usage of Buffer usage to the new class.

chris-codaio:
> > What new code does this inject into the isolate and make available to the packs code in a new way (around the isolate boundary)? I'm worried about the expansion of attackable surface area.
> 
> @chris-codaio I briefly looked into the new thunk_bundle. browserify doesn't introduce new packages for the bundle. The only change at this moment is to inject a new Buffer class and replace all usage of Buffer usage to the new class.

Is that because there's a reference to `Buffer`?  What if we accidentally add a reference to `setImmediate`?

huayang-codaio:
> Is that because there's a reference to Buffer? What if we accidentally add a reference to setImmediate?

@chris-codaio browserify isn't replacing setTimeout etc because it's available on both nodejs & browser. We do that with a shim which does not apply for the thunk bundle. After all as long as we don't register an IVM stub to the setTimeout outside the context, it should not do anything bad but would only behave differently from expectation.

jonathan-codaio (commented):
Is it possible to whitelist what gets polyfilled in browserify? So, just start with polyfilling Buffer for now, and we could add specific other items as needed?

huayang-codaio:
> Is it possible to whitelist what gets polyfilled in browserify? So, just start with polyfilling Buffer for now, and we could add specific other items as needed?

@jonathan-codaio I am not sure if that's configurable with browserify, but that's doable with esbuild. If we want to take the approach to whitelist polyfills, I can do that with something similar to the `timers_shim.js`. 

I think the debate is if we really care about the security of the thunk code. My personal feeling is that the pack code can do whatever we do in the thunk code so that securing the thunk code isn't the most critical thing. The most straightforward attack that I can think of is to inject some malware code with browserify that may break packs or steal info from packs. That's basically what can happen with any code. 

IMO the boundary we really care about and want to secure is the bootstrap setup code that defines what resources are given to the ivm context.

jonathan-codaio:
Wondering if we should just follow the timers_shim.js approach to inject this in a clear, isolated way for now?

I agree though that browserify-ing the thunk or anything doesn't seem like it changes anything security-wise. It's not injecting any node primitives into the isolate at all, right? It's just creating some pure-JS polyfills that are subject to the same sandboxing as any other code.

huayang-codaio:
sure. we can start with that. I don't see any potential dependency of node modules except Buffer in the near term. the whitelist approach should work.

huayang-codaio:
closing this PR as most changes will be discarded.

PR #1284, opened 2021-08-10
by default compile with no timer shims

huayang-codaio:


jonathan-codaio (approved):
Fine for now, but fragile we should clean this up after.

huayang-codaio:
will make that more verbal later. maybe with an enum.

PR #1285, opened 2021-08-11
refactor enableTimers from a boolean to more specific timerStrategy

huayang-codaio:
In an earlier quick fix I made `enableTimers` to be undefined, false, true. Each has a different behavior on the timer's shim. 

This is confusing. This PR makes the option more specific by adding an enum of the available options.


testing/compile.ts
```diff
@@ -103,10 +110,23 @@ async function uglifyBundle({
   fs.writeFileSync(`${outputBundleFilename}.map`, uglifyOutput.map);
 }
 
+function getInjections({timerStrategy = TimerShimStrategy.Disable}: CompilePackBundleOptions) {
```
jonathan-codaio, line 113: Nit, maybe getTimerInjections, since this is specific to timers?

testing/compile.ts
```diff
@@ -103,10 +110,23 @@ async function uglifyBundle({
...
+  switch (timerStrategy) {
+    case TimerShimStrategy.Disable:
+      return [];
```
jonathan-codaio, line 116: When I saw the option "Disable", I thought that mean, disable timers, but I think it means, don't use a shim? In which case I think `TimerShimStrategy.None` is maybe clearer?

testing/compile.ts
huayang-codaio: oh I meant to have this method return all injections. it's just that we only have timer shim right now.

jonathan-codaio (approved):

PR #1289, opened 2021-08-13
rename image to image reference

spencer-codaio:
per https://staging.coda.io/d/Packs-IA-go-packs_d36WK4zgrYx/My-Tasks_suDJO#View-of-My-Bugs_tu8_z/r131&modal=true renaming just the type name now and will change value later
@coda/packs


schema.ts
```diff
@@ -31,13 +31,14 @@ export enum ValueHintType {
   Person = 'person',
   Percent = 'percent',
   Currency = 'currency',
-  Image = 'image',
+  // TODO(spencer): change value to reflect new enum name and update ImageAttachment to Image
```
jonathan-codaio, line 34: I think it's actually unlikely we'll change the value, we'd have to go rebuild all the packs that use it and bump all their versions, and make the `coda` code handle both values during the transition, so may not be worth the pain.

jonathan-codaio (approved):
You'll want to make the corresponding change on the `packs` side too and merge them close together, so that somebody making a different SDK change after this isn't blocked on needing to update the repo.

If you haven't done this dance before, you can update package.json in `packs` or `coda` with the commit hash of your branch commit as long as it's pushed to the remote, and make the changes in other repo based on your yet-to-be-merged SDK. Then you just change the commit hash to the post-merge hash after you eventually merge the SDK change.

schema.ts
spencer-codaio: ah okay i can remove this then

PR #1290, opened 2021-08-13
inject buffer for the thunk bundle

huayang-codaio:
Per feedback from https://github.com/coda/packs-sdk/pull/1274, we don't want to browserify the entire thunk bundle. For now, the only node module used by thunk is the Buffer object. 

This PR injects Buffer with a shim and avoided browserify'ing the whole bundle. 

thunk_bundle.js size increase from 106K to 170K.


huayang-codaio:
PTAL @chris-codaio @jonathan-codaio

jonathan-codaio (approved):
This seems reasonable to me.

dist/runtime/common/marshaling/marshal_buffer.js
```diff
@@ -5,7 +5,7 @@ const constants_1 = require("./constants");
 const constants_2 = require("./constants");
 function marshalBuffer(val) {
     // Buffer is not provided by IVM. If the bundle is not browserified, global.Buffer will be undefined.
-    if (global.Buffer && val instanceof Buffer) {
+    if (val instanceof Buffer) {
```
chris-codaio, line 8: Is the comment above here still valid?

Makefile
```diff
@@ -66,6 +66,7 @@ compile:
 	${ROOTDIR}/node_modules/.bin/esbuild ${ROOTDIR}/runtime/thunk/thunk.ts \
 		--bundle \
 		--outfile=${ROOTDIR}/bundles/thunk_bundle.js \
+		--inject:${ROOTDIR}/testing/injections/buffer_shim.js \
```
chris-codaio, line 69: Why are you injecting code from a `testing` directory into the bundle?

PR #1295, opened 2021-08-16
shim crypto in pack bundles

huayang-codaio:
crypto is another library that's available in both nodejs and browser so that it's not by default browserified. 

Shimming crypto is however a bit tricky. The implementation of crypto.getRandomValues is more secured than using Math.random(). (https://stackoverflow.com/questions/5651789/is-math-random-cryptographically-secure) 

That leaves us two options:
- implement a fake getRandomValues method with a reasonable algorithm.
- expose crypto.getRandomValues (actually the node version) through IVM interface. 

I chose to use a fake getRandomValues method with `mersenne-twister` to support our current need. So that we don't increase the exposure of native libraries to the VM context. 

The other issue here is that esbuild won't put injections into IVM globals. In this particular case, crypto is usually used as `global.crypto` in modules. I had to inject it into global manually but that definitely causes a global leak once the shim is applied. I think this is okay since the shim is only used to build the pack bundle. As far as the existing `packs/` bundles, `getRandomValues` is the only used crypto method.


jonathan-codaio (commented):
Seems reasonable.

Since this is unlikely to be the last time we need to add a shim, just wanted to clarify a general principle. Shimming like this is essentially always safe, because we're adding pure-JS polyfills that are subject to all the normal constraints of the sandbox, right? It's only if we ever decided to address something by injecting an additional value into the IVM context that we'd have to carefully scrutinize what's newly exposed to the sandbox?

huayang-codaio:
@jonathan-codaio That's my understanding as well. The shim code shares the access scope of the pack code and the thunk code: we don't apply any shim to bootstrap code that's running in the native environment (not IVM). 

@chris-codaio please correct us if anything is missing.

jonathan-codaio (approved):

.mocharc.json
```diff
@@ -8,7 +8,8 @@
   "recursive": true,
   "reporter": "dot",
   "slow": 75,
-  "timeout": 10000,
+  "timeout": 20000,
```
huayang-codaio, line 11: fyi @coda/packs I had to increase the mocha timeout since some compiler tests run slower on CI.

PR #1301, opened 2021-08-17
Make category optional for legacy packs

johnli-codaio:
PTAL @huayang-codaio @jonathan-codaio @coda/packs 

CC @varney-codaio

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ generated-documentation:
 
 .PHONY: typedoc
 typedoc:
-	${ROOTDIR}/node_modules/.bin/typedoc index.ts --out ${ROOTDIR}/docs --excludeExternals --excludePrivate --excludeProtected --gitRevision --gitRemote main
+	${ROOTDIR}/node_modules/.bin/typedoc index.ts --out ${ROOTDIR}/docs --excludeExternals --excludePrivate --excludeProtected --gitRevision main
 
 .PHONY: docs
 docs: generated-documentation typedoc
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ generated-documentation:
 
 .PHONY: typedoc
 typedoc:
-	${ROOTDIR}/node_modules/.bin/typedoc index.ts --out ${ROOTDIR}/docs --excludeExternals --excludePrivate --excludeProtected --gitRevision main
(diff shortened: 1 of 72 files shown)


johnli-codaio:
https://github.com/coda/packs-sdk/pull/1301/files#diff-8033e9b81c9757b4f6b0dffdbd5f34f7fbc5a59761f876c863768a9da20ea1b7L329

wayyyy down here is the 1 line diff

docs/enums/DefaultConnectionType.html
```diff
@@ -86,7 +86,7 @@ <h3>Proxy<wbr>Actions<wbr>Only</h3>
 					<div class="tsd-signature tsd-kind-icon">Proxy<wbr>Actions<wbr>Only<span class="tsd-signature-symbol">:</span> <span class="tsd-signature-symbol"> = 3</span></div>
 					<aside class="tsd-sources">
 						<ul>
-							<li>Defined in <a href="https://github.com/coda/packs-sdk/blob/main/types.ts#L46">types.ts:46</a></li>
+							<li>Defined in <a href="https://github.com/kr-project/packs-sdk/blob/main/types.ts#L46">types.ts:46</a></li>
```
jonathan-codaio, line 89: Hmm, this diffs are undesirable. I think you may need to update your remote to point to the new `coda` org in the url?

johnli-codaio: good find.

jonathan-codaio: Alternatively you could maybe add the `--gitRemote` flag to the `typedoc` target in the Makefile so it's not dependent on any developer's particular git configuration.

huayang-codaio: good catch. this is so easy to be ignored. let me add a linter rule I think ..

Makefile
```diff
@@ -93,7 +93,7 @@ generated-documentation:
 
 .PHONY: typedoc
 typedoc:
-	${ROOTDIR}/node_modules/.bin/typedoc index.ts --out ${ROOTDIR}/docs --excludeExternals --excludePrivate --excludeProtected --gitRevision main
+	${ROOTDIR}/node_modules/.bin/typedoc index.ts --out ${ROOTDIR}/docs --excludeExternals --excludePrivate --excludeProtected --gitRevision --gitRemote main
```
jonathan-codaio, line 96: You still want `--gitRevision main` but you want to add a `--gitRemote` value that points at the url of the repo.

jonathan-codaio (approved):

PR #1306, opened 2021-08-18
Don't error out if falsy values are returned in the result object schema

alan-codaio:
Part of https://staging.coda.io/d/Packs-IA-go-packs_d36WK4zgrYx/Bugs-Feature-requests_suyN0#Open-Bugs_tu6FU/r61&modal=true, where if the id field returned by a formula was falsy (i.e. 0, ''), then the result validation would fail.

Now we only error out if the ID field is specifically empty (null or undefined).

### DIFF
--- a/testing/validation.ts
+++ b/testing/validation.ts
@@ -338,7 +338,7 @@ function validateObject<ResultT extends Record<string, unknown>>(
     }
   }
 
-  if (schema.id && schema.id in result && !result[schema.id]) {
+  if (schema.id && schema.id in result && (result[schema.id] === null || result[schema.id] === undefined)) {
     errors.push({
       message: `Schema declares "${schema.id}" as an id property but an empty value was found in result.`,
     });
--- a/testing/validation.ts
+++ b/testing/validation.ts
@@ -338,7 +338,7 @@ function validateObject<ResultT extends Record<string, unknown>>(
     }
   }
 
-  if (schema.id && schema.id in result && (result[schema.id] === null || result[schema.id] === undefined)) {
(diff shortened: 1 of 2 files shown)


alan-codaio:
Oh hmmm... is this WAI?

testing/validation.ts
```diff
@@ -338,7 +338,7 @@ function validateObject<ResultT extends Record<string, unknown>>(
     }
   }
 
-  if (schema.id && schema.id in result && !result[schema.id]) {
+  if (schema.id && schema.id in result && (result[schema.id] === null || result[schema.id] === undefined)) {
```
jonathan-codaio, line 341: Nit, prefer objectUtils.isDefined()

jonathan-codaio (approved):

jonathan-codaio:
> Oh hmmm... is this WAI?

Ah, hmm. Well, `0` is probably ok but `''` seems invalid for an id?

alan-codaio:
> > Oh hmmm... is this WAI?
> 
> Ah, hmm. Well, `0` is probably ok but `''` seems invalid for an id?

What are the downstream effects of providing such a value in the id field? Other than maybe overzealous deduping that may occur due to the multiple objects having the same id perhaps?

jonathan-codaio:
> > > Oh hmmm... is this WAI?
> > 
> > 
> > Ah, hmm. Well, `0` is probably ok but `''` seems invalid for an id?
> 
> What are the downstream effects of providing such a value in the id field? Other than maybe overzealous deduping that may occur due to the multiple objects having the same id perhaps?

It might work if you literally had only one value with `''` as the id, but it's much more likely to indicate an issue.

jonathan-codaio (approved):

PR #1311, opened 2021-08-19
Avoid showing `[object Object]` in status error

huayang-codaio:
Alternative is to show JSON.stringify(body). 

however it may not be well structured for display purpose. For example Coda API 401 looks like 

image

So let's simply show this is a fetcher status error with code.


spencer-codaio (commented):
hmm my preference is towards being more verbose / show the full body to have more info available. Or is your thinking we need some more work on handling long error messages in the debug log to show this?

huayang-codaio:
That makes sense. Codans can see the errors from system logs. but indeed external developers don't have any info.

spencer-codaio (approved):

api.ts
```diff
@@ -77,7 +77,7 @@ export class StatusCodeError extends Error {
   options: FetchRequest;
   response: StatusCodeErrorResponse;
   constructor(statusCode: number, body: any, options: FetchRequest, response: StatusCodeErrorResponse) {
-    super(`${statusCode} - ${body}`);
+    super(`Fetch failed with status code: ${statusCode} - ${JSON.stringify(body, null, 4)}`);
```
jonathan-codaio, line 80: I think `util.inspect()` is preferable. Expands a full object into a string, without transforming it.

huayang-codaio: util is a node library, while this is used in thunk, non-browserified. I don't want to inject `util` just for this.

jonathan-codaio: Ah ok, makes sense, too bad.

PR #1312, opened 2021-08-19
Augment the setTimeout() shim to block synchronously before yielding with a Promise.

jonathan-codaio:
Context for those not already in the discussion, our own Coda Internal pack makes a Coda API mutation request, and then needs to poll the getMutationStatus endpoint until the mutation is applied. It's using setTimeout() to do this today, to sleep in between hitting that endpoint. Future packs that do Coda API mutations may need this too, as well as user packs with similar constraints.

Alex found the suggestion of using `Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, timeout);` as a way to synchronously block in `node`. By adding this to our shim, we seem to get the best of both worlds, where the timeout is still respected, however because it's not actually scheduled at the system level at the system level, when the isolate times out and gets torn down, this blocking call should get torn down immediately too.

Because other tasks can't happen in between as one would generally expect with setTimeout(), we'll still leave this shim disabled by default, so developers will have to see and acknowledge our caveat first.

PTAL @chris-codaio @alexd-codaio @huayang-codaio @coda/packs

### DIFF
--- a/cli/build.ts
+++ b/cli/build.ts
@@ -26,9 +26,13 @@ export async function handleBuild({outputDir, manifestFile, minify, timerStrateg
   }
 }
 
-export async function build(manifestFile: string): Promise<string> {
+export async function build(
+  manifestFile: string,
+  {timerStrategy}: {timerStrategy?: TimerShimStrategy} = {},
+): Promise<string> {
   const {bundlePath} = await compilePackBundle({
     manifestPath: manifestFile,
+    timerStrategy,
   });
 
   return bundlePath;
--- a/dist/cli/build.d.ts
+++ b/dist/cli/build.d.ts
@@ -7,5 +7,7 @@ interface BuildArgs {
     timerStrategy: TimerShimStrategy;
 }
 export declare function handleBuild({ outputDir, manifestFile, minify, timerStrategy }: Arguments<BuildArgs>): Promise<void>;
-export declare function build(manifestFile: string): Promise<string>;
+export declare function build(manifestFile: string, { timerStrategy }?: {
+    timerStrategy?: TimerShimStrategy;
+}): Promise<string>;
 export {};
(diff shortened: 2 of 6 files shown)


huayang-codaio:
have you tested if this works in the VM? 

maybe a unittest in execution_test.ts

```
await executeFormulaOrSyncWithVM({formulaName: 'Timer', params: [1], bundlePath});
```

huayang-codaio:
oh I just tested this works. 

```
➜  packs-sdk git:(hg-status-error-message) ✗ ./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake.ts Timer 1 --vm --timerStrategy=fake
1
```

nice

huayang-codaio (approved):

jonathan-codaio:
> have you tested if this works in the VM?
> 
> maybe a unittest in execution_test.ts
> 
> ```
> await executeFormulaOrSyncWithVM({formulaName: 'Timer', params: [1], bundlePath});
> ```
> 
> Fake pack has a Timer formula uses setTimeout

Oh yep, good call on the test. How can I get the timers_shim in a unittest?

huayang-codaio:
in execution_test.ts, right now it's doing `bundlePath = await buildBundle(`${__dirname}/packs/fake`);`

just replace that with `compilePackBundle` with timerStrategy = fake.

jonathan-codaio:
> in execution_test.ts, right now it's doing `bundlePath = await buildBundle(`${__dirname}/packs/fake`);`
> 
> just replace that with `compilePackBundle` with timerStrategy = fake.

Oh yep, just figured it out, thanks.

jonathan-codaio:
Ok I tested this on all our current packs and it seems to work great. Solves the issue with the Coda Internal pack needing to sleep before polling mutation. Solves a similar issue with the Mode pack that I think was broken before this because it wasn't building with the shim.

The Barcode pack still does not work, this change does not change that issue (a zlib callback seems to be undefined during the timeout).

huayang-codaio:
Fantastic!

PR #1313, opened 2021-08-20
Better simulate structure of request-promise StatusCodeError.

jonathan-codaio:
Looks like there is another property `error` of request-promise's StatusCodeError that many of our existing packs rely on to inspect specific error messages. They probably shouldn't use this field, it's redundant with `body`, but we can clean that up later.

While I'm here, I'm making the message exactly reproduce the message from request-promise, seems safer.

https://github.com/request/promise-core/blob/master/lib/errors.js#L22

PTAL @patrick-codaio @huayang-codaio @coda/packs


huayang-codaio:
Thanks for fixing.

jonathan-codaio:
> Thanks for fixing.

Stay offline, enjoy your OOO time 😛

huayang-codaio:
Just taking a break from watching my wife shopping. Haha.

patrick-codaio (approved):

PR #1323, opened 2021-08-24
support bad pack ids whitelist for identity name validation

huayang-codaio:
instead of https://github.com/coda/packs/pull/2681, let's just whitelist all identity names from the SFDC pack to avoid breaking existing enterprise users.

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -498,7 +498,9 @@ function isValidObjectId(component: string): boolean {
   return Base64ObjectRegex.test(component);
 }
 
-const BAD_PACK_IDS = [1090];
+const BAD_PACK_IDS = [
+  1090, // salesforce pack has a large number of dynamic identity ids that include empty spaces.
+];
 
 // These sync tables already violate the object id constraints and should be cleaned up via upgrade.
 const BAD_SYNC_TABLE_NAMES = [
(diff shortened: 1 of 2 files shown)


chris-codaio (commented):
As an alternative, you could request permission to and fix all of the docs that would be impacted by this?  Might be easier to do that now than later when there's much more usage.

huayang-codaio:
@chris-codaio hmm, I maybe to do that but will need help on a few things:
- need SRE help to figure out all the affected docs. 
- need advice to how change the sync table key in a doc. @jonathan-codaio how do we do this? I suppose this is different from recreating a table but needs to manipulate the table key live right?

testing/upload_validation.ts
```diff
@@ -498,6 +498,10 @@ function isValidObjectId(component: string): boolean {
   return Base64ObjectRegex.test(component);
 }
 
+const BAD_PACK_IDS = [
```
jonathan-codaio, line 501: Nit, can we cal this PackIdsWithBadSyncTableNames, for clarity?

yarn.lock
```diff
@@ -384,14 +384,6 @@
     "@typescript-eslint/typescript-estree" "4.29.3"
     debug "^4.3.1"
 
-"@typescript-eslint/scope-manager@4.29.2":
```
jonathan-codaio, line 387: Weird that there's diffs here given no change to package.json. Mind confirming and checking this in separately if these are due to a bad previous commit of yarn.lock?

jonathan-codaio (commented):
Cool, given the discussion here this sounds like the right thing to do for now. We should fix all of the packs together as a tech debt task in the future.

https://codaproject.slack.com/archives/C01835XEBSS/p1629845658491100

yarn.lock
huayang-codaio: yeah, this is from a previous bad commit. without updating yarn.lock build is going to complain about unclean repo.

PR #1332, opened 2021-08-27
fix date param type map

huayang-codaio:
https://staging.coda.io/d/Packs-IA-go-packs_d36WK4zgrYx/Bugs-Feature-requests_suyN0#Open-Bugs_tu6FU/r115

param type of date should be mapped to string when it's passed to execute method. 

`TypeMap` was used by both params and results. but it seems like at the moment we are using `ValueType` for result type. `TypeMap` is only used to type `PackFormulaResult` which isn't precise anyway (it doesn't respect `formula.resultType` right now). So it looks like fixing param type of date doesn't have any further side effects. 

I have no idea of the generated doc files. but they are generated in main.

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -62,13 +62,13 @@ interface TypeMap {
   [Type.string]: string;
   [Type.object]: object;
   [Type.boolean]: boolean;
-  [Type.date]: Date;
+  [Type.date]: string;
   [Type.html]: string;
   [Type.image]: string;
 }
 
 export type PackFormulaValue = $Values<Omit<TypeMap, Type.object>> | PackFormulaValue[];
-export type PackFormulaResult = $Values<TypeMap> | PackFormulaResult[];
+export type PackFormulaResult = $Values<Omit<TypeMap, Type.date>> | PackFormulaResult[];
 
 export type TypeOf<T extends PackFormulaResult> = T extends number
   ? Type.number
--- a/dist/api_types.d.ts
+++ b/dist/api_types.d.ts
@@ -30,12 +30,12 @@ interface TypeMap {
     [Type.string]: string;
     [Type.object]: object;
     [Type.boolean]: boolean;
-    [Type.date]: Date;
+    [Type.date]: string;
     [Type.html]: string;
     [Type.image]: string;
 }
 export declare type PackFormulaValue = $Values<Omit<TypeMap, Type.object>> | PackFormulaValue[];
-export declare type PackFormulaResult = $Values<TypeMap> | PackFormulaResult[];
+export declare type PackFormulaResult = $Values<Omit<TypeMap, Type.date>> | PackFormulaResult[];
 export declare type TypeOf<T extends PackFormulaResult> = T extends number ? Type.number : T extends string ? Type.string : T extends boolean ? Type.boolean : T extends Date ? Type.date : T extends object ? Type.object : never;
 export declare enum ParameterType {
     String = "string",
--- a/dist/bundle.d.ts
+++ b/dist/bundle.d.ts
@@ -233,12 +233,12 @@ export interface TypeMap {
 	[Type.string]: string;
 	[Type.object]: object;
 	[Type.boolean]: boolean;
-	[Type.date]: Date;
+	[Type.date]: string;
 	[Type.html]: string;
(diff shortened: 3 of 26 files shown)


api_types.ts
```diff
@@ -62,13 +62,13 @@ interface TypeMap {
...
   [Type.boolean]: boolean;
-  [Type.date]: Date;
+  [Type.date]: string;
   [Type.html]: string;
   [Type.image]: string;
 }
 
 export type PackFormulaValue = $Values<Omit<TypeMap, Type.object>> | PackFormulaValue[];
-export type PackFormulaResult = $Values<TypeMap> | PackFormulaResult[];
+export type PackFormulaResult = $Values<Omit<TypeMap, Type.date>> | PackFormulaResult[];
```
jonathan-codaio, line 71: Hmm, is this saying that if you try to use `makeFormula` to define a formula that returns `Type.date` it won't compile? Seems like up until now we technically supported that case, right?

If so, was hoping we wouldn't be disallowing that, and could somehow reconcile dates as parameters being strings while dates as return values still being Dates.

api_types.ts
```diff
@@ -62,13 +62,13 @@ interface TypeMap {
   [Type.string]: string;
   [Type.object]: object;
   [Type.boolean]: boolean;
-  [Type.date]: Date;
+  [Type.date]: string;
```
jonathan-codaio, line 65: Do you know if this affects when a date is used in an object schema property? I think that's unrelated but I forget sometimes.

jonathan-codaio (commented):
Sorry about the documentation diffs, not sure why my last PR removed them, but it looks like they come back easily enough here: https://github.com/coda/packs-sdk/pull/1333

api_types.ts
jonathan-codaio: Oh you're right, we changed `makeFormula()` to require declaring a primitive type as the result type (ValueType) and if you want to get a rich value like a date you have to use codaType (ValueHintType). So this is probably fine.

Can you just patch this in to `coda` and `packs` and make sure there aren't any TS errors with this change?

api_types.ts
huayang-codaio: good point. let me try that.

huayang-codaio:
Closing this one. Confirmed with Eric that no repro.

jonathan-codaio:
Fascinating. Do we have any theories on where the serialized formula params get converted back into Dates?

huayang-codaio:
I spent some time reading the code but it was not clear to me where the conversion happens.

jonathan-codaio:
Oh here we go: https://github.com/coda/coda/blob/main/modules/protos/protos/helpers/calc_service.ts#L42

Looks like we marshal from the grpc request into JS objects based on the declared types. For v1 execution, those JS objects will be used as-is. For v2, your new marshaling will make sure a Date stays a Date.

huayang-codaio:
ah, thanks for the quick find. I was searching for `Date.parse` lol 

but I suspect it's also a Date for v1 executions. otherwise formulas like google calendar events sync would not be able to execute, see below (`toISOString` isn't a function on strings).

```
    execute: async ([dateRange, calendar, search, showDeleted, showHiddenInvitations, excludeDeclined], context) => {
      const [minDate, maxDate] = dateRange;
      const {continuation} = context.sync;
      const calendarId = await helpers.getCalendarId(calendar, context);

      const params = objectUtils.removeEmpty({
        timeMin: minDate.toISOString(),
        timeMax: maxDate.toISOString(),
        singleEvents: true,
        orderBy: 'startTime',
        pageToken: continuation?.nextPageToken,
        q: search,
        showDeleted,
        showHiddenInvitations,
      });
```

huayang-codaio:
oh, maybe your code is pointing v1 executions using Date. ignore more previous comment in that case.

jonathan-codaio:
Yeah it's Date in both cases. Was just saying that after the proto value is marshaled to a Date, in the v1 case it stays a date because that post-marshaled value is used as-is when executed in the calc server. In the v2 case it stays a date because your new lambda marshaling code ensures that it does even while making the extra hops to lambda and IVM.

huayang-codaio:
cool cool. thanks for explaining!

PR #1334, opened 2021-08-31
Teach our marshaling the difference between null and undefined

patrick-codaio:
Fixes https://go/bug/19250. Specifically, the issue was caused by pack code that assumed that an un-set boolean param would be `undefined` when it came through as `null`. This happened because JSON has no native way to represent `undefined`, and so we hack.

### DIFF
--- a/runtime/common/marshaling/index.ts
+++ b/runtime/common/marshaling/index.ts
@@ -7,7 +7,11 @@ import {unmarshalDate} from './marshal_dates';
 import {unmarshalError} from './marshal_errors';
 import {unmarshalNumber} from './marshal_numbers';
 
+// JSON has no native way to represent `undefined`.
+const HACK_UNDEFINED_JSON_VALUE = '__UNDEFINED__';
+
 const MaxTraverseDepth = 100;
+
 const customMarshalers = [marshalError, marshalBuffer, marshalNumber, marshalDate];
 
 const customUnmarshalers: Array<(val: any) => any> = [unmarshalError, unmarshalBuffer, unmarshalNumber, unmarshalDate];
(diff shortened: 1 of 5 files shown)


test/marshaling_test.ts
```diff
@@ -17,6 +17,7 @@ describe('Marshaling', () => {
     assert.deepEqual([1], transform([1]));
     assert.deepEqual({a: 1}, transform({a: 1}));
     assert.deepEqual(undefined, transform(undefined));
+    assert.deepEqual([undefined], transform([undefined]));
```
huayang-codaio, line 20: maybe also add one for null.

runtime/common/marshaling/index.ts
```diff
@@ -7,7 +7,11 @@ import {unmarshalDate} from './marshal_dates';
 import {unmarshalError} from './marshal_errors';
 import {unmarshalNumber} from './marshal_numbers';
 
+// JSON has no native way to represent `undefined`.
+const HACK_UNDEFINED_JSON_VALUE = '__UNDEFINED__';
```
huayang-codaio, line 11: can we make this more specific to coda like `__CODA_UNDEFINED__`, just in case someone accidentally used the keyword.

huayang-codaio (commented):
thank you for catching & fixing this!

test/marshaling_test.ts
jonathan-codaio: +1

jonathan-codaio (commented):
Many thanks. Other than existing comments, LGTM.

Amazing how bad JSON is as a marshaller for POJOs.

test/marshaling_test.ts
patrick-codaio: Thanks for suggesting more tests - I found out that there was also a problem with {a:undefined}. To solve that, I needed to add another traversal :/

runtime/common/marshaling/index.ts
patrick-codaio: great idea, done.

huayang-codaio (approved):

PR #1343, opened 2021-09-01
Move typedoc config to a file

ekoleda-codaio:
Will be adding more options as we move to MkDocs, and this makes them easier to manage. Also set the git remote via config, so you don't need to manually modify it when working off of a fork.

### DIFF
--- a/docs/classes/PackDefinitionBuilder.html
+++ b/docs/classes/PackDefinitionBuilder.html
@@ -128,7 +128,7 @@
 						<li class="tsd-description">
 							<aside class="tsd-sources">
 								<ul>
-									<li>Defined in <a href="https://github.com/coda/packs-sdk/blob/main/builder.ts#L51">builder.ts:51</a></li>
+									<li>Defined in builder.ts:51</li>
 								</ul>
 							</aside>
 							<h4 class="tsd-parameters-title">Parameters</h4>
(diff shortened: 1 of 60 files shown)


docs/classes/PackDefinitionBuilder.html
```diff
@@ -128,7 +128,7 @@ <h3>constructor</h3>
 						<li class="tsd-description">
 							<aside class="tsd-sources">
 								<ul>
-									<li>Defined in <a href="https://github.com/coda/packs-sdk/blob/main/builder.ts#L51">builder.ts:51</a></li>
```
jonathan-codaio, line 131: Hmm, looks a bit busted, we seem to have lost all of the source code links?

jonathan-codaio (commented):
How does moving the git remote to the .js file help with forks/branches?

I would think we'd just want/need to set it via a flag or environment variable in the `make` command. Something like `BRANCH=foo make typedoc` or `REMOTE=foo.git make typedoc`.

docs/classes/PackDefinitionBuilder.html
ekoleda-codaio: 🤦 I read the diff backwards when I checked it before. I misunderstood the `gitRemote` option, it needs already existing remote, not a URL. Reverted.

jonathan-codaio (approved):

PR #1344, opened 2021-09-01
Basic MkDocs setup

ekoleda-codaio:
For the moment MkDocs is just serving the complete HTML that comes from TypeDoc, but that will be changed to markdown in a later PR. Out GitHub pages config will need to change to pull from gh-pages:/ instead of main:/docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,8 @@ publish-docs:
 		echo "The documentation can only be published from main at head."; \
 		exit 1; \
 	fi
+	# Build the docs and push them to the gh-pages brance.
+	# See: https://www.mkdocs.org/user-guide/deploying-your-docs/#github-pages
 	${PIPENV} run mkdocs gh-deploy
 
 .PHONY: test
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ YARN_CACHE_DIR=~/.yarncache
 ISOLATED_VM_VERSION_COMMAND="require('./node_modules/isolated-vm/package.json').version"
 ISOLATED_VM_VERSION=$(shell node -p -e $(ISOLATED_VM_VERSION_COMMAND))
 
+PIPENV := PYTHONPATH=${ROOTDIR} PIPENV_IGNORE_VIRTUALENVS=1 pipenv
+
 # Aliases
 bs: bootstrap
 
@@ -21,9 +23,14 @@ _bootstrap-node:
 	yarn config set cache-folder ${YARN_CACHE_DIR}
 	yarn install
 
+.PHONY: _bootstrap-python
+_bootstrap-python:
+	PIPENV sync
+
 .PHONY: bootstrap
 bootstrap:
 	$(MAKE) MAKEFLAGS= _bootstrap-node
+	$(MAKE) MAKEFLAGS= _bootstrap-python
 	echo
 	echo '  make bootstrap complete!'
 	echo
@@ -104,8 +111,16 @@ typedoc:
 docs: typedoc generated-documentation
 
 .PHONY: view-docs
-view-docs: typedoc
-	open ${ROOTDIR}/docs/index.html
+view-docs: docs
+	expect -c 'spawn PIPENV run mkdocs serve; expect "Serving on"; exec open "http://localhost:8000"; interact'
+
+.PHONY: publish-docs
+publish-docs:
+	if [ -z ${shell git status -uno | grep "Your branch is up to date with 'origin/main'"} ]; then \
+		echo "The documentation can only be published from main at head."; \
+		exit 1; \
+	fi
+	PIPENV run mkdocs gh-deploy
 
 .PHONY: test
 test:
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ _bootstrap-node:
 
 .PHONY: _bootstrap-python
 _bootstrap-python:
-	PIPENV sync
+	${PIPENV} sync
 
 .PHONY: bootstrap
 bootstrap:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -9,6 +9,8 @@ YARN_CACHE_DIR=~/.yarncache
 ISOLATED_VM_VERSION_COMMAND="require('./node_modules/isolated-vm/package.json').version"
 ISOLATED_VM_VERSION=$(shell node -p -e $(ISOLATED_VM_VERSION_COMMAND))
 
+PIPENV := PYTHONPATH=${ROOTDIR} PIPENV_IGNORE_VIRTUALENVS=1 pipenv
```
jonathan-codaio, line 12: Is this making any assumptions about what system python is running? Or does the system-installed pipenv seem sufficient to produce a hermetic python 3.9 environment?

Context is that we have problems when relying on system python, it can differ a lot from machine to machine, plus on the eng team about half the team uses linux devboxes on AWS and the rest use macs. On the `coda` side we have to set up a standardized pipenv as part of the bootstrap process. I guess if the built-in pipenv is sufficient to bootstrap isn't a consistent python 3.9 environment then this will work.

Makefile
```diff
@@ -21,9 +23,14 @@ _bootstrap-node:
 	yarn config set cache-folder ${YARN_CACHE_DIR}
 	yarn install
 
+.PHONY: _bootstrap-python
+_bootstrap-python:
+	PIPENV sync
```
jonathan-codaio, line 28: Hmm, does this work? Would need to be `${PIPENV}` to evaluate the variable right?

Makefile
```diff
@@ -103,8 +110,16 @@ typedoc:
...
+view-docs: docs
+	expect -c 'spawn PIPENV run mkdocs serve; expect "Serving on"; exec open "http://localhost:8000"; interact'
+
+.PHONY: publish-docs
+publish-docs:
+	if [ -z ${shell git status -uno | grep "Your branch is up to date with 'origin/main'"} ]; then \
+		echo "The documentation can only be published from main at head."; \
+		exit 1; \
+	fi
+	PIPENV run mkdocs gh-deploy
```
jonathan-codaio, line 122: This step is assuming that hosted docs live in gh-pages and can't be configured to respect an alternate directory?

mkdocs.yml
```diff
@@ -0,0 +1,32 @@
...
+  - admonition
+  - attr_list
+  - pymdownx.details
+  - pymdownx.highlight:
+      linenums: true
+  - pymdownx.keys
+  # Snippets inclusion.
+  # See: https://facelessuser.github.io/pymdown-extensions/extensions/snippets/
+  - pymdownx.snippets:
+      base_path: documentation/examples
```
jonathan-codaio, line 16: What exactly does this do with the example snippets?

jonathan-codaio (commented):
Exciting stuff!

mkdocs.yml
ekoleda-codaio: Nothing as per this PR, but in my demo I was using it to pull the IDE snippets into guides, to single-source some of the content. I can strip this file down for the initial commit and then add to it we build out the docs if you'd prefer.

Makefile
jonathan-codaio: I guess we check for a valid python here: https://github.com/coda/coda/blob/main/setup-env.sh#L164 I think that implies that we still install it somewhere else in `make bs` though.

In my experience using python requires some flavor of setup-env.sh script to bootstrap one's environment, we'll see how far we can get without one but it kind of feels inevitable. Totally fine if we don't get all of it in one PR though, we can just do individual meaningful pieces as long as they don't break `make build`.

mkdocs.yml
jonathan-codaio: Yeah let's just tackle that separately.

Makefile
```diff
@@ -103,8 +110,16 @@ typedoc:
...
+view-docs: docs
+	PYTHONPATH=${ROOTDIR} PIPENV_IGNORE_VIRTUALENVS=1 expect -c 'spawn pipenv run mkdocs serve; expect "Serving on"; exec open "http://localhost:8000"; interact'
+
+.PHONY: publish-docs
+publish-docs:
+	if [ -z ${shell git status -uno | grep "Your branch is up to date with 'origin/main'"} ]; then \
+		echo "The documentation can only be published from main at head."; \
+		exit 1; \
+	fi
+	${PIPENV} run mkdocs gh-deploy
```
jonathan-codaio, line 122: What does this actually do?

ekoleda-codaio: The description is here:

https://www.mkdocs.org/user-guide/deploying-your-docs/#github-pages

When I tested it in my fork it worked perfectly, pushing the built docs into the `gh-pages` branch.

Makefile
ekoleda-codaio: Digging around, it seems like `make bs` just fails if the Python version isn't a match with the Pipfile:

https://github.com/coda/coda/blob/main/scripts/dev/check-python-version.sh

We can try to replicate that in this repo, but I agree that we probably add that in later as needed.

mkdocs.yml
ekoleda-codaio: Done

Makefile
jonathan-codaio: This works on Mac but unfortunately this is what I'm getting when I run `make bs` on my linux devbox:

```
 admin@devbox-jonathan  ~/src/packs-sdk   ek-mkdocs  make bs
yarn config v1.22.5
success Set "cache-folder" to "/home/admin/.yarncache".
Done in 0.08s.
yarn install v1.22.5
[1/5] Validating package.json...
[2/5] Resolving packages...
success Already up-to-date.
Done in 0.39s.
Warning: Python 3.9 was not found on your system…
You can specify specific versions of Python with:
  $ pipenv --python path/to/python
make[1]: *** [Makefile:28: _bootstrap-python] Error 1
make: *** [Makefile:33: bootstrap] Error 2
```

So can't proceed with this quite as-is, either need to make all of this a manual step for now to unblock you while we figure this out, or need to deal with our linux env not having python 3.9 available by default.

jonathan-codaio (approved):
Works on linux 🎉 !

ekoleda-codaio:
Yay! For the record, the 3.9 requirement was arbitrarily added by `pipenv`, and I confirmed that `mkdocs` works correctly on 3.7 as well ([according to pypi](https://pypi.org/project/mkdocs/) it requires 3.6+). Unfortunately the Pipfile doesn't allow you to specify a range of python minor versions (like 3.6+) so I've changed it to only require the "3" major version.

PR #1345, opened 2021-09-01
Documentation content checkpoint: finish types.ts.

jonathan-codaio:
Chip chip chippin' away.

Looks like you might have to expand types.ts manually to see the diff.


types.ts
```diff
@@ -187,95 +234,202 @@ interface BaseAuthentication {
...
+   * user to the {@link authorizationUrl}.
+   */
   additionalParams?: {[key: string]: any};
 
-  // Some OAuth providers will return the API domain with the OAuth response.
-  // This is the key in the OAuth response json body that points to the endpoint.
+  /**
+   * In rare cases, OAuth providers will return the specific API endpoint domain for the user as
+   * part of the OAuth token exchange response. If so, this is the property in the OAuth
+   * token exchange response JSON body that points to the endpoint.
```
ekoleda-codaio, line 366: Is this value later accessed via `context.endpoint`? If so, let's mention that and link to that property.

types.ts
```diff
@@ -187,95 +234,202 @@ interface BaseAuthentication {
...
+    /**
+     * A placeholder value for the text input where the user will enter a password.
+     */
     placeholderPassword?: string;
 
-    // Some auth providers pass apiKeys in the username and do not require a password
+    /**
+     * If true, only a username input will be shown to the user.
+     * Some services pass API keys in the username field and do not require a password
+     * (not recommended!).
```
ekoleda-codaio, line 402: I feel like this could confuse the developer reading this, since they are likely stuck with whatever the API provider has chosen and can't use an alternative. I'd suggest removing.

types.ts
```diff
@@ -349,57 +526,129 @@ export type VariousSupportedAuthentication =
   | MultiQueryParamTokenAuthentication
   | WebBasicAuthentication;
 
+/**
+ * @ignore
+ */
 export type VariousSupportedAuthenticationTypes = $Values<Pick<VariousSupportedAuthentication, 'type'>>;
 
+/**
+ * Definition for a custom column type that you apply to any column in any Coda table.
```
ekoleda-codaio, line 535: "you apply" => "users can apply"

types.ts
```diff
@@ -349,57 +526,129 @@ export type VariousSupportedAuthentication =
...
+ * A column format tells Coda to interpret the value in a cell by executing a formula
+ * using that value, typically looking up data related to that value from a third-party API.
+ * For example, the Weather pack has a column format "Current Weather"; when applied to a column,
+ * if you type a city or address into a cell in that column, that location will be used an input
```
ekoleda-codaio, line 539: "will be used an input" => "will be used as the input"

types.ts
```diff
@@ -349,57 +526,129 @@ export type VariousSupportedAuthentication =
...
+ * to a formula that fetches the current weather at that location, and the resulting object with
+ * weather info will be shown in the cell.
+ *
+ * A column format is just a wrapper around a formula defined in the {@link formulas} section
+ * of your pack definition. It tells Coda to execute that particular formula using the value
+ * of the cell as input.
+ *
+ * The formula referenced by a Format must have exactly one required parameter.
```
ekoleda-codaio, line 547: "Format" => "format"

types.ts
```diff
@@ -349,57 +526,129 @@ export type VariousSupportedAuthentication =
...
+ *
+ * You may optionally specify one or more {@link matchers}, which are regular expressions
+ * that can be matched against values that users paste into table cells, to determine if
+ * this Format is applicable to that value. This is a discovery mechanism to help users
+ * realize that there is a pack that may augment their experience of working with such values.
+ *
+ * For example, if you're building a Wikipedia pack, you may write a matcher regular expression
+ * that looks for Wikipedia article URLs, if you have a formula that can fetched structured data
+ * given an article URL. This would help users discover that there is a pack that can fetch
```
ekoleda-codaio, line 556: Does the user need to have the pack installed for the matcher to run? Or are they run for all packs?

types.ts
```diff
@@ -349,57 +526,129 @@ export type VariousSupportedAuthentication =
...
+ */
 export interface Format {
+  /**
+   * The name of this column format. This will show to users in the column type chooser.
+   */
   name: string;
+  /** @deprecated Namespaces are being removed from the product. */
   formulaNamespace: string;
+  /**
+   * The name of the formula to invoke for values in columns using this format.
```
ekoleda-codaio, line 569: This implies that the formula powering the column format must also be available as public formula (vs the private formuals that sync tables use). If so, perhaps add that note somewhere.

types.ts
```diff
@@ -349,57 +526,129 @@ export type VariousSupportedAuthentication =
...
   formulaName: string;
   /** @deprecated No longer needed, will be inferred from the referenced formula. */
   hasNoConnection?: boolean;
+  /**
+   * A brief, optional explanation of how users should use this format, for example, what kinds
+   * of values they should put in columns using this format.
+   */
   instructions?: string;
+  /**
+   * Any regular expressions that match URLs that the formula implementing this format
```
ekoleda-codaio, line 581: "Any" => "A list of"

types.ts
```diff
@@ -349,57 +526,129 @@ export type VariousSupportedAuthentication =
...
 export interface Quota {
   monthlyLimits?: Partial<{[quotaLimitType in QuotaLimitType]: number}>;
   // TODO(alexd): Deprecate
   maximumSyncInterval?: SyncInterval;
   sync?: SyncQuota;
 }
 
+/**
+ * @deprecated Define these in the pack management UI instead.
+ *
```
ekoleda-codaio, line 641: Should these also include `@ignore`? Or is that duplicative?

types.ts
```diff
@@ -416,6 +665,10 @@ export type BasicPackDefinition = Omit<PackVersionDefinition, 'version'>;
  * heart of the implementation of a Pack.
  */
 export interface PackVersionDefinition {
+  /**
+   * The semantic version of the pack. This must be valid semantic version of the form `1`, `1.2`, or `1.2.3`.
+   * When uploading a pack version, the semantic version must be greater than any previously uploaded version.
+   */
```
ekoleda-codaio, line 671: Should we mention somewhere, perhaps in the interface-level description, the need / behavior of these versions in the Web IDE.

types.ts
```diff
@@ -426,12 +679,37 @@ export interface PackVersionDefinition {
...
   networkDomains?: string[];
 
   // User-facing components
+
+  /**
+   * @deprecated
+   */
   formulaNamespace?: string; // TODO: @alan-fang remove
+  /**
+   * Definitions of this pack's formulas. See {@link Formula}.
```
ekoleda-codaio, line 700: Perhaps note that this is where buttons are also defined, since they are just formulas with isAction: true?

ekoleda-codaio (changes requested):
Love to see this shaping up! Did a detailed review again.

types.ts
jonathan-codaio: I kept them non-ignored but deprecated with the thinking that these are indeed concepts that are an active part of building a pack, they've just moved from the code to the pack management UI. They'll still show up in TypeScript even if we hide them from TypeDoc, so I wanted the explanation to be readily viewable to a user who stumbles across these fields.

types.ts
jonathan-codaio: Up to you. The web IDE uses BasicPackVersionDefinition which doesn't have a version at all, so my thinking was that a web IDE user wouldn't come across this interface on their own.

types.ts
jonathan-codaio: Done.

types.ts
```diff
@@ -133,17 +136,61 @@ export interface NoAuthentication {
   type: AuthenticationType.None;
 }
 
+/**
+ * Configuration for a step that will run after the user sets up a new account
+ * that fetches a set of endpoint domains that the user has access to and prompts
+ * the user to select the one that should apply to this account.
+ *
+ * The selected endpoint domain is bound to this account and used as the root domain
+ * of HTTP requests made by the fetcher.
```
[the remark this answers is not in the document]
ekoleda-codaio: Should we include a sentence about it here as well?

types.ts
```diff
@@ -133,17 +136,61 @@ export interface NoAuthentication {
...
+ * packs. A Jira account may have access to multiple Jira instances; after authorizing
+ * the user account, this step makes an API call to fetch all of the Jira instances
+ * that the user has access to, which are rendered as options for the user,
+ * and the endpoint domain of the select option (of the form <instance>.atlassian.net)
+ * is stored along with this account.
+ */
 export interface SetEndpoint {
   type: PostSetupType.SetEndpoint;
+  /**
+   * An arbitrary name for this step, to distinguish from other steps of the same type
```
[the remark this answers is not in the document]
ekoleda-codaio: To me "arbitrary" makes it sound unimportant, and if it's unimportant then why even let me set it at all. That said, not a big deal.

types.ts
ekoleda-codaio: In that case I'd remove or downplay the language about this being a discovery mechanism until that new regex matcher work is done.

types.ts
ekoleda-codaio: Ah, I didn't realize there was a separate type for the Web IDE.

ekoleda-codaio (approved):
Some small suggestions, but otherwise approved.

types.ts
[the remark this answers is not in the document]
jonathan-codaio: Makes sense, added.

types.ts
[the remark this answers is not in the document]
jonathan-codaio: Yeah, I can see that. Though I'm ok communicating that the value is unimportant, it truly doesn't matter what the value is, as long as it exists and is unique. FWIW I find arbitrary to be a pretty useful term in programming, but could just be me.

types.ts
jonathan-codaio: Makes sense, downplayed.

PR #1346, opened 2021-09-02
Switch typedoc to markdown output

ekoleda-codaio:


ekoleda-codaio:
Abandoning this PR, since I can't get the diffbase I want.

PR #1347, opened 2021-09-03
Switch typedoc output to markdown

ekoleda-codaio:
Markdown output allows for the MkDocs chrome and navigation to be used on the generated reference docs. Moving the path of the reference docs to `/docs/reference/sdk` to avoid future conflicts with hand-written documentation. The interesting files are at the top and the bottom.


documentation/generated/examples.json
```diff
@@ -4,7 +4,7 @@
     "triggerTokens": [
       "addColumnFormat"
     ],
-    "exampleFooterLink": "https://coda.github.io/packs-sdk/classes/PackDefinitionBuilder.html#addColumnFormat",
+    "exampleFooterLink": "https://coda.github.io/packs-sdk/reference/sdk/classes/PackDefinitionBuilder#addColumnFormat",
```
jonathan-codaio, line 7: Thanks!

jonathan-codaio (approved):
If submitting this is going to blow away the old reference urls, we may want to coordinate merging this with a push to the alpha adhoc, so that alpha users don't have their SDK reference links break on them for very long.

PR #1349, opened 2021-09-03
Remove formulaNamespace from Pack.formulas

alan-codaio:
Removes the convention of Pack formulas being under separate namespaces, so that `pack.formulas` is a flat array of formulas. 

Re: consistency breaking, I have a PR [here](https://github.com/coda/coda/pull/70451) that bumps the SDK without updating any of the logic, and that seems to pass consistency. So we can separately investigate what part of the branching logic is actually throwing an error, and clean up the SDK now.


jonathan-codaio (approved):
🪓 

Very excited. WCGW

PR #1358, opened 2021-09-07
Slightly better error message when no formula found.

jonathan-codaio:
User was trying to run a sync from the CLI and it wasn't finding the sync because of a typo, and it made them think they needed to add dummy formulas to the pack to make it work.

We should also document whether you need to specify the name of the sync table or the name of the sync formula (I'm not actually sure off the top of my head).

PTAL @huayang-codaio @coda/packs

### DIFF
--- a/bundles/thunk_bundle.js
+++ b/bundles/thunk_bundle.js
@@ -4477,11 +4477,11 @@ function isDynamicSyncTable(syncTable) {
 // runtime/common/helpers.ts
 init_buffer_shim();
 function findFormula(packDef, formulaNameWithNamespace) {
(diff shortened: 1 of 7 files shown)


huayang-codaio (approved):
I am not against the changes in this PR but I feel like the user's issue is related to https://github.com/coda/packs-sdk/blob/main/testing/execution.ts#L161

We probably throw an error there when the CLI name doesn't map to either a formula or a sync formula. Sorry that was my bad didn't handling the failure case.

jonathan-codaio:
Makes sense, how about with the latest changes?

huayang-codaio:
That looks perfect. Thank you!

bala-codaio (commented):
If the user is already authenticated and authorized then disclosing the non existence of formula or sync table is good. 

However, if the user is not authorized yet then we will be leaking existence of a formula or table and will be an information disclosure issue.

PR #1363, opened 2021-09-09
Don't attempt to build the gh-pages branch

ekoleda-codaio:
Confirmed the config is syntactically valid using `circleci config validate`.


jonathan-codaio (approved):
Nice

PR #1365, opened 2021-09-10
Use yarn link for better imports in sample code 

ekoleda-codaio:
The existing sample code (in `/documentation/(examples|snippets)` imports from `index.ts` and has to comment out the real import to `@codahq/packs-sdk`:

```ts
import * as coda from '../../../index';

// BEGIN
// **This import statement gives you access to all parts of the Coda Packs SDK. Uncomment to start!**
// import * as coda from '@codahq/packs-sdk';
```

The `// BEGIN` comment and related logic helped hide the local import in the Web IDE, but can't be leveraged by MkDocs when including sample code in the documentation ([example](https://coda-packs-sdk.readthedocs-hosted.com/en/ek-getstarted/samples/hello_world/)). This setup also relies on the developer noticing the import is commented out and remembering to uncomment it.

This PR uses `yarn link` to provide `@codahq/packs-sdk` via a symlink to itself, allowing us to write sample code that can be used as-is by developers and which compiles against local code. I also added a Make rule that compiles the sample code to make sure it is working correctly.


jonathan-codaio (approved):
Hot damn. What a great solution.

I wonder if we should get rid of the packs-examples repo now that we're able to do this, and those full working packs can just be another sub-directory? (We also use that repo as the basis for `coda init` but I'm sure we could find an alternate approach.)

Should we also get rid of the `// BEGIN` markers in a followup or filter them out somehow? Seems like those are also going to make their way into documentation website which isn't ideal.

ekoleda-codaio:
It might make sense to consolidate the repos, but I'm much more likely to include these small example snippets in the docs then the full, multi-file samples. And like you said, that repo is currently a little bound up with the CLI, so we'd have to do some work to unwind it. I wouldn't prioritize it I guess.

I thought about removing all of the `// BEGIN` markers, but we'd certainly want them in the snippets (don't want to inject another import) and there may be some examples where it makes sense too. The comment isn't ideal, but isn't that bad either.

PR #1377, opened 2021-09-16
Add a CLI get started guide

ekoleda-codaio:
Staged here: https://coda-packs-sdk.readthedocs-hosted.com/en/ek-cli/get-started/cli/


documentation/examples/hello_cli/manifest.ts
```diff
@@ -0,0 +1,13 @@
+import type {PackVersionDefinition} from '@codahq/packs-sdk';
+import {formats} from './formulas';
+import {formulas} from './formulas';
+import {syncTables} from './formulas';
+
+export const manifest: PackVersionDefinition = {
```
jonathan-codaio, line 6: Just curious on your opinion thus far of the newPack() builder pattern, and whether we should use that for the CLI as well? No need to change anything now, just raising the question.

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+    ```sh
+    mkdir my-pack
+    cd my-pack
+    ```
+
+1. Install the Packs SDK.
+
+    ```sh
+    npm install @codahq/packs-sdk
```
jonathan-codaio, line 37: Do we want to recommend they set up a node project with `npm init` first and install dependencies locally?

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+## Set up the Pack structure
+
+1. Create a directory for your new Pack.
+
+    ```sh
+    mkdir my-pack
+    cd my-pack
+    ```
+
+1. Install the Packs SDK.
```
jonathan-codaio, line 34: Weird that these are all numbered `1.` but they seem to render as sequentially in the rendered output. Is that Markdown fanciness so you don't have renumber lists as you iterate?

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+Your directory should now contain the following files:
+
+* `formulas.ts` - A place to define the Coda formulas, including those that power sync tables and column formats.
+* `manifest.ts` - Defines the metadata and contents of your Pack.
+* `node_modules` - The dependencies downloaded from NPM (standard for Node.js projects).
+* `package-lock.json` - The versions of the dependencies downloaded from NPM (standard for Node.js projects).
+* `package.json` - The project's dependencies from NPM (standard for Node.js projects).
+* `schemas.ts` - A place to define the schemas (structured data types) used by your Pack.
+* `types.ts` - A place to define TypeScript types for the data used by your Pack.
```
jonathan-codaio, line 59: As we go, feel free to suggest a better file layout if you like. This is what has worked best in the `packs` repo but that doesn't mean we have to recommend it externally.

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+## Add functionality to you Pack
```
jonathan-codaio, line 61: Nit, s/you/your

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+1. Edit `manifest.ts` to set the formula namespace to `HelloWorld`:
+
+    === "manifest.ts"
+        ```ts hl_lines="8"
+        --8<-- "hello_cli/manifest.ts"
+        ```
+
+    !!! note
+        The value in `formulaNamespace` isn't normally visible to the user, as Coda prefers to show the Pack's icon instead. However when a formula is copied from the formula editor and pasted as plain text the namespace is used instead of the icon.
```
jonathan-codaio, line 80: The plan is to fully remove formulaNamespace from the SDK, and we're actually finally quite close to doing that. And external users will never actually see a namespace, since we're only going to allow installing user-created packs in docs that have the icon functionality. When pasting a formula as plain text in a doc using the icon, the text is actually the pack name and not the namespace.

So I guess we should find the least awkward way to say that this field is deprecated (in a brand new product...) and its value is irrelevant. The pack builder does a decent job obfuscating it because it just provides a default namespace without the maker needing to know about it.

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+If everything works correctly this should output `Hello world!`.
+
+## Upload your Pack
+
+So far everything you've built only exists on your local machine, and Coda has no knowledge of it. To see it working in a real doc you'll need to upload your Pack to Coda's servers.
+
+### Register an API token
+
+The `coda` CLI uses the Coda API under the hood to upload your code, and likewise needs an API token to operate. Registering an API token is a one-time setup step that must be done for each new Pack.
```
jonathan-codaio, line 98: It actually only has to be done once per repo/environment/directory. As long there's a Coda API key in the current directory or a parent directory, we'll find and use it.

So to work within the `packs` repo, for example, which has 80+ packs, you still only have to do `coda init` once.

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+1. Register an API key for Pack uploads:
+
+    ```shell
+    npx coda register --codaApiEndpoint=https://{{coda.domain}}
```
jonathan-codaio, line 103: The `--codaApiEndpoint` flag can be omitted, that's only needed if you're working with a non-prod environment. Goal for post-alpha is nobody except Coda engineers needs to know about the flag.

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+    ```
+
+1. When prompted to create a new API token, type `y` and hit enter.
+
+    This will open your browser to the API token creation dialog.
+
+1. In the **Name** field enter "Hello World Pack" and then click **Generate API token**.
+
+1. Scroll down the page to the **Coda API tokens** section, find the token you just created, and click the **Copy token** link.
```
jonathan-codaio, line 112: Ugh, we should fix this long-standing bug so you don't have to scroll. Mind filing that at go/packs?

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+1. Switch back to your terminal, paste your token into the prompt, and hit enter.
+
+This will create a new file `.coda.json` in your working directory that contains the API token.
+
+??? warning "Don't check in .coda.json"
+    If you use a version control system you will likely want to make sure this file isn't check in, as the token within provides access to your account. For example, if using Git, add `.coda.json` to your [`.gitignore` file](https://git-scm.com/docs/gitignore).
```
jonathan-codaio, line 119: Does `coda init` create a `.gitignore`? If it doesn't, we should file a task to make sure it does. If it does, we should maybe call that out.

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+### Create the Pack
+
+Now that you have the access configured you can create the new Pack on Coda's servers. Like registering an API token, this is also a one-time setup step that needs to be done for each Pack.
```
jonathan-codaio, line 123: Would remove `Like registering an API token` per the above comment.

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+```shell
+npx coda create manifest.ts --name "Hello World" --description "My first Pack." --codaApiEndpoint=https://{{coda.domain}}
```
jonathan-codaio, line 126: May want to note that `--name` and `--description` are optional, and you can always add/edit them in the UI later.

As above, can remove `--codaApiEndpoint`

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+    
+    ```ts
+    // Some boilerplate when failed to load source code
+    ```
+
+    This is expected, since the CLI will only upload the built Pack and not the source code.
+
+## Install and use the Pack
+
+Your new Pack is now available to use in all your docs, and you can install it just like a Coda-made Pack.
```
jonathan-codaio, line 150: Nit, I wonder if "Coda-made Pack" will cease to mean as much going forward when most packs are user-created.

docs/get-started/cli.md
```diff
@@ -4,4 +4,219 @@ title: On your local machine
...
+
+### Create the Pack
+
+Now that you have the access configured you can create the new Pack on Coda's servers. Like registering an API token, this is also a one-time setup step that needs to be done for each Pack.
+
+```shell
+npx coda create manifest.ts --name "Hello World" --description "My first Pack." --codaApiEndpoint=https://{{coda.domain}}
+```
+
+This will create a new, empty Pack on Coda's servers and output its URL in the Pack Studio. It stores the Pack's ID in the new file `.coda-pack.json`.
```
jonathan-codaio, line 129: May want to call it that this command will also output the URL of the pack management UI page.

jonathan-codaio (commented):
Awesome, rendered docs look so good.

documentation/examples/hello_cli/manifest.ts
ekoleda-codaio: I've been concerned about how the Pack Editor and CLI use two different "dialects", and the inability to easily copy and paste between them. I'll start a discussion in Slack to explore this more.

docs/get-started/cli.md
ekoleda-codaio: Done: https://go/packsbug/196

docs/get-started/cli.md
ekoleda-codaio: No, it doesn't seem to do that today. Filed here: https://go/packsbug/197

docs/get-started/cli.md
ekoleda-codaio: Done

docs/get-started/cli.md
ekoleda-codaio: Done with name and description. See above regarding endpoint.

docs/get-started/cli.md
ekoleda-codaio: I call out the URL to the Pack Studio (name TBD apparently). Is that what you are talking about?

docs/get-started/cli.md
ekoleda-codaio: Good point. Fixed here and in `web.md`

docs/get-started/cli.md
jonathan-codaio: Nice, that works.

docs/get-started/cli.md
jonathan-codaio: Definitely makes sense as a general approach. My main thing is I'd like to frame it in a way so that the maker doesn't get worried about having to understand this concept as being part of pack development, when it's basically irrelevant to their work and will disappear soon. In this case I think it's appropriate to talk about the future.

Maybe can say something like, "In older Coda docs, the namespace is used to write a formula using the formula builder, e.g. MyNamespace::MyFormula. In most docs, and soon all docs, the namespace will no longer be needed, and the formula builder will display the Pack's icon instead. Once all Coda docs have this behavior, you will no longer need to specify a namespace in the Pack definition at all."

docs/get-started/cli.md
jonathan-codaio: Ok.

docs/get-started/cli.md
jonathan-codaio: Oh, got it.

docs/get-started/cli.md
ekoleda-codaio: As per our offline discussion, I've removed mention of the `formulaNamespace` field altogether and don't instruct developers to change the value.

jonathan-codaio (approved):
Sweet

PR #1378, opened 2021-09-17
Remove context.logger, in favor of console.log.

jonathan-codaio:
Sending companion `packs` and `coda` changes at the same time.

PTAL @huayang-codaio @patrick-codaio @coda/packs CC @ekoleda-codaio


huayang-codaio (approved):
minion_celebrate

Breaking change though. But good to be done before the packathon.

PR #1379, opened 2021-09-20
Auto scroll to API settings when opening account settings via `coda register`

alan-codaio:
Bug here: https://staging.coda.io/d/Packs-IA-go-packs_d36WK4zgrYx/My-Tasks_suDJO#View-of-My-Bugs_tu8_z/r196&modal=true

Our accounts page already has an abstraction that handles this! Tested by opening https://dev.coda.io:8080/account?openDialog=CREATE_API_TOKEN&scopeType=pack#apiSettings. 

https://user-images.githubusercontent.com/79662144/134059447-9c451e45-10ca-40f6-a1b9-aefcb1588433.mov

### DIFF
--- a/cli/register.ts
+++ b/cli/register.ts
@@ -21,7 +21,7 @@ export async function handleRegister({apiToken, codaApiEndpoint}: Arguments<Regi
     if (!shouldOpenBrowser.toLocaleLowerCase().startsWith('y')) {
       return process.exit(1);
     }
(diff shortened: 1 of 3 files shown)


patrick-codaio (approved):

jonathan-codaio (approved):
Aha, interesting. Cool, this works.

I had thought we'd need/want to fix this by having the Generate API Token button make sure to scroll the page to where you can find your token. But if we can scroll there in advance, that works too.

PR #1391, opened 2021-09-22
fixing the packs-sdk build

patrick-codaio:
`make build` and `make docs` were not working properly.


jonathan-codaio (commented):
I think we probably actually want to invest in configuring the linter to relax certain rules for our documentation directories. We will probably use some non-Coda-style patterns for examples that are a little more digestible for non-programmers.

patrick-codaio:
I made eslint less strict about what happens in documentation/examples/...

docs/reference/sdk/functions/assertCondition.md
```diff
@@ -1,18 +1 @@
...
-| `message?` | `string` |
-
-#### Returns
-
-asserts condition
-
-#### Defined in
-
-[helpers/ensure.ts:25](https://github.com/coda/packs-sdk/blob/main/helpers/ensure.ts#L25)
+REFLECTION_MEMBER
```
ekoleda-codaio, line 1: It looks like perhaps upgrading the typescript markdown library broke some of the docs?

ekoleda-codaio:
I can't repro this at head either, `make docs` and 'make lint` run without error.

patrick-codaio:
@ekoleda-codaio `make bs` then `make build` breaks on head for me. 

I'll look into the broken docs...

huayang-codaio:
I wonder if we would just want to disable linter for documentations/ at all.

huayang-codaio (approved):
I can repro this issue. Thanks for fixing. 

btw. can you double check runtime/isolated-vm/out/isolated_vm.node should be updated? It's not clear to me what led to the change.

huayang-codaio (commented):
oh, didn't realize that some .md files are broken.

jonathan-codaio:
Anybody have a preferred way forward here? Should we revert the offending typedoc-plugin-markdown? Or do they have a fix already that we can fast-forward to?

ekoleda-codaio:
It seem we could either a) pin typedoc and the plugin to the last working version of b) turn off the option `allReflectionsHaveOwnDocument`, which works around the bug.

patrick-codaio:
There is no fix in their github that i can see.

Also, I reverted to the old 3.10.4 dependency and regenerated the docs but that didn't fix this issue, so maybe i have a local env issue...

ekoleda-codaio:
I've opened an issue on their repo: https://github.com/tgreyuk/typedoc-plugin-markdown/issues/253

jonathan-codaio:
I'd say maybe let's just pin to the old version until the issue is resolved.

patrick-codaio:
@ekoleda-codaio nice find! I set `allReflectionsHaveOwnDocument: false` here now. WDYT?

ekoleda-codaio:
It doesn't look great IMO (https://coda-packs-sdk.readthedocs-hosted.com/en/patrick-fix-build-docs/reference/sdk/) but it's a fine workaround until they fix the bug.

PR #1392, opened 2021-09-22
Allow users to specify workspaceId when creating a Pack via CLI

alan-codaio:
Just plumbing this in so users don't have to use their default WS to create Packs!

### DIFF
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -137,6 +137,11 @@ if (require.main === module) {
           alias: 'd',
           describe: 'A description of the Pack. Can be set later in the UI.',
         },
+        workspaceIdOrUrl: {
+          string: true,
+          alias: 'w',
+          describe: 'The workspace ID, or workspace URL that you want your Pack to be created under.',
+        },
         codaApiEndpoint: {
           string: true,
           hidden: true,
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -137,7 +137,7 @@ if (require.main === module) {
           alias: 'd',
           describe: 'A description of the Pack. Can be set later in the UI.',
         },
-        workspaceIdOrUrl: {
+        workspace: {
           string: true,
           alias: 'w',
           describe: 'The workspace ID, or workspace URL that you want your Pack to be created under.',
--- a/cli/create.ts
+++ b/cli/create.ts
@@ -1,6 +1,7 @@
 import type {Arguments} from 'yargs';
 import {PACK_ID_FILE_NAME} from './config_storage';
 import {createCodaClient} from './helpers';
+import { ensureExists } from '../helpers/ensure';
 import {formatEndpoint} from './helpers';
 import {formatError} from './errors';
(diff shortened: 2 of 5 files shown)


cli/create.ts
```diff
@@ -17,16 +17,18 @@ interface CreateArgs {
   codaApiEndpoint: string;
   name?: string;
   description?: string;
+  workspaceId?: string;
```
jonathan-codaio, line 20: Should we allow users to pass a url or an id, and if it's a url we parse out the workspace id? Small things like this help a bit with usability because users often don't know how to get an id, but are generally comfortable copy/pasting urls from the browser if they know which page.

Also I think coda.ts needs to be updated to define the new flag. We could call it `workspace` so it's a little clearer you can pass a url or an id.

alan-codaio: yeah good callout, let me update!

cli/coda.ts
```diff
@@ -137,6 +137,11 @@ if (require.main === module) {
           alias: 'd',
           describe: 'A description of the Pack. Can be set later in the UI.',
         },
+        workspaceIdOrUrl: {
```
jonathan-codaio, line 140: Nit, maybe just call this `workspace`

cli/create.ts
```diff
@@ -64,3 +68,13 @@ export async function createPack(
     return printAndExit(errors.join('\n'));
   }
 }
+
+function parseWorkspaceIdOrUrl(workspaceIdOrUrl: string | undefined): string | undefined {
+  const urlMatch = workspaceIdOrUrl?.match(/.*\/workspaces\/ws-[a-zA-z0-9].{10}/);
```
jonathan-codaio, line 73: Hm, I don't think you need the `.` in there? That's saying 10 of any character. I think you want `[a-zA-z0-9]{10}` which is 10 of those specific characters.

Do we know that WS ids definitely don't contain dashes or underscores?

cli/create.ts
```diff
@@ -64,3 +68,13 @@ export async function createPack(
...
+  if (urlMatch) {
+    const parsedUrl = ensureExists(workspaceIdOrUrl).split('/');
+    return parsedUrl[parsedUrl.findIndex(part => part === 'workspaces') + 1]
```
jonathan-codaio, line 76: You can just create a capturing group for the workspace id using parens:

```
const match = /.*\/workspaces\/(ws-[a-zA-z0-9]{10})/.exec(workspaceIdOrUrl);
if (match) {
  return match[1];
}
```

cli/create.ts
alan-codaio: hmmm let me check. I might have overgeneralized from my single example..

alan-codaio: ok yeah looks like they have dashes or underscores:

https://github.com/coda/coda/blob/main/modules/js-core/utils/ids.ts#L1

jonathan-codaio (approved):

PR #1393, opened 2021-09-22
Revert dependabot commits to fix doc generation

patrick-codaio:
I kept reverting dependabot commits until `make bs && make build` worked. It seems the proper "fix" (https://github.com/coda/packs-sdk/pull/1391) uncovers a bug in the typedoc-plugin-markdown library


ekoleda-codaio (approved):
Thanks for the fix! Reading the comments in the typedoc-plugin-markdown history I _think_ you could just revert the `typedoc` version, but maybe it's safer to roll back everything until it works.

jonathan-codaio (commented):
Yeah I think you can just revert the typedoc one, right?

patrick-codaio:
Y'all are right @ekoleda-codaio @jonathan-codaio , it was only the typedoc. I could have SWORN I tried that before doing all this... oh well, updated.

PR #1394, opened 2021-09-22
Put all in-line examples in code blocks

ekoleda-codaio:
Improves how they are displayed in the rendered docs. [Preview](https://coda-packs-sdk.readthedocs-hosted.com/en/ek-example/reference/sdk/functions/makeFormula/).

### Before
Before

### After
After


jonathan-codaio (approved):
Ahh, cool deal

PR #1398, opened 2021-09-23
Revert "use fs move instead of rename (#1396)"

huayang-codaio:
This reverts commit da2d5139c93c4a1c4f49aa4c18a4ee867733c59f.

this is causing other issues.


alan-codaio (approved):
Rubber stamping ✅

PR #1399, opened 2021-09-23
Add enums to better support options for ScaleSchema

alan-codaio:
Realized that we have these options for `CurrencySchema` and `ScaleSchema` in main repo, but not here, which is probably a bad experience if users don't know what to put for `currencyCode` or `icon`.


schema.ts
```diff
@@ -110,10 +150,33 @@ export interface SliderSchema extends NumberSchema {
   step?: number | string;
 }
 
+export enum IconSet {
```
jonathan-codaio, line 153: Do we use this icon set for anything other than scales in `coda`? Just wondering if maybe we want to call it ScaleIconSet.

schema.ts
```diff
@@ -96,10 +96,50 @@ export enum CurrencyFormat {
   Financial = 'financial',
 }
 
+export enum CurrencyCode {
```
jonathan-codaio, line 99: What happens if you use a currency code that's not in this list? Wondering if it works if we might just want to document that it's an ISO-whatever currency code rather than trying to keep these in sync as we add more to the product.

schema.ts
alan-codaio: It's has other usage in Coda (An example use case is apparently we render the star icon in `paid_feature_banner.tsx`)

schema.ts
alan-codaio: We fall back to USD if they provide an unsupported currency code. I see, you think I should just add a comment saying that we fallback to USD if the currency is not supported?

jonathan-codaio: Yeah let's leave it as a string but add a doc comment that says something like:

```
/**
  A three-letter ISO 4217 currency code, e.g. USD or EUR. If the currency code is not supported by Coda, the value will be rendered using USD.
  */
```

schema.ts
jonathan-codaio: Hmm ok. Let's call it ScaleIconSet, I'd be surprised if we found another use for this really specific set of icons.

I think in the core product we want to get rid of this tiny subset of icons and use our full icon set, at which point we'd go back to using a string, and probably need to publish our icons8 icon names somewhere.

alan-codaio: yeah sounds good. It did seem a little weird for the non scale usages in coda repo.

schema.ts
```diff
@@ -99,6 +99,10 @@ export enum CurrencyFormat {
 export interface CurrencySchema extends NumberSchema {
   codaType: ValueHintType.Currency;
   precision?: number;
+  /**
+    A three-letter ISO 4217 currency code, e.g. USD or EUR.
```
jonathan-codaio, line 103: Oops, my example had bad syntax. Make sure each line starts with ` * `

jonathan-codaio (approved):

PR #1400, opened 2021-09-23
properly handle null response for object type

huayang-codaio:
report https://codaproject.slack.com/archives/C02ERJRBA21/p1632429365476100

basically we default an object type response to `{}` even if it's an optional field. on the other hand, the recursive check is using `typeof currentValue === 'object'`, while `typeof null` is `object`. 

so if you return null for a field that's object type, it will be come `{}` in the actual output. 

before:

```
-> % ./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake.ts NullPerson 10000 --vm
[ {} ]
```

after: 

```
-> % ./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake.ts NullPerson 10000 --vm
[ null ]
```


bundles/thunk_bundle.js
```diff
@@ -4452,6 +4452,9 @@ var AttributionNodeType;
 init_buffer_shim();
 var import_clone = __toModule(require_clone());
 
+// helpers/object_utils.ts
+init_buffer_shim();
```
jonathan-codaio, line 4456: Interesting, where did this diff come from?

jonathan-codaio: Ah, it's just an artifiact of introducing a new import?

test/packs/fake.ts
```diff
@@ -78,6 +78,16 @@ export const manifest: PackDefinition = createFakePack({
...
       },
     }),
+    makeFormula({
+      resultType: ValueType.Array,
+      items: fakePersonSchema,
+      name: 'NullPerson',
+      description: '',
+      parameters: [],
+      execute: ([]) => {
+        return [null];
```
jonathan-codaio, line 88: What was the behavior before the fix?

bundles/thunk_bundle.js
huayang-codaio: hmm. I am not sure but I tried to rebuild and the result is not changed. I guess it's just a hard coded stuff from esbuild whenever something is imported.

test/packs/fake.ts
huayang-codaio: the output was `[{}]` before this change.

jonathan-codaio: Yah makes sense. Any idea what the before/after looks like in an actual doc though? I actually don't know what `[null]` maps to in a doc.

huayang-codaio: oh you meant in a doc. I think empty object shows up as an empty circle.

image

null object is not one of the value type so it shouldn't be returned by itself.

The issue with null is that it fails validation with a confusing message. For example, the object schema may be requiring a field `id` of the object. Before this PR, the error message looks like `id is required by is undefined`. After this PR, I haven't tried yet but would expect some different error.

jonathan-codaio: Yep all makes sense and seems like a good fix.

Sorry to be paranoid, but mind trying this in a doc when you have a chance (not high priority obviously)? I'm a bit scared of changes to the key mapping just because we might have packs relying on fringe behavior that's not good but has just been there for a long time, and because I don't know how we handle null/undefined flowing through the pack result parsing code. Our TS doesn't include null/undefined as pack formula result values, so the `coda` code may or may not be handling them properly.

I suspect it's actually fine, I just want to be sure that we're not introducing new exceptions.

dragonghy: sure. happy to give this a thorough test. let me hold this pr until next. agree to be cautious on the format changes.

dragonghy: ooops lol I can actually use my personal github account to reply!

huayang-codaio:
seems like the user perceived response is identical before/after. 

image

testing pack:

```

// This line creates your new Pack.

pack.addFormula({
  // This is the name that will be called in the formula builder. Remember, your formula name cannot have spaces in it.
  name: "NullString",
  description: "A Hello World example.",
  parameters: [],
  cacheTtlSecs: 0,
  execute: function () {
    return null;
  },
  resultType: coda.ValueType.String
});

pack.addFormula({
  // This is the name that will be called in the formula builder. Remember, your formula name cannot have spaces in it.
  name: "ListOfNullString",
  description: "A Hello World example.",
  parameters: [],
  cacheTtlSecs: 0,
  execute: function () {
    return [null];
  },
  resultType: coda.ValueType.Array,
  items: { type: coda.ValueType.String },
});


pack.addFormula({
  // This is the name that will be called in the formula builder. Remember, your formula name cannot have spaces in it.
  name: "NullList",
  description: "A Hello World example.",
  parameters: [],
  cacheTtlSecs: 0,
  execute: function () {
    return null;
  },
  resultType: coda.ValueType.Array,
  items: { type: coda.ValueType.String },
});

pack.addFormula({
  // This is the name that will be called in the formula builder. Remember, your formula name cannot have spaces in it.
  name: "NullObject",
  description: "A Hello World example.",
  parameters: [],
  cacheTtlSecs: 0,
  execute: function () {
    return null;
  },
  resultType: coda.ValueType.Object,
  schema: coda.makeObjectSchema({
    type: coda.ValueType.Object,
    properties: {
      col: { type: coda.ValueType.String },
    }
  }),
});

pack.addFormula({
  // This is the name that will be called in the formula builder. Remember, your formula name cannot have spaces in it.
  name: "NullObjectWithRequiredField",
  description: "A Hello World example.",
  parameters: [],
  cacheTtlSecs: 0,
  execute: function () {
    return null;
  },
  resultType: coda.ValueType.Object,
  schema: coda.makeObjectSchema({
    type: coda.ValueType.Object,
    properties: {
      col: { type: coda.ValueType.String, required: true },
    }
  }),
});

pack.addFormula({
  // This is the name that will be called in the formula builder. Remember, your formula name cannot have spaces in it.
  name: "NullObjectWithRequiredFieldInSubObject",
  description: "A Hello World example.",
  parameters: [],
  cacheTtlSecs: 0,
  execute: function () {
    return [{child: null}];
  },
  resultType: coda.ValueType.Array,
  items: coda.makeObjectSchema({
    type: coda.ValueType.Object,
    properties: {
      child: {
        type: coda.ValueType.Object,
        properties: {
          id: { type: coda.ValueType.String, required: true },
        },
      },
    }
  }),
});
```

huayang-codaio:
take one step back. it's probably more essential to see if the response can be strongly typed.

huayang-codaio:
wow fixing the result type turns out to be more complicated than I thought. still exploring options..

huayang-codaio:
closing this one since I've added the strict typescript check for return values.

PR #1405, opened 2021-09-25
Properly support `ScaleSchema`, `SliderSchema`, `CurrencySchema` typing to be compatible with our value formats

alan-codaio:
It seems as though we never properly supported our `ScaleSchema` to be compatible with our value format configurations. `ScaleSchema` has the following typing:
```
{
  codaType: ValueHintType.Scale;
  maximum: number;
  icon: string;
}
```
but in multiple places we allow Pack makers to simply provide the `codaType`. When users only provide the `codaType`, our value format parser tries to parse the `maximum` and `icon` when they are undefined. This corrupted configuration causes the doc to RTE.

The second issue here is that even if users _do_ provide things like `icon` and `maximum` in their schema definitions, those properties get ignored because we only use number schema, which is of shape:
```
{
  type: ValueType.Number;
  codaType?: NumberHintTypes;
  ...baseSchemaProperties;
}
```
So zod strips these (erroneously) extraneous fields from the parsed metadata. 

The same issue occurs with `SliderSchema` and `CurrencySchema` too, but it doesn't cause the doc to RTE since we fallback to default values there.


api.ts
```diff
@@ -547,11 +546,11 @@ export function makeFormula<ParamDefsT extends ParamDefs>(
       break;
     }
     case ValueType.Number: {
-      const {resultType: unused, codaType, ...rest} = definition;
```
jonathan-codaio, line 564: This seems like a breaking change to start ignoring `codaType` at the top level of the formula definition. That was introduced as a shorthand so makers wouldn't have to confront the `schema` field in most cases. It's being used in a lot of pack defs so we can't just ignore. I think the answer is probably to use the schema if it's present, otherwise use `codaType` in a synthetic schema like it's currently doing.

api.ts
```diff
@@ -624,7 +623,7 @@ type StringFormulaDefV2<ParamDefsT extends ParamDefs> = BaseFormulaDefV2<ParamDe
 
 type NumericFormulaDefV2<ParamDefsT extends ParamDefs> = BaseFormulaDefV2<ParamDefsT, number> & {
   resultType: ValueType.Number;
-  codaType?: NumberHintTypes;
```
jonathan-codaio, line 641: As above, this is a breaking change, so we probably need to allow both the `codType` (short-hand) and `schema` (a full schema definition). Bonus points if we can allow exactly one or the other.

schema.ts
```diff
@@ -56,19 +56,16 @@ export const StringHintValueTypes = [
   ValueHintType.Markdown,
   ValueHintType.Url,
 ] as const;
-export const NumberHintValueTypes = [
+export const SimpleNumberHintValueTypes = [
```
jonathan-codaio, line 59: What's the distinction of simple hint types vs others? Date/Time/DateTime/Percent all have extra schema fields too.

schema.ts
```diff
@@ -96,21 +93,26 @@ export enum CurrencyFormat {
   Financial = 'financial',
 }
 
-export interface CurrencySchema extends NumberSchema {
+export interface SimpleNumberSchema extends BaseNumberSchema {
+   // These need no other values for Coda's value formats.
```
jonathan-codaio, line 97: Not sure that's true? Date and Time have a `format` attribute, and DateTime has `dateFormat` and `timeFormat`.

And Percent needs to be associated with NumericSchema somehow (as well as ValueType.Number with no hint type).

alan-codaio: by this, you mean a schema like
```
{
  type: ValueType.Number;
  codaType: ValueHintType.Percent;
}
```

Is an okay format?

api.ts
alan-codaio: Yeah that makes sense. I think the side effect of keeping that synthetic schema is that
```
{type: ValueType.Number, codaType}
```
is not valid for scale schemas (maximum and icon are required, and our coda parsing code has no fallback for creating the corresponding value format here. So we'd need the following extra steps:
- make `icon` and `maximum` optional properties
- in `coda` we can add a code change to default fallback to `icon: star` and `maximum: 5` or something (as long as the doc doesn't break if the result is greater than the maximum value). How does that sound?

Or alternatively we can add icon and maximum into the synthetic schema when the provided codaType is scale.

schema.ts
alan-codaio: oops, sorry for the messiness, will fix this

api.ts
jonathan-codaio: Ah interesting, the scale properties are required by the similar ones for slider are optional. Yeah, we should do what you suggest, make the scale properties optional and add defaults on the `coda` side. 

There's potentially a way to use existing defaults on the `coda` side, because if you just add a new slider column it applies defaults for you, we should try to trigger that defaulting or reuse the same default values.

schema.ts
jonathan-codaio: Yeah, that's a valid format, and so is

```
{ type: ValueType.Number, codaType: ValueHintType.Percent, precision: 3}
```

testing/upload_validation.ts
```diff
@@ -480,12 +474,86 @@ const booleanPropertySchema = zodCompleteObject<BooleanSchema & ObjectSchemaProp
...
+  precision: z.number().optional(),
+  currencyCode: z.string().optional(),
+  format: z.nativeEnum(CurrencyFormat).optional(),
   ...basePropertyValidators,
+})
+
+// TODO(alan): create base date, time, and datetime schemas to reuse across numeric schema
+// properties and string schema properties.
+
+const datePropertySchema = zodCompleteObject<DateSchema & ObjectSchemaProperty>({
```
jonathan-codaio, line 517: We need to define zod schemas for Date/Time/DateTime when the codaType is a string too, right, otherwise those fields will get stripped out when using a string formula?

Totally fine not to define a base zod schema, you can copy-paste the zod schemas but change the codaType to valueHintType.String.

schema.ts
```diff
@@ -79,24 +79,46 @@ export interface BooleanSchema extends BaseSchema {
...
 }
 
-export interface NumericSchema extends NumberSchema {
+export interface NumericSchema extends BaseNumberSchema {
   codaType?: ValueHintType.Percent; // Can also be undefined if it's a vanilla number
   precision?: number;
   useThousandsSeparator?: boolean;
 }
 
+export interface NumericDateSchema extends BaseNumberSchema {
```
jonathan-codaio, line 94: We probably want to eliminate BaseDateSchema then and rename DateSchema-->StringDataSchema perhaps? We have the same issues with those 3 string-based schemas.

api.ts
jonathan-codaio: This is still a breaking change since existing packs are using `codaType` to define a hint type at the top level of the formula definition. So we need to accept both `codaType` and `schema`. You provide `codaType` if you want the current behavior, a simple shorthand if you just want a type hint, but we don't support other attributions of the schema. If you want those, you need to use the `schema` attribute.

To make this either-or crystal clear, we should probably make the typing here something like:

```
BaseFormulaDefV2<ParamDefsT, number> & {
  resultType: ValueType.Number;
  execute(params: ParamValues<ParamDefsT>, context: ExecutionContext): Promise<number> | number;
} & 
  ({schema?: NumberSchema} | {codaType?: NumberHintTypes});
```

It's gnarly that we accept both, but we'd break existing packs if we didn't, and early feedback was that providing `schema` just to give a hint type was too confusing for novice users.

testing/upload_validation.ts
alan-codaio: Yeah I guess I was trying to preserve the BaseDateSchema as much as possible.. will update!

api.ts
alan-codaio: Sorry, just getting to these feedbacks. Yeah that makes sense to me. I actually think it'll be fine, now that we're going to make those `ScaleSchema` fields optional.

api.ts
```diff
@@ -638,9 +640,9 @@ type StringFormulaDefV2<ParamDefsT extends ParamDefs> = BaseFormulaDefV2<ParamDe
...
   execute(params: ParamValues<ParamDefsT>, context: ExecutionContext): Promise<number> | number;
-};
+} & 
+  ({schema?: NumberSchema} | {codaType?: NumberHintTypes});
```
jonathan-codaio, line 645: Do we need to do this for StringFormulaDefV2 too?

testing/upload_validation.ts
```diff
@@ -480,12 +475,83 @@ const booleanPropertySchema = zodCompleteObject<BooleanSchema & ObjectSchemaProp
...
+const currencyPropertySchema = baseNumberPropertySchema.extend({
+  type: zodDiscriminant(ValueType.Number),
+  codaType: zodDiscriminant(ValueHintType.Currency),
+  precision: z.number().optional(),
+  currencyCode: z.string().optional(),
+  format: z.nativeEnum(CurrencyFormat).optional(),
+  ...basePropertyValidators,
+})
+
+const numericDatePropertySchema = zodCompleteObject<NumericDateSchema & ObjectSchemaProperty>({
```
jonathan-codaio, line 515: We need similar zod schemas for the 3 string-based date/time schemas too, right?

jonathan-codaio (commented):
Should add a note about these changes in the CHANGELOG under the existing 0.5.0 header too.

Hopefully you can merge this before @huayang-codaio 's bigger TS change to avoid merge conflicts.

huayang-codaio:
ha, I was wondering what these are when I was doing my PR. 

> Hopefully you can merge this before @huayang-codaio 's bigger TS change to avoid merge conflicts. 

I am not rushing to merge but will need to do a bit clean up. Probably will attempt to merge by the end of today. I don't see many conflicts though by reading the changes in this PR.

testing/upload_validation.ts
alan-codaio: I was going to tackle them in their own followup PR, but you're right that these should probably be done in the same change 😅

test/upload_validation_test.ts
```diff
@@ -806,6 +843,10 @@ describe('Pack metadata Validation', () => {
               'alphanumeric characters, underscores, and dashes, and no spaces.',
             path: 'syncTables[0].schema.identity.name',
           },
+          {
```
alan-codaio, line 846: I cannot figure out why this test is now also failing the schema parsing around the getter's item schema of all things, especially because correcting the identity name (an unrelated field) results in no errors occurring.

schema.ts
```diff
@@ -174,16 +199,22 @@ export enum DurationUnit {
...
-export interface DurationSchema extends StringSchema<ValueHintType.Duration> {
+export interface DurationSchema extends BaseStringSchema<ValueHintType.Duration> {
   precision?: number;
   maxUnit?: DurationUnit;
 }
 
-export interface StringSchema<T extends StringHintTypes = StringHintTypes> extends BaseSchema {
+export interface BaseStringSchema<T extends StringHintTypes = StringHintTypes> extends BaseSchema {
   type: ValueType.String;
   codaType?: T;
```
jonathan-codaio, line 209: BaseNumberSchema doesn't have `codaType?: T` -- any reason these are inconsistent?

api.ts
```diff
@@ -364,7 +364,7 @@ export interface PackFormulaDef<ParamsT extends ParamDefs, ResultT extends PackF
 export interface StringFormulaDef<ParamsT extends ParamDefs> extends CommonPackFormulaDef<ParamsT> {
   execute(params: ParamValues<ParamsT>, context: ExecutionContext): Promise<string> | string;
   response?: {
-    schema: StringSchema;
+    schema: BaseStringSchema;
```
jonathan-codaio, line 367: How come this changed to BaseStringSchema, shouldn't it still be the StringSchema union type?

api.ts
```diff
@@ -402,8 +402,8 @@ export type BooleanPackFormula<ParamDefsT extends ParamDefs> = BaseFormula<Param
 export type StringPackFormula<
   ParamDefsT extends ParamDefs,
   ResultT extends StringHintTypes = StringHintTypes,
-> = BaseFormula<ParamDefsT, SchemaType<StringSchema<ResultT>>> & {
-  schema?: StringSchema<ResultT>;
+> = BaseFormula<ParamDefsT, SchemaType<BaseStringSchema<ResultT>>> & {
+  schema?: BaseStringSchema<ResultT>;
```
jonathan-codaio, line 406: Ditto for these.

api.ts
```diff
@@ -667,15 +671,15 @@ interface BaseFormulaDefV2<ParamDefsT extends ParamDefs, ResultT extends string
 
 type StringFormulaDefV2<ParamDefsT extends ParamDefs> = BaseFormulaDefV2<ParamDefsT, string> & {
   resultType: ValueType.String;
-  codaType?: StringHintTypes;
   execute(params: ParamValues<ParamDefsT>, context: ExecutionContext): Promise<string> | string;
-};
+} & 
+({schema?: BaseStringSchema} | {codaType?: StringHintTypes});
```
jonathan-codaio, line 676: Ditto here. It's inconsistent with what's happening with NumberSchema.

test/upload_validation_test.ts
jonathan-codaio: I can take a look.

jonathan-codaio: Aha, it's because makeSyncTable() copies the identity into the schema, which gets copied into the getter's item schema. So the item schema is actually failing validation for the same reason--it contains an identity whose name has spaces. Our zod error transforming is properly surfacing in the invalid identity name error at the top level, but it's not doing it for child fields. I think this is ok, maybe long term something to try to fix, but the items schema really does have an error so I think it's ok that we're flagging an error on it in this test.

testing/upload_validation.ts
```diff
@@ -480,18 +480,126 @@ const booleanPropertySchema = zodCompleteObject<BooleanSchema & ObjectSchemaProp
...
   codaType: z.enum([...StringHintValueTypes]).optional(),
   ...basePropertyValidators,
 });
 
+const stringPropertySchema = z.union([
+  stringDatePropertySchema,
+  stringTimePropertySchema,
+  stringDateTimePropertySchema,
+  durationPropertySchema,
+  baseStringPropertySchema
```
jonathan-codaio, line 600: I think this is a little over-permissive. Because this validator allows all of the string hint types, for the date/time/duration one's it's actually redundant with the specific validators you've already declared, and allows schemas with those hint types to ALSO match this validator, which is more permissive. So I think we should probably have a fall-through validator that's more strict about which codaTypes it will validate. I made that change along with a few others here, you can just borrow from this commit (or I can commit it to your branch or you can cherrypick it)

https://github.com/coda/packs-sdk/commit/8249aa6763985634ce2c1d77d0f749bbdb7e0c3f

testing/upload_validation.ts
```diff
@@ -480,18 +480,126 @@ const booleanPropertySchema = zodCompleteObject<BooleanSchema & ObjectSchemaProp
...
+  ...basePropertyValidators,
+})
+
+const numericPropertySchema = baseNumberPropertySchema.extend({
+  codaType: zodDiscriminant(ValueHintType.Percent).optional(),
+  precision: z.number().optional(),
+  useThousandsSeparator: z.boolean().optional(),
+})
+
+const scalePropertySchema = baseNumberPropertySchema.extend({
```
jonathan-codaio, line 494: I think it's unnecessary that these are extending baseNumberPropertySchema, and it's inconsistent with similar validators that just declare all of the fields without extending. In this case, I don't think you're getting anything from the base validator since you redeclare everything. So would just eliminate the extends. I made that change here:

https://github.com/coda/packs-sdk/commit/8249aa6763985634ce2c1d77d0f749bbdb7e0c3f

https://github.com/coda/packs-sdk/commit/8249aa6763985634ce2c1d77d0f749bbdb7e0c3f

testing/upload_validation.ts
```diff
@@ -480,18 +480,126 @@ const booleanPropertySchema = zodCompleteObject<BooleanSchema & ObjectSchemaProp
...
+
+const numberPropertySchema = z.union([
+  numericPropertySchema, 
+  scalePropertySchema, 
+  sliderPropertySchema, 
+  currencyPropertySchema,
+  numericDatePropertySchema,
+  numericTimePropertySchema,
+  numericDateTimePropertySchema,
+  baseNumberPropertySchema,
```
jonathan-codaio, line 550: I don't think you need this, it's overly-permissive for the same reason I mentioned below for baseStringPropertySchema, and you've already covered this case with numericPropertySchema. I removed it in https://github.com/coda/packs-sdk/commit/8249aa6763985634ce2c1d77d0f749bbdb7e0c3f

jonathan-codaio (commented):
It looks like your Prettier isn't working, so there's a lot of formatting issues, probably want to get that fix to avoid a bunch of formatting-only diffs.

testing/upload_validation.ts
alan-codaio: Just so I understand, this is just to make the typing cleaner (and not have overlapping types)? And it shouldn't really affect any edge cases, right?

jonathan-codaio: Well, the way it was before was actually negatively impacting edge cases. The union validator had validators for things like:

```
type: ValueType.String;
codaType: ValueHintType.Date;
format?: string;
```

and

```
type: ValueType.String;
codaType?: ValueHintType;
```

Suppose that `format` became a required field for schemas with date hint types. If you pass a schema with ValueHintType.Date but no format, zod should fail. But because baseStringPropertySchema looked like the bottom schema, it would allow that invalid schema through, because while it would fail the date validator, it would match the base validator.

In other words, need to make sure the types and validators are not just properly inclusive, but also properly exclusive.

api.ts
```diff
@@ -399,11 +399,8 @@ export type BooleanPackFormula<ParamDefsT extends ParamDefs> = BaseFormula<Param
   schema?: BooleanSchema;
 };
 
-export type StringPackFormula<
-  ParamDefsT extends ParamDefs,
-  ResultT extends StringHintTypes = StringHintTypes,
-> = BaseFormula<ParamDefsT, SchemaType<StringSchema<ResultT>>> & {
-  schema?: StringSchema<ResultT>;
+export type StringPackFormula<ParamDefsT extends ParamDefs> = BaseFormula<ParamDefsT, SchemaType<StringSchema>> & {
```
jonathan-codaio, line 402: @huayang-codaio Should check that this doesn't bust anything with the strong typing for return values he just added.

testing/upload_validation.ts
```diff
@@ -421,20 +436,10 @@ const commonPackFormulaSchema = {
...
-    codaType: z.enum([...NumberHintValueTypes]).optional(),
-    description: z.string().optional(),
-  }).optional(),
-});
-
 const stringPackFormulaSchema = zodCompleteObject<Omit<StringPackFormula<any>, 'execute'>>({
   ...commonPackFormulaSchema,
   resultType: zodDiscriminant(Type.string),
-  schema: zodCompleteObject<StringSchema>({
+  schema: zodCompleteObject<BaseStringSchema>({
```
jonathan-codaio, line 442: This needs to be using stringPropertySchema, for the same reason that you changed numericPackFormulaSchema to use numberPropertySchema, right?

jonathan-codaio (commented):
Home stretch!

api.ts
huayang-codaio: seems like `SchemaType<StringSchema>` is identical to string here anyway.

huayang-codaio:
cool as long as packs build passes. :P

testing/upload_validation.ts
alan-codaio: ugh, sorry, missed that too. I think I did a find-replace in vscode and it overeagerly renamed a bunch of things.

jonathan-codaio: I think you just need to address this one, and then we're good.

alan-codaio: ah good catch! updated!

testing/upload_validation.ts
```diff
@@ -430,16 +445,6 @@ const commonPackFormulaSchema = {
...
-  schema: zodCompleteObject<NumberSchema>({
-    type: zodDiscriminant(ValueType.Number),
-    codaType: z.enum([...NumberHintValueTypes]).optional(),
-    description: z.string().optional(),
-  }).optional(),
-});
-
 const stringPackFormulaSchema = zodCompleteObject<Omit<StringPackFormula<any>, 'execute'>>({
   ...commonPackFormulaSchema,
   resultType: zodDiscriminant(Type.string),
```
jonathan-codaio, line 450: I mean that line 451 probably needs to say `schema: stringPropertySchema.optional()`, in order for string formula return values to pick up all the new fields that you just added support for in object schema properties.

You correctly did this for `numericPackFormulaSchema`. On line 566, you changed the schema validator to `schema: numberPropertySchema.optional(),` so for the same reason, you need to make a similar change for string formulas, right?

alan-codaio: Ah right, sorry, keep losing track of some of these. Updated

jonathan-codaio (approved):

PR #1406, opened 2021-09-28
Update to Zod 3.9.1, use strict objects

oleg-codaio:
See https://github.com/coda/coda/pull/62037 - trying to figure out what's failing to validate. It might be best to just use strict objects by default (instead of stripping) so that we can tell and advise users if they're using, for instance, a misspelled property. And for any migrations we can make use of `.optional()` properties.

PTAL @jonathan-codaio

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "uuid": "^8.3.2",
     "xml2js": "^0.4.23",
     "yargs": "^17.1.1",
-    "zod": "^3.8.2"
+    "zod": "^3.9.1"
   },
   "devDependencies": {
     "@types/browserify": "^12.0.37",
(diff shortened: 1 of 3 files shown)


oleg-codaio:
Actually, this looks like a regression in Zod.

oleg-codaio:
See https://github.com/colinhacks/zod/issues/668

spencer-codaio:
will this fix https://github.com/coda/packs-sdk/pull/1413 as well?

spencer-codaio:
hmm it looks like this should be fixed in 3.9.3 according to that thread so the diffs that are there must be something else

oleg-codaio:
Yeah, worth investigating. Could be another bug in Zod.

PR #1419, opened 2021-09-28
fix validate no changes to do all commands that generate possible diff

spencer-codaio:
https://github.com/coda/packs-sdk/pull/1417 is failing even though it does not pass `make bs; make build` test locally. This is because our `vaidate-no-changes` step in CI only does `compile` which doesn't actually try to generate docs. Changing to use also do the `docs` step.

validated that on that branch with this step it fails and doesn't without

@jonathan-codaio cc @coda/packs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ publish-local: build
 	cp -r dist/* ../packs/node_modules/@codahq/packs-sdk/dist/
 
 .PHONY: validate-no-changes
-validate-no-changes: compile
+validate-no-changes: compile docs
 	$(eval UNTRACKED_FILES := $(shell git status --short))
 	$(eval CHANGED_FILES := $(shell git diff --name-only))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -156,7 +156,7 @@ publish-local: build
 	cp -r dist/* ../packs/node_modules/@codahq/packs-sdk/dist/
 
 .PHONY: validate-no-changes
-validate-no-changes: compile
+validate-no-changes: compile docs
```
spencer-codaio, line 159: this worked without doing the whole `build` step so we won't have to re-lint but lmk if there was another reason you wanted to do build @jonathan-codaio

jonathan-codaio (approved):
Yeah either way is fine. Using `build` means we don't have to keep the sub-steps of `build` in sync with those of `validate-no-changes`, but this changes infrequently.

PR #1421, opened 2021-09-29
enforce strict return type of formulas

huayang-codaio:
This PR is a best-effort attempt to enforce strict return type of the formulas. 

The high level idea is simple: just pass the types all the way down to every where that an execute function may need return type. Cool. 

There are two blockers to maximum leverage the typescript.

The final formula will normalize the schema (basically pascal case + replace : with _), making it impossible to add typing. closest stuff is `Capitalize<K>` but isn't close enough. So all the sealed formulas will have general result type. 

On the other hand, some properties have a field `fromKey` which will be used to replace the original key. So if the key is `x` and the fromKey is `y`, the execute function is supposed to return an object with `y` instead of `x`. This is also impossible to type because typescript can't get key from object values. If we migrate fromKey to something similar like `rename` (e.g. key is `y` and `rename` is `x`), we can type those keys.  

For the second case, this PR adds strong typing for all properties w/o fromKey. If there's at least one property with fromKey, `{[key: string]: any}` will be added to the result type in addition to the strong typed keys. 

## Testing

packs repo testing: https://github.com/coda/packs/commit/6d8724835aa364101f2bd425fd3d13e4bf8fbada
coda repo testing: https://github.com/coda/coda/commit/0a22a4c6db93d10bf5c3aaf98824b68816b166f7

All the changes only include typescript changes so that I'd expect no impact in functionality. 

I have fixed bunch of mis-typed return types in packs-sdk and packs repo. 

PTAL @jonathan-codaio @spencer-codaio @patrick-codaio @alan-codaio @coda/packs 

(I may still need to refactor a bit or add some explanation in the code)


jonathan-codaio (commented):
Oof, this is a nightmarish amount of TS :)

So this PR adds strong typing for formulas that return string/number/boolean/arrays of the scalars? And for formulas that return objects or arrays of objects, it just strongly types that they actually return objects?

I didn't quite follow the point about this PR doing some strong typing for schema fields without fromKey. As you say, since we do schema normalization, how can this do strong typing for those fields? Strong typing seems really hard in general because you are allowed to return both normalized and un-normalized values.

So, given that, I would punt for now on attempting to do any strong typing for object properties.

huayang-codaio:
> So this PR adds strong typing for formulas that return string/number/boolean/arrays of the scalars? And for formulas that return objects or arrays of objects, it just strongly types that they actually return objects?

This pr strong types all `string/number/boolean/arrays/objects` as long as there's no `fromKey`. 

The challenge of fromKey is that it requires the user to return keys that's not exactly the keys in schema (but in schema value fromKey). In that case my approach is to allow arbitrary key as long as some property has fromKey. 

On the other hand, there are two levels of string typing.
1) strong type the formula that user produces, 
2) strong type the normalized formula

doing 1) will give users a good idea if their return value doesn't match their schema. and that's what this PR is doing. 

2) is impossible due to normalization..

jonathan-codaio (approved):
A little bit intimidated by all the TS here (not that it was particularly clean before), but if this doesn't break anything in the `packs` repo, I'm willing to give this a shot.

jonathan-codaio:
We should describe this in the CHANGELOG though. This can go out as part of the existing 0.5.0 version we haven't published yet.

huayang-codaio:
cool. I will update changelog. 

The TS change is massive. I'll give it more testing & refactoring today.

PR #1422, opened 2021-09-30
allow string for Date codaType

huayang-codaio:
not sure why we require a codaType.Date to be Date in the first place but many packs are just returning just a string value.


schema.ts
```diff
@@ -273,7 +273,7 @@ export function isArray(val?: Schema): val is ArraySchema {
 type PickOptional<T, K extends keyof T> = Partial<T> & {[P in K]: T[P]};
 
 interface StringHintTypeToSchemaTypeMap {
-  [ValueHintType.Date]: Date;
+  [ValueHintType.Date]: Date | string;
```
jonathan-codaio, line 276: It's actually ` | number` too.

jonathan-codaio (approved):

PR #1425, opened 2021-10-01
Allow RegExp objects in column matchers

patrick-codaio:
This is an intermediate state, we will eventually remove the ability to have Strings in this field.

Planned sequence:
- Merge this SDK update
- Update SDK in Coda app
- Update all existing Coda packs to use RegExp objects (with a careful release of the updates)
- Announce to pack beta devs that the String option is going away
- Remove the String option from matchers type


ekoleda-codaio (approved):
LGTM, but I don't know the infra that well.

jonathan-codaio (approved):

PR #1428, opened 2021-10-04
Require uploaded format matchers to be regexes

patrick-codaio:


testing/upload_validation.ts
```diff
@@ -76,8 +84,32 @@ export async function validatePackVersionMetadata(metadata: Record<string, any>)
       validated.error.errors.flatMap(zodErrorDetailToValidationError),
     );
   }
+  const packMetadata = validated.data as PackVersionMetadata;
+  validateColumnFormats(packMetadata.formats);
```
jonathan-codaio, line 88: I think you probably just want to have zod do this validation, by adding a refiner to the existing `matchers` validator here:

https://github.com/coda/packs-sdk/blob/main/testing/upload_validation.ts#L629

Could be as simple as `matchers: z.array(z.string().regex(<whatever>)),`

jonathan-codaio: Or possibly using `refine` to add the extra check that `new RegExp()` works on the value too.

patrick-codaio: Wow, zod does it all. Done!

testing/upload_validation.ts
```diff
@@ -50,6 +50,14 @@ import {isNil} from '../helpers/object_utils';
 import {makeSchema} from '../schema';
 import * as z from 'zod';
 
+/**
+ * The uncompiled column format matchers will be expected to be actual regex objects,
+ * and when we compile the pack / stringify it to json, we will store the .toString()
+ * of those regex objects. This regex is used to hydrate the stringified regex back into
+ * a real RegExp object.
+ */
+export const PACKS_VALID_COLUMN_FORMAT_MATCHER_REGEX = /^\/(.*)\/([a-z]+)?$/;
```
jonathan-codaio, line 59: I assume this is exported because you plan to import on the `coda` side?

testing/upload_validation.ts
```diff
@@ -772,6 +780,22 @@ function validateFormulas(schema: z.ZodObject<any>) {
     });
 }
 
+function validateFormatMatchers(values: string[]): boolean {
```
jonathan-codaio, line 783: Hmm, I think this should be accepting a string and not an array now, because it's getting called once per matcher now, instead of once for all the matchers together.

testing/upload_validation.ts
patrick-codaio: Yup, for the runtime protection. Happy to just copy/paste it if that feels cleaner to you.

testing/upload_validation.ts
patrick-codaio: Ah, I had put the `.refine` outside the `.array()` call. Fixed.

jonathan-codaio (approved):

PR #1439, opened 2021-10-05
default CLI execute in vm

huayang-codaio:
simple change. default CLI execute with vm option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Formula return types are now strong-typed (except if you are using the fromKey attribute of object properties).
 
+- CLI `coda execute` now defaults to run with vm. To execute without vm, use `--no-vm`.
+
 ## 0.4.6
 
 - Bug fix: Executing sync table formulas via CLI now validates results correctly.
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -30,6 +30,7 @@ if (require.main === module) {
         vm: {
           boolean: true,
           desc: 'Execute the requested command in a virtual machine that mimics the environment Coda uses to execute Packs.',
+          default: true,
         },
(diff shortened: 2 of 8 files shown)


jonathan-codaio (approved):
Exciting.

I assume this works quite reliably now and stack traces are actionable?

huayang-codaio:
> I assume this works quite reliably now and stack traces are actionable?

@jonathan-codaio Except that that path may look weird because the bundle is built in a tmp folder. I spent a some time before but there isn't a easy way to fix it unless 1) we build the bundle in current folder, or 2) patch the source map package...

```
huayang@Huayangs-MBP [14:30:28] [~/code/packs-sdk] [hg-default-vm-execute]
-> % ./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake.ts Throw           
Error: here
    at throwError (../../../../../../private/var/folders/Users/huayang/code/packs-sdk/test/packs/fake.ts:24:9)
    at Object.execute (../../../../../../private/var/folders/Users/huayang/code/packs-sdk/test/packs/fake.ts:67:9)
    at doFindAndExecutePackFunction (/Users/huayang/code/packs-sdk/bundles/thunk_bundle.js:4796:22)
    at Object.findAndExecutePackFunction (/Users/huayang/code/packs-sdk/bundles/thunk_bundle.js:4786:18)
    at <unknown> (<isolated-vm>:1:13)
```

jonathan-codaio:
Seems non-blocking. Let's go ahead and merge and try it out.

Maybe we can clean up the stack traces but if it's really just path cleanliness that seems more like a backlog item than anything that should block us from starting to use this.

huayang-codaio:
I'll create a backlog item for this in the tech debt section.

PR #1442, opened 2021-10-07
[packs sdk documentation] upload docs script

spencer-codaio:
foundational script for publishing the docs that are generated by `mkdocs` into the developer documentation S3 bucket.

The file path is explained [here](https://staging.coda.io/d/Scalable-Packs-Storage_dYhioPL2vsk/High-Level-Design_suq7L#_luZc1) under "bucket structure."

image

needs this PR on the infra side for the proper access IAM https://github.com/coda/infra/pull/2125

@coda/packs / @jonathan-codaio @ekoleda-codaio @chris-codaio


Makefile
```diff
@@ -124,8 +127,21 @@ docs: typedoc generated-documentation
 view-docs:
 	PYTHONPATH=${ROOTDIR} PIPENV_IGNORE_VIRTUALENVS=1 expect -c 'spawn pipenv run mkdocs serve; expect "Serving on"; exec open "http://localhost:8000"; interact'
 
+###############################################################################
+### Deployment of documentation ### 
+
+# This step generates all the documentation for the SDK using mkdocs and dumps the contents in /site
+.PHONY: build-docs
+build-docs:
+	${PIPENV} run mkdocs build;
```
jonathan-codaio, line 136: Nit, do you need the semicolon?

Makefile
```diff
@@ -2,6 +2,9 @@ MAKEFLAGS = -s ${MAX_PARALLEL_MAKEFLAG}
 SHELL = /bin/bash
 ROOTDIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
+# TODO(spencer): this should probably be the semantic version of the sdk corresponding to the package.json
```
jonathan-codaio, line 5: Any reason not to just do this now?

Makefile
```diff
@@ -124,8 +127,21 @@ docs: typedoc generated-documentation
...
+# This step generates all the documentation for the SDK using mkdocs and dumps the contents in /site
+.PHONY: build-docs
+build-docs:
+	${PIPENV} run mkdocs build;
+
 .PHONY: publish-docs
-publish-docs:
+publish-docs: docs build-docs
+# TODO(spencer): change GIT SHA to the version from package.json
+	(cd ${ROOTDIR}; tsnode documentation/documentation_publisher.ts push ${CURRENT_GIT_SHA1})
```
jonathan-codaio, line 141: The binary is `ts-node`. Also this will only work if node_modules/.bin is on the path, so unless something is setting the path, probably want to reference it as node_modules/.bin/ts-node

Makefile
```diff
@@ -124,8 +127,21 @@ docs: typedoc generated-documentation
 view-docs:
 	PYTHONPATH=${ROOTDIR} PIPENV_IGNORE_VIRTUALENVS=1 expect -c 'spawn pipenv run mkdocs serve; expect "Serving on"; exec open "http://localhost:8000"; interact'
 
+###############################################################################
+### Deployment of documentation ### 
+
+# This step generates all the documentation for the SDK using mkdocs and dumps the contents in /site
+.PHONY: build-docs
```
jonathan-codaio, line 134: Might want to call this `build-mkdocs`. There's already a `docs` target that "builds" documentation, so we should probably just be specific about what each target actually does.

Alternatively, do we want to add this step to the `docs` target, or is unnecessary to do on each commit?

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,132 @@
...
+const program = new Command();
+
+// only works after `program` has been parsed
+function isVerbose() {
+  return program.opts().verbose;
+}
+
+function handleError(e: Error) {
+  console.error(e);
+  process.exit(1);
```
jonathan-codaio, line 24: We have a printAndExit utility that essentially does this, if you want to reuse.

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,132 @@
+/* eslint-disable no-console */
```
jonathan-codaio, line 1: Can use the `print()` utility in this repo that's just a synonym for console.log, that captures the intent of writing to the terminal and avoids needing to deal with the linter.

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,132 @@
...
+import AWS from 'aws-sdk';
+import {Command} from 'commander';
```
jonathan-codaio, line 3: Nit, I know `coda` just eliminated yargs in favor of Commander, but we only use yargs in this repo right now. Would be nice to migrate the the whole repo to Commander, but we have a lot bigger fish to fry right now, so would suggest just using yargs so we don't have to deal with two libraries for commands, and we can migrate the whole repo all at once at some later date.

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,132 @@
...
+    const fullFilePath = path.join(BaseGeneratedDocsPath, basePath);
+    const fileStats = fs.lstatSync(fullFilePath);
+    if (fileStats.isFile()) {
+      const contentType = mime.getType(fullFilePath) || undefined;
+      const fileData = fs.createReadStream(fullFilePath);
+      if (isVerbose()) {
+        console.log(`Uploading ${basePath} to S3 as a file`);
+      }
+      return s3
+        .putObject({
```
jonathan-codaio, line 64: No idea if it makes sense here, but in the `packs` repo we use the `s3 sync` command, which can handle syncing an entire directory full of content to/from S3 all in one go. We use this in the CI to pull down existing packs and do diffs against the latest build.

https://github.com/coda/packs/blob/main/scripts/get-current-packs.sh#L22

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,132 @@
...
+        files.map(fileName => {
+          const nestedFilePath = path.join(basePath, fileName);
+          return pushDocsDirectory(nestedFilePath);
+        }),
+      );
+    }
+  }
+
+  try {
+    console.log(`Pushing the current packs-sdk documentation ${key} to ${env}...`);
```
alan-codaio, line 88: I believe we also have a `print` helper, but if your code isn't triggering our no-console lint rule then it's probably fine to leave as is?

documentation/documentation_publisher.ts
jonathan-codaio: Yeah, these things tend to sit in an intermediate state for months. It's not great if you're trying to make simple changes to logic and you find yourself having to learn a whole new library just because the repo is inconsistent.

spencer-codaio: ok done

spencer-codaio:
ok updated to using aws s3 sync and the standard yargs stuff

image

will follow-up with adding the makefile handling for `activate` command

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,110 @@
...
+  return `${EnvironmentKeyRoot}/${environment}`;
+}
+
+async function pushDocsToEnv(env: string) {
+  const s3 = getS3Service(env);
+  const bucket = getS3Bucket(env);
+  const key = getS3ConfigKey();
+
+  async function pushDocsDirectory(): Promise<any> {
+    await exec(`aws s3 sync --profile ${env} --region ${AwsRegion} ${BaseGeneratedDocsPath} s3://${bucket}/${key}`);
```
jonathan-codaio, line 49: `spawnSync` may work here too, not sure if there's a need to make async vs not

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,110 @@
...
+    print(`Pushing the current packs-sdk documentation ${key} to ${env}...`);
+    await pushDocsDirectory();
+    print(`The current packs-sdk documentation was pushed to ${env}:${key} successfully.`);
+  } catch (err: any) {
+    handleError(err);
+  }
+}
+
+async function pushDocumentation() {
+  await Promise.all(Environments.map(env => pushDocsToEnv(env)));
```
jonathan-codaio, line 67: Probably want these to get pushed in a deterministic order, and let one succeed before trying the others? In which case, could use spawnSync and just use a regular `for` loop.

`prod` needs to get split out into a separate step I think though, in order for CircleCI to halt and wait for approval first?

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,110 @@
...
+
+function handleError(e: Error) {
+  printAndExit(e.message + ` ${e.stack || ''}`);
+}
+
+function getS3Service(env: string): S3 {
+  const config: S3.ClientConfiguration = {
+    signatureVersion: 'v4',
+    region: AwsRegion,
+    credentials: new AWS.SharedIniFileCredentials({profile: env}),
```
jonathan-codaio, line 25: Have you been able to test this locally? If so, is there a way to print a failure message if the credentials are invalid/missing, that explains what a developer should do to set up their environment to test this?

spencer-codaio: yeah i've been testing this locally and sure thing! I believe `prodaccess-admin` inserts the appropriate credentials into the ~/.aws/credentials file so i can try once my credentials are expired to see what the error is

documentation/documentation_publisher.ts
spencer-codaio: oh interesting didnt know a sync version existed will try that

documentation/documentation_publisher.ts
spencer-codaio: sure that sounds reasonable for pushing them in order.

the reason they all get pushed at the same time (including prod) is because the runtime config flow is 1) push everything 2) progressively active the new version for each environment with activating the new prod waiting on approval. Will see how feasible we can use the "active" piece with Cloudfront serving and then revisit the flow here if not.

spencer-codaio: although I guess if we're making it possible to serve all versions at once we actually want the push step to be hidden too since it might accidentally expose the latest version that hasn't been approved if someone is able to enumerate the URL so maybe just cleaner to only push once we approve too

spencer-codaio: ok this is resolved with the new logic

spencer-codaio:
New flow removes the "activating" logic and just relies on "pushing" as the way to release. Then in the CI workflow, we'll "push" progressively from head -> staging -> wait for approval -> prod.

image

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,90 @@
...
+import yargs from 'yargs';
+const exec = promisify(childExec);
+
+const AwsRegion = 'us-west-2';
+const BaseGeneratedDocsPath = 'site';
+const DocumentationBucket = 'developer-documentation';
+const PacksSdkBucketRootPath = 'packs';
+
+function handleError(e: Error) {
+  printAndExit(e.message + ` ${e.stack || ''}`);
```
jonathan-codaio, line 18: Total nice, `${e.message} ${e.stack || ''}`

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,90 @@
+import AWS from 'aws-sdk';
+import type {Arguments} from 'yargs';
+import S3 from 'aws-sdk/clients/s3';
+import {exec as childExec} from 'child_process';
+import {print, printError, printWarn} from '../testing/helpers';
+import {printAndExit} from '../testing/helpers';
+import {promisify} from 'util';
+import {version} from '../package.json';
+import yargs from 'yargs';
+const exec = promisify(childExec);
```
jonathan-codaio, line 10: Nit, newline above.

package.json
```diff
@@ -15,7 +15,7 @@
     "url": "https://github.com/coda/packs-sdk"
   },
   "dependencies": {
-    "@types/sinon": "^10.0.4",
+    "@types/aws-sdk": "^2.7.0",
```
jonathan-codaio, line 18: Can this go in devDependencies with all the other types?

documentation/documentation_publisher.ts
```diff
@@ -0,0 +1,90 @@
...
+  async function pushDocsDirectory(key: string): Promise<any> {
+    await exec(`aws s3 sync --profile ${env} --region ${AwsRegion} ${BaseGeneratedDocsPath} s3://${bucket}/${key}`);
+  }
+
+  try {
+    // If folder already exists, warn since we are uploading all the documentation wholesale instead of "syncing."
+    const obj = await s3.listObjectsV2({Bucket: bucket, MaxKeys: 1, Prefix: versionedKey}).promise();
+    if (!obj.Contents?.length) {
+      printWarn(`Folder already exists for ${version}.`);
+    }
```
jonathan-codaio, line 60: I'm not sure if this is that useful if we don't stop you if you're going to overwrite. I think it's fine to omit this, but if we want to warn people then we should error out in this case unless you pass a commandline flag to force the upload.

jonathan-codaio (commented):
Nice, quite clean now. Just minor stuff.

documentation/documentation_publisher.ts
jonathan-codaio: That's fine, mind having it error out here then, and have a yargs flag to force it to continue?

jonathan-codaio (approved):
How does one use `--forceUpload` in practice though, do you just add it to `make publish-docs-head --forcePublish` etc?

What happens if the CI stalls because of this? Assuming the plan is that the Codan just runs it manually on a local machine to overwrite and the CI is never supposed to force upload. Would just be good to comment about the intended usage, maybe near the flag definition or something.

PR #1443, opened 2021-10-07
Add a prompt to release the latest version for `coda release` if one is not provided by the manifest or the CLI

alan-codaio:
Now that it's possible to have versionless manifests, we need to handle this edge case in `coda release` where a version is not supplied on the CLI nor is one supplied by the manifest. In this case we grab the latest Pack version and attempt to release that instead.

I wonder if we should just make the manifest an optional argument now?

Tested output:
```
ts-node ../packs-sdk/cli/coda.ts release manifest.ts 
No version specified in your manifest. Do you want to release the latest version of the Pack (0.0.2)? (y/n)
n
```

### DIFF
--- a/cli/release.ts
+++ b/cli/release.ts
@@ -1,4 +1,5 @@
 import type {Arguments} from 'yargs';
+import type {BasicPackDefinition} from '..';
 import type {PackVersionDefinition} from '..';
 import {build} from './build';
 import {createCodaClient} from './helpers';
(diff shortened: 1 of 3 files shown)


cli/release.ts
```diff
@@ -41,6 +42,8 @@ export async function handleRelease({
     );
   }
 
+  const codaClient = createCodaClient(apiKey, formattedEndpoint);
+
   // TODO(alan/jonathan): Deal with the case of a pack that doesn't specify a version at all.
```
jonathan-codaio, line 47: Can resolve this TODO.

cli/release.ts
```diff
@@ -49,13 +52,24 @@ export async function handleRelease({
     try {
       const bundleFilename = await build(manifestFile);
       const manifest = await importManifest<PackVersionDefinition>(bundleFilename);
-      packVersion = manifest.version as string;
+      packVersion = manifest.version as string | undefined;
```
jonathan-codaio, line 55: Nit, this is my fault, but maybe these lines would be clearer as something like:

```
const manifest = await importManifest<BasicPackDefinition | PackVersionDefinition>(bundleFilename);
packVersion = 'version' in manifest && manifest.version;
```

cli/release.ts
```diff
@@ -49,13 +52,24 @@ export async function handleRelease({
...
     } catch (err: any) {
       return printAndExit(`Got an error while building your pack to get the current pack version: ${formatError(err)}`);
     }
   }
 
-  const codaClient = createCodaClient(apiKey, formattedEndpoint);
+  if (!packVersion) {
+    const {items: versions} = await codaClient.listPackVersions(packId, {limit: 1});
+    const [latestPackVersionData] = versions;
```
jonathan-codaio, line 63: I think we need to handle the case where the are no version yet, can just printAndExit an error message.

cli/release.ts
alan-codaio: Ah, did you make `importManifest` templated in your last PR? Sorry, I must've missed that.

PR #1453, opened 2021-10-13
Restore deleted documentation

ekoleda-codaio:
Most of our handwritten documentation was accidentally deleted in #1405.


ekoleda-codaio:
My guess is that some types were removed in that PR, but re-running tsdoc didn't delete the old generated files. So perhaps @alan-codaio deleted the entire docs directory and re-ran tsdoc, not realizing there were hand-written docs in there.

PR #1458, opened 2021-10-14
Add libraries guide

ekoleda-codaio:
Staged: https://coda-packs-sdk.readthedocs-hosted.com/en/ek-libraries/guides/advanced/libraries/

### DIFF
--- a/docs/guides/advanced/libraries.md
+++ b/docs/guides/advanced/libraries.md
@@ -1,3 +1,50 @@
 # Using libraries
 
-TODO
+JavaScript has a rich and diverse set of libraries available, making it easy to re-use the work of others and more quickly develop an application. You can utilize many of these libraries when building Packs, but there are some important limitations to be aware of.
+
+## Only available in the CLI
+
+At the moment you can only add libraries to your Pack when building with the Packs command line tool (CLI). If you are using the browser-based Pack Studio and want to use a library you'll have to migrate to the CLI workflow. See the [CLI getting started guide][get_started_cli] for information on how to install and setup the CLI.
+
+## Use NPM
+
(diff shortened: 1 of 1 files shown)


docs/guides/advanced/libraries.md
```diff
@@ -1,3 +1,50 @@
...
+[lodash]: https://lodash.com/
+
+## Library compatibility
+
+Some JavaScript libraries are written assuming the code is being run in either a web browser or on a Node.js server, and the Packs execution environment isn't exactly either. The execution environment provides access to the [standard built-in JavaScript objects][mdn_standard], but doesn't support [Web APIs][mdn_web] or [Node.js APIs][node_apis]. These include but are not limited to:
+
+- Browser: `window`, `document`, `XMLHttpRequest`, `fetch()`
+- Node.js: `fs`, `http`
+
+Unfortunately there isn't an easy way to determine beforehand if a given library will fail due to an unavailable API. At the moment the best approach is to try the library and see if it runs successfully in the Packs environment.
```
jonathan-codaio, line 40: You may want to expand on this (could be in a followup), the maker can e.g. use `coda execute` which will as of this week by default run your code in our VM with all the restrictions in place, and you should be able to see at that point what works and what doesn't.

Another potential followup would be to talk about how setTimeout is disabled by default, e.g. what gets explained here: https://github.com/coda/packs-sdk/blob/main/testing/injections/timers_disabled_shim.js#L5

ekoleda-codaio: I've added a line about `coda execute`, and it's a good callout. I started investigating the timers shims, but found some issues, so I'll address in a follow-up PR.

PR #1460, opened 2021-10-15
Flesh out the Todoist sample

ekoleda-codaio:
The idea here was to create a canonical integration Pack, extending the sample we already have for Todoist. I added:

- Getter formulas
- Column formats with matchers
- Action formula that does and update and returns the schema to update the sync table.

I also stripped down the comments and did some other changes to align with the style guide.

Once approved, I'll carve out individual features from this Pack to create smaller samples to highlight specific features.


documentation/examples/todoist/todoist.ts
```diff
@@ -1,100 +1,404 @@
-// This import statement gives you access to all parts of the Coda Packs SDK.
 import * as coda from "@codahq/packs-sdk";
 
-// This line creates your new Pack.
+
+// Constants.
+
+const ProjectUrlPatterns: RegExp[] = [
+  new RegExp("^https://todoist.com/app/project/([0-9]+)$"),
```
jonathan-codaio, line 7: Any reason to use `new RegExp()` instead of the slash notation? Is this part of the style guide?

documentation/examples/todoist/todoist.ts
```diff
@@ -1,100 +1,404 @@
...
+pack.setUserAuthentication({
+  type: coda.AuthenticationType.OAuth2,
+  // Settings pulled from the Todoist OAuth guide.
+  // https://developer.todoist.com/guides/#oauth
+  authorizationUrl: "https://todoist.com/oauth/authorize",
+  tokenUrl: "https://todoist.com/oauth/access_token",
+  scopes: ["data:read_write"],
+
+  // Determines the display name of the connected account.
+  getConnectionName: coda.makeMetadataFormula(async context => {
```
jonathan-codaio, line 32: You actually don't need `coda.makeMetadataFormula`, we changed it so you can just use a function directly and the builder will wrap it with `makeMetadataFormula` implicitly.

documentation/examples/todoist/todoist.ts
```diff
@@ -1,100 +1,404 @@
...
+});
 
-// When using the fetcher, this is the domain of the API that your pack makes
-// fetcher requests to.
-pack.addNetworkDomain("api.todoist.com");
+// Schemas
+
+// A reference to a synced Project. Usually you can use
+// `coda.makeReferenceSchemaFromObjectSchema` to generate these from the primary
+// schema, but that doesn't work in this case since a Project itself can contain
```
jonathan-codaio, line 48: Ah so these are references to their own table? That's unfortunate we can't demonstrate references more simply, alas, for another example.

documentation/examples/todoist/todoist.ts
```diff
@@ -1,100 +1,404 @@
...
+  identity: {
+    name: "Task",
+  },
+});
+
+/**
+ * Convert a Project API response to a Project schema.
+ */
+function toProjectSchema(project: any, withReferences=false) {
+  let result: any = {
```
jonathan-codaio, line 182: Can maybe avoid the `: any`, or does TS complain on line 191 since you're assigning into an inferred literal type?

documentation/examples/todoist/todoist.ts
```diff
@@ -1,100 +1,404 @@
...
+  featured: ["project", "url"],
+  identity: {
+    name: "Task",
+  },
+});
+
+/**
+ * Convert a Project API response to a Project schema.
+ */
+function toProjectSchema(project: any, withReferences=false) {
```
jonathan-codaio, line 181: Super nit, maybe `toProject()`? Pedantic, but it's not returning a schema, it's returning a project object that conforms to the schema.

documentation/examples/todoist/todoist.ts
```diff
@@ -1,100 +1,404 @@
...
+      name: "Not found",  // If sync'ed, the real name will be shown instead.
+    };
+  }
+  return result;
+}
+
+/**
+ * Convert a Task API response to a Task schema.
+ */
+function toTaskSchema(task: any, withReferences=false) {
```
jonathan-codaio, line 202: Ditto, `toTask()`

documentation/examples/todoist/todoist.ts
ekoleda-codaio: No, it was just to avoid having to escape so many forward slashes in the URL pattern.

documentation/examples/todoist/todoist.ts
ekoleda-codaio: Ah, right. Updated.

documentation/examples/todoist/todoist.ts
ekoleda-codaio: I think in the narrow samples I cut from this I will omit the same-table references for simplicity, but having worked through this challenge thought it was worth leaving them in, in case other devs run into this need.

documentation/examples/todoist/todoist.ts
ekoleda-codaio: Done.

jonathan-codaio (approved):

PR #1461, opened 2021-10-18
Fix code snippets that don't include `BEGIN` comment

ekoleda-codaio:
In #1365 I removed the `// BEGIN` comment from samples that no longer required it. However the logic of the snippet extractor assumed it would always be present, leading to some mangled samples. This PR updates the logic to accommodate samples that don't have the comment.


jonathan-codaio (commented):
Hmm yeah makes sense to add some kind of enforcement around BEGIN usage. Is the issue that `getCodeFile()` is used for both snippets (which should have BEGINs) and other examples (which should not)? If so, should we add a second input to indicate which file type it is, and throw an error if no BEGIN is found?

PR #1463, opened 2021-10-18
add lint git hook

spencer-codaio:
mostly takes files + process from the `coda` repo

image
triggers on push now

@coda/packs

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,18 @@ module.exports = {
   settings: {},
   rules: {
     '@typescript-eslint/restrict-plus-operands': 'error',
+    'max-len': [
+      'error',
+      {
+        code: 120,
+        tabWidth: 2,
+        ignoreUrls: true,
+        ignoreStrings: true,
+        ignoreTemplateLiterals: true,
+        ignoreRegExpLiterals: true,
+        ignorePattern: '^import ',
+      },
+    ],
   },
   overrides: [
     {


huayang-codaio:
Oh I recall why I didn't proceed earlier because I was using a npm to run hooks but we didn't want to propagate the npm dependency to downstream. Using a local script works.

.eslintignore
```diff
@@ -0,0 +1,7 @@
+.circleci/*
+.githooks/*
+.idea/*
+.vscode/*
+node_modules/*
+./documentation/snippets
+./documentation/examples
```
jonathan-codaio, line 7: We definitely want these to get linted, we just want them to have their own rule overrides. I'm trying to understand, is the custom .eslintrc.js not working properly or something?

documentation/.eslintrc.js
```diff
@@ -1,21 +1,23 @@
...
-        'max-len': ['error', { 'code': 80 }],
-        'quotes': ['error', 'double'],
+        '@typescript-eslint/no-unused-vars': [
+          'error',
+          {
+            varsIgnorePattern: 'response|datasourceUrl|MySchema',
+            argsIgnorePattern: 'context|param',
+          },
+        ],
+        'max-len': ['error', {code: 80}],
```
huayang-codaio, line 15: 80 is too small. can we do 

```
    'max-len': [
      'error',
      {
        code: 120,
        tabWidth: 2,
        ignoreUrls: true,
        ignoreStrings: true,
        ignoreTemplateLiterals: true,
        ignoreRegExpLiterals: true,
        ignorePattern: '^import ',
      },
    ],

```

huayang-codaio (approved):
i have no other comments.

documentation/.eslintrc.js
spencer-codaio: just to make sure, this is for the `documentation` linter that you want to make the change for? cc @ekoleda-codaio

huayang-codaio: oops mb. meant for the main one. and the main one has no configure for max-len. new configurations can be added now or later. I don't have a preference.

PR #1465, opened 2021-10-18
Divide samples into "Topic" and "Full" categories

ekoleda-codaio:
As per [go/sample-code-style-guide](http://go/sample-code-style-guide), shorter, topical samples are better for the editor side panel, while longer, more realistic samples are best read in the full page documentation. This PR introduces the concept of categories to the example metadata, mapping those types of samples to "Topic" and "Full" respectively. Both types of samples will be shown in the docs, but separated out in the nav. In a later PR I'll change the editor side panel logic to only include the topical samples.

This PR also:
- Introduces a new link type, `SamplePage`, which auto-generates a footer link that points to the full page for the sample in the documentation.
- Update the names of the topical examples to be titles ("Sync Tables") instead of method names (`addSyncTable()`).

Staged: https://coda-packs-sdk.readthedocs-hosted.com/en/ek-examplecat/samples/topic/authentication/


documentation/types.ts
```diff
@@ -41,10 +42,16 @@ export interface CompiledExample {
...
 }
 
 export enum UrlType {
+  SamplePage = 'SamplePage',
   SdkReferencePath = 'SdkReferencePath',
   Web = 'Web',
 }
+
+export enum ExampleCategory {
+  Topic = 'topic',
```
jonathan-codaio, line 55: Nit, the enum value convention is inconsistent between this and the UrlType enum right above it. Maybe capitalize these?

ekoleda-codaio: Fixed.

PR #1480, opened 2021-10-20
[hosted SDK documentation] add CI flow for publishing docs to S3

spencer-codaio:
# Summary
adds the CI flow for publishing docs to S3.

The process only happens on the new `release` branch. The idea is that whenever we're ready to publish a new NPM public version, we would merge into release and the CI would handle publishing to NPM as well by running that make command (this PR doesn't include that change yet)

This is what the CI workflow looks like
image

1. first validate that the directory is clean
2. build the docs and save that info for later
3. each step for each environment uploads the docs into S3 for the respective environment.
4. We wait for approval before proceeding with the prod push step.

PTAL @huayang-codaio cc @ekoleda-codaio @coda/packs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,6 +159,10 @@ workflows:
       - build:
           context:
             - dockerhub
+          filters:
+            branches:
+              ignore:
+                - release
 
   build-and-deploy-docs:
     jobs:


Makefile
```diff
@@ -200,8 +207,8 @@ publish-local: build
 
 .PHONY: validate-no-changes
 validate-no-changes: clean compile docs
-	$(eval UNTRACKED_FILES := $(shell git status --short))
-	$(eval CHANGED_FILES := $(shell git diff --name-only))
+	$(eval UNTRACKED_FILES := $(shell git status --short dist/))
```
spencer-codaio, line 210: this filters to just the `dist` directory

.circleci/config.yml
```diff
@@ -37,10 +72,147 @@ jobs:
...
+
+    steps:
+      - checkout
+      - bootstrap_node
+      - bootstrap_aws_cli
+      - attach_workspace:
+          at: ~/repo
+      - run:
+          name: Deploy Adhoc
+          command: 'make publish-docs-adhoc FLAGS=--forceUpload'
```
spencer-codaio, line 107: we force upload since we are only doing this on release, so we should be careful about merging into release instead of breaking the pipeline whenever we need to make changes to an existing version

.circleci/config.yml
```diff
@@ -37,10 +72,147 @@ jobs:
...
+            - util
+          circle_token: ${CIRCLECI_TOKEN}
+          requires:
+            - build
+      - build-docs:
+          context:
+            - dockerhub
+          requires:
+            - build      
+      # TODO(spencer): figure out why we aren't serving CDN on adhoc, technically this step isn't needed
```
huayang-codaio, line 184: I am not sure why it's necessary to push to adhoc. or if we meant to push to all envs (including adhoc), what do you mean by we are not serving CDN on adhoc?

.circleci/config.yml
```diff
@@ -37,10 +72,147 @@ jobs:
...
+      - push-adhoc-docs:
+          context:
+            - dockerhub
+          requires:
+            - build-docs
+      - push-head-docs:
+          context:
+            - dockerhub
+          requires:
+            - push-adhoc-docs
```
huayang-codaio, line 195: there's no manual approval here anyway. how about do it parallel with deploy adhoc. so only require build-docs?

.circleci/config.yml
```diff
@@ -37,10 +72,147 @@ jobs:
...
+      - push-staging-docs:
+          context:
+            - dockerhub
+          requires:
+            - push-head-docs
```
huayang-codaio, line 200: ditto.

.circleci/config.yml
```diff
@@ -37,10 +72,147 @@ jobs:
...
+      - utils/slack-notify-waiting-for-approval:
+          slack_bot_token: ${PUSH_REMINDER_BOT_TOKEN}
+          requires:
+            - push-staging-docs
```
huayang-codaio, line 204: here we can require all staging/head/adhoc

.circleci/config.yml
```diff
@@ -37,10 +72,147 @@ jobs:
...
+            - dockerhub
+          type: approval
+          requires:
+            - push-staging-docs
+      - push-prod-docs:
+          context:
+            - dockerhub
+          requires:
+            - wait-for-approval
+      # TODO: add publish NPM package with approval?
```
huayang-codaio, line 218: ah, good idea. so eventually we can leverage this to publish to npm? like the direction.

Makefile
huayang-codaio: seems good. but can you help me understand why this filter is necessary?

.circleci/config.yml
huayang-codaio: can you say a bit more about "we should be careful about merging into release instead of breaking the pipeline whenever we need to make changes to an existing version"?

Makefile
```diff
@@ -37,6 +37,13 @@ _bootstrap-python:
 _bootstrap-githooks: clean-githooks
 	-(cd ${ROOTDIR}; scripts/dev/git-hooks.sh --install)
 
+# This should only be needed for CI rather than locally,
+# so we don't run it in `make bs`
+.PHONY: bootstrap_aws_creds
+bootstrap_aws_creds:
+	mkdir -p ~/.aws
+	echo ${AWSBase64Creds} | base64 -d > ~/.aws/credentials
```
huayang-codaio, line 45: where does `AWSBase64Creds` come from?

huayang-codaio:
amazing work! left some suggestions / questions inline.

Makefile
spencer-codaio: yeah for some reason installing pipenv + all the dependencies is changing the `requirements.txt` file and couldn't figure out why... I'm assuming this was the original intent anyways given the below message says "dist directory not clean"

spencer-codaio: oh wait i think i removed the pip dependency from this step tho so let me try removing

.circleci/config.yml
```diff
@@ -1,4 +1,15 @@
 version: 2.1
+orbs:
+  aws-cli: circleci/aws-cli@1.3.0
+  jq: circleci/jq@1.9.1
+  utils: coda/utils@0.5.0
```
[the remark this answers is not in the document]
spencer-codaio: haha yeah a bunch of neat stuff here

Makefile
spencer-codaio: right now we just insert it as an environment variable in circleCI for each repo but i think we're moving towards using circleCI contexts https://circleci.com/docs/2.0/contexts/

PR #1501, opened 2021-10-29
Add a script that indicates which identifiers need TypeDoc documentation.

jonathan-codaio:
Output looks like:

```
TemporaryBlobStorage                     Interface                      api_types.ts
storeBlob                                Method                         undefined
storeUrl                                 Method                         undefined
getConnectionName                        Property                       types.ts
getConnectionUserId                      Property                       types.ts
type                                     Property                       types.ts
DefaultValueType                         Type alias                     api_types.ts
ExternalObjectPackFormula                Type alias                     compiled_types.ts
ExternalPackFormat                       Type alias                     compiled_types.ts
```

Only 380 to go! Many are enum members though which hopefully can go fast.


ekoleda-codaio (approved):
Thanks for putting this together!

PR #1502, opened 2021-10-30
fix runtime bootstrap upon naming conflict

huayang-codaio:
This is a tricky bug. if the exported bundle has an export symbol `coda` or `pack`, it will conflict with the global symbol `coda` and `pack` and such confuses `registerBundle`. 

This PR adds `global.` prefix to make sure that the `registerBundle` is setting to the global symbol. (of course if the bundle has an export called global we are screwed. hopefully nobody would do so)

also adding a v2 pack for testing.

### DIFF
--- a/dist/runtime/bootstrap/index.js
+++ b/dist/runtime/bootstrap/index.js
@@ -157,7 +157,7 @@ async function registerBundle(isolate, context, path, stubName) {
     // compiling the bundle allows IVM to map the stack trace.
     const bundle = fs_1.default.readFileSync(path).toString();
     // bundle needs to be converted into a closure to avoid leaking variables to global scope.
-    const script = await isolate.compileScript(`(() => { ${bundle} \n global.${stubName} = exports })()`, {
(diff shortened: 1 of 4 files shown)


jonathan-codaio (commented):
Can you say more about the case where a pack has an export named `pack` -- don't all packs have that by definition? How come this wasn't breaking before?

huayang-codaio:
@jonathan-codaio 

that code used to be 

```
  const script = await isolate.compileScript(`(() => { ${bundle} \n ${stubName} = exports })()`, {
    filename: path,
  });
```

after injecting the bundle code and stubName (`pack`), it looks like 

```
(() => { $
  const pack = coda.newPack();
  ...
  exports.pack = pack;  // this is "export const pack"
  ...

  pack = exports;
})()
```

huayang-codaio:
maybe a better way is to do 

```
(() => { 
  ${stubName} = (() => {
    ${bundle} \n 
    return exports;
  })() 
})()
```

jonathan-codaio (approved):

PR #1503, opened 2021-11-01
Small pass of TSDoc content, largely for schema elements.

jonathan-codaio:
Starting to churn through the outstanding undocumented SDK identifiers. Hoping to make this fairly streamlined, open to suggestions to facilitate that. One idea might be to collaborate on the PRs, I can take a first pass like this with the meat of the content, as articulate as I can, and rather than providing PR comments with text clarifications, you could edit the code directly on the branch and push them up? I mostly completely defer to your superior stylistic preferences, and I think the only things we'd nee to take time to discuss would be anything that relates to correctness of the content.


ekoleda-codaio (approved):
Sounds like a good plan; don't let the perfect be the enemy of the good :-) This PR LGTM.

PR #1509, opened 2021-11-03
More TSDoc, for some enums.

jonathan-codaio:
Another small pass.

I also exempted some enum values where it seemed redundant to document the individual values, which seemed relatively self-explanatory.

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -3,6 +3,14 @@ import type {ArraySchema} from './schema';
 import type {Continuation} from './api';
 import type {MetadataFormula} from './api';
 
+/**
+ * Markers used internally to represent data types for parameters and return values.
+ * It should not be necessary to ever use these values directly.
+ *
+ * When defining a parameter, use {@link ParameterType}. When defining
+ * a formula return value, or properties within an object return value,
+ * use {@link ValueType}.
+ */
 export enum Type {
   string,
   number,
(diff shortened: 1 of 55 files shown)


api_types.ts
```diff
@@ -370,6 +424,22 @@ export interface SyncExecutionContext extends ExecutionContext {
 
 // A mapping exists in coda that allows these to show up in the UI.
 // If adding new values here, add them to that mapping and vice versa.
+/**
+ * Special "live" date range values that can be used as the {@link defaultValue}
```
ekoleda-codaio, line 428: Aside: It's odd that for all other types of `defaultValue` it just affects the display and the parameter is undefined, but in this one special case it causes the parameter to be populated with real defaults.

api_types.ts
```diff
@@ -370,6 +424,22 @@ export interface SyncExecutionContext extends ExecutionContext {
...
+ * for a date array parameter.
+ *
+ * Date array parameters are meant to represent date ranges. A date range can
+ * be a fixed range, e.g. April 1, 2020 - May 15, 2020, or it can be a "live"
+ * range, like "last 30 days".
+ *
+ * At execution time, a date range will always be passed to a pack as an
+ * array of two specific dates, but for many use cases, it is necessary
```
ekoleda-codaio, line 436: "always ... two specific dates" - What about in the non-builder case? Are you limited to only two dates in the list?

schema.ts
```diff
@@ -274,6 +274,11 @@ export interface SliderSchema extends BaseNumberSchema<ValueHintType.Slider> {
   step?: number | string;
 }
 
+/**
+ * Icons that can be used with a {@link ScaleSchema}.
+ *
+ * For example, to render a star rating, use a {@link ScaleSchema} with `icon: ScaleIconSet.Star`.
```
ekoleda-codaio, line 280: Nitpick: Can you prefix the enum with `coda.`, since that's the default in the style guide.

ekoleda-codaio (changes requested):

api_types.ts
```diff
@@ -82,19 +90,65 @@ export type TypeOf<T extends PackFormulaResult> = T extends number
...
+  /**
+   * Indicates a parameter that is a Coda date value (which includes time and datetime values).
+   */
   Date = 'date',
+  /**
+   * Indicates a parameter that is a Coda rich text value that should be passed to the pack as HTML.
+   */
   Html = 'html',
+  /**
+   * *Not yet supported.* Indicates a parameter that is a Coda image. Eventually, such a parameter will
```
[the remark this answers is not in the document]
jonathan-codaio: Whoa, didn't realize. We had all this debate about people wanting to use images in their packathon packs but they were getting the image name and not the image url due to how our formula engine coerces values, so I thought this didn't work. Maybe their problem was that they were just using String parameter definitions instead of Image parameters?

api_types.ts
jonathan-codaio: Hmm yeah I agree that `defaultValue` handling is funky in general. I'm not sure this is actually different though? In both cases it populates the initial Coda formula with a value. In this case, the value populated is just dynamically evaluated based on the current date, rather than a hardcoded value. But the user will observe this in the context of a builder UI showing date range options, and doesn't have to know that under the hood it was just an initial CFL clause.

schema.ts
jonathan-codaio: Done.

ekoleda-codaio (approved):

PR #1510, opened 2021-11-03
More typedoc, mostly for schema types.

jonathan-codaio:

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -102,7 +102,7 @@ export class StatusCodeError extends Error {
    */
   statusCode: number;
   /**
-   * The body of the HTTP response, parsed as a JavaScript object if the contents were JSON data.
+   * The parsed body of the HTTP response.
    */
   body: any;
   /**
--- a/api.ts
+++ b/api.ts
@@ -54,9 +54,21 @@ export {FetchRequest} from './api_types';
  * and the Coda UI will display the message.
  */
 export class UserVisibleError extends Error {
+  /** @hidden */
   readonly isUserVisible = true;
+  /** @hidden */
   readonly internalError: Error | undefined;
 
+  /**
+   * Use to construct a user-visible error.
+   *
+   * @example
+   * ```
+   * if (!url.startsWith("http")) {
+   *   throw new coda.UserVisibleError("Please provide a valid url.");
+   * }
+   * ```
+   */
   constructor(message?: string, internalError?: Error) {
     super(message);
     this.internalError = internalError;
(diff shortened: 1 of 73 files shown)


schema.ts
```diff
@@ -200,33 +213,74 @@ export type NumberSchema =
...
 export interface NumericDateSchema extends BaseNumberSchema<ValueHintType.Date> {
+  /** Instructs Coda to render this value as a date. */
   codaType: ValueHintType.Date;
-  // A Moment date format string, such as 'MMM D, YYYY', that corresponds to a supported Coda date column format.
+  /**
+   * A Moment date format string, such as 'MMM D, YYYY', that corresponds to a supported Coda date column format.
+   *
+   * Only applies when this is used as a sync table property.
+   */
   format?: string;
```
jonathan-codaio, line 247: This API is currently kind of crappy. The main app represents format options for date/time/datetime columns as a preset list Moment format strings, like 'MMM D, YYYY'. In practice, this is an enum, since there are a specific set of valid values.

The reason this was added was for Cross-doc, which gets the format string of the source table via the API, and then passes it through as the schema of the destination table. The API doesn't attempt to enumerate all the formats, it just knows that they're strings. And theoretically DocX folks could add new format strings at any time. So we were trying to eliminate the need to update the API or SDK when that happens.

But for now it means that unless you use one of the magic format strings that the app supports, your formatting will be ignored. (And it will already be ignored if this isn't a sync table property, since canvas values don't support formatting yet.)

I don't know if we want to aim to convert this to an enum, or explain this, or list the valid values somewhere, or something else.

api.ts
```diff
@@ -68,21 +80,45 @@ interface StatusCodeErrorResponse {
...
+  /**
+   * The name of the error, for identiciation purposes.
+   */
   override name: string = 'StatusCodeError';
+  /**
+   * The HTTP status code, e.g. `404`.
+   */
   statusCode: number;
+  /**
+   * The raw body of the HTTP response.
```
ekoleda-codaio, line 105: In my testing for the [fetcher guide](https://coda.github.io/packs-sdk/guides/advanced/fetcher/#errors) I found that this contains the parsed body, just like `response.body`. If this is WAI then I think "raw" gives the wrong impression.

schema.ts
```diff
@@ -183,13 +183,26 @@ export type NumberHintTypes = typeof NumberHintValueTypes[number];
 export type ObjectHintTypes = typeof ObjectHintValueTypes[number];
 
 interface BaseSchema {
+  /**
+   * A explanation of this object schema property shown to the user in the UI.
+   *
+   * If your pack has a object schema with many properties, it may be useful to
```
ekoleda-codaio, line 189: Typo: "a object schema" => "an object schema"

schema.ts
ekoleda-codaio: I agree that an enum is problematic. The Sheets API also had a string format, and we just documented that format via handwritten docs:

https://developers.google.com/sheets/api/guides/formats

Once we have that we can come back and update these reference docs to point to that page.

api.ts
jonathan-codaio: Ah good to know. Yeah this whole object is confusing (the confusion comes from the request-promise behavior we're trying to keep compatibility with). That sounds right. In which case, I think `error` is just an alias for `body`. I believe `response.body` is supposed to be unparsed, and indeed it looks like we try to ensure that it is: https://github.com/coda/packs-sdk/blob/main/api.ts#L98

schema.ts
jonathan-codaio: Thanks, done.

schema.ts
jonathan-codaio: Aha great, glad there's precedence for this.

api.ts
```diff
@@ -102,11 +102,11 @@ export class StatusCodeError extends Error {
    */
   statusCode: number;
   /**
-   * The raw body of the HTTP response.
+   * The body of the HTTP response, parsed as a JavaScript object if the contents were JSON data.
```
ekoleda-codaio, line 105: I imagine we also do the XML parsing here? And the binary -> Buffer? If so it might be best to just say that we parse it. Ideally we'd link to the fetcher guide that talks about response parsing, but I think at the moment we don't have a good way to link from TypeDoc to mkdocs, given the variable leading path.

api.ts
ekoleda-codaio: I found it extra confusing that `response.body` is parsed, but `error.response.body` isn't parsed. Is the compatibility with request-promise still needed, or is that legacy?

ekoleda-codaio (changes requested):

api.ts
jonathan-codaio: Good point, done.

api.ts
jonathan-codaio: We have a bunch of old packs relying on this. Ideally we would fix them but it's pretty hard to track down all of the obscure behavior that they're relying on, in particular because of the weak typing.

ekoleda-codaio (approved):

PR #1512, opened 2021-11-04
add custom authentication type

spencer-codaio:
# Summary
Adds a custom authentication type for users to use templating for arbitrary insertion of system or user-provided secrets. This handles the Plaid pack's use case instead of overloading the WebBasic authentication type like we currently do.

[Accompanying Packs PR](https://github.com/coda/coda/pull/63879)

context here: https://staging.coda.io/d/Packs-Project-Tracker-go-pak_dL5PQJ5pk51/My-tasks-stories_suQf8#My-tasks_tujrl/r319&modal=true

@coda/packs / @jonathan-codaio


chris-codaio:
Is there a spec for this?

types.ts
```diff
@@ -79,6 +79,11 @@ export enum AuthenticationType {
    * See https://en.wikipedia.org/wiki/Basic_access_authentication
    */
   WebBasic = 'WebBasic',
+  /**
+   * Authenticate using custom templated values, which you insert into the request. The user provides all the
```
jonathan-codaio, line 83: Maybe something like:

`Authenticate in a custom way by having one or more arbitrary secret values inserted into the request url or body using template replacement. See {@link CustomAuthentication}.`

types.ts
```diff
@@ -413,6 +418,30 @@ export interface WebBasicAuthentication extends BaseAuthentication {
   };
 }
 
+export interface CustomAuthParameter {
+  /**
+   * The name used to refer to the parameter. Should match the string for the key in the params mapping.
```
jonathan-codaio, line 423: Not sure what the "params mapping" is. Maybe `A name used to refer to this parameter. Use this name to generate the template replacement string. See {@link CustomAuthentication}.`

types.ts
```diff
@@ -413,6 +418,30 @@ export interface WebBasicAuthentication extends BaseAuthentication {
...
+  name: string;
+
+  /**
+   * A description shown to the user indicating what value they should provide for this parameter.
+   */
+  description: string;
+}
+
+/**
+ * Authenticate using custom templated parameters, which you insert into the request. The user provides all the
```
jonathan-codaio, line 434: Maybe something like:

Authenticate in a custom way by having one or more arbitrary secret values inserted into the request url or body using template replacement.

Some APIs use non-standard authentication schemes which often require secret credentials to be put in specific places in the request URL or request body. Custom authentication supports many of these cases by allowing you as the pack author to define one or more secret values that the user must provide (or that you as the pack author must provide, if using system authentication), and then you can specify where these values are to be inserted into the request using a syntax using curly braces inspired by template engines.

The syntax to use for inserting a secret credential is `{{<paramName>-<invocationToken>}}`. That is, curly braces surrounding the name of one of the parameters you define, with a suffix that is the unique invocation token for the current formula invocation. The invocation token is part of the {@link ExecutionContext} and is required for security reasons.

@example
```
// Suppose you're using an API that wants a secret id in the request URL,
// and a secret value in the request body. You can define a Custom authentication
// configuration with two params, one called "secretId" and one called "secretValue".
// The user will be prompted to specify a value for each of these when setting up an account.
// In the `execute` body of your formula, you can specify where those values are
// inserted into the request using curly brace {{ }} template replacement syntax.
// The string that you put inside the curly braces is the name of a parameter
// (in this case either "secretId" or "secretValue"), then a dash, then the value
// of `context.invocationToken`. This makes the actual template replacement string
// unique to this formula invocation, which is important for security reasons.
execute: async function([], context) {
  let secretIdTemplateName = "secretId-" + context.invocationToken;
  let urlWithSecret = "/api/entities/{{" + secretIdTemplateName + "}}"

  let secretValueTemplateName = "secretValue-" + context.invocationToken;
  let bodyWithSecret = JSON.stringify({
    key: "{{" + secretValueTemplateName + "}}",
    otherBodyParam: "foo",
  });
 
  let response = await context.fetcher.fetch({method: "GET", url: urlWithSecret, body: bodyWithSecret});
  ...
}
```

types.ts
```diff
@@ -413,6 +418,30 @@ export interface WebBasicAuthentication extends BaseAuthentication {
...
+ * parameters you specify, and you have access to them through a templating engine.
+ */
+export interface CustomAuthentication extends BaseAuthentication {
+  type: AuthenticationType.Custom;
+  /**
+   * Configuration for labels to show in the UI when the user sets up a new acount.
```
jonathan-codaio, line 440: This would be more than just UI labels, it defines what parameters are eligible for use at all.

testing/upload_validation.ts
```diff
@@ -343,6 +344,24 @@ const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny>
     }).optional(),
     ...baseAuthenticationValidators,
   }),
+  [AuthenticationType.Custom]: zodCompleteStrictObject<CustomAuthentication>({
+    type: zodDiscriminant(AuthenticationType.Custom),
+    params: z
+      .array(
+        zodCompleteStrictObject<CustomAuthentication['params'][number]>({
+          name: z.string(),
+          description: z.string().optional(),
```
jonathan-codaio, line 353: The TypeScript for this field isn't optional. Should make these match, whatever you preference is.

testing/fetcher.ts
```diff
@@ -230,6 +231,30 @@ export class AuthenticatingFetcher implements Fetcher {
...
+      case AuthenticationType.Custom: {
+        const request = {
+          url,
+          body,
+          form,
+          headers,
+        };
+        const {params} = this._credentials as CustomCredentials;
+
+        let requestWithTemplateSubstitutions = JSON.stringify(request);
```
jonathan-codaio, line 243: Hmm, I might just do replacement on each request component separately. Fields a bit weird/fragile to make a big string out of disparate fields and replace across all of them.

testing/fetcher.ts
```diff
@@ -230,6 +231,30 @@ export class AuthenticatingFetcher implements Fetcher {
...
+        if (requestWithTemplateSubstitutions) {
+          // For awful APIs that require auth parameters in the request body, we have
+          // this scheme where we do template substitution of the body using an unguessable
+          // random token as part of the template key.
+          Object.entries(params).forEach(([key, value]) => {
```
jonathan-codaio, line 248: I wonder if we care to look at the params from the auth definition or not. For the most part, the set of params from the auth def should be exactly the same as the keys in the secret credentials, since the latter is derived from the former. But if you're changing your auth definition, they will diverge until you update the secrets. I suppose during that transition we could continue to use the old secrets verbatim, vs ignoring any keys that aren't present in the current auth definition. I think either way your requests probably won't work until the secrets are updated. So possibly fine as-is.

types.ts
spencer-codaio: oh this is amazing thanks for the suggestion! The example looks great -- any thoughts on whether we should include a link to real API examples of when you might need this? (i.e. link to plaid and briefly 1-2 sentence describe it requires secret in both the url and body)

testing/upload_validation.ts
spencer-codaio: oops good catch, will make it required because it's the only indication in the UI for user auth for what the user should enter here lol

testing/fetcher.ts
spencer-codaio: hmm ok i was debating between the two approaches. Making it explicit SGTM. Another question I had here was how to handle template substitution inside the URL. Currently how we do it for web basic is we replace the URI encoded version of `{{hello-1234}}`. My inclination was that we should support replacing both the URI encoded flavor as well as the normal `{{}}` variant. LMK if you have other opinions

testing/fetcher.ts
spencer-codaio: Yeah this is how we currently do it for `MultiQueryParam` too where we store the actual "expected keys" from the current definition of the pack as well as the "stored key-value pairs" to handle when these values change. 

I'm not sure if we have any flow of invalidating/reauthenticating a user/system account connection tho when thes values change.. perhaps something for us to think about as a smooth item later on cc @alicia-codaio @punit-codaio

testing/fetcher.ts
huayang-codaio: +1 don't feel very comfortable do a replacement on the full string. at least don't want to worry about the corner cases.

types.ts
spencer-codaio: thoughts on 
```
* Authenticate in a custom way by having one or more arbitrary secret values inserted into the request url, body,
* headers, or the form data using template replacement. See {@link CustomAuthentication}.
```
given im replacing also those pieces in the request?

types.ts
spencer-codaio: added link to the overall CustomAuthParameter doc string and updated to reflect your sentiment more

types.ts
spencer-codaio: yeah good call this was an old string. updated to be more descriptive

testing/fetcher.ts
spencer-codaio: oh yeah good call on url-encoding the secret value

testing/fetcher.ts
```diff
@@ -230,6 +231,49 @@ export class AuthenticatingFetcher implements Fetcher {
...
+        let bodyWithSubstitutions = body;
+        let formWithSubstitutions = JSON.stringify(form);
+        let headersWithSubstitutions = JSON.stringify(headers);
+
+        const {params} = this._credentials as CustomCredentials;
+
+        Object.entries(params).forEach(([key, value]) => {
+          if (urlWithSubstitutions) {
+            urlWithSubstitutions = urlWithSubstitutions.replaceAll(
+              `{{${key}-${this._invocationToken}}}`,
```
jonathan-codaio, line 245: I would make a helper method for this, just to eliminate the risk of a typo in one but not all of these, that would be very hard to catch.

types.ts
```diff
@@ -413,6 +418,74 @@ export interface WebBasicAuthentication extends BaseAuthentication {
...
+ * user or system authentication). When constructing a network request, you may indicate where these values should
+ * be inserted by our fetcher service using the syntax described below (similar to templating engines).
+ *
+ * To insert the credentials, simply put `{{<paramName>-<invocationToken>}}` as a string anywhere in your request,
+ * where `<paramName>` is the name of the parameter defined in the params mapping and `<invocationToken>` is the
+ * secret invocation-specific token provided within the {@link ExecutionContext}. The invocation
+ * token is required for security reasons.
+ *
+ * @example
+ * // Suppose you're using an API that requires a secret id in the request URL,
```
jonathan-codaio, line 452: Would just verify how this renders in the generated TypeDoc. I don't know what non-code looks like below an @example tag, if it looks code you could maybe remove the `//` prefixes here, or if it doesn't render well, could put the `//` comments inside the triple backticks to make it part of the code (that was my original intention, but if it doesn't need to be part of the code, it probably doesn't need comment markers at all).

testing/fetcher.ts
```diff
@@ -230,6 +231,49 @@ export class AuthenticatingFetcher implements Fetcher {
...
       }
+      case AuthenticationType.Custom: {
+        let urlWithSubstitutions = url;
+        let bodyWithSubstitutions = body;
+        let formWithSubstitutions = JSON.stringify(form);
+        let headersWithSubstitutions = JSON.stringify(headers);
+
+        const {params} = this._credentials as CustomCredentials;
+
+        Object.entries(params).forEach(([key, value]) => {
```
jonathan-codaio, line 242: I think we have enough test scaffolding in place that it shouldn't be awful to add a unittest case for this. Given the complexity here, worth having the coverage.

jonathan-codaio (commented):
Cool

types.ts
spencer-codaio: oh good call-out -- i had verified via the vscode generated comment but will validate on typedoc

testing/fetcher.ts
spencer-codaio: done. Just for my info, is this just a testing file that is supposed to mimic what we do in the `coda` repo?

chris-codaio:
Can we make sure we sniff the response from the server and remove any copies of the auth material in the body & headers?  That way we protect against reflection attacks to get the auth data.  Would love to see a threat model on this.

jonathan-codaio:
> Can we make sure we sniff the response from the server and remove any copies of the auth material in the body & headers? That way we protect against reflection attacks to get the auth data. Would love to see a threat model on this.

There isn't actually any new functionality here, we're just taking what was already done for WebBasic and making a separate auth type enum for it so that's clearer.

testing/fetcher.ts
jonathan-codaio: Yeah this just mimics what's done on the `coda` side. It's used by the CLI simulator, so if you run `coda execute path.ts --fetch` it'll use this code when making fetcher requests from the CLI.

It's annoying that it's grown to have so much code duplication, but the alternative would be try to abstract this auth handling and centralize it all the SDK, and have the `coda` call it, but the `coda` side is different enough that it hasn't felt like it's been worth prioritizing that.

chris-codaio:
> > Can we make sure we sniff the response from the server and remove any copies of the auth material in the body & headers? That way we protect against reflection attacks to get the auth data. Would love to see a threat model on this.
> 
> There isn't actually any new functionality here, we're just taking what was already done for WebBasic and making a separate auth type enum for it so that's clearer.

Agreed that functionality already exists, but what's new is we're opening this up to third parties and could therefore be susceptible to an attack.

testing/fetcher.ts
spencer-codaio: ahh gotcha ok so it does get exposed to the user tho in some form. thanks for the context!

jonathan-codaio:
> > > Can we make sure we sniff the response from the server and remove any copies of the auth material in the body & headers? That way we protect against reflection attacks to get the auth data. Would love to see a threat model on this.
> > 
> > 
> > There isn't actually any new functionality here, we're just taking what was already done for WebBasic and making a separate auth type enum for it so that's clearer.
> 
> Agreed that functionality already exists, but what's new is we're opening this up to third parties and could therefore be susceptible to an attack.

That makes sense.

We already sanitize every response header except for a small list of well-known headers from a runtime config, for all authentication types. For this template replacement thing, we sanitize the body before we ever log it: https://github.com/coda/coda/blob/main/modules/server/calc-server/private/authentication_helper.ts#L208. When we first added it, we did some threat modeling all together that led us to use the unique invocation token to protect against injection attacks.

The remaining risk would be if services echo back secrets in response bodies. I can't think of a great reason why any service would do this or why services that require this template replacement would be more likely than other services to do this. I suppose they could echo back the request body as part of the response. In which case, if we really want to be cautious, should we just do a find-and-replace of the secrets in the body?

chris-codaio:
> > > > Can we make sure we sniff the response from the server and remove any copies of the auth material in the body & headers? That way we protect against reflection attacks to get the auth data. Would love to see a threat model on this.
> > > 
> > > 
> > > There isn't actually any new functionality here, we're just taking what was already done for WebBasic and making a separate auth type enum for it so that's clearer.
> > 
> > 
> > Agreed that functionality already exists, but what's new is we're opening this up to third parties and could therefore be susceptible to an attack.
> 
> That makes sense.
> 
> We already sanitize every response header except for a small list of well-known headers from a runtime config, for all authentication types. For this template replacement thing, we sanitize the body before we ever log it: https://github.com/coda/coda/blob/main/modules/server/calc-server/private/authentication_helper.ts#L208. When we first added it, we did some threat modeling all together that led us to use the unique invocation token to protect against injection attacks.
> 
> The remaining risk would be if services echo back secrets in response bodies. I can't think of a great reason why any service would do this or why services that require this template replacement would be more likely than other services to do this. I suppose they could echo back the request body as part of the response. In which case, if we really want to be cautious, should we just do a find-and-replace of the secrets in the body?

Yeah, that sounds perfect (and should be pretty trivial to add I think?).  If this ends up being painful to build or maintain, then may not be worth it in the end.

jonathan-codaio:
Works for me. @spencer-codaio for this auth type, want to just do a find-and-replace for each of the secret values on the body before returning/parsing it?

spencer-codaio:
@jonathan-codaio yeah sure i can follow-up on this. Where would that need to happen?

types.ts
```diff
@@ -413,6 +418,76 @@ export interface WebBasicAuthentication extends BaseAuthentication {
...
+ * Authenticate for custom, non-standard API authentication schemes by inserting one or more arbitrary secret values
+ * into the request (the body, URL, headers, or form data) using template replacement.
+ *
+ * Some APIs use non-standard authentication schemes which often require secret credentials to be put in specific places
+ * in the request URL or request body. Custom authentication supports many of these cases by allowing you as the pack
+ * author to define one or more secret values that the user or you as the pack author must provide (depending on
+ * user or system authentication). When constructing a network request, you may indicate where these values should
+ * be inserted by our fetcher service using the syntax described below (similar to templating engines).
+ *
+ * {% raw %}
```
spencer-codaio, line 446: this is needed because otherwise `mkdocs` tries to templatize the `{{...}}` and errors out in rendering the markdown.

testing/fetcher.ts
spencer-codaio: done added unit tests

index.ts
```diff
@@ -11,6 +11,14 @@ export type {PackVersionDefinition} from './types';
...
 export type {SystemAuthentication} from './types';
 export type {WebBasicAuthentication} from './types';
+export type {CustomAuthentication} from './types';
+export type {CustomHeaderTokenAuthentication} from './types';
+export type {NoAuthentication} from './types';
+export type {HeaderBearerTokenAuthentication} from './types';
+export type {CodaApiBearerTokenAuthentication} from './types';
+export type {QueryParamTokenAuthentication} from './types';
+export type {MultiQueryParamTokenAuthentication} from './types';
+export type {AWSSignature4Authentication} from './types';
```
jonathan-codaio, line 21: Nit, please alphabetize.

mkdocs.yml
```diff
@@ -2,7 +2,7 @@
 # See: https://www.mkdocs.org/user-guide/configuration/
 site_name: Coda Packs SDK
 # TODO(spencer): Update with new url
-site_url: !ENV [MK_DOCS_SITE_URL, "https://coda.io/developers/temp-developer-docs/packs/latest/"]
```
jonathan-codaio, line 5: Was this intentional?

testing/auth.ts
```diff
@@ -130,6 +133,36 @@ class CredentialHandler {
     print('Credentials updated!');
   }
 
+  handleCustom(
+    paramDefs: Array<{
```
jonathan-codaio, line 137: This can be `CustomAuthParameter[]` right?

testing/auth.ts
```diff
@@ -130,6 +133,36 @@ class CredentialHandler {
...
+      name: string;
+      description: string;
+    }>,
+  ) {
+    assertCondition(this._authDef.type === AuthenticationType.Custom);
+    if (paramDefs.length === 0) {
+      printAndExit(
+        `Please define one or more entries for "params" in the defaultAuthentication section of this Pack definition.`,
```
jonathan-codaio, line 145: Nit, this can be used with systemAuthentication too, right?

testing/fetcher.ts
```diff
@@ -39,6 +40,15 @@ interface RequestError {
   error?: string;
 }
 
+function getTemplateReplacementValueForKey(key: string, invocationToken: string): string {
+  return `{{${key}-${invocationToken}}}`;
+}
+
+// Mirrors our utility replaceAll function in `coda` repo.
+function replaceAll(str: string, find: string, replace: string): string {
```
jonathan-codaio, line 48: Looks like TypeScript has a native `replaceAll`.

spencer-codaio: Yeah this was very confusing to me. I was using that before and getting no type errors, but on running the tests I was getting a `replaceAll is not a function`, so I switched to fill it like we do in `coda`. Any ideas on why that would be the case?

testing/auth.ts
spencer-codaio: oh yeah hmm this is what all of our things do right now. I can update them all to reference the authentication set in the pack definition, esp. since we swtiched over to the declarative non-manifest version for our examples

mkdocs.yml
spencer-codaio: yeah meant to comment on this but this makes the local mkdocs serving work by defaulting to null when this isn't set

testing/auth.ts
spencer-codaio: good catch

index.ts
spencer-codaio: done and separated out into the authentication types themselves like we do with the others

testing/fetcher.ts
jonathan-codaio: Looks like it was added in ES2021, so we'd probably need to change our target in tsconfig.json to use that. I don't know why VSCode and the compiler aren't complaining though, they should presumably be looking at that version.

Fine to leave this as is but I'd add a comment to remove it when we switch the target to ES2021.

testing/auth.ts
jonathan-codaio: Ah yeah that's confusing, let's just clean them all up to say something more general about the authentication configuration in the pack defintion.

jonathan-codaio (commented):
Looks like the latest changes didn't pushed up. Would also want to do the find-and-replace on the secrets in the responseBody before parsing it as well.

az-codaio:
This may be more of a UI problem, but do we give the user any indication of what they're actually authenticating to / where their credentials are going?

Looking at the example image in https://github.com/coda/coda/pull/63879, I see that we prompt the user for their Auth Token  / Auth Secret / Auth Parameter, but we don't tell them how these credentials are being used.

jonathan-codaio:
> This may be more of a larger UI discussion, but do we give the user any indication of what they're actually authenticating to / where their credentials are going?
> 
> Looking at the example image in [coda/coda#63879](https://github.com/coda/coda/pull/63879), I see that we prompt the user for their Auth Token / Auth Secret / Auth Parameter, but we don't tell them how these credentials are being used. Wondering if it's possible for a pack author to setup a dummy domain that collects credentials using this auth type.

The user would get this screen in response to installing a specific named pack, and normally it would have the icon of the pack on it (that screenshot just has our default icon). So for example when you install the Shopify pack and it asks you for an API key, it's really clear that you're providing Shopify credentials for use with a Shopify integration.

One thing we plan to do but haven't done is show and emphasize the networkDomains from the pack def, so it's clear what domain your credentials will be used on.

az-codaio:
> One thing we plan to do but haven't done is show and emphasize the networkDomains from the pack def, so it's clear what domain your credentials will be used on.

Displaying `networkDomain` is what I had in mind; something similar to how iOS will prompt the user that an app is attempting to load another site to sign in. My understanding is that the pack author controls the icon, so it doesn't seem like something users can rely on.

Simulator Screen Shot - iPhone 13 - 2021-11-09 at 14 22 22

jonathan-codaio:
> > One thing we plan to do but haven't done is show and emphasize the networkDomains from the pack def, so it's clear what domain your credentials will be used on.
> 
> Displaying `networkDomain` is what I had in mind; something similar to how iOS will prompt the user that an app is attempting to load another site to sign in. My understanding is that the pack author controls the icon, so it doesn't seem like something users can rely on.
> 
> Simulator Screen Shot - iPhone 13 - 2021-11-09 at 14 22 22

Yeah, the networkDomain is our central security feature so we do plan to emphasize it pretty widely. Haven't done it though, but feel free to hold us accountable on this if you think we should be farther along on this.

jonathan-codaio (approved):

PR #1513, opened 2021-11-04
Don't pass the codaType to the formula in the pack definition

alan-codaio:
`codaType` is a schema property, not a formula one, so for `makeFormula` it should be used to only construct the `schema`, but should be omitted from the formula definition. This was causing successive builds of Packs that had formulas with return types to fail our upload validation step.

Tested by updating devserver and manually checking the failure case.


api.ts
```diff
@@ -605,9 +605,13 @@ export function makeFormula<
     case ValueType.String: {
       // very strange ts knows that fullDefinition.codaType is StringHintTypes but doesn't know if
       // fullDefinition is StringFormulaDefV2.
-      const {onError: _, resultType: unused, ...rest} = fullDefinition as StringFormulaDefV2<ParamDefsT>;
       const codaType = 'codaType' in fullDefinition ? fullDefinition.codaType : undefined;
+      if ('codaType' in fullDefinition) {
+        delete fullDefinition.codaType;
```
jonathan-codaio, line 610: It's a little dangerous to alter a parameter (fullDefinition is the exact object the user passed) in case the user uses it in multiple calls (very unlikely, but would result in very confusing behavior).

I wonder if we can do something like:

```
      const def: StringFormulaDefV2<ParamDefsT> = {
        ...fullDefinition,
        codaType: 'codaType' in fullDefinition ? fullDefinition.codaType : undefined,
      };
      const {onError: _, resultType: unused, codaType, ...rest} = def;
```

which would extract `codaType` from `rest` while also making `codaType` a local variable for use below.

jonathan-codaio (commented):
Thanks!

I think you could probably add a simple test for this that calls `makeFormula()` with a `codaType`, and ensures that `codaType` is not return value (but is in the return value's `schema` object).

api.ts
alan-codaio: Sure, let me try that out. I guess our upload validation won't complain if `codaType` is present but undefined, we'll see if the problem is resolved with this approach.

 My reasoning around using `delete` was that both the entry name and value are removed from the object

jonathan-codaio (approved):

PR #1514, opened 2021-11-04
Create a way to store build options in the .coda-pack.json file. Use it to store the timerStrategy.

jonathan-codaio:
I tried to add this to the pack definition as discussed on Slack, but unfortunately there's a chicken-and-egg problem with that, because we can't reliably pull values out of the TS until after the bundle is built, but we need the timer strategy as an input to generate the build in the first place.

So since we already have a .coda-pack.json file per pack, I added the ability to store persistent options in that file, and read from that when doing the build.

PTAL @huayang-codaio @patrick-codaio @coda/packs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 - **Breaking Change** Column Formats must now use only real Regex objects in their `matchers` array.
 
+- If your pack uses fake timers (to simulate setTimeout) you can now store this as a persistent
+  build option. Previously, you had to remember to include the flag --timerStrategy=fake any time
+  you used any of the `coda` CLI commands. Now you can run
+  `coda setOption path/to/pack.ts timerStrategy fake` and it will store the option persistently
+  in your `.coda-pack.json` file and use it for all builds.
+
 ## 0.5.0
 
 - **Breaking Change** `context.logger` has been removed. It has been redundant with `console.log`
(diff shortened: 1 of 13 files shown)


cli/set_option.ts
```diff
@@ -0,0 +1,40 @@
...
+
+function validateOption(option: string, value: string): PackOptions {
+  const validOptions = Object.values(PackOptionKey) as string[];
+  if (!validOptions.includes(option)) {
+    return printAndExit(`Unsupported option "${option}". Value options are: ${validOptions.join(', ')}`);
+  }
+  const key = option as PackOptionKey;
+  switch (key) {
+    case PackOptionKey.timerStrategy:
+      const validValues = Object.values(TimerShimStrategy) as string[];
```
patrick-codaio, line 32: totally optional, but with your pattern, I think you could do this in a generic way outside the switch: `const validValues = Object.values(PackOptions[key]) as string[]`. Even your printAndExit is nicely generic.

patrick-codaio (approved):

cli/set_option.ts
jonathan-codaio: Hmm good thought, not sure it works though since `PackOptions[key]` isn't a runtime list of valid values, but merely a TypeScript type.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 - **Breaking Change** Column Formats must now use only real Regex objects in their `matchers` array.
 
+- If your pack uses fake timers (to simulate setTimeout) you can now store this as a persistent
+  build option. Previously, you had to remember to include the flag --timerStrategy=fake any time
+  you used any of the `coda` CLI commands. Now you can run
+  `coda setOption path/to/pack.ts timerStrategy fake` and it will store the option persistently
```
huayang-codaio, line 8: cool!

PR #1515, opened 2021-11-04
Switch to iife format for pack and thunk bundles

huayang-codaio:
# Context

We used to manually add a closure when throwing a bundle into IVM context. 

```
  const script = await isolate.compileScript(
    `(() => { ${stubName} = (() => { ${bundle} \n return exports; })(); })()`,
    {
      filename: path,
    },
  );
  await script.run(context);
```

## why manual closure
The reason of doing so is that isolated-vm evaluation has a different behavior from loading / eval / requiring a module in NodeJS. Imagine that we have a simple module like this:

```
-> % cat test.js
'use strict';
function hello() {
  console.log('hello world');
}
module.exports = hello;
```

When the code is `eval` or required in node, no symbol leaks to global. 

```
> require('./test')
[Function: hello]
> global.hello
undefined
> hello
Uncaught ReferenceError: hello is not defined
> module.hello
undefined

> const code = fs.readFileSync('./test.js'); eval(code.toString());
[Function: hello]
> hello
Uncaught ReferenceError: hello is not defined
> global.hello
undefined
> module.hello
undefined
```

This behavior is however different in isolated-vm. if we do 

```
  const script = await isolate.compileScript(testJsBundle, {filename: path});
  await script.run(context);
```

`hello` will leak to `global.hello`. `isolated-vm` does support something like `evalClosure` which will bound local symbols. However  `evalClosure` doesn't allow us specify the filename of the code so that we can't map stack back with source map. (note that [the documentation](https://github.com/laverdet/isolated-vm#contextevalclosureignoredcode-arguments-options) says evalClosure supports scriptOrigin but the code doesn't actually do it. as far as I can tell after looking into the cplus implementation of [isolate_handle.cc](https://github.com/laverdet/isolated-vm/blob/d5b4948250f9c011c3e5a63ea3287ce72bdff2ed/src/module/isolate_handle.cc) used for script and [context_handle.cc](https://github.com/laverdet/isolated-vm/blob/27cb89d3c3c4ef433af22ac8efa38a26f1caabda/src/module/context_handle.cc) used for eval). This limits our option to only `compileScript`.

## why does it break source map.

`(() => { ${stubName} = (() => { ${bundle} \n return exports; })(); })()` isn't compiled with sourcemap. it shifts the offset a few characters away. this is crucial to minified bundles. 

## what's IIFE & CJS

checkout https://betterprogramming.pub/what-are-cjs-amd-umd-esm-system-and-iife-3633a112db62 to see what the difference is between IIFE & CJS

in short, IIFE generates something like 

```
module.exports = (() => { <actual code> })();
```

# What does this PR do

In this PR does 3 things:
- generate thunk bundle and pack bundle in IIFE format (these two bundles are loaded into IVM)
- bump pack-sdk version to 0.7.0 (see below)
- add an option `requiresManualClosure` that allows us to avoid manual closure if the pack bundle is compiled with packs-sdk after 0.7.0. see https://github.com/coda/coda/pull/63913 for the coda repo changes to make this possible. 

# Testing

Assuming that I'll bump this in the coda repo, there are a few things that we need to test.

- [x] legacy pack bundles still run without a problem.
- [x] new pack bundles will run with correct stack trace. 
- [x] new pack bundles will run on old runtime. (this isn't a big deal though)
- [x] test with devapp. (see https://github.com/coda/coda/pull/63913)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 0.7.0
-- Pack bundle format is changed to IIFE. Previously compiled bundles will still be supported.
+- Pack bundle format is changed to IIFE to fix occasional stacktrace misinterpretation. Previously compiled bundles will still be supported but may suffer from inaccurate sourcemap issue until the pack is built with the new SDK.
 
 ## 0.6.0
 - **Breaking Change** Column Formats must now use only real Regex objects in their `matchers` array.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 0.7.0
-- **Breaking Change** Pack bundle format is changed to IIFE. Previously compiled bundles will still be supported for a short period of time.
+- Pack bundle format is changed to IIFE. Previously compiled bundles will still be supported.
 
 ## 0.6.0
 - **Breaking Change** Column Formats must now use only real Regex objects in their `matchers` array.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
-## 0.6.0
+## 0.7.0
+- **Breaking Change** Pack bundle format is changed to IIFE. Previously compiled bundles will still be supported for a short period of time.
 
+## 0.6.0
 - **Breaking Change** Column Formats must now use only real Regex objects in their `matchers` array.
 
 ## 0.5.0
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ dist/test
 yarn-error.log
 local-docs
 site
+.coda-pack.json
--- a/Makefile
+++ b/Makefile
@@ -85,17 +85,27 @@ compile-isolated-vm:
 		else echo "isolated-vm version matches, skipping."; \
 	fi
 
-.PHONY: compile
-compile:
-	${ROOTDIR}/node_modules/.bin/tsc
+.PHONY: compile-thunk
+compile-thunk:
+	# this bundle is loaded into ivm, better to use iife to avoid local symbols leak to global.
 	${ROOTDIR}/node_modules/.bin/esbuild ${ROOTDIR}/runtime/thunk/thunk.ts \
 		--bundle \
 		--outfile=${ROOTDIR}/bundles/thunk_bundle.js \
 		--inject:${ROOTDIR}/testing/injections/buffer_shim.js \
-		--format=cjs \
+		--format=iife \
+		--global-name=module.exports \
 		--banner:js="'use strict';"
+
+.PHONY: compile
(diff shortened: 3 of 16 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-## 0.6.0
+## 0.7.0
+- **Breaking Change** Pack bundle format is changed to IIFE. Previously compiled bundles will still be supported for a short period of time.
```
jonathan-codaio, line 2: What change would we make that would break previously compiled bundles? Can we support them forever, it's just that they'll continue to have broken source maps?

Otherwise we have to update every first-party pack again, and ask all beta users to update their packs.

jonathan-codaio (commented):
Nice work on this thorny problem.

Just for my knowledge, what kinds of issues were we seeing with source maps? Was this a long-standing issue with entirely broken source maps? Or just occasional issues? Or did things break when you added the extra closure recently?

huayang-codaio:
> Just for my knowledge, what kinds of issues were we seeing with source maps? Was this a long-standing issue with entirely broken source maps? Or just occasional issues? Or did things break when you added the extra closure recently?

oh, it's an occasional issue. since the code is shift a few characters, the source mapping sometimes will end up with missing or incorrect information. when the code is not minified (e.g. web editor), the source mapping works most of the time. When the bundle is minified (e.g. CLI packs), some frames might be missing or incorrect. Sometimes it's not obvious. For example, the stack below is from staging.

```
{
  category: 'Internal',
  type: 'Unknown error',
  message: "Cannot read property 'shipment' of undefined",
  details: 'Runtime error',
  stack: "TypeError: Cannot read property 'shipment' of undefined\n" +
    '    at parseTrackingInfoFromTrackResponse (packs/ups/main/helpers.ts:16:37)\n' +
    '    at <unknown> (packs/ups/main/formulas.ts:22:14)\n' +
    '    at Generator.next (<anonymous>:null:null)\n' +
    '    at fulfilled (null:null:null)\n'
}
```

This line `at <unknown> (packs/ups/main/formulas.ts:22:14)` is misinterpreted due to the sourcemap offset.

CHANGELOG.md
huayang-codaio: we can probably support them forever. maybe sometime we'll be running on pack bundles with sdk <0.7.0. We can remove the support then. 

with that begin said, maybe I'll remove the "breaking change" tag here to avoid confusion.

jonathan-codaio (approved):
Assuming aggressive testing, LGTM :)

huayang-codaio:
devapp tested see https://github.com/coda/coda/pull/63913.

PR #1517, opened 2021-11-05
Support numbers with simple autocomplete.

jonathan-codaio:


ekoleda-codaio (approved):
Thanks for the quick response!

PR #1519, opened 2021-11-05
TSDoc for ExecutionContext and friends.

jonathan-codaio:

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -351,25 +351,81 @@ export interface Network {
 const ValidFetchMethods = ['GET', 'PATCH', 'POST', 'PUT', 'DELETE'] as const;
 export type FetchMethodType = typeof ValidFetchMethods[number];
 
-// Copied from https://developer.mozilla.org/en-US/docs/Web/API/Request
+/**
+ * An HTTP request used with the {@link Fetcher}.
+ *
+ * The structure largely follows https://developer.mozilla.org/en-US/docs/Web/API/Request
+ */
 export interface FetchRequest {
+  /** The HTTP method/verb (e.g. GET or POST). */
   method: FetchMethodType;
+  /**
+   * The URL to connect to. This is typically an absolute URL, but if your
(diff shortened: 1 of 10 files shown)


api_types.ts
```diff
@@ -351,40 +351,168 @@ export interface Network {
...
   status: number;
+  /**
+   * The body of the response.
+   *
+   * If the response contains JSON data, either because the Content-Type header is application/json
+   * or if the data is JSON-parseable, this will be a parsed JavaScript object.
+   * Similarly, if the response headers are text/xml or application/xml, this will be a parsed
+   * JavaScript object using the `xml2js` library.
+   *
+   * If implicit parsing is undesirable, you may use {@link isBinaryResponse} on the request
```
ekoleda-codaio, line 422: Should you mention that it returns a Node.js Buffer in this case?

api_types.ts
```diff
@@ -351,40 +351,168 @@ export interface Network {
...
+ * You will be returned a URL that you can then return with your formula response, and
+ * Coda will ingest the data from that URL into permanent storage.
+ */
 export interface TemporaryBlobStorage {
+  /**
+   * Fetches the data at the given URL, applying user authentication credentials as appropriate,
+   * and stores it in Coda-hosted temporary storage. Returns a URL for the temporary file
+   * that you should return in your formula response.
+   *
+   * The URL expires after 15 minutes by default, but you may pass a custom expiry, however
```
ekoleda-codaio, line 482: Is it really 15 minutes? In my tests a few weeks back it seemed to be closer to 15 seconds.

api_types.ts
jonathan-codaio: Yep good point, updated.

api_types.ts
jonathan-codaio: As best I can tell it's 15 minutes: https://github.com/coda/coda/blob/main/modules/server/doc-bulk-storage/document_bulk_storage.ts#L567

ekoleda-codaio (approved):

PR #1522, opened 2021-11-09
Cleanup AWS authentication, add support for assuming IAM role

chris-codaio:
I'd love to be able to use AssumeRole for authentication of packs we need internally - much better than having to manage credential rotation manually on an ongoing basis.

Thoughts on this change?  It is technically a breaking change, but I don't think anyone is using this auth type currently?  I'm happy to do the work to merge this in elsewhere and build support for the assume role auth type.

PTAL @jonathan-codaio @ekoleda-codaio @huayang-codaio


testing/auth.ts
```diff
@@ -68,7 +68,8 @@ export function setupAuth(manifestDir: string, packDef: BasicPackDefinition, opt
     case AuthenticationType.OAuth2:
       ensureExists(packDef.defaultAuthentication, 'OAuth2 only works with defaultAuthentication, not system auth.');
       return handler.handleOAuth2();
-    case AuthenticationType.AWSSignature4:
+    case AuthenticationType.AWSAccessKey:
+    case AuthenticationType.AWSAssumeRole:
```
jonathan-codaio, line 72: How hard would it be to implement support in the simulator for these? Trying to avoid creating new tech debt since we have a truckload of it already.

jonathan-codaio (approved):
Yeah to my  knowledge nothing is using this today, so fine to do the breaking change.

Minding adding a line to CHANGELOG.md though noting the change and the new auth type?

testing/auth.ts
chris-codaio: Access Key shouldn't be too much work.  Assume Role is tougher as it requires the caller to already be authenticated to AWS using a principal that the role to be assumed trusts - but we could build it with that assumption.

chris-codaio: Started working on this and it bloated quite a bit.  I'll plan to tackle the fetcher updates mostly as a followup PR.

PR #1534, opened 2021-11-09
make type optional for makeObjectSchema

huayang-codaio:
since makeObjectSchema would only accept object type anyway.

https://golinks.io/bug/19692

@coda/packs


jonathan-codaio (approved):
🎉

PR #1535, opened 2021-11-10
More TSDoc

jonathan-codaio:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
 ## 0.7.1
+
 - Update internal authentication mechanisms for interacting with AWS. Not currently available externally.
+- `makeObjectSchema` no longer requires you to redundantly specify `type: ValueType.Object` in your schema definition.
 
 ## 0.7.0
+
 - Pack bundle format is changed to IIFE to fix occasional stacktrace misinterpretation. Previously compiled bundles will still be supported but may suffer from inaccurate sourcemap issue until the pack is built with the new SDK.
 
 ## 0.6.0
+
 - **Breaking Change** Column Formats must now use only real Regex objects in their `matchers` array.
 
 - If your pack uses fake timers (to simulate setTimeout) you can now store this as a persistent
--- a/api.ts
+++ b/api.ts
@@ -397,6 +397,9 @@ export interface PackFormulas {
   readonly [namespace: string]: Formula[];
 }
 
+/**
+ * Base type for the inputs for creating a pack formula.
+ */
 export interface PackFormulaDef<ParamsT extends ParamDefs, ResultT extends PackFormulaResult>
   extends CommonPackFormulaDef<ParamsT> {
   execute(params: ParamValues<ParamsT>, context: ExecutionContext): Promise<ResultT> | ResultT;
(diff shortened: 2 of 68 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
 ## 0.7.1
+
```
huayang-codaio, line 2: oh, just for my understanding, is the space for formatting or for npm to parse release logs properly?

api.ts
```diff
@@ -488,11 +509,26 @@ export function isSyncPackFormula(fn: BaseFormula<ParamDefs, any>): fn is Generi
   return Boolean((fn as GenericSyncFormula).isSyncFormula);
 }
 
+/**
+ * The return value from the formula that implements a sync table. Each sync formula invocation
+ * returns one reasonable size page of results. The formula may also return a continuation, indicating
+ * that the sync formula should be invoked again to get a next page of results. Sync functions
+ * are called in a loop until there is no continuation returned.
```
huayang-codaio, line 516: not sure if `in a loop` could be made more clear.

api.ts
```diff
@@ -779,7 +845,33 @@ export interface MetadataFormulaObjectResultType {
...
+ * share the same structure, so all of these supporting features are defined as `MetadataFormulas`.
+ * You typically do not need to define a `MetadataFormula` explicitly, but rather can simply define
+ * the JavaScript function that implements the formula. Coda will wrap this function with the necessary
+ * formula boilerplate to make it look like a complete Coda formula.
+ *
+ * All metadata functions are passed an {@link ExecutionContext} as the first parameter,
+ * and the optional second parameter is a string whose purpose and value varies depending on
+ * the use case. For example, a metadata formula that implements parameter autocomplete will
+ * be passed the user's current search if the user has started typing to search for a result.
+ * Not all metadata formulas make use of this second parameter.
```
ekoleda-codaio, line 873: Doesn't `autocomplete` have a 3rd parameter as well, for the previous parameter values? Not sure you need to mention it here, but for completeness you might.

huayang-codaio (approved):

api.ts
```diff
@@ -1162,6 +1301,25 @@ export function makeSyncTableLegacy<
   return makeSyncTable({name, identityName: '', schema, formula, connectionRequirement, dynamicOptions});
 }
 
+/**
+ * Creates a dynamic sync table definition.
+ *
+ * @example
+ * ```
+ * coda.makeDynamicSyncTable({
```
ekoleda-codaio, line 1309: In a world where we promote the `pack.addThing()` paradigm over the manifest, is there any utility for the `makeThing()` functions?

schema.ts
```diff
@@ -696,6 +703,16 @@ export type SchemaType<T extends Schema> = T extends BooleanSchema
...
 
+/**
+ * Utility that examines a JavaScript value and attempts to infer a schema definition
+ * that describes it.
+ *
+ * It is vastly preferable to define a schema manually. A clear and accurate schema is one of the
+ * fundamentals of a good pack. However, for data that is truly dynamic for which a schema can't
+ * be known in advance nor can a function be written to generate a dynamic schema from other
+ * inputs, it may be useful to us this helper to sniff the return value and generate a basic
+ * inferred schema from it.
```
ekoleda-codaio, line 714: Are there any limitations on the object passed in? Can it contain arrays of sub-objects, for instance? How does select the ID and primary?

schema.ts
```diff
@@ -726,12 +743,54 @@ export function generateSchema(obj: ValidTypes): Schema {
...
+ * ```
+ */
 export function makeSchema<T extends Schema>(schema: T): T {
   return schema;
 }
 
 export const PlaceholderIdentityPackId = -1;
 
+/**
+ * A wrapper for creating a schema definition for an object value.
```
ekoleda-codaio, line 775: Other than the new ability to skip `type: coda.ValueType.Object`, does this provide any utility beyond `makeSchema`?

CHANGELOG.md
jonathan-codaio: Oh I don't think it matters, my VS Code autoformatter just added these.

api.ts
jonathan-codaio: Sure, changed it to "repeatedly", but happy to adjust the language as desired.

api.ts
jonathan-codaio: Yep you're right, added a section about that.

api.ts
jonathan-codaio: No and yes. If people exclusively use the builder, then we don't need these. If people still want to have an out so that they don't use the builder, we would still want these, they provide some boilerplate and default values that are useful. Right now the builder methods just call these. I think that's fine if that continues to be the case, there aren't meant to be too many of these, just one per building block.

schema.ts
jonathan-codaio: I don't think there's any limitations, it works recursively so should handle nested objects just fine.

It doesn't infer an `id` or `primary`, so I added a sentence about that.

schema.ts
jonathan-codaio: It does some validation at import time, and because it's restricted to objects, the TypeScript inference is strong, TS will enforce that you're providing an object schema definition rather than just any valid schema definition. That would still work with `makeSchema` if you included `type: ValueType.Object`, but people seemed to find value in eliminating that boilerplate. But, I'm not opposed to consolidating these.

huayang-codaio:
Thanks!

ekoleda-codaio (approved):
Thanks!

PR #1557, opened 2021-11-16
Fix CustomAuthentication typedoc

ekoleda-codaio:
Without the `{% raw %}` directives the surrounding `{{`'s were getting lost. I also reformatted the code to be max 80 chars, so it fits better on the web page.

Staged: https://coda-packs-sdk--1557.com.readthedocs.build/en/1557/reference/sdk/interfaces/CustomAuthentication/


jonathan-codaio (approved):
Thanks! Just noticed this as well.

PR #1558, opened 2021-11-16
Flesh out formulas guide.

ekoleda-codaio:
Also added some content to the fetcher guide on rate limits.

Staged:
- https://coda-packs-sdk--1558.com.readthedocs.build/en/1558/guides/blocks/formulas/
- https://coda-packs-sdk--1558.com.readthedocs.build/en/1558/guides/advanced/fetcher/#rate-limits


docs/guides/blocks/formulas.md
```diff
@@ -11,7 +11,9 @@ Formulas are one of the most basic building blocks in Coda; used to calculate va
 
 ## Structure of a formula
 
-TODO
+A formula definition consists of a set of key-value pairs, which specify the various settings for the formula. Most of these settings are metadata, such as the name, description, parameters and result type. The actual code that is run each time the formula recalculated is specified using in the `execute` key.
+
+<img src="../../../images/formula_structure.png" srcset="../../../images/formula_structure_2x.png 2x" class="screenshot" alt="Mapping of formula coda to formula editor.">
```
jonathan-codaio, line 16: GH won't let me comment on line 21 but existing code there has an extra "the" before "letters".

docs/guides/blocks/formulas.md
```diff
@@ -30,20 +32,74 @@ Formulas can accept parameters, which is the primary way for them to access data
 
 ## Results
 
-TODO
+All formulas must return a result, which is a single value matching the type specified in the `resultType` property of the formula definition. See the [Data types guide][data-types] for more information on the type of values that can be returned.
+
+
+## Authentication
+
+Formulas can use [authentication][authentication] to fetch private data. When using [system authentication][system_auth] there is no change to the user experience, but when you use [user authentication][user_auth] the formula editor will be automatically updated to prompt for a connected account. The connected account will appear as the first parameter of the formula, but it's value is not passed to your formula's `execute` method.
```
jonathan-codaio, line 40: Nit, "its" in "it's value"

docs/guides/blocks/formulas.md
```diff
@@ -30,20 +32,74 @@ Formulas can accept parameters, which is the primary way for them to access data
...
+      // ...
+      execute: async function ([url], context) {
+        // ...
+      },
+    });
+    ```
+
+By default if a Pack defines an authentication method then all formulas will require a connected account. You can change this behavior using the [`connectionRequirement`][connectionRequirement] property of the formula definition. Setting it to `Optional` makes the user account optional, and `None` removes the account parameter completely.
+
+While it's possible to use multiple accounts within a document, each instance of the formula must use a specific account. It isn't possible to have a formula that uses the account of whichever user is currently viewing the document. It's important for collaboration that all users in a Coda doc have access to the same data.
```
jonathan-codaio, line 71: This last part isn't actually true, it depends on whether the formula is an action or not. If a formula has `isAction: true` you can call it with "User's private account".

docs/guides/blocks/formulas.md
```diff
@@ -30,20 +32,74 @@ Formulas can accept parameters, which is the primary way for them to access data
...
 
 
 ## Caching
 
-TODO
+For performance reasons all formula results are cached by default. If your formula is called again with the same parameters, the result will be loaded from the cache instead of re-running your code. Building or releasing a new version of your Pack will invalidate this cache, ensuring you get fresh results using your new code.
```
jonathan-codaio, line 76: Again, only non-action formulas are cached by default.

Also, "the result will be loaded from the cache" makes it seem like it's guaranteed that subsequent invocations will get loaded from the cache, but they only _might_ be loaded from the cache, if it's still within the TTL and the cache hasn't grown too large.

docs/guides/blocks/formulas.md
```diff
@@ -30,20 +32,74 @@ Formulas can accept parameters, which is the primary way for them to access data
...
 
 
 ## Recalculation
 
-TODO
+Coda has complex logic that determines when formula results need to be recalculated. Formulas will always recalculate when the parameter values change, but it will also happen periodically for other reasons including routine maintenance. In general you shouldn't make any assumptions about how often your formula will be run. If running your formula is expensive in some way (API cost, etc) make sure to use [caching](#caching) and [rate limits][fetcher_rate_limits] to limit the impact of recalculation.
+
+Building or releasing a new version of your Pack doesn't automatically cause existing formulas to recalculate, so users may still see old results for a while. In the **Settings** tab of the Pack's side panel there is a "Refresh now" button (**⟳**) that allows users to recalculate all formulas using the Pack. Pack makers also have access to the **Auto-refresh when version changes** toggle, which will do this automatically for the current window for each new Pack version.
+
+<img src="../../../images/settings_recalc.png" srcset="../../../images/settings_recalc_2x.png 2x" class="screenshot" alt="Recalculation settings in the Pack side panel">
```
jonathan-codaio, line 90: Just a reminder that these settings will all move into the new debug drawer in a week or two so we'll want to replace the screenshot at that time.

docs/guides/advanced/fetcher.md
```diff
@@ -303,7 +303,7 @@ try {
 
 ## Authentication
 
-The authentication you configure for your Pack is automatically applied to fetcher requests, with no extra code needed. For example, if you have setup `HeaderBearer` authentication, an `Authorization` header with the user's token will be automatically added to your fetcher requests. This is only done for formulas that use a connected account: those that have a `connectionRequirement` of `REQUIRED`, or `OPTIONAL` and the user opted to selected an account.
+The [authentication][authentication] you configure for your Pack is automatically applied to fetcher requests, with no extra code needed. For example, if you have setup `HeaderBearer` authentication, an `Authorization` header with the user's token will be automatically added to your fetcher requests. This is only done for formulas that use a connected account: those that have a `connectionRequirement` of `REQUIRED`, or `OPTIONAL` and the user opted to selected an account.
```
jonathan-codaio, line 306: Total nit, I generally think of "setup" as a noun and "set up" as a verb.

docs/guides/blocks/formulas.md
ekoleda-codaio: Good catch, fixed.

docs/guides/blocks/formulas.md
ekoleda-codaio: You can tell by now this is my weak spot 😄.  Fixed.

docs/guides/blocks/formulas.md
ekoleda-codaio: Ya, my thought with this guide was for it to describe the behavior of "regular" formulas (not actions or syncs), and then in the actions guide to describe the ways in which action formulas differ from regular formulas. I felt like it was easier to be definitive in this guide vs. having to qualify "except in this case" for the other formula types. I've added a note at the top of this guide to make that distinction clearer, but let me know if you think it would be better to include the caveats throughout.

docs/guides/blocks/formulas.md
ekoleda-codaio: Ditto on actions.

Adjusted the wording to indicate that it attempts to load from the cache.

docs/guides/blocks/formulas.md
ekoleda-codaio: Ah, good to know. I'll keep an eye out.

docs/guides/advanced/fetcher.md
ekoleda-codaio: Good catch, fixed.

docs/guides/blocks/formulas.md
jonathan-codaio: Ok let's see how it goes. Can add the caveats in later if users are missing the distinction.

jonathan-codaio (approved):

PR #1559, opened 2021-11-16
Add exports and documentation for some of the entities in the typedoc build warnings.

jonathan-codaio:
Doesn't look like my plan is going to work to create a synthetic entrypoint for getting these previously unreachable entities into the TypeDoc documentation. It looks like when you specify more than one entrypoint, TypeDoc completely changes its output, and prefixes each markdown file with the name of the entry point, e.g. index.PackDefinitionBuilder, which is ugly and breaks any existing links. It also creates module pages for each entrypoint. I didn't immediately see a way to affect this behavior.

So I just started exporting some of the intermediate types, there aren't a ton so far. There's still a bunch we're warning about, my hope was to get rid of all the warnings but maybe we'll just leave some, and others that it's warning about like `Quota` are things we want to delete from the SDK entirely.


ekoleda-codaio (approved):

ekoleda-codaio:
> There's still a bunch we're warning about, my hope was to get rid of all the warnings but maybe we'll just leave some, and others that it's warning about like `Quota` are things we want to delete from the SDK entirely.

There is a typedoc config field you can use to suppress the warnings for types that are intentionally excluded: https://typedoc.org/guides/options/#intentionallynotexported

jonathan-codaio:
> > There's still a bunch we're warning about, my hope was to get rid of all the warnings but maybe we'll just leave some, and others that it's warning about like `Quota` are things we want to delete from the SDK entirely.
> 
> There is a typedoc config field you can use to suppress the warnings for types that are intentionally excluded: https://typedoc.org/guides/options/#intentionallynotexported

Ah great! I'll use that in a future PR, then we can enforce no warnings either.

PR #1561, opened 2021-11-17
add automated flow for publishing to npm when push to release

spencer-codaio:
any concerns with this approach? wanted to get this in to test with releasing the new version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,7 +180,7 @@ jobs:
           command: 'make publish-docs-prod FLAGS=--forceUpload'
       # TODO: add a probe step here to validate?
   
-  publish-npm:
+  publish-to-npm:
     executor: default_build_environment
     working_directory: ~/repo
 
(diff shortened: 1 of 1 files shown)


jonathan-codaio (commented):
Ah, so the idea is that when we update the `release` branch it'll update the docs and npm together? That seems reasonable.

The snag is authenticating with npm. What's the role user that would do this, and are credentials set up for that? Probably solvable problems, except currently our npm settings require an OTP code for publishing a new package. We could talk about relaxing that, or see if there's a way to provide that in the approval step.

spencer-codaio:
yep! thats the idea so that they're kept in sync.

ahh okay will keep this on the burner then and can see if i figure out a way after doing the manual release. OTP seems difficult to handle...

jonathan-codaio:
> yep! thats the idea so that they're kept in sync.
> 
> ahh okay will keep this on the burner then and can see if i figure out a way after doing the manual release. OTP seems difficult to handle...

We can choose to relax it, just have to put some thought into what the risks are.

Though given that this is an open-source repo, we may want a sanity-check on what gets published to npm rather than blindly auto publishing the contents of the `release` branch in case somebody finds a way to attack that.

spencer-codaio:
hmm yeah seems like more overhead than worth right now -- maybe after deploy it just sends you a slack message reminding you to publish to NPM or something like that. Will close for now

PR #1565, opened 2021-11-17
making format formulaNamespace optional

huayang-codaio:
This will make format easier to use before we are able to get rid of `formulaNamespace` completely. 

after:

image

### DIFF
--- a/dist/bundle.d.ts
+++ b/dist/bundle.d.ts
@@ -2781,7 +2781,7 @@ export interface Format {
 	 */
 	name: string;
 	/** @deprecated Namespaces are being removed from the product. */
-	formulaNamespace: string;
+	formulaNamespace?: string;
 	/**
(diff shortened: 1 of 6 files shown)


jonathan-codaio (commented):
LG, but mind sending the corresponding `coda` change and running consistency on it, so we can gauge the effects first? Just want to avoid people being unable to bump the SDK without running into those changes.

alan-codaio:
IIRC nothing should be amiss until we actually remove the property here. But no harm in running consistency just in case!

huayang-codaio:
I am testing the change with a consistency check. If this breaks anything, I feel like it will likely happen once a pack actually uses a format without specifying formulaNamespace. This won't be caught by the consistency check.

huayang-codaio:
https://github.com/coda/coda/pull/64482 consistency run passes.

jonathan-codaio (approved):
🎉

PR #1569, opened 2021-11-18
add empty defaults for pack executions

spencer-codaio:
ran into an `ensureExists` error in a pack where i had to dig into the dist for packs-sdk to figure out what was wrong. Providing some reasonable empty defaults to catch cases where typing doesn't cover the protection since we don't do strict typing so users don't get an arcane error.

also side note is that this was only in the `Internal` logs and everything else in logs said it was succesful, except the actual execution was not -- wondering how we can audit to make sure that doesn't happen
image

I tested this with the formulas in `fake.ts` not sure if i should add unit tests somewhere too

PTAL @huayang-codaio cc @jonathan-codaio  @coda/packs


api.ts
```diff
@@ -1356,7 +1359,7 @@ export function makeSyncTable<
     const {result, continuation} = await wrappedExecute(params, context);
```
huayang-codaio, line 1359: this errors out for me as well. if undefined is returned, the unpack will throw.

spencer-codaio: ahh ok will add -- how did you hit that case in the `fake.ts` testing?

huayang-codaio: i changed Students sync table to return undefined in the default branch. and then 

```
./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake.ts Students random-string
```

huayang-codaio (approved):

api.ts
spencer-codaio: ah cool that makes sense

PR #1570, opened 2021-11-18
Add documentation for RequestHandlerTemplate and ResponseHandlerTemplate.

jonathan-codaio:
This the very last part of the SDK that I think needs reference documentation.

This is a bit of a dumpster fire of a feature though, in terms of its understandibility, consistency, and completeness. I'm also complete fine saying that we hide this and say it's unsupported until we decide to invest in it more.

These allow you to define formulas for common fetcher tasks without writing code. It looks like it was part of the original packs code from about 3 years ago and I haven't personally worked with it before, I had to teach myself how this seems to work while writing this. We appear to use it for some early packs like Figma and GitHub.

The aspiration to allow people to write packs declaratively is reasonable and it's long been a goal of Coda to let people write packs who don't know how to code, but since we're code-based for now it would be reasonable to say that this current legacy offering isn't worth putting out there yet.

### DIFF
--- a/handler_templates.ts
+++ b/handler_templates.ts
@@ -35,7 +35,7 @@ import {withQueryParams} from './helpers/url';
  *         return value.toLowerCase();
  *       },
  *     },
- *     queryParams: ["id", "format"],
+ *     queryParams: ["format"],
  *   },
  * });
  * ```
--- a/handler_templates.ts
+++ b/handler_templates.ts
@@ -13,22 +13,135 @@ import {isArray} from './schema';
 import {isObject} from './schema';
 import {withQueryParams} from './helpers/url';
 
-type ParamMapper<T> = (val: T) => T;
-
+/**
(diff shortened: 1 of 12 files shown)


handler_templates.ts
```diff
@@ -13,22 +13,135 @@ import {isArray} from './schema';
...
+ *   request: {
+ *     method: "GET",
+ *     url: "https://example.com/api/widgets/{id}",
+ *     nameMapping: {outputFormat: "format"},
+ *     transforms: {
+ *       format: function(value) {
+ *         return value.toLowerCase();
+ *       },
+ *     },
+ *     queryParams: ["id", "format"],
```
ekoleda-codaio, line 38: Should "id" be in this list, given that it's used in the URL and not passed as a query param?

ekoleda-codaio (approved):

handler_templates.ts
jonathan-codaio: Ah yep good catch.

PR #1571, opened 2021-11-19
add the missing schema description in normalization

huayang-codaio:
https://golinks.io/pak/454
[bug-19930](https://golinks.io/bug/19930)

Turns out that the schema description is dropped in the schema normalization. This means that they are missing in uploaded pack bundles & metadata. I don't plan to re-upload any packs for this change. But new versions will have schema description in place.


jonathan-codaio (approved):
Aha, good catch. So the field already existed in the metadata type def?

huayang-codaio:
Yes it exists in the metadata def.

PR #1572, opened 2021-11-19
restrict param autocomplete to only string & number, and respect param type in autocomplete shape

huayang-codaio:
https://staging.coda.io/d/_dL5PQJ5pk51#_tu43U/r462&modal=true
https://golinks.io/bug/20266

autocomplete was using a vague type of `string | number`. This PR makes it more strict based on the param type. 

image


api_types.ts
```diff
@@ -70,7 +70,7 @@ export const imageArray: ArrayType<Type.image> = {
 };
 
 // Mapping from our type enum to the JS types they are manifested as.
-interface TypeMap {
+export interface TypeMap {
```
huayang-codaio, line 73: this is ignored in typedoc.js already.

huayang-codaio:
hold on. I need to piggyback more into this..

huayang-codaio:
There's a separate ticket of fixing return type of functional autocomplete. 

It actually takes more work than I expected. Will do that in a separate PR.

api.ts
```diff
@@ -1044,29 +1051,30 @@ export interface SimpleAutocompleteOption {
...
+      });
     } else if (typeof option === 'number') {
-      return {
+      metadataResults.push({
         value: option,
         display: option.toString(),
-      };
+      });
     }
-    return option;
```
jonathan-codaio, line 1068: Does this silently drop options that are strings or numbers? Should we error on that instead?

huayang-codaio: let me do that.

api.ts
```diff
@@ -267,7 +268,9 @@ export function wrapMetadataFunction(
 /** Options you can specify when defining a parameter using {@link makeParameter}. */
 export type ParameterOptions<T extends ParameterType> = Omit<ParamDef<ParameterTypeMap[T]>, 'type' | 'autocomplete'> & {
   type: T;
-  autocomplete?: MetadataFormulaDef | Array<string | number | SimpleAutocompleteOption>;
+  autocomplete?: T extends ParameterType.Number | ParameterType.String
```
jonathan-codaio, line 271: Does this mean we won't allow you to define an autocomplete for a parameter that's not a string or a number? I guess that's consistent with what we technically support right now, just wonder if it will break anybody who did it for a boolean or date field and it kinda works for whatever reasons due to quirks of CFL. But I guess that's ok if we don't intend for that to work.

jonathan-codaio (approved):

api.ts
huayang-codaio: yeah, I thought about it. It's unclear what the expected autocomplete return type would be if the parameter type isn't string or number. So I just restrict this to number / string. 

It won't break existing pack bundles but would give TS errors if non string/number parameter has an autocomplete. I'll verify the packs repo first before I merge this PR.

huayang-codaio: tested in packs repo that make validate works.

PR #1573, opened 2021-11-19
Fix broken @link references.

jonathan-codaio:
This should fix almost all the busted ones, the remaining ones are already fixed by https://github.com/coda/packs-sdk/pull/1570 (friendly ping on that one).


api.ts
```diff
@@ -1182,6 +1182,23 @@ export function makeObjectFormula<ParamDefsT extends ParamDefs, SchemaT extends
   }) as ObjectPackFormula<ParamDefsT, SchemaT>;
 }
 
+/**
+ * A set of options used internally by {@link makeDynamicSyncTable}
```
ekoleda-codaio, line 1189: Isn't this also used by developers when they add dynamic schemas to static sync tables?

ekoleda-codaio (approved):

api.ts
jonathan-codaio: Ah yep, updated.

PR #1574, opened 2021-11-19
Test to enforce that we don't have broken @links.

jonathan-codaio:


ekoleda-codaio (approved):
Great idea!

PR #1576, opened 2021-11-22
Remove type in examples of createObjectSchema

ekoleda-codaio:
Now that it's no longer a required field, remove it for simplicity.


jonathan-codaio (approved):
🪓 🎉

PR #1590, opened 2021-11-23
add typing to metadata function return type.

huayang-codaio:
The original trigger is that the autocomplete functions don't do return value checks. As I investigated, it turns out metadata function return values (autocomplete is a metadata function) are not typed at all. 

Our typescript tree has grown over sophisticated. It quickly blown head if we were to make everything perfect. This is my 3 attempt where I had to use some force casting to get things working. 

packs-sdk testing (we indeed need to fix some return types): https://github.com/coda/packs/commit/99494b564025cbe468c3ec950dedae1e1ca92f99
coda repo testing: TODO


runtime/isolated-vm/package.json
```diff
@@ -54,5 +54,5 @@
     "test": "node test.js || nodejs test.js"
   },
   "types": "isolated-vm.d.ts",
-  "version": "4.3.6"
+  "version": "4.3.5"
```
az-codaio, line 57: Looks like this is downgrading the bump from https://github.com/coda/packs-sdk/pull/1583?

huayang-codaio: thanks for catching. will revert.

jonathan-codaio:
Is there a bug or user issue report that this is specifically addressing?

At a certain point, as you note, the ever-more-complex TypeScript makes the SDK less helpful rather than more helpful, so I don't think we should feel completely obligated to be hyper-accurate.

huayang-codaio:
This is the ticket: https://staging.coda.io/d/_dL5PQJ5pk51#_tu43U/r464&modal=true

The complain is that we don't do any return type checking for autocomplete functions. My other PR adds type checking if the autocomplete is in simple/plain format but none has been added if the autocomplete turns a function or a full formula.

jonathan-codaio (commented):
Yeah, interesting.

@ekoleda-codaio Thoughts on this? Curious as to when you filed https://staging.coda.io/d/Packs-Project-Tracker-go-pak_dL5PQJ5pk51/Data_sun7J#All-Tasks_tu43U/r464&modal=true if you were expecting this would be caught in TypeScript vs elsewhere?

The typing here is quite useful in that helps narrow the currently very generic metadata formula type depending on the use case. It comes at the expense of yet more TS generics that make the SDK even harder to understand.

I think it's a partial solution to the problem of misunderstanding what to return for an autocomplete function. If you don't use TS, you could wind up the same situation. I'm wondering if we should instead be validating the return values for autocomplete (and possibly other metadata formulas) and logging errors for the pack maker if they don't match expectations?

ekoleda-codaio:
I guess you could interpret that bug two ways:

1) Returning a string array should work, and the bug is that it's broken.
2) Returning a string array shouldn't work, and the bug is that we didn't catch it at build time.

I was in the mindset of the former. String arrays worked for static autocomplete, they should work for dynamic autocomplete as well. This is the bug I think we should fix.

Regarding the TS enforcement of types, I do think it is a bit of a double edged sword. It makes it harder to make a mistake, but the error messages are so _incredibly_ opaque that they often aren't actionable. By contrast, the errors we manually throw (like when you forgot to declare a network domain) are much more readable and actionable. 

Regarding folks that don't use TS, I don't know if we need to worry about them too much. The Web IDE forces you to use TypeScript, and the `coda init` flow sets it up as well. Non-TS projects can only be created using the CLI, for folks that specifically chose to deviate from the recommended path.

jonathan-codaio:
I'm not opposed to doing the same transformation for autocomplete return values at execution time (converting arrays into objects with display/value keys). The reason it works for hardcoded values is that we handle all that in the SDK, at pack creation time we convert your hardcoded value into a formula that does the right thing. We can do a similar conversion at execution time.

huayang-codaio:
I like the idea of runtime validation. TBH it's not really safe to purely rely on TS on the server side anyway. On the other side, if we are still using TS, ppl would consider it a bug if some API is not typed (like PR). 

In the ideal world, I would rely on one approach of typing and make it complete on all the APIs. This could be either TS or JS+typedoc whatever. At this moment, I assume that we are still stick with TS for the 2022 launch. 

Specifically about TS, maybe it worths a revisit of the type definitions and some point. It's probably over complicated at this moment.

jonathan-codaio:
> I like the idea of runtime validation. TBH it's not really safe to purely rely on TS on the server side anyway. On the other side, if we are still using TS, ppl would consider it a bug if some API is not typed (like PR).
> 
> In the ideal world, I would rely on one approach of typing and make it complete on all the APIs. This could be either TS or JS+typedoc whatever. At this moment, I assume that we are still stick with TS for the 2022 launch.
> 
> Specifically about TS, maybe it worths a revisit of the type definitions and some point. It's probably over complicated at this moment.

Cool, so should we have the calc server transform autocomplete results of the form `['a', 'b']` to `[{display: 'a', value: 'a'}, {display: 'b', value: 'b'}]` before returning them so that Eric's code starts working?

If we start doing that and considering that valid, the TS here would actually be too restrictive, we could maybe just table the TS changes if Eric's immediate issue is fixed by the transformation.

huayang-codaio:
oh, sure. let me check the calc server part. 

on the other side, the TS in this PR will be allowing the ['a', 'b'] format. I assume an autocomplete can accept `number[] | string[] | {display: string, value: number | string} | MetadataFunction | MetadataFormula`. This is that the TS was trying to do but it was never implemented on the return type.

huayang-codaio:
I have sent https://github.com/coda/packs-sdk/pull/1606 and https://github.com/coda/coda/pull/65146 to address the runtime issue. Do we still want to proceed this TS change?

jonathan-codaio:
> I have sent #1606 and [coda/coda#65146](https://github.com/coda/coda/pull/65146) to address the runtime issue. Do we still want to proceed this TS change?

Awesome! Let's skip the TS change for now, I think the `coda` side change is a better overall outcome and we're ok without the extra TS for now.

PR #1605, opened 2021-12-03
Allow specifying SetEndpoint.getOptionsFormula without using makeMetadataFormula.

jonathan-codaio:
PTAL @ekoleda-codaio @huayang-codaio @coda/packs


types.ts
```diff
@@ -203,6 +212,12 @@ export enum PostSetupType {
  */
 export type PostSetup = SetEndpoint;
 
+/**
+ * Simplified configuration for {@link PostSetup} that a pack developer can specify when calling
+ * {@link setUserAuthentication} or {@link setSystemAuthentication}.
+ */
+export type PostSetupDef = SetEndpointDef;
```
huayang-codaio, line 219: Not related to this PR: I suppose that we are implementing this convention that `xxxDef` indicates pack bundle input. I realized in a TS refactor that this assumption might be violated in a few places. we may want to fix them when we are at it.

huayang-codaio (approved):

types.ts
jonathan-codaio: Yeah, this would be worth doing. I started using the Def syntax informally when splitting out user-entered types from the final persisted types, but as the pattern becomes more common it seems worth formalizing this. (Can also rename these in a better way and formalize that.)

PR #1606, opened 2021-12-03
support autocomplete function returning string list

huayang-codaio:
autocomplete string list is not recognized browser client at this moment. https://github.com/coda/coda/pull/65146 should fix that. 

meanwhile this PR transforms autocomplete functions to always return a structured list in the compile time.


jonathan-codaio (commented):
Hmm, can you say a bit more about this? Is this an alternate approach to the `coda` side changes, and we would do one but not both? Or are you suggesting we do both?

huayang-codaio:
> Hmm, can you say a bit more about this? Is this an alternate approach to the coda side changes, and we would do one but not both? Or are you suggesting we do both?

if we pick one, I'd say we go with this. The coda PR is mostly for any unknown pack that's currently broken for string list autocomplete. I don't know if there's any one. We can void it if it's not a concern at all.

jonathan-codaio:
Yeah, the `coda` fix actually seems a bit more robust because with the compile-time fix, if you were to use

```
autocomplete: makeMetadataFormula(async context => ...)
```

we wouldn't wrap your formula because it's already wrapped and you could still return a string array and we'd get a runtime error. We could do even more auto-wrapping and try to detect this case, but the `coda` change seems simpler and captures more cases.

huayang-codaio:
good point. let's go with the coda side fix then.

PR #1608, opened 2021-12-04
make connection requirement optional for autocomplete metadata formula

huayang-codaio:
Metadata formulas are default to requiring a connection. this isn't true for autocomplete formulas. This PR changes it to optional instead. 

On the other hand, I wonder if we should just make metadata formulas default to use a required connection. I can totally see a developer playing with it without a connection. Why we decided to make it required in the first place? 

tested in devapp that this fixes the issue.


jonathan-codaio:
I like this change and I think this is or is nearly the right end state. Are we sure it's backwards-compatible though? It seems like it's going to the connection requirement on a bunch of existing packs. Maybe it's ok because it's changing it from required to optional and that's effective a no-op? I think it's worth testing this with a couple existing packs that use autocomplete with connections.

I think it was a mistake to make connections required by default for metadata formulas in general. I think at the time it seemed like a sensible thing to do because most of the use cases were things like getConnectionName that made no sense without a connection. As use cases evolved to autocomplete and then things like getSchema without connections, it doesn't make sense anymore and is limiting us.

huayang-codaio:
it's only going to affect new pack bundles, where changing connectionRequirement from required to optional should be okay (imo). 

let me test this with the g-cal pack and see if it works well.

huayang-codaio:
@jonathan-codaio so the packs are working well. G-calendar & cross-doc (with dynamic tables)

however, changing connection requirement makes a version bump. for autocomplete, it's likely just a minor bump. for most of other metadata functions, that can lead to a major bump. 

there's no way to skip version bumps for v1 packs today since they are managed as any other v2 pack. good news is that we can delay bumping these packs until the next necessary change. 

wdyt?

api.ts
```diff
@@ -1016,7 +1019,9 @@ export function makeMetadataFormula(
       makeStringParameter('formulaContext', 'Serialized JSON for metadata', {optional: true}),
     ],
     examples: [],
-    connectionRequirement: options?.connectionRequirement || ConnectionRequirement.Required,
+    // alternatively should we change this default to ConnectionRequirement.Optional? I can imagine someone
+    // testing/playing with a metadata formula without a connection.
+    connectionRequirement: options?.connectionRequirement || ConnectionRequirement.Optional,
```
huayang-codaio, line 1024: if this is going to happen in the future, better sooner than later.

api.ts
```diff
@@ -1016,7 +1019,9 @@ export function makeMetadataFormula(
       makeStringParameter('formulaContext', 'Serialized JSON for metadata', {optional: true}),
     ],
     examples: [],
-    connectionRequirement: options?.connectionRequirement || ConnectionRequirement.Required,
+    // alternatively should we change this default to ConnectionRequirement.Optional? I can imagine someone
```
jonathan-codaio, line 1022: Can remove comment.

jonathan-codaio (commented):
Ok let's give this a shot. Please update the SDK version to 0.8.0 since this is breaking and add an entry to CHANGELOG. We'll probably want to announce this in #ext-packs-all as well since this is breaking.

When we update the packs repo with this let's release a couple packs like GCal that use metadata formulas so we can catch any potential issues early.

huayang-codaio:
will do.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 0.8.0
```
jonathan-codaio, line 1: I think package.json needs to get bumped.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+- **Breaking Change** Connection requirement of metadata formulas are now set to optional. If your pack is using sematic versions, this will likely lead to a major version bump in your next release.
```
jonathan-codaio, line 3: Nit, just for comprehensive, maybe "The connection requirement for metadata formulas now defaults to optional instead of required."

jonathan-codaio (approved):

huayang-codaio:
deployed crossdoc in prod and it's looking good.

PR #1609, opened 2021-12-07
Relax continuation typing to support a nested hash

nigel-codaio:
Relax continuation typing to support a nested hash.

See https://codaproject.slack.com/archives/C02ERJRBA21/p1638833332320300 for context.


jonathan-codaio (approved):
Thanks! Just need to run `make build` and commit all the junk.

PR #1634, opened 2021-12-14
Adjust snippets

ekoleda-codaio:
Some small changes to snippets I've been wanting to make:

- Add an OAuth2 snippet
- Remove `headers` from Fetcher snippets, since they aren't used often and very often need to be deleted.
- No more fake method names as the display names for snippets. `addStringFormula` isn't a method name, but looks like one. Instead using `addFormula:string` instead, while keeping the previous `addStringFormula` as a hidden trigger string.


jonathan-codaio (approved):
Worth having a separate dedicated snippet that demonstrates custom headers?

ekoleda-codaio:
> Worth having a separate dedicated snippet that demonstrates custom headers?

It's a tough once, since the majority of the complexity actually exists when you use the fetcher. I've wanted to build a full sample for it, but had trouble finding a good API to base it on.

PR #1635, opened 2021-12-15
add an optional description field for sync formula

huayang-codaio:
ticket: https://staging.coda.io/d/_dL5PQJ5pk51#_tu43U/r460&modal=true
slack: https://codaproject.slack.com/archives/C02K6EABVMK/p1636475056153300

I guess I need to update the packs repo with a description to every sync table and dynamic sync table after? 

@coda/packs @jonathan-codaio @patrick-codaio @spencer-codaio @alan-codaio 

cc @alicia-codaio


api.ts
```diff
@@ -1232,6 +1234,12 @@ export interface SyncTableOptions<
    * from an e-commerce platform should be called 'Products'. This name must not contain spaces.
    */
   name: string;
+  /**
+   * The description of the sync table. This is shown to users in the Coda UI.
+   * This should describe what the sync table does in more detailed languages. For example, the
```
jonathan-codaio, line 1239: Nit, "language" singular.

jonathan-codaio (approved):
Cool. Just confirming that the resulting description is present in the metadata that the browser users when rendering the Explore panel, so that we're able put this in the UI? I always get a bit confused about whether a given input field is going to make its way into the output metadata or whether some more plumbing is required.

Since it's optional, doesn't seem like you have to immediately go update the `packs` repo, but agree that it would best for usability if we added descriptions soon-ish. I don't think that's higher priority than other bugs though. I would just test that this all works as expected with a user-created pack.

huayang-codaio:
@jonathan-codaio yes, per Alicia, the description will show on the UI.

image

jonathan-codaio:
Yep I just wanted to clarify that we didn't need to do any more plumbing in the SDK to have the description available for rendering in the UI.

huayang-codaio:
will test it for sure.

patrick-codaio (approved):

huayang-codaio:
The description will make its way to the UI. 

Weird thing is that I somehow need to `make clean` devbox for the packs-sdk change to be picked up in the browser pack editor. I think it used to only take a rebuild of the website bundle. 

image

PR #1637, opened 2021-12-17
update documentation on release

spencer-codaio:


.circleci/config.yml
```diff
@@ -86,26 +86,54 @@ jobs:
           command: make docs
       - run:
           name: "MkDocs"
-          command: make build-mkdocs
+          # TODO(spencer): Update with new url
+          command: MK_DOCS_SITE_URL="https://head.coda.io/developers/temp-developer-docs/packs/latest/" make build-mkdocs
```
jonathan-codaio, line 90: This is just bringing it up to date with `main` before you update the url? Aka we should just be rubber-stamping this?

spencer-codaio: yeah i just noticed it's out of date on staging since the workflow hasnt been run in a while

PR #1640, opened 2021-12-17
Default the --fetch flag to false.

jonathan-codaio:
I think this should do the trick?

PTAL @ekoleda-codaio


huayang-codaio (approved):
makes sense!

PR #1642, opened 2021-12-20
Add CustomAuth digest transform

nigel-codaio:
I'm building a Pack that uses a form of custom digest auth.   It requires sending a header which is a hex digest hash of `<session-id>:<secret-key>:<secret-token>`.   Given secrets aren't accessible to the Pack invocation code, this requires extending the current `CustomAuth` schema with an optional hex digest transform that is applied after substitution.

Example:
```ts
const customAuthWithDigest = {
  type: AuthenticationType.Custom,
  params: [
     {name: 'secretValue', description: 'Description for param1'},
     {name: 'secretToken', description: 'Description for param2'},
  ],
  digestTransform: {
    algorithm: 'sha1',
    header: 'X-Some-Header',
  }
};
```

You'd then add a header:
```ts
{
  'X-Some-Header': `session-key:{{secretValue-${context.invocationToken}}}:{{secretToken-${context.invocationToken}}}`
}
```

After auth token substitution this becomes:
```ts
{
  'X-Some-Header': `session-key:secret-value:secret-token`
}
```

We'd then hash this using the requested digest algorithm (in this example `sha1`) to return:
```ts
{
  'X-Some-Header': '2c74b0b90597a5cf66b38f1d93b6cf19c379e750'
}
```

### DIFF
--- a/testing/fetcher.ts
+++ b/testing/fetcher.ts
@@ -24,6 +24,7 @@ import type {TokenCredentials} from './auth_types';
 import {URL} from 'url';
 import type {WebBasicCredentials} from './auth_types';
 import {assertCondition} from '../helpers/ensure';
+import {createHash} from 'crypto';
 import {ensureExists} from '../helpers/ensure';
 import {ensureNonEmptyString} from '../helpers/ensure';
 import {ensureUnreachable} from '../helpers/ensure';
@@ -282,6 +283,7 @@ export class AuthenticatingFetcher implements Fetcher {
         let headersWithSubstitutions = JSON.stringify(headers);
 
         const {params} = this._credentials as CustomCredentials;
+        const {digestTransform} = this._authDef;
 
         Object.entries(params).forEach(([key, value]) => {
           if (urlWithSubstitutions) {
@@ -322,11 +324,20 @@ export class AuthenticatingFetcher implements Fetcher {
           }
         });
 
+        const newHeaders: FetchRequest['headers'] = headersWithSubstitutions ? JSON.parse(headersWithSubstitutions) : undefined;
+        if (newHeaders && digestTransform) {
+          const {algorithm, header} = digestTransform;
+          const headerValue = newHeaders[header];
+          if (headerValue) {
+            newHeaders[header] = createHash(algorithm).update(headerValue, 'utf-8').digest('hex')
+          }
+        }
+
         return {
           url: urlWithSubstitutions,
           body: bodyWithSubstitutions,
           form: formWithSubstitutions ? JSON.parse(formWithSubstitutions) : undefined,
-          headers: headersWithSubstitutions ? JSON.parse(headersWithSubstitutions) : undefined,
+          headers: newHeaders,
         };
       }
       case AuthenticationType.QueryParamToken: {
--- a/types.ts
+++ b/types.ts
@@ -480,6 +480,20 @@ export interface CustomAuthParameter {
 }
 
 /**
+* An optional header hext digest transform for the {@link CustomAuthentication} authentication type.
+*/
+export interface CustomAuthHeaderDigestTransform {
+  /**
+   * The header on which the hex digest should be applied.
+   */
+  header: string;
+  /**
+   * The hash digest algorithm to apply to the header value.
+   */
+   algorithm: 'md5' | 'sha1' | 'sha256' | 'sha512'
+} 
+
+/**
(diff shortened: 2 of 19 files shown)


spencer-codaio:
hmm interesting.. wondering if this should actually be on a per-param level instead of an auth-wide specification? Could imagine packs wanting multiple custom secrets from user but only some need to be serialized according to some format. and others need to be inserted as plain text. Thoughts?

also cc @jonathan-codaio

nigel-codaio:
@spencer-codaio what did you have in mind?  Not following how you'd make this a per-param level thing.   The only alternate I could come up with for a per-param level things would be to support a list of headers you'd want to encode as in:
```ts
const customAuthWithDigest = {
  ...,
  digestTransform: {
    algorithm: 'sha1',
    headers: ['X-Some-Header1', 'X-Some-Header2', ...]
  }
};
```

That said, I couldn't find a use-case for that as all the examples I'd found of this style of auth use a single auth header.    Trivial to extend if we wanted to go that route now or later.

types.ts
```diff
@@ -479,6 +479,20 @@ export interface CustomAuthParameter {
   description: string;
 }
 
+/**
+* An optional header hext digest transform for the {@link CustomAuthentication} authentication type.
```
jonathan-codaio, line 483: Nit, hext-->hex. Should probably include example usage and effect in the doc comment too.

jonathan-codaio (commented):
Adding @ekoleda-codaio to weigh in. We have to consider developer-facing SDK changes pretty carefully at this point, everything we add now has to be well-documented and supported forever. Extending the SDK for specific use cases has gotten us into a lot of trouble and we've spent many months trying to document and rationalize prior single-use-case functionality.

Can you say a bit more about the use case and the API you're working with? Is this notion of a digest a commonly-used pattern or is it something specific to this API? Are there are types of header transforms for which we might want to try to be future-proof?

testing/fetcher.ts
```diff
@@ -322,11 +324,20 @@ export class AuthenticatingFetcher implements Fetcher {
           }
         });
 
+        const newHeaders: FetchRequest['headers'] = headersWithSubstitutions ? JSON.parse(headersWithSubstitutions) : undefined;
+        if (newHeaders && digestTransform) {
+          const {algorithm, header} = digestTransform;
+          const headerValue = newHeaders[header];
+          if (headerValue) {
+            newHeaders[header] = createHash(algorithm).update(headerValue, 'utf-8').digest('hex')
```
ekoleda-codaio, line 332: Would we want to verify that the has algorithm is a known value? The type definition won't enforce this if a Pack is built using vanilla JavaScript.

ekoleda-codaio, line 332: I'd suggest making the digest encoding variable as well, to make this a bit more flexible.

types.ts
```diff
@@ -546,6 +560,12 @@ export interface CustomAuthentication extends BaseAuthentication {
    * replacement inside the constructed network request.
    */
   params: CustomAuthParameter[];
+  /**
+   * An optional header hex digest transform that will be performed once template replacement is complete. 
+   * The header {@link CustomAuthHeaderDigestTransform.header} will be transformed using the specified hex
+   * digest {@link CustomAuthHeaderDigestTransform.algorithm}.
+   */
+  digestTransform?: CustomAuthHeaderDigestTransform
```
ekoleda-codaio, line 568: I wonder if we should follow a pattern similar to `postSetup`, where we have an array of transform steps, with the only implemented step type at the moment being this one. That would allow devs to use multiple digest transforms on a single request and give us a place to hang future oddball transforms that are needed.

nigel-codaio:
I was initially looking at this to access the [ill-documented BlueIris security API ](https://www.houselogix.com/docs/blue-iris/BlueIris/json.htm)(used by my home automation and security camera system).   I did more search and it looks to be a poor/butchered implementation of a form of [Digest Access Authentication](https://en.wikipedia.org/wiki/Digest_access_authentication) but doesn't follow the spec at all and ends up using session tokens embedded in the payload, not the headers.    Given it's a special snowflake, I'm just going to close this and figure out an alternate proxy-based approach to access my system.

ekoleda-codaio (commented):
Handling the seemingly infinite nuance in auth flows with config only is certainly going to be a challenge. I don't know of any other APIs that follow this exact pattern, but most of my experience is with OAuth2, which is a different beast. Some suggestions below.

ekoleda-codaio:
Looks like this was closed as I was typing 😄 Disregard!

PR #1652, opened 2021-12-21
Support a configurable scope delimiter for OAuth2.

jonathan-codaio:
SDK side of changes to address https://staging.coda.io/d/Quality-Tracker_d-GJF-DmEUK/My-Issues_sud-l#My-issues_tui9H/r19756&modal=true. Sending the `coda` side shortly too.

@ekoleda-codaio what's the best way to test this--is there a Todoist example that would validate that this works out of the box?

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -336,6 +336,7 @@ const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny> 
     authorizationUrl: z.string(),
     tokenUrl: z.string(),
     scopes: z.array(z.string()).optional(),
+    scopeDelimiter: z.string().optional(),
     tokenPrefix: z.string().optional(),
     additionalParams: z.record(z.any()).optional(),
     endpointKey: z.string().optional(),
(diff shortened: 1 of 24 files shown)


patrick-codaio (approved):

jonathan-codaio:
Ok tested this with Todoist and it worked, after I fixed a longstanding bug in the yargs default value for extraOAuthScopes that results in an empty string scope being added that messes up the join.

ekoleda-codaio:
Sorry for the late reply here @jonathan-codaio. There wasn't an out-of-the box sample, but modifying Todoist was how I discovered it.

Regarding the extra space, it sounds the same as https://golinks.io/bug/20748, so you might want to close that as well.

testing/upload_validation.ts
```diff
@@ -336,6 +336,7 @@ const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny>
     authorizationUrl: z.string(),
     tokenUrl: z.string(),
     scopes: z.array(z.string()).optional(),
+    scopeDelimiter: z.string().optional(),
```
huayang-codaio, line 339: I forgot why I didn't merge https://github.com/coda/packs-sdk/pull/1273 but one comment from Chris was to limit this to a subset of single characters.

chris-codaio: +1 - worry about parsing this down the line and missing text.  Let's require this to be one of a known set of single character separators.

ekoleda-codaio: I imagine that space and comma would give us 99% coverage.

jonathan-codaio: Thanks for the suggestion, sending a followup to restrict to specific delimiters only.

PR #1668, opened 2021-12-28
let password fallback to empty space (to be consistent with coda repo)

huayang-codaio:
https://codaproject.slack.com/archives/C02ERJRBA21/p1640727998430700

otherwise password is set to undefined and thus we have `<password>:undefined` in the header. 

```
> var a = {x: 1, y: undefined}
undefined
> a
{ x: 1, y: undefined }
> var {x, y} = a
undefined
> y
undefined
```

### DIFF
--- a/test/auth_test.ts
+++ b/test/auth_test.ts
@@ -784,7 +784,7 @@ describe('Auth', () => {
         sinon.assert.calledOnceWithExactly(mockMakeRequest, {
           body: undefined,
           form: undefined,
-          headers: {Authorization: 'Basic c29tZS11c2VybmFtZTp1bmRlZmluZWQ=', 'User-Agent': 'Coda-Test-Server-Fetcher'},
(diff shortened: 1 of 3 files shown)


test/auth_test.ts
```diff
@@ -784,7 +784,7 @@ describe('Auth', () => {
         sinon.assert.calledOnceWithExactly(mockMakeRequest, {
           body: undefined,
           form: undefined,
-          headers: {Authorization: 'Basic c29tZS11c2VybmFtZTp1bmRlZmluZWQ=', 'User-Agent': 'Coda-Test-Server-Fetcher'},
```
huayang-codaio, line 787: this is actually encoded `...:undefined` :)

gita-codaio (approved):

gita-codaio:
thank you!

PR #1676, opened 2022-01-03
Remove stray quotes from our custom auth example

patrick-codaio:
Fixes https://coda.github.io/packs-sdk/reference/sdk/interfaces/CustomAuthentication/ to match https://coda.github.io/packs-sdk/guides/advanced/authentication/#custom-tokens (which is the correct one). While I'm here, I changed the quotes to double quotes since Eric picked that for our documentation style guide.

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -509,8 +509,8 @@ export interface CustomAuthParameter {
  * // Suppose you're using an API that requires a secret id in the request URL,
  * // and a different secret value in the request body. You can define a Custom
  * // authentication configuration with two params:
- * // params: [{name: 'secretId', description: 'Secret id'},
- * //          {name: 'secretValue', description: 'Secret value'}])
+ * // params: [{name: "secretId", description: "Secret id"},
+ * //          {name: "secretValue", description: "Secret value"}])
  * // The user or the pack author will be prompted to specify a value for each
  * // of these when setting up an account.
  * // In the `execute` body of your formula, you can specify where those values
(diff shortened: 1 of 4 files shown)


types.ts
```diff
@@ -525,7 +525,7 @@ export interface CustomAuthParameter {
  *   let secretIdTemplateName = "secretId-" + context.invocationToken;
  *   let urlWithSecret = "/api/entities/{{" + secretIdTemplateName + "}}"
  *   let secretValueTemplateName = "secretValue-" + context.invocationToken;
- *   let secretHeader = 'Authorization  {{"' + secretValueTemplateName + '"}}';
+ *   let secretHeader = "Authorization  {{" + secretValueTemplateName + "}}";
```
spencer-codaio, line 528: don't you need to use different quotes inside the string (single quotes) or escape?

spencer-codaio (approved):
thanks for making uniform!

ekoleda-codaio (approved):
Good catch!

types.ts
patrick-codaio: There's a chance we should be escaping somehow, but the extra quotes were making the code not work. FWIW, https://coda.github.io/packs-sdk/guides/advanced/authentication/#custom-tokens doesn't have them and your urlWithSecret here doesn't have them.

Here's some example code I wrote that worked:
```
      const authHeaders = {
        'x-client': '8100a127-10c2-4f01-a18f-28b1b550637f',
        'x-api-user': `{{userId-${context.invocationToken}}}`,
        'x-api-key': `{{apiToken-${context.invocationToken}}}`,
      }

      let tagResponse = await context.fetcher.fetch({
        method: "GET",
        url: "https://habitica.com/api/v3/tags",
        headers: authHeaders,
      });
```

spencer-codaio: oh wait nvm i was reading this wrong lol my bad

spencer-codaio: somehow i thought there were "s inside a pair of "s but there isnt. thanks!

PR #1684, opened 2022-01-04
Do not allow giant pack versions

patrick-codaio:
Context: https://codaproject.slack.com/archives/CGLP9V98C/p1641251870120700

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -862,9 +862,28 @@ const syncTableSchema = z.union([genericDynamicSyncTableSchema, genericSyncTable
 // (Zod doesn't let you call .extends() after you've called .refine(), so we're only refining the top-level
 // schema we actually use.)
 const unrefinedPackVersionMetadataSchema = zodCompleteObject<PackVersionMetadata>({
-  version: z
-    .string()
-    .regex(/^\d+(\.\d+){0,2}$/, 'Pack versions must use semantic versioning, e.g. "1", "1.0" or "1.0.0".'),
+  version: z.string().refine(
+    version => {
+      const match = version.match(/^(\d+)(\.\d+)?(\.\d+)?$/);
+      if (!match || match.length < 2) {
+        return false;
+      }
+      for (let i = 1; i < match.length; i++) {
+        // Unless all 3 semantic version fields are specified, there will be blank match groups.
+        if (match[i] === undefined) {
+          break;
+        }
+        // The first group won't have a prefix '.'
+        const versionNumber = i === 1 ? match[i] : match[i].substring(1);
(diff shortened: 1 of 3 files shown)


testing/upload_validation.ts
```diff
@@ -862,9 +862,28 @@ const syncTableSchema = z.union([genericDynamicSyncTableSchema, genericSyncTable
 // (Zod doesn't let you call .extends() after you've called .refine(), so we're only refining the top-level
 // schema we actually use.)
 const unrefinedPackVersionMetadataSchema = zodCompleteObject<PackVersionMetadata>({
-  version: z
-    .string()
-    .regex(/^\d+(\.\d+){0,2}$/, 'Pack versions must use semantic versioning, e.g. "1", "1.0" or "1.0.0".'),
+  version: z.string().refine(
+    version => {
+      const match = version.match(/^(\d+)(\.\d+)?(\.\d+)?$/);
```
jonathan-codaio, line 867: Would it work to just change each of these from `\d+` to, say, `\d{1,10}` or whatever a compatible character limit is? (Acknowledging that we'll wind up disallowing some large castable values.)

Alternatively, can you just add a `refine()` after the `regex()` call, and just call `split('.')` on it since we know it's syntactically valid at that point?

`return value.split('.').all(num => Number(num) <= ...);`

And doing the refine separately allows for an error message specific to this case.

patrick-codaio: Great idea to split it up. Done!

jonathan-codaio (approved):

testing/upload_validation.ts
```diff
@@ -864,7 +864,12 @@ const syncTableSchema = z.union([genericDynamicSyncTableSchema, genericSyncTable
 const unrefinedPackVersionMetadataSchema = zodCompleteObject<PackVersionMetadata>({
   version: z
     .string()
-    .regex(/^\d+(\.\d+){0,2}$/, 'Pack versions must use semantic versioning, e.g. "1", "1.0" or "1.0.0".'),
+    .regex(/^\d+(\.\d+){0,2}$/, 'Pack versions must use semantic versioning, e.g. "1", "1.0" or "1.0.0".')
+    .refine(
+      // Version numbers must not be bigger than a postgres integer.
+      version => version.split('.').filter(part => Number(part) > 2_147_483_647).length === 0,
```
spencer-codaio, line 870: oh just curious what is the `_` syntax in the number?

patrick-codaio: https://2ality.com/2018/02/numeric-separators.html 

Fun fact, the guy who got that included into JS is a friend of mine :)

PR #1685, opened 2022-01-05
add some meta tags

spencer-codaio:
will test by checking the upload to readdocs from the pr

### DIFF
--- a/documentation/theme/main.html
+++ b/documentation/theme/main.html
@@ -1,21 +1,16 @@
-{% extends "base.html" %}
+{% extends "base.html" %} 
 
+{% block extrahead %}
 <!-- TODO: should shape these up and should cater to specific page -->
 <meta name="description" content="Documentation for the SDK to build Coda Packs." />
-<meta
-  name="image"
-  content="https://codaio.imgix.net/docs/OZNRRq68EI/blobs/bl-O5NyMFtXuk/0a49697fee3d51730d4c66e8726be2c3deb1801e8c751f58d40e49351425eeb8a124e13cb0a2647549663df99233d80f7f92990d0e6c501ae36d40d72bd74cf558386f8204c3e64814673be573f0599673831ce7d11429c11753d276cfc758ec92cb5079?auto=format%2Ccompress&fit=max&w=1080&dpr=1"
-/>
+<meta name="image" content="https://cdn.coda.io/external/img/doc-banner-compressed-2x-v2.jpg" />
 
 <!-- Twitter tags -->
 <meta name="twitter:card" content="summary" />
 <meta name="twitter:site" content="@coda_hq" />
 <meta name="twitter:title" content="Coda Packs SDK" />
 <meta name="twitter:description" content="Documentation for the SDK to build Coda Packs." />
-<meta
-  name="twitter:image"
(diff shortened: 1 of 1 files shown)


documentation/theme/main.html
```diff
@@ -1,6 +1,32 @@
...
 
-{% block announce %} 
-  The Pack SDK is currently in a limited beta.
-  <a href="https://coda.io/packsbeta">Click here</a> to learn more and apply.
-{% endblock %}
+<!-- TODO: should shape these up and should cater to specific page -->
+<meta name="description" content="Documentation for the SDK to build Coda Packs." />
+<meta
+  name="image"
+  content="https://codaio.imgix.net/docs/OZNRRq68EI/blobs/bl-O5NyMFtXuk/0a49697fee3d51730d4c66e8726be2c3deb1801e8c751f58d40e49351425eeb8a124e13cb0a2647549663df99233d80f7f92990d0e6c501ae36d40d72bd74cf558386f8204c3e64814673be573f0599673831ce7d11429c11753d276cfc758ec92cb5079?auto=format%2Ccompress&fit=max&w=1080&dpr=1"
```
jonathan-codaio, line 7: Ah this is a bit fragile, since it's a cover image from doc that could get deleted at any time. We'd want to use an image that's uploaded directly to our CDN. Do you know if it already exists there?

If it doesn't, you can upload it directly pretty easy by logging into the coda-common AWS account and going to S3 and finding the images bucket. You may need to access in #it-help for temporary access.

spencer-codaio: ah good call ill upload

spencer-codaio: do we garbage collect images from docs? for some reason i thought they were just around forever

jonathan-codaio: I don't actually know for sure. Presumably we do after we stop counting stuff towards your attachments doc size. But certainly we're likely to if we delete the entire doc, and I think that's the Beta doc which is now obsolete so it could go away on us.

spencer-codaio: oh interesting but yeah makes sense that its fragile

documentation/theme/main.html
```diff
@@ -1,6 +1,25 @@
...
-  <a href="https://coda.io/packsbeta">Click here</a> to learn more and apply.
+{% block extrahead %}
+<!-- TODO: should shape these up and should cater to specific page -->
+<meta name="description" content="Documentation for the SDK to build Coda Packs." />
+<meta name="image" content="https://cdn.coda.io/external/img/doc-banner-compressed-2x-v2.jpg" />
+
+<!-- Twitter tags -->
+<meta name="twitter:card" content="summary" />
+<meta name="twitter:site" content="@coda_hq" />
+<meta name="twitter:title" content="Coda Packs SDK" />
```
jonathan-codaio, line 11: To my chagrin, the external name was decided to be "Coda Pack SDK" (singular), so should update throughout the meta tags.

jonathan-codaio (approved):

documentation/theme/main.html
spencer-codaio: oh that seems off to me but ok will change

PR #1687, opened 2022-01-05
remove hidden option of the pack parameter

huayang-codaio:
https://staging.coda.io/d/Quality-Tracker_d-GJF-DmEUK/My-Issues_sud-l#My-issues_tui9H/r20767&modal=true

A few sanity checks before removing this field:
- packs repo doesn't use this field. 
- the field isn't actually loaded into `coda/modules/common/packs/formula.ts:generateFormulaDefinitions`, which means that it's being used on client at this moment. 
- after trying load the hidden option into client (as in `generateFormulaDefinitions`), here are a few observations:
  - hidden parameters will be hidden from the structured builder (e.g. a button). This behavior seems useless for third-party packs. It's currently only being used by https://github.com/coda/coda/blob/main/modules/common/packs/formula.ts#L147 as far as I can tell. This usage is directly related to packs. 
  - it breaks the browser on regular formulas if a parameter is hidden. hidden parameters still show up and can be updated. snapshot workflow is able to pick up a hidden parameter but browser just seems to stuck there forever. 
 
So, my conclusion is that the removal of hidden field is safe and won't be a regression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - An optional "description" field is added to sync table definition, that will be used to display in the UI.
 - You no longer need to use the `--fetch` flag with `coda execute` to use a real fetcher. Set `--no-fetch` to use a mock fetcher (the old default behavior).
 - OAuth2 authentication now supports a `scopeDelimiter` option for non-compliant APIs that use something other than a space to delimit OAuth scopes in authorization URLs.
+- Deprecated `hidden` field is now fully removed on formula parameter.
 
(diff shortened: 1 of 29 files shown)


huayang-codaio:
not sure if it's worth to be put into the change log since nobody uses it. let me know.

jonathan-codaio (approved):
Nit, mind adding a bullet point to the CHANGELOG just for completeness?

patrick-codaio (approved):

PR #1689, opened 2022-01-06
inject patches in `coda init`

huayang-codaio:
This PR allows `coda init` to patch `mold-source-map` using the one from packs-example (https://github.com/coda/packs-examples/pull/94). This avoids deprecation warnings from this zombie package.

### DIFF
--- a/cli/init.ts
+++ b/cli/init.ts
@@ -8,13 +8,18 @@ const GitIgnore = `.coda.json
 .coda-credentials.json
 `;
 
+function updateMoldSourceMap() {
+  // unfortuanately Windows has no grep.
+  const packageFileName = 'node_modules/mold-source-map/package.json';
+  const lines = fs.readFileSync(packageFileName).toString().split('\n');
+  const validLines = lines.filter(line => !line.includes('"main":'));
+  fs.writeFileSync(packageFileName, validLines.join('\n'));
+}
+
 function addPatches() {
   spawnProcess(`npm set-script postinstall "npx patch-package"`);
 
-  spawnProcess(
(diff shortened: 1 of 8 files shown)


cli/init.ts
```diff
@@ -30,6 +39,9 @@ export async function handleInit() {
     .join(' ');
   spawnProcess(`npm install --save-dev ${devDependencyPackages}`);
 
+  // https://staging.coda.io/d/Quality-Tracker_d-GJF-DmEUK/Bugs_sucBW#Bugs_tuq-W/r20768&modal=true
```
patrick-codaio, line 42: Are you sure you want a link to staging in a public repo?

patrick-codaio (approved):

cli/init.ts
huayang-codaio: ah, good catch. totally forgot this is a public repo..

jonathan-codaio (commented):
I'm having a little trouble keep track of the similar PRs :) Is this one we still want to proceed with?

Seems fine to me.

huayang-codaio:
yeah, this one allows `coda init` to let the developer pack repo to patch packages.

jonathan-codaio (approved):

huayang-codaio:
ok so I have tested 
- node 16, 14.17.3 (which comes with npm@7+) and they worked. 
- node 14.18.2, 14.17.1, 14.13.1 (which comes with npm@6) and upgraded npm to @7. they worked as well. 

at the end of day, it's just a mess to try to auto create patches for developers.

jonathan-codaio:
That makes sense. Thanks for trying!

PR #1690, opened 2022-01-06
bump to node 16

huayang-codaio:
not sure if there's anything preventing us from bumping to node 16. I've seen tickets on NodeJS 16 so that I feel it's time to bump the build env to node 16.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ orbs:
 executors:
   default_build_environment:
     docker:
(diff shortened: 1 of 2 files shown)


jonathan-codaio (commented):
Hmm I think we kept it at 14 because that's what Lambda runs on? Probably safest to keep that in sync.

huayang-codaio:
ah, yes. thanks for calling this out.

PR #1694, opened 2022-01-11
get rid of client-oauth2

huayang-codaio:
according to https://staging.coda.io/d/Quality-Tracker_d-GJF-DmEUK/My-Issues_sud-l#My-issues_tui9H/r20745&modal=true, client-oauth2 doesn't work for some oauth providers since it passes client secret via headers. 

this PR replaces client-oauth2 with direct requests to oauth providers (logic dup'ed from the packs-auth service). 

tested by 
- todoist pack
- fake_v2 pack with google auth. 
- force injecting a refresh and make sure it works.

### DIFF
--- a/test/packs/fake_v2.ts
+++ b/test/packs/fake_v2.ts
@@ -16,7 +16,7 @@ pack.setUserAuthentication({
   },
 
   async getConnectionName(context) {
-    const response = await context.fetcher!.fetch({
+    const response = await context.fetcher.fetch({
       method: 'GET',
       url: 'https://www.googleapis.com/oauth2/v2/userinfo',
       headers: {
@@ -33,7 +33,7 @@ pack.addFormula({
   parameters: [],
   resultType: coda.ValueType.String,
   execute: async ([], context) => {
-    const response = await context.fetcher!.fetch({
+    const response = await context.fetcher.fetch({
       method: 'GET',
       url: 'https://www.googleapis.com/oauth2/v2/userinfo',
       headers: {
--- a/test/packs/fake_v2.ts
+++ b/test/packs/fake_v2.ts
@@ -2,6 +2,48 @@ import * as coda from '../..';
 
 export const pack = coda.newPack();
 
+pack.addNetworkDomain('googleapis.com');
+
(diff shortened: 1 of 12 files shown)


testing/oauth_server.ts
```diff
@@ -33,30 +40,57 @@ export function launchOAuthServerFlow({
...
-    redirectUri: makeRedirectUrl(port),
-    query: additionalParams,
+  const scope = requestedScopes ? requestedScopes.join(scopeDelimiter || ' ') : requestedScopes;
+  const redirectUri = makeRedirectUrl(port);
+  const callback = async (code: string): Promise<TokenCallbackResponse> => {
+    const params = {
+      grant_type: 'authorization_code',
+      code,
+      client_id: clientId,
+      client_secret: clientSecret,
```
ekoleda-codaio, line 50: It looks like this implementation always passes the `client_secret` as a POST form parameter, which doesn't match the production environment. There, it first attempts to pass it as a form parameter, and if that fails tries to pass it in an `Authorization: Basic` header. While long term we may want to make this an explicit option, for now we should likely replicate the logic in prod.

test/packs/fake_v2.ts
```diff
@@ -2,6 +2,48 @@ import * as coda from '../..';
...
+  tokenUrl: 'https://accounts.google.com/o/oauth2/token',
+  scopes: ['https://www.googleapis.com/auth/calendar', 'profile', 'email'],
+  additionalParams: {
+    access_type: 'offline',
+    prompt: 'consent select_account',
+    session: false,
+  },
+
+  async getConnectionName(context) {
+    const response = await context.fetcher!.fetch({
```
jonathan-codaio, line 19: Super nit, don't need the `!`s anymore.

testing/oauth_server.ts
jonathan-codaio: Yeah let's make sure we handle the fallback here.

huayang-codaio: sure we can do that.

jonathan-codaio (approved):

PR #1715, opened 2022-01-14
Add more best practices

ekoleda-codaio:
Preview: https://coda-packs-sdk--1715.com.readthedocs.build/en/1715/guides/best-practices/


docs/guides/best-practices.md
```diff
@@ -6,12 +6,16 @@ While it's easy to get started building a Pack, there are lots of options to exp
...
+- [x] Ensure each formula has a few [examples defined][formulas_examples]. These show up in the generated documentation for the formula.
+- [x] Set a [suggested value][parameter_suggested_value] for required parameters. This allows users to drag the formula from the side panel and see a result immediately.
+- [x] For each formula that has only a single required parameter, consider exposing it as a [column format][column_formats]. This makes it easier for users to use the functionality in a table.
+- [x] Consider the [caching behavior][formulas_caching] of your formulas. Use longer cache times for formulas where the results are deterministic, and shorter cache times for ones that fetch external data.
 
 
 ## Actions
 
 - [x] Follow the guidance on [action naming][actions_naming].
+- [x] Ensure each action formula [returns a result][actions_result]. This allows the doc to react to the action being completed.
```
jonathan-codaio, line 18: It might be worth elaborating on this, e.g. "For an action with no meaningful value to return, simply returning a non-empty value like 'OK' by convention is useful and appropriate."

jonathan-codaio (approved):
i wonder if we should disable GitHub's auto-collapsing of diffs from the hand-written markdown files.

docs/guides/best-practices.md
ekoleda-codaio: This is already included in the section being linked to:

> In some cases there is no meaningful result, in which case the convention is to return the string "OK".

PR #1716, opened 2022-01-14
More narrowly target linguist

ekoleda-codaio:
There is a mixture of handwritten and auto-generated files in the `/docs` directory. Adjust the linguist settings to more narrowly target the generated portions.


jonathan-codaio (approved):
Thanks!

PR #1719, opened 2022-01-14
add script + update readme

spencer-codaio:
add temp script while CI is broken for the relase flow due to the security issue and update some docs 

cc @ekoleda-codaio @coda/packs @jonathan-codaio

### DIFF
--- a/scripts/update-docs.sh
+++ b/scripts/update-docs.sh
@@ -1,7 +1,4 @@
-###############################################################################
-# This is temporary script to update the docs while our CI is disabled due to 
-# security.
-###############################################################################
+# temporary script before we move to CI
 
 # head
 MK_DOCS_SITE_URL="https://head.coda.io/packs/build/latest/" make build-mkdocs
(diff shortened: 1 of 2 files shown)


scripts/update-docs.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# staging
+MK_DOCS_SITE_URL="https://staging.coda.io/packs/build/latest/" make build-mkdocs
+make publish-docs-staging FLAGS=--forceUpload
+
+# prod
+MK_DOCS_SITE_URL="https://coda.io/packs/build/latest/" make build-mkdocs
+make publish-docs-prod FLAGS=--forceUpload
+
+echo "\nFinishing pushing to all environments!"
```
jonathan-codaio, line 18: I would echo in between each just to checkpoint progress.

jonathan-codaio (approved):

scripts/update-docs.sh
```diff
@@ -0,0 +1,19 @@
+###############################################################################
+# This is temporary script to update the docs while our CI is disabled due to 
+# security.
```
jonathan-codaio, line 3: Would avoid calling too much attention in a public repo.

spencer-codaio: ah right i can jst remove

scripts/update-docs.sh
spencer-codaio: sounds good -- there is output but sounds right to make it explicit

PR #1722, opened 2022-01-15
Add constants for working with SVGs and dark mode

oleg-codaio:
Using for https://github.com/coda/coda/pull/66395.

IMO having constants like this would make it easier for us deprecate this feature because we'll explicitly break Packs from compiling if we ever decide to remove those constants.

PTAL @coda/packs


jonathan-codaio (approved):
CC @ekoleda-codaio 

Could also consider putting them in an enum to namespace them and pollute the top-level autocomplete namespace a little less, but I don't feel strongly.

oleg-codaio:
Thoughts on the namespace instead? IMO an enum isn't the right thing to use to group a collection of constants.

jonathan-codaio (approved):

PR #1723, opened 2022-01-15
Fix up inclusion due to mold-source-map issue

oleg-codaio:
Running into an issue here, seeing if it will help:

https://app.circleci.com/pipelines/github/coda/coda/186790/workflows/7654be99-3688-4b20-94ff-03821c8ac2d9/jobs/1023964

PTAL @coda/packs

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,6 @@
     "type": "git",
     "url": "https://github.com/coda/packs-sdk"
   },
-  "scripts": {
-    "postinstall": "npx patch-package"
-  },
   "dependencies": {
     "@aws-crypto/sha256-js": "^2.0.1",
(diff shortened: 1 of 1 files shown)


jonathan-codaio (commented):
FYI @huayang-codaio 

This seemed to be needed downstream for another use case but I don't know the details.

huayang-codaio:
I think I have fixed https://github.com/coda/coda/pull/66447. Seems like yarn.lock was simply not updated. Not related to postinstall (which I don' think actually affects downstream).

huayang-codaio:
okay confirmed that npm 6 will complain about missing files of the patch. https://github.com/coda/packs-sdk/pull/1742 to fix the issue and bump npm version to 0.8.1

PR #1737, opened 2022-01-19
Small doc updates

ekoleda-codaio:
A few small doc updates to incorporate recent feedback from developers:
- Explicitly mention that OAuth 1.0a isn't supported.
- Include a troubleshooting section for when you forget `isAction: true`.
- Add the `any` type to the dynamic sync table template, to avoid TS errors when adding new properties outside of the declaration.


docs/guides/development/troubleshooting.md
```diff
@@ -46,6 +46,14 @@ This error indicates that the domain you provided in an `addNetworkDomain()` cal
 These errors can occur when installing the SDK locally, specifically during the installation of the `isolated-vm` dependency. They indicate that certain required developer tools are not present on the machine. See the [help documentation][isolated_vm_requirements] for this library for more information on how to install these missing packages on your machine.
 
 
+### Invalid action
+
+- `Unable to execute invalid action`
+
+This error will appear at the bottom of the screen after pressing a button, and indicates that the formula used in the button isn't a valid action. In order for a formula to be used as an action within a button, it must be defined with `isAction: true`. See the [Actions guide][actions_create] for more information.
```
jonathan-codaio, line 53: Probably fine for this context, but worth knowing that any time the formula in a button doesn't evaluate to an "action" this same error message will happen. So if you write your own action formula that uses a valid pack action formula but do other bad things with the formula, you could get this same error.

documentation/snippets/dynamic_sync_table.ts
```diff
@@ -20,7 +20,7 @@ pack.addDynamicSyncTable({
   getSchema: async function (context) {
     let datasourceUrl = context.sync.dynamicUrl!;
     // TODO: Fetch metadata about the data source and get the list of fields.
-    let properties = {
+    let properties: any = {
```
jonathan-codaio, line 23: Could probably use `coda.ObjectSchemaProperties` for correctness. Are you intentionally trying to avoid that complexity?

docs/guides/development/troubleshooting.md
ekoleda-codaio: Ya, I tried working in the idea of the "top-level" formula needing to be an action, but couldn't find a wording that felt right. I think this level of detail should be fine for now, but worth watching out for.

jonathan-codaio (approved):

PR #1742, opened 2022-01-19
remove patch and update changelog

huayang-codaio:
production patch will lead to a patch failure if downstream is using npm 6 or earlier. let's remove this patch since it's not a big deal for this repo. 

(alternative is to make the patch dev only but I don't think we care about the deprecated warning inside this repo)


jonathan-codaio (approved):
Good find re: npm 6

oleg-codaio:
Thanks!

PR #1747, opened 2022-01-25
allow fetch request body to be Buffer

huayang-codaio:
https://staging.coda.io/d/Quality-Tracker_d-GJF-DmEUK/Bugs_sucBW#Bugs_tuq-W/r21180&view=modal

The goal is to allow packs send binary in the fetcher request using Buffer.

However, the caveat is that `Buffer` isn't available in the web builder. As I was testing, monaco will assume the type to be any when it doesn't recognize `Buffer`. 

image

While it used to be:

image

Alternatively, we can make the body as `string | Uint8Array` which is available from browser and NodeJS. I need to check if it's supported in isolated-vm though. Thoughts? 

@jonathan-codaio @ekoleda-codaio @coda/packs


huayang-codaio:
Maybe ArrayBuffer or Blob are more standard here. let me give them a shot. https://tr.javascript.info/fetch-basics

ekoleda-codaio:
Whatever we choose, I think it should be bidirectional, and currently Packs receive binary data (`isBinaryResponse`) as a `Buffer`. The source of binary data is mostly `Image` parameters, which currently need to be fetched from the codahosted.io domain, and are returned as a `Buffer`.

huayang-codaio:
After more research, I feel like this is getting bigger in scope. @jonathan-codaio Buffer is used in other places like `BlobStorage` (which isn't working for the web builder as well). I wonder if we want to making a breaking change to use ArrayBuffer for any binary data here so that's compatible for both Node & browser. 

Need to update all the typings in coda repo as well.

huayang-codaio:
I feel like it's a cleaner direction and will bridge the gap between browser and node for binary support (where a few tickets mentioned).

jonathan-codaio:
Yeah I think the issue here is simply typing in Monaco in the browser. The actual runtime type is `Buffer`: packs execute in node and we use `Buffers`, full stop. There isn't actually a browser compatibility issue to bridge. I think it's simply a "missing type def" issue with Monaco, since we're using Monaco with "browser types" or something? Can we just add a fill for a node-compatible `Buffer` type def in Monaco?

huayang-codaio:
to clarify, two directions. 

1. see if we can make monaco supporting a "Buffer" type, maybe by adding a browserify Buffer class here. 
2. replace Buffer with ArrayBuffer. this is a breaking change but the ArrayBuffer is indeed supported by both NodeJs and browser.

alexd-codaio:
FYI you could use `Uint8Array` in the browser, which is a subclass of Buffer per the docs.

image

alexd-codaio:
Also, https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/buffer.d.ts#L531

huayang-codaio:
yes `Uint8Array` is another alternative. but to be clear, I meant to replace Buffer with `Uint8Array` because `Uint8Array` in browser isn't exactly a subclass of Buffer.

huayang-codaio:
@alexd-codaio yeah, all these types work in NodeJS. My goal is to make monaco recognize the type in the browser so that pack builders don't see them as `any` type. Specifically `Buffer` isn't a type in browser yet.

chris-codaio:
Is `Buffer` actually available in the sandbox?

huayang-codaio:
> Is Buffer actually available in the sandbox?

Buffer is not available in the sandbox. 

```
> var context = isolate.createContextSync()
undefined
> context.evalSync('new Buffer()')
Uncaught ReferenceError: Buffer is not defined
    at <isolated-vm>:1:1
    at (<isolated-vm boundary>)
    at REPL27:1:9
```

We are using the browserify buffer in fact.

huayang-codaio:
The standard ArrayBuffer and Uint8Array both are available in isolated vm

```
> context.evalSync('new Uint8Array()')
undefined
> context.evalSync('new ArrayBuffer()')
undefined
```

jonathan-codaio:
What do we think the "right" answer is for developers? Is it to use `Buffer` as the interface because it's what you would expect when working within binary data in node? That seems reasonable to me and reasonable that we polyfill it using browserify in the sandbox.

In which case, I hope all we need to do is add a typedef for `Buffer` to Monaco? We don't need a buffer class to actually execute in the browser. Or the issue that since we bundle the code in the browser, are we not using browserify, and so we don't even get the opportunity to polyfill `Buffer` when using the web editor>

huayang-codaio:
I am not sure what's the "right" answer here. but it's definitely an easier way to avoid a breaking change at this moment. I had a prototype working to inject `Buffer` into the monaco context. This would make things working for now. 

image

jonathan-codaio:
Yeah my preference would be to inject a Buffer typedef within `MonacoEditorContext._tryImportExtraLibs()`. Assuming it's just a type issue, that should work.

I don't know if there are issues at bundle time too though.

huayang-codaio:
https://github.com/coda/coda/pull/66908 should add Buffer to monaco. This PR is ready to go.

jonathan-codaio:
Great! Just confirming with the 2 PRs together, code using buffer methods builds and executes properly?

huayang-codaio:
ah good question. it will build. but I am not sure if it runs. let me check.

huayang-codaio:
okay it throws Buffer is not defined in runtime. hmm.

jonathan-codaio:
Ah yeah, was afraid of that. It's probably not getting browserified then since I think we might only do that in the CLI.

huayang-codaio:
two approaches in my mind:
1. compile buffer into the browser bundle. 
2. inject global Buffer in the pack runtime. 

I think 2) is more elegant since we don't need to put all the dependencies into the browser bundle, while the buffer is in the runtime anyway just not injected into global yet.

huayang-codaio:
This works now. 

image

jonathan-codaio:
Hmm, do we know why `Buffer` isn't included in isolated-vm, is it just because it's not part of the "core" JS they intend to support? @chris-codaio any security concerns jump out about injecting Buffer into the runtime?

To be clear, I think this is the best solution, barring security issues.

huayang-codaio:
Yeah, Buffer is specific to NodeJS I think. not part of core JS. 

To clarify, we already have Buffer in the runtime. It's only not at the global level. (e.g. global.Buffer)

testing/fetcher.ts
```diff
@@ -329,9 +329,9 @@ export class AuthenticatingFetcher implements Fetcher {
             );
           }
 
-          if (bodyWithSubstitutions) {
+          if (typeof bodyWithSubstitutions === 'string') {
             bodyWithSubstitutions = replaceAll(
-              bodyWithSubstitutions,
+              bodyWithSubstitutions as string,
```
oleg-codaio, line 334: Are you sure you need the `as string` here? TS should be able to infer it. Same above.

huayang-codaio: for some reason TS was not able to infer the type. I thought it was VS code but `git push` also complains on the linter part.

huayang-codaio: image

huayang-codaio:
```
huayang@Huayangs-MBP [17:40:15] [~/code/packs-sdk] [hg-buffer-body *]
-> % make build 
testing/fetcher.ts:296:89 - error TS2339: Property 'replace' does not exist on type 'string | Buffer'.
  Property 'replace' does not exist on type 'Buffer'.

296             bodyWithTemplateSubstitutions = ensureExists(bodyWithTemplateSubstitutions).replace(
                                                                                            ~~~~~~~


Found 1 error.

make: *** [compile] Error 2
```

huayang-codaio: actually, the error is for the one above. this one can be removed.

jonathan-codaio (approved):
If we're already including Buffer in the sandbox, then this works for me. @chris-codaio please opine if you have other thoughts, thanks!

chris-codaio:
I'm not aware of any security concern around Buffer polyfill.

PR #1774, opened 2022-02-04
Add lightweight GitHub sample

ekoleda-codaio:


documentation/examples/github/full.ts
```diff
@@ -0,0 +1,159 @@
...
+      url: "https://api.github.com/user",
+    });
+    return response.body.login;
+  }
+});
+
+// A schema that defines a repo object.
+const RepoSchema = coda.makeObjectSchema({
+  properties: {
+    id: { type: coda.ValueType.Number },
```
huayang-codaio, line 36: have we been using extra spaces in curly braces in the ts files in examples? doesn't seem the case in other examples. 

just checking if it's new or old. I think (to me) it's the VS new version defaults to add spaces between curly braces. 

Some settings like this
image

huayang-codaio (approved):

documentation/examples/github/full.ts
ekoleda-codaio: I don't think I've done it consistently, but I've been trying to do it more since that's how the Web IDE auto-formats the code. I'll do a pass on the code in a separate PR to standardize this and add a lint rule.

PR #1775, opened 2022-02-04
Update social metadata

ekoleda-codaio:
Improve the social cards for the documentation. Specifically, Slack wasn't showing the page's title in the card. Updated the social metadata to use the page's title (preferring the longer title in the h1) and the page's custom meta description (if defined).

Before and after:
Screen Shot 2022-02-04 at 10 57 01 AM

### DIFF
--- a/documentation/theme/main.html
+++ b/documentation/theme/main.html
@@ -1,23 +1,36 @@
 {% extends "base.html" %}
 
 {% block extrahead %}
+
 {% set title = config.site_name %}
-{% if page and page.meta and page.meta.title %}
-  {% set title = page.meta.title ~ " - " ~ title %}
+{% if page and page.toc|first is defined %}
+  {% set title = page.toc.items[0].title ~ " - " ~ title %}
+{% elif page and page.meta and page.meta.title %}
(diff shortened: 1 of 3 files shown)


documentation/theme/main.html
```diff
@@ -1,23 +1,39 @@
-{% extends "base.html" %} 
+{% extends "base.html" %}
 
 {% block extrahead %}
-<!-- TODO: should shape these up and should cater to specific page -->
-<meta name="description" content="Documentation for the SDK to build Coda Packs." />
+
+{% set title = config.site_name %}
+{% if page.toc|first is defined %}
```
huayang-codaio, line 6: where is the template language defined? looks like jinja2 template.

huayang-codaio (approved):

documentation/theme/main.html
ekoleda-codaio: Ya, mkdocs uses Jinja2: https://www.mkdocs.org/dev-guide/themes/#creating-a-custom-theme

huayang-codaio: thanks for the pointer!

PR #1776, opened 2022-02-07
Update lint rules for sample code

ekoleda-codaio:
Specifically:

- Require whitespace about curly braces (to match auto-formatting done by the Web IDE).
- Require trailing commas (for consistency).
- Require semicolons (for consistency).

Ran `make lint-fix` to fix all the code that broke these rules.


huayang-codaio (approved):
Thank you!

PR #1782, opened 2022-02-07
restrict coda-api-auth network domains

huayang-codaio:
I am adding network domains to the add-connection UI dialog, except the coda api auth type. In which case, user would assume that it only talks to coda api.

However I realized that this is not enforced in our validation. 

@coda/packs @patrick-codaio

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -936,6 +936,19 @@ function validateFormulas(schema: z.ZodObject<any>) {
       },
       {message: 'A formula namespace must be provided whenever formulas are defined.', path: ['formulaNamespace']},
     )
+    .refine(
+      metadata =>
+        !(
+          metadata.defaultAuthentication?.type === AuthenticationType.CodaApiHeaderBearerToken &&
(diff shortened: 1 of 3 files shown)


testing/upload_validation.ts
```diff
@@ -936,6 +936,19 @@ function validateFormulas(schema: z.ZodObject<any>) {
       },
       {message: 'A formula namespace must be provided whenever formulas are defined.', path: ['formulaNamespace']},
     )
+    .refine(
+      metadata =>
+        !(
+          metadata.defaultAuthentication?.type === AuthenticationType.CodaApiHeaderBearerToken &&
+          metadata.networkDomains?.some(
+            (domain: string) => !['coda.io', 'codahosted.io', 'localhost', 'calc-grpc-proxy'].includes(domain),
```
huayang-codaio, line 944: `'localhost', 'calc-grpc-proxy'` are for the debug pack, which is the only pack that can use a non-public domain.

patrick-codaio (approved):

PR #1789, opened 2022-02-15
add StringEmbedSchema

spencer-codaio:
adding an embed schema for embed values to allow for force embed

validated working in `coda`
image

https://github.com/coda/coda/pull/67937

@coda/packs @jonathan-codaio @alan-codaio  cc @ekoleda-codaio


schema.ts
```diff
@@ -410,6 +410,20 @@ export interface StringDateSchema extends BaseStringSchema<ValueHintType.Date> {
...
 }
 
+export interface StringEmbedSchema extends BaseStringSchema<ValueHintType.Embed> {
+  /** Instructs Coda to render this value as an embed. */
+  codaType: ValueHintType.Embed;
+  /**
+   * Toggle whether to try to force embed the content in Coda. Should be kept to false for most cases.
+   *
+   * By default, we use an external provider that supports and normalizes embeds for different sites. If you are trying
+   * to embed a non-common site or one that is not supported by this provider,
```
jonathan-codaio, line 420: Nit, "uncommon" vs "non-common". Also wonder if we should just say we use Iframely, otherwise how would somebody determine whether their site is supported or not?

spencer-codaio: oops thanks lol. 

yeah that's a good call. was curious to hear @ekoleda-codaio's thoughts we had another thread about sharing process of asking iframely to support providers. was wondering if we could link to that page here but wasnt sure where we were putting it

ekoleda-codaio: I do plan to have a page on embeds, including info about iframely, but it doesn't exist yet. Perhaps just mention iframely specifically now, and I'll replace it with a link to that guide when it exists.

spencer-codaio: sounds good!

jonathan-codaio (approved):

PR #1790, opened 2022-02-17
Add content to support page

ekoleda-codaio:


docs/support.md
```diff
@@ -1,3 +1,43 @@
...
+    <a href="https://community.coda.io/new-topic?category=developers-central/coda-packs" class="md-button md-button--primary">Post a new question</a>
+  </p>
+</form>
+
+
+## Problems or bugs
+
+If you are experiencing a problem with your Pack or have identified a bug you should reach out to the Coda Support team. Send an email to [help+packs@coda.io][support_email], and include the following information:
+
+- The ID of your Pack (visible in the Pack Studio URL).
```
jonathan-codaio, line 32: Could also just say to send the url of your pack (packs studio url and gallery listing url would both have the id in it). Seems people generally have a hard time figuring out how mentally to extract content from urls even if it seems obvious.

jonathan-codaio (approved):

docs/support.md
ekoleda-codaio: Good call, updated to mention the Pack Studio URL to make it explicit.

PR #1793, opened 2022-02-18
Add GA configuration

ekoleda-codaio:
Needed to track metrics for 3.0.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -71,6 +71,11 @@ extra:
     - icon: fontawesome/brands/facebook
       link: https://www.facebook.com/codahq/
 
+  # Analytics configuration.
+  analytics:
+    provider: google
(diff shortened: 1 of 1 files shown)


jonathan-codaio:
Ah, this was a critical point when discussing how we were going to host these docs.

My understanding is that in order to put GA tags on the page, we have to show the cookie consent warning. I believe that is not implemented on these static pages, so we can't use GA yet. This was a conscious decision.

CC @harisiva @spencer-codaio to confirm my understanding here

harisiva-codaio:
@jonathan-codaio -- that's right. I assumed that we had the cookie banner on this page already; my bad. We can add the banner, and then turn on GA collection.

ekoleda-codaio:
Understood. Closing this PR while we figure out next steps.

PR #1796, opened 2022-02-23
Add ValueHintType.Email for new Email column type

hewitt-codaio:


index.ts
```diff
@@ -136,11 +136,15 @@ export {CurrencyFormat} from './schema';
 export type {CurrencySchema} from './schema';
 export type {DurationSchema} from './schema';
 export {DurationUnit} from './schema';
+export type {EmailDisplayType} from './schema';
```
jonathan-codaio, line 139: Nit, I think you want to do a regular export on these two enums, so their values can be referenced too.

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+
+/**
+ * A schema representing a return value or object property that is an email address.
+ *
+ * An email address can be represented visually as an icon, an icon plus the email address, or
+ * the just the email address.  Auto-complete against previously typed domain names is
+ * also an option in the user interface.
+ */
+export interface EmailSchema extends BaseStringSchema<ValueHintType.Email> {
+  /** Instructs Coda to render this value as a scale. */
```
jonathan-codaio, line 428: Nit, copy/paste bug `as an email`.

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+   * Display email address only.
+   */
+  EmailOnly = 'emailOnly',
+}
+
+/**
+ * A schema representing a return value or object property that is an email address.
+ *
+ * An email address can be represented visually as an icon, an icon plus the email address, or
+ * the just the email address.  Auto-complete against previously typed domain names is
```
jonathan-codaio, line 424: Total nit, I think we use `autocomplete` without a dash elsewhere in the documentation.

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+ * Display types that can be used with a {@link LinkSchema}.
+ */
+export enum LinkDisplayType {
+  /**
+   * Display icon only.
+   */
+  IconOnly = 'IconOnly',
+
+  /**
+   * Display url.
```
jonathan-codaio, line 446: Nit, documentation style guide is to use `URL` capitalized.

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+}
+
+/**
+ * Display types that can be used with a {@link LinkSchema}.
+ */
+export enum LinkDisplayType {
+  /**
+   * Display icon only.
+   */
+  IconOnly = 'IconOnly',
```
jonathan-codaio, line 443: These values should probably match the `coda` values exactly (which I think use `iconOnly`, `url`, etc).

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+  Card = 'Card',
+
+  /**
+   * Display the referenced web page as an embed.
+   */
+  Embed = 'Embed',
+}
+
+/**
+ * A schema representing a return value or object property that is an hyperlink.
```
jonathan-codaio, line 467: Nit, `a hyperlink`.

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+ *
+ * The link can be displayed in the UI in five different ways, as per the above enumeration.
```
jonathan-codaio, line 469: Total nit, I'd probably just say `in multiple ways` so the comment doesn't become stale if we add more.

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+ */
+export interface LinkSchema extends BaseStringSchema<ValueHintType.Url> {
+  /** Instructs Coda to render this value as a scale. */
```
jonathan-codaio, line 472: Nit, `as a hyperlink`.

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+  codaType: ValueHintType.Url;
+  /** How the url should be displayed in the UI. */
```
jonathan-codaio, line 474: Nit, `URL` all caps.

jonathan-codaio (commented):
Awesome! Thanks so much for this. Mostly just documentation nits. We're disciplined about the doc comments since they generate what appears at https://coda.io/packs/build/latest/reference/sdk/

jonathan-codaio:
And I think if you just run `make build` after you're done with your final changes, the CI should be happy.

schema.ts
hewitt-codaio: Thank you for the nudge to verify.  I believe in this case they are capitalized, though.  Declarations [here](https://github.com/coda/coda/blob/8cb3fa17d3605a000c98f4429501e4c3bcb3bc74/modules/common/serialized-types/formats.ts#L91).

hewitt-codaio:
> And I think if you just run `make build` after you're done with your final changes, the CI should be happy.

My `make build` is failing at the end, possibly related to my git origin, which may explain the CI failures.  Will look into it more tomorrow.

```
Writing /Users/coda/code/packs-sdk/index.ts -> /Users/coda/code/packs-sdk/dist/bundle.d.ts
Checking generated files...
Compiler option "skipLibCheck" is disabled to properly check generated output
Done in 7.11s
isolated-vm version matches, skipping.
Please config your git origin to git@github.com:coda/packs-sdk.git
make: *** [typedoc] Error 1
```

jonathan-codaio (commented):
Ah yes, regarding `make build`, it turns out the documentation generating library uses the git origin to generate link urls (or at least it did at one point) so that check is in place to make sure the links don't break. You may have an origin from before the organization rename, if you update it it will probably start working.

jonathan-codaio:
Oh and last tiny thing, let's bump the version in package.json to `0.9.0` and add an entry to CHANGELOG.md that briefly mentions the existence of ValueHintType.Email and that ValueHintType.Url will now create Link columns, a minor but technically breaking change.

schema.ts
```diff
@@ -392,6 +397,84 @@ export interface ScaleSchema extends BaseNumberSchema<ValueHintType.Scale> {
...
+
+  /**
+   * Display the referenced web page as a card.
+   */
+  Card = 'Card',
+
+  /**
+   * Display the referenced web page as an embed.
+   */
+  Embed = 'Embed',
```
spencer-codaio, line 463: hmm not for this PR to address but wondering if this should be consolidated with the Embed type..

schema.ts
hewitt-codaio: Apologies for the inconsistency - endeavoring to follow "when in Rome" in both places, but the Romans unfortunately don't agree in this case.  Will make the schema match our API.

Does the [x-tsEnumNames here](https://github.com/coda/coda/blob/d55a47d9efd4f9a6394c62e8fa87eb56fce9d176/modules/server/api/oas/v1/schemas/columns.yaml#L236) accomplish the mapping, or must it be done elsewhere?

jonathan-codaio: The mapping from packs SDK values back to Coda internal values happens in this file, you can see where it assumes the enum values are the same by using castEnum() in a few places:

https://github.com/coda/coda/blob/main/modules/common/packs-helpers/index.ts#L340

So the new switch case for the new ValueHintType would need to explicitly map each enum value instead.

(x-tsEnumNames is a little thing that just generates the TypeScript enum name in generated code, the `Foo` in `Foo = 'bar'` in the enum declaration)

jonathan-codaio (approved):

PR #1802, opened 2022-03-01
Add "link" command to associate with an existing Pack ID.

dweitzman-codaio:
Updates .coda-pack.json to point at a pre-existing Pack that was created in the browser.

### DIFF
--- a/cli/create.ts
+++ b/cli/create.ts
@@ -42,7 +42,7 @@ export async function createPack(
   // flow if they don't have a Coda API token.
   const apiKey = getApiKey(codaApiEndpoint);
   if (!apiKey) {
-    printAndExit('Missing API key. Please run `coda register <apiKey>` to register one.');
+    printAndExit('Missing API key. Please run `coda register` to register one.');
   }
 
   if (!fs.existsSync(manifestFile)) {
--- a/cli/link.ts
+++ b/cli/link.ts
@@ -22,8 +22,6 @@ interface LinkArgs {
 
 export async function handleLink({manifestDir, codaApiEndpoint, packIdOrUrl}: ArgumentsCamelCase<LinkArgs>) {
   const formattedEndpoint = formatEndpoint(codaApiEndpoint);
-  // TODO(dweitzman): Redirect to the `coda register` if there's not
-  // an existing Coda API token.
   // TODO(dweitzman): Add a download command to fetch the latest code from
   // the server and ask people if they want to download after linking.
   const apiKey = getApiKey(codaApiEndpoint);
--- a/cli/link.ts
+++ b/cli/link.ts
@@ -3,26 +3,21 @@ import {createCodaClient} from './helpers';
 import {formatEndpoint} from './helpers';
 import {getApiKey} from './config_storage';
 import {getPackId} from './config_storage';
+import {isResponseError} from '../helpers/external-api/coda';
 import {printAndExit} from '../testing/helpers';
 import {promptForInput} from '../testing/helpers';
 import {storePackId} from './config_storage';
 
 // Regular expression that matches coda.io/p/<packId> or <packId>.
-const PackUrlRegex = /^(?:https:\/\/(?:[^/]*)coda.io(?:\:[0-9]+)?\/p\/)?([0-9]+)(:?[^0-9]*)$/;
+const PackEditUrlRegex = /^https:\/\/(?:[^/]*)coda.io(?:\:[0-9]+)?\/p\/([0-9]+)(:?[^0-9].*)?$/;
+const PackGalleryUrlRegex = /^https:\/\/(?:[^/]*)coda.io(?:\:[0-9]+)?\/packs\/[^/]*-([0-9]+)$/;
+const PackPlainIdRegex = /^([0-9]+)$/;
+const PackRegexes = [PackEditUrlRegex, PackGalleryUrlRegex, PackPlainIdRegex];
 
 interface LinkArgs {
   manifestDir: string;
   codaApiEndpoint: string;
   packIdOrUrl: string;
-}
-
-export function parsePackIdOrUrl(packIdOrUrl: string): number | null {
-  const match = packIdOrUrl.match(PackUrlRegex);
-  if (!match) {
-    return null;
-  }
-  const matchedNumber = match[1];
-  return Number(matchedNumber);
 }
 
 export async function handleLink({manifestDir, codaApiEndpoint, packIdOrUrl}: ArgumentsCamelCase<LinkArgs>) {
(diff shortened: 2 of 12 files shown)


cli/link.ts
```diff
@@ -0,0 +1,65 @@
...
+import {createCodaClient} from './helpers';
+import {formatEndpoint} from './helpers';
+import {getApiKey} from './config_storage';
+import {getPackId} from './config_storage';
+import {printAndExit} from '../testing/helpers';
+import {promptForInput} from '../testing/helpers';
+import {storePackId} from './config_storage';
+
+// Regular expression that matches coda.io/p/<packId> or <packId>.
+const PackUrlRegex = /^(?:https:\/\/(?:[^/]*)coda.io(?:\:[0-9]+)?\/p\/)?([0-9]+)(:?[^0-9]*)$/;
```
jonathan-codaio, line 11: For usability, I think we should support pack gallery urls too, so users don't have to guess which kind of url they should provide. Those look like https://coda.io/packs/mermaid-10537. Could just have an array of regexs or something.

cli/link.ts
```diff
@@ -0,0 +1,65 @@
...
+export function parsePackIdOrUrl(packIdOrUrl: string): number | null {
+  const match = packIdOrUrl.match(PackUrlRegex);
+  if (!match) {
+    return null;
+  }
+  const matchedNumber = match[1];
+  return Number(matchedNumber);
+}
+
+export async function handleLink({manifestDir, codaApiEndpoint, packIdOrUrl}: ArgumentsCamelCase<LinkArgs>) {
```
jonathan-codaio, line 28: Nit, would put this above parsePackIdOrUrl so it's the first function in the file, by convention we try to keep the "public" thing up top so you don't have to scan down too far to find the meat of the file.

cli/link.ts
```diff
@@ -0,0 +1,65 @@
...
+  const formattedEndpoint = formatEndpoint(codaApiEndpoint);
+  // TODO(dweitzman): Redirect to the `coda register` if there's not
+  // an existing Coda API token.
+  // TODO(dweitzman): Add a download command to fetch the latest code from
+  // the server and ask people if they want to download after linking.
+  const apiKey = getApiKey(codaApiEndpoint);
+  if (!apiKey) {
+    printAndExit('Missing API key. Please run `coda register <apiKey>` to register one.');
```
jonathan-codaio, line 36: Nit, would probably leave out `<apiKey>` from the message--if you run it without an API key, it prompts you with what you need to do and opens a browser tab with a deep link to generate a well-scoped key.

cli/link.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  const codaClient = createCodaClient(apiKey, formattedEndpoint);
+  const packId = parsePackIdOrUrl(packIdOrUrl);
+  if (packId === null) {
+    printAndExit(`packIdOrUrl must be a pack ID or URL, not ${packIdOrUrl}`);
```
jonathan-codaio, line 42: Huh, TypeScript must have recently gotten smarter about inferring type narrowing in cases like this. You used to need to add a `return` here otherwise the code below wouldn't have been properly narrowed to know that `packId` is non-null. I think that since `printAndExit()` is typed as returning `never`, a recent version of TS is able to narrow the type even without a return.

Regardless, I'd recommend using `return` in all of these cases to make it clearer that these lines are terminal.

cli/link.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  // Verify that the user has edit access to the pack. Currently only editors
+  // can call getPack().
+  await codaClient.getPack(packId);
```
jonathan-codaio, line 47: If the user doesn't have access, does this throw? We should catch that and print an actionable error message rather than showing a stack trace to the user.

jonathan-codaio (commented):
Amazing!

jonathan-codaio (approved):
Nice!

cli/create.ts
```diff
@@ -42,7 +42,7 @@ export async function createPack(
   // flow if they don't have a Coda API token.
   const apiKey = getApiKey(codaApiEndpoint);
   if (!apiKey) {
-    printAndExit('Missing API key. Please run `coda register <apiKey>` to register one.');
```
huayang-codaio, line 45: nice

cli/link.ts
```diff
@@ -0,0 +1,83 @@
...
+  codaApiEndpoint: string;
+  packIdOrUrl: string;
+}
+
+export async function handleLink({manifestDir, codaApiEndpoint, packIdOrUrl}: ArgumentsCamelCase<LinkArgs>) {
+  const formattedEndpoint = formatEndpoint(codaApiEndpoint);
+  // TODO(dweitzman): Redirect to the `coda register` if there's not
+  // an existing Coda API token.
+  // TODO(dweitzman): Add a download command to fetch the latest code from
+  // the server and ask people if they want to download after linking.
```
huayang-codaio, line 28: I am not sure if we upload the original source code with CLI upload. We might just upload the compiled bundle. 

But if this is the case where a developer wants to convert to CLI from web editor, this can be a great feature.

cli/link.ts
```diff
@@ -0,0 +1,83 @@
...
+interface LinkArgs {
+  manifestDir: string;
+  codaApiEndpoint: string;
+  packIdOrUrl: string;
+}
+
+export async function handleLink({manifestDir, codaApiEndpoint, packIdOrUrl}: ArgumentsCamelCase<LinkArgs>) {
+  const formattedEndpoint = formatEndpoint(codaApiEndpoint);
+  // TODO(dweitzman): Redirect to the `coda register` if there's not
+  // an existing Coda API token.
```
huayang-codaio, line 26: nits. probably remove this TODO. the solution in this PR (e.g. print&exit) seems good enough.

cli/link.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  const existingPackId = getPackId(manifestDir, codaApiEndpoint);
+  if (existingPackId) {
+    if (existingPackId === packId) {
+      return printAndExit(`Already associated with pack ${existingPackId}. No change needed`, 0);
+    }
+
+    const input = promptForInput(
+      `Overwrite existing deploy to pack https://coda.io/p/${existingPackId} with https://coda.io/p/${packId} instead? Press "y" to overwrite or "n" to cancel: `,
```
huayang-codaio, line 63: like this.

huayang-codaio (approved):
congrats on your first packs-sdk PR!

cli/link.ts
dweitzman-codaio: Done

PR #1823, opened 2022-03-01
"whoami" command for determining the owner of the API token

dweitzman-codaio:
Test plan:

Ran `npx ts-node ./cli/coda.ts whoami`, `npx ts-node ./cli/coda.ts whoami <valid token>`, and  `npx ts-node ./cli/coda.ts whoami <invalid token>`

I haven't tested a variety of token types (scoped vs non-scoped)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### 0.8.3
 
-- Added `whoami` command to see API token details.
-- Added `link` command to set up upload for an existing Pack ID instead of creating a new one.
+- Added `coda whoami` command to see API token details.
+- Added `coda link` command to set up upload for an existing Pack ID instead of creating a new one.
 
 ### 0.8.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
+### 0.8.3
+
+- Added `whoami` command to see API token details.
+- Added `link` command to set up upload for an existing Pack ID instead of creating a new one.
+
 ### 0.8.2
+
 - Added `StringEmbedSchema` to handle configuration on how embed values appear in docs
 
 ### 0.8.1
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -86,7 +86,7 @@ if (require.main === module) {
     })
     .command({
       command: 'whoami [apiToken]',
-      describe: 'Looks up the api token owner',
+      describe: 'Looks up information about the API token that is registered in this environment',
       builder: {
         codaApiEndpoint: {
           string: true,
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -14,6 +14,7 @@ import {handleRelease} from './release';
 import {handleSetOption} from './set_option';
 import {handleUpload} from './upload';
 import {handleValidate} from './validate';
+import {handleWhoami} from './whoami';
 import yargs from 'yargs';
 
 if (require.main === module) {
@@ -82,6 +83,18 @@ if (require.main === module) {
         },
       },
       handler: handleRegister,
+    })
+    .command({
+      command: 'whoami [apiToken]',
+      describe: 'Looks up the api token owner',
+      builder: {
+        codaApiEndpoint: {
+          string: true,
+          hidden: true,
+          default: DEFAULT_API_ENDPOINT,
+        },
+      },
+      handler: handleWhoami,
     })
     .command({
       command: 'build <manifestFile>',
--- a/cli/whoami.ts
+++ b/cli/whoami.ts
@@ -1,0 +1,37 @@
+import type {ArgumentsCamelCase} from 'yargs';
+import {createCodaClient} from './helpers';
+import {formatEndpoint} from './helpers';
+import {getApiKey} from './config_storage';
+import {isResponseError} from '../helpers/external-api/coda';
+import {printAndExit} from '../testing/helpers';
+import {tryParseSystemError} from './errors';
+
+interface WhoamiArgs {
+  apiToken?: string;
+  codaApiEndpoint: string;
+}
+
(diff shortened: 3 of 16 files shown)


cli/coda.ts
```diff
@@ -83,6 +84,18 @@ if (require.main === module) {
       },
       handler: handleRegister,
     })
+    .command({
+      command: 'whoami [apiToken]',
```
huayang-codaio, line 88: seems like this should be `whoami <apiToken>` since it's optional.

huayang-codaio (approved):
lgtm

huayang-codaio:
nits. it's always good to attach a test plan in the PR description.

cli/coda.ts
jonathan-codaio: I think `[ ]` does indicate optional args.

cli/coda.ts
```diff
@@ -83,6 +84,18 @@ if (require.main === module) {
...
+      describe: 'Looks up the api token owner',
```
jonathan-codaio, line 89: We're pretty disciplined about language/style/punctuation in developer-facing documentation. Would probably do `Looks up information about the API token that is registered in this environment.`

cli/whoami.ts
```diff
@@ -0,0 +1,37 @@
...
+  apiToken?: string;
+  codaApiEndpoint: string;
+}
+
+export async function handleWhoami({apiToken, codaApiEndpoint}: ArgumentsCamelCase<WhoamiArgs>) {
+  const formattedEndpoint = formatEndpoint(codaApiEndpoint);
+  if (!apiToken) {
+    apiToken = getApiKey(codaApiEndpoint);
+    if (!apiToken) {
+      return printAndExit('Missing API key. Please run `coda register` to register one.');
```
jonathan-codaio, line 19: Nit, let's be consistent about "key" vs "token" in developer-facing language, so would use "token" here.

cli/whoami.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+  }
+
+  const client = createCodaClient(apiToken, formattedEndpoint);
+
+  try {
+    const {name, loginId, workspace} = await client.whoami();
+
+    printAndExit(`You are ${name} (${loginId}) in workspace ${workspace.id}`, 0);
```
jonathan-codaio, line 28: Is it returning a workspace for you in practice? I can't remember exactly but I thought maybe the token would only include a workspace if it's scoped to a doc. I don't think a workspace makes any sense if it's a general API token or scoped to a pack.

We should also print info about the scope: is it read-only vs read-write. Is it scoped to all packs or just a specific pack id. Is it scoped to something doc-related, in which case indicate that the token cannot be used with the packs CLI.

And nit, `return`.

jonathan-codaio (commented):
Nice.

cli/coda.ts
huayang-codaio: oh my mistake.

cli/whoami.ts
dweitzman-codaio: Updated here and for all the other places this error appears (create, link, release, upload)

cli/coda.ts
dweitzman-codaio: Changed the wording but didn't include the trailing period for consistency with all the other one-sentence command descriptions

cli/whoami.ts
dweitzman-codaio: `whoami` always returns an workspace but doesn't return information about what a token is scoped to or whether it's read-only vs read-write:

https://github.com/coda/coda/blob/b446d841c21cf3d078fc2cb4da7051af2d9c9b5f/modules/server/api/private/actions/v1/whoami.ts#L49

Can I determine any of that scoping information by parsing the token itself, and if so would it be appropriate to parse the token in client-side SDK code?

huayang-codaio: we don't seem to have an API doing that. the packs-sdk itself doesn't have the creds to decode the API token.

cli/coda.ts
dweitzman-codaio: I essentially copied the `register` command for that format. Seems like it's not super intuitive to a user at a glance how `<requiredArgument>` and `[optionalArgument]` in the yargs library work

dweitzman-codaio:
> nits. it's always good to attach a test plan in the PR description.

Done, I think. not sure where exactly the edit has to be for github to incorporate it into the squashed description

cli/whoami.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const errors = [`Unexpected error while checking owner of API token: ${err}`, tryParseSystemError(err)];
+    return printAndExit(errors.join('\n'));
+  }
+}
+
+export function formatWhoami(user: PublicApiUser) {
+  const {name, loginId, tokenName} = user;
+
+  return `You are ${name} (${loginId}) using token "${tokenName}"`;
```
jonathan-codaio, line 42: Ah right, we only expose whether it is scoped, and not what the scope is. I can't remember why we do this. What you have here is good, let's also just note if the token is scoped or not.

dweitzman-codaio: Added a "scoped" or "non-scoped" adjective before the word "token" and updated the corresponding test

dweitzman-codaio:
Updated CHANGELOG.md and bumped the package version

CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
+### 0.8.3
+
+- Added `whoami` command to see API token details.
+- Added `link` command to set up upload for an existing Pack ID instead of creating a new one.
```
jonathan-codaio, line 4: Nit, I would do `coda whoami` and `coda link` for clarity--there's a lot of functionality in the SDK beyond just the CLI tool so it will be helpful to be specific.

jonathan-codaio (approved):
Nice!

CHANGELOG.md
dweitzman-codaio: Done

PR #1835, opened 2022-03-04
make deploy docs available on any commit to main

spencer-codaio:
realizing the `release` branch flow is a big pain to work with and with Github tags/releases that we're going to start using to tag new NPM releases i don't see the need for it. 

Instead, the new flow gives the option for any commit on main to deploy docs, as long as you first approve the doc deploy flow. It still maintains the extra approval to deploy to prod so you can manually verify on head/staging that docs are updated appropriately.

cc @coda/packs  @ekoleda-codaio @jonathan-codaio

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,6 +180,16 @@ jobs:
 workflows:
   version: 2
 
+  commit-validation:
+    jobs:
+      - build:
+          context:
+            - dockerhub
+          filters:
+            branches:
+              ignore:
+                - main
+
   commit-validation-and-deploy-docs:
     jobs:
       - build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,36 +179,31 @@ jobs:
 
 workflows:
   version: 2
-  commit_validation:
-    jobs:
-      - build:
-          context:
-            - dockerhub
-          filters:
-            branches:
-              ignore:
-                - release
 
-  build-and-deploy-docs:
+  commit-validation-and-deploy-docs:
     jobs:
       - build:
           context:
             - dockerhub
-          filters:
-            branches:
-              only:
-                - release
+          # filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -179,36 +179,31 @@ jobs:
...
+  commit-validation-and-deploy-docs:
     jobs:
       - build:
           context:
             - dockerhub
-          filters:
-            branches:
-              only:
-                - release
+          # filters:
```
spencer-codaio, line 188: remove this before merging after test

spencer-codaio:
image
this is what it'll look like on `main`. I omitted the slack notification for the first one so as to not spam everyone making commits to main as this should only be used for a small # of merges into maini

patrick-codaio (approved):
So if I'm understanding this right, on every commit to main, circle rebuilds the docs, but doesn't deploy them anywhere until someone finds the circle job page, and even then it only pushes to head/staging until a second confirmation. I like it.

spencer-codaio:
@patrick-codaio so actually it wont even build the docs until the first confirmation, and then it only pushes to prod with the second manual confirmation!

the only thing it does without any confirmation is the normal "make build, validate no changes" flow

jonathan-codaio:
So for the 90% of changes to `main` that don't need to deploy docs, the Circle workflow just sits there dangling forever? Just wondering if that's ok, and when Circle would know it can garbage collect the workflow and its artifacts. Seems worth checking with internal tools folks about this.

spencer-codaio:
@jonathan-codaio yeah.. i just realized this build step is stuck lol... gonna check to see how to make that not required

spencer-codaio:
also oops i realized i removed the normal validation for every branch that's not main fixing that too

spencer-codaio:
image
validated with ITI that it is prefectly fine

PR #1842, opened 2022-03-11
Delete errant file

ekoleda-codaio:
Some cruft that I must have accidentally committed during a PR.


spencer-codaio (approved):
goodbye `~` it was nice knowing ya

PR #1845, opened 2022-03-17
Add a type map for AuthenticationType to AuthenticationMetadata

dweitzman-codaio:


compiled_types.ts
```diff
@@ -54,6 +66,32 @@ export type AuthenticationMetadata = DistributiveOmit<
...
+  T,
+  'getConnectionName' | 'getConnectionUserId' | 'postSetup'
+> & {
+  getConnectionName?: MetadataFormulaMetadata;
+  getConnectionUserId?: MetadataFormulaMetadata;
+  postSetup?: PostSetupMetadata[];
+};
+
+/** @hidden */
+export interface AuthenticationMetadataTypeMap {
```
jonathan-codaio, line 80: While you're here, mind also just adding a sibling type map that maps from the type to the full interface, e.g. `[AuthenticationType.QueryParamToken]: QueryParamTokenAuthentication;` just so we have that for future convenience?

dweitzman-codaio: Added in types.ts, which seemed like the logical place

jonathan-codaio (approved):

PR #1848, opened 2022-03-18
Document column formats in the canvas

ekoleda-codaio:


docs/guides/blocks/column-formats.md
```diff
@@ -98,6 +98,15 @@ Currently only URL patterns are fully supported, and the Pack must declare a cor
 Only Packs already installed in the document will have their matchers used. If multiple column formats both match the cell input then Coda will choose one arbitrarily. If the user doesn't want to use the column format they can manually change it afterwords.
 
 
+## Link formatting
+
+If your column format has [matchers](#matchers) defined it can also affect how links are displayed on the canvas. When editing the display settings for a link, if it matches a column format there will be a new option to display it using the Pack.
+
+<img src="../../../images/column_format_canvas.png" srcset="../../../images/column_format_canvas_2x.png 2x" class="screenshot" alt="Selecting the account in the column format options">
```
huayang-codaio, line 105: would it help to add an image for connected pack?

huayang-codaio (approved):

docs/guides/blocks/column-formats.md
ekoleda-codaio: Do you mean what the result looks like after clicking **Connect to Todoist**?

huayang-codaio: Yeah. Just wondering.

ekoleda-codaio: I tried it out, but I don't think it adds much value, and will end up being quite different depending on the Pack. Thanks for the suggestion though!

PR #1851, opened 2022-03-21
update workflow to run via manual trigger

spencer-codaio:
also allows for release type input

apparently it doesn't show up in the "Actions" tab unless you configure it to run on `workflow_dispatch`

### DIFF
--- a/.github/actions/release.yml
+++ b/.github/actions/release.yml
@@ -7,9 +7,9 @@ on:
         required: true
         type: choice
         options:
-        - patch
-        - minor
-        - major 
+          - patch
+          - minor
+          - major 
 
 jobs:
   release:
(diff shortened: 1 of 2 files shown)


.github/actions/release.yml
```diff
@@ -1,3 +1,16 @@
+
+on: 
+  workflow_dispatch:
+    inputs:
+      increment:
+        description: 'Release version increment'     
+        required: true
+        type: choice
+        options:
+        - patch
```
jonathan-codaio, line 10: Nit, I think these need to be indented?

spencer-codaio: oop done

jonathan-codaio (approved):

PR #1852, opened 2022-03-21
Document array of objects

ekoleda-codaio:
Multiple users have been confused about how to return an array of objects. Added this to the Data types guide, plus:

- Included sections for `String`, `Number`, and `Boolean` for completeness.
- Moved the `Object` section earlier, so I had already discussed object before I talk about arrays of objects.

### DIFF
--- a/docs/guides/basics/data-types.md
+++ b/docs/guides/basics/data-types.md
@@ -39,28 +39,54 @@ For a formula the value type is specified in the `resultType` property, and for 
 [View all types][ValueType]{ .md-button }
 
 
-### Arrays
+### Strings
 
-Lists of data can be returned using the [`Array`][Array] value type. You must also specify the value type of the items in the array, using the `items` property.
+Text values can be returned as [JavaScript strings][mdn_string] using the [`String`][String] value type.
 
 ```ts
 pack.addFormula({
   // ...
-  resultType: coda.ValueType.Array,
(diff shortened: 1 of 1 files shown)


docs/guides/basics/data-types.md
```diff
@@ -39,28 +39,54 @@ For a formula the value type is specified in the `resultType` property, and for
...
+    return true;
+  },
+});
+```
 
 
 ### Objects
 
-Structured data can be returned using the [`Object`][Object] value type. These objects must conform to an existing schema, as specified in the `schema` property. See the [Schemas guide][schemas] for more information about defining and using schemas.
+Structured data can be returned as [JavaScript objects][mdn_object] using the [`Object`][Object] value type. These objects must conform to an existing schema, as specified in the `schema` property. See the [Schemas guide][schemas] for more information about defining and using schemas.
```
jonathan-codaio, line 89: "existing schema" is maybe a bit confusing--is this trying to say a "known schema" or "explicitly-defined schema", i.e. "you can't just return an object you also have to describe the schema"? Would maybe be a bit more specific.

jonathan-codaio (approved):

docs/guides/basics/data-types.md
ekoleda-codaio: Changed to "defined schema", which I think captures it OK.

PR #1855, opened 2022-03-28
allow multiple properties to reuse the same fromKey

huayang-codaio:
https://staging.coda.io/d/Quality-Tracker_d-GJF-DmEUK/Bugs_sucBW#Bugs_tuq-W/r19767&view=modal

nothing special. probably just a use case that we never considered. 

@coda/packs @jonathan-codaio


bundles/thunk_bundle.js
```diff
@@ -4300,6 +4300,9 @@ module.exports = (() => {
   init_buffer_shim();
   var import_clone = __toESM(require_clone());
 
+  // helpers/object_utils.ts
+  init_buffer_shim();
```
jonathan-codaio, line 4304: Any idea why this changed?

jonathan-codaio (approved):
For some reason I always find the key-mapping code the hardest to understand thing. But yeah makes sense and glad there's testing of it.

bundles/thunk_bundle.js
huayang-codaio: I am not sure. I tried make bs and rebase and nothing was changed. I can play with the revision history a bit before merging this.

huayang-codaio: oh turns out whenever we import something `testing/injections/buffer_shim.js` will be injected and that leads to `init_buffer_shim`. 

it's still a bit weird why `dist/bundles/thunk_bundle.js` is adding that line. could be an esbuild feature or that we are not using esbuild inject properly. but I can't quickly find a way to avoid this line.

PR #1856, opened 2022-03-28
make default parameter to work in CLI

huayang-codaio:
https://staging.coda.io/d/Quality-Tracker_d-GJF-DmEUK/Bugs_sucBW#Bugs_tuq-W/r19740&view=modal

before:

```
-> % ./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake_v2.ts formulaWithOptionalParameter                                   
{ Name: undefined }
```

after:

```
-> % ./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake_v2.ts formulaWithOptionalParameter
{ Name: 'Bob' }
```

@jonathan-codaio @coda/packs

### DIFF
--- a/dist/testing/coercion.js
+++ b/dist/testing/coercion.js
@@ -9,21 +9,28 @@ function coerceParams(formula, args) {
     const { parameters, varargParameters } = formula;
     const coerced = [];
     let varargIndex = 0;
-    for (let i = 0; i < args.length; i++) {
+    for (let i = 0; i < parameters.length; i++) {
+        const paramValue = args[i];
         const paramDef = parameters[i];
-        if (paramDef) {
-            coerced.push(coerceParamValue(paramDef, args[i]));
+        if (paramValue !== undefined) {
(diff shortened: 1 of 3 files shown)


jonathan-codaio (commented):
I flagged in the bug for @ekoleda-codaio to weigh in on whether this change is actually desirable, given that the "default" value isn't really a default in practice.

huayang-codaio:
oh do we not use default value in optional parameters actually?

ekoleda-codaio:
Copying my note from the bug:

> Ya, I wouldn’t vote for implementing this as described. This does seem like another case where the name defaultValue set wrong expectations. If the parameter was called suggestedValue I doubt people would expect that behavior on the CLI. Does underscore the need to refactor this field though.

huayang-codaio:
oh so defaultValue is indeed suggestedValue. I just realized it. 

let's close that ticket. the real ticket might be just renaming defaultValue to suggested value.

PR #1859, opened 2022-03-30
wrap getSchema to allow skipping the always-exist array

huayang-codaio:
https://staging.coda.io/d/Quality-Tracker_d-GJF-DmEUK/Bugs_sucBW#Bugs_tuq-W/r20975&view=modal

`getSchema` always returns an array of schemas. to make it easier, this PR allows the developer to return the item schema and we'll wrap it if it's not an array. 

@jonathan-codaio @patrick-codaio @coda/packs 

before:

```
-> % ./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake_v2.ts FooDyns --getSchema
{
  type: 'array',
  items: {
    type: 'object',
    primary: 'name',
    id: 'name',
    properties: { name: [Object] }
  }
}
```

after:

```
-> % ./node_modules/.bin/ts-node cli/coda.ts execute test/packs/fake_v2.ts FooDyns --getSchema
{
  type: 'array',
  items: {
    type: 'object',
    primary: 'name',
    id: 'name',
    properties: { name: [Object] }
  }
}
```


test/packs/fake_v2.ts
```diff
@@ -121,20 +121,24 @@ pack.addSyncTable({
...
-  getDisplayUrl: coda.makeMetadataFormula(async () => 'display url'),
-  getSchema: coda.makeMetadataFormula(async () => {
-    return coda.makeSchema({
-      type: coda.ValueType.Array,
-      items: {type: coda.ValueType.Object, properties: {}},
-    });
-  }),
+  getName: async () => 'name',
+  getDisplayUrl: async () => 'display url',
+  getSchema: async () => {
```
huayang-codaio, line 126: tested before / after.

helpers/object_utils.ts
```diff
@@ -32,3 +32,10 @@ export function isNil<T>(obj: T | null | undefined): obj is null | undefined {
 export function deepCopy<T>(obj: T): T {
   return JSON.parse(JSON.stringify(obj));
 }
+
+export function isPromise<T>(obj: any): obj is Promise<T> {
+  // This method should generally be avoided but can be used for high throughput / performance
```
patrick-codaio, line 37: If we're trying to match internal coda comment styling, this advice to callers of the function would be best as a jsdoc /**...*/ comment above line 36.

cli/coda.ts
```diff
@@ -37,6 +37,11 @@ if (require.main === module) {
           desc: 'Execute the requested command in a virtual machine that mimics the environment Coda uses to execute Packs.',
           default: true,
         },
+        getSchema: {
```
jonathan-codaio, line 40: Nice.

Although, I wonder if we want to generalize how to invoke the many metadata formulas attached to a sync table. Should the flag instead be something like `--metadataFormula=getSchema`? (Find to handle this separately.) WDYT @ekoleda-codaio ?

test/packs/fake_v2.ts
```diff
@@ -121,20 +121,24 @@ pack.addSyncTable({
...
+    return {
+      type: coda.ValueType.Object,
```
jonathan-codaio, line 128: Is there also a test case that returns the array version?  Want to make sure we're exercising both.

api.ts
```diff
@@ -267,6 +268,36 @@ export function wrapMetadataFunction(
...
+    return {
+      type: ValueType.Array,
+      items: schema,
+    } as ResultT;
+  }
+}
+
+export function wrapGetSchema(getSchema: MetadataFormula | undefined): MetadataFormula | undefined {
+  return (
+    getSchema && {
```
jonathan-codaio, line 284: Nit, mind doing

```
if (!getSchema) {
  return;
}
```

So we can unnest the big logic block.

api.ts
```diff
@@ -267,6 +268,36 @@ export function wrapMetadataFunction(
   return typeof fnOrFormula === 'function' ? makeMetadataFormula(fnOrFormula) : fnOrFormula;
 }
 
+function transformSchema<ResultT extends PackFormulaResult>(schema?: any): ResultT {
```
jonathan-codaio, line 271: Nit, maybe `transformToArraySchema` for clarity.

api.ts
```diff
@@ -1487,7 +1518,7 @@ export function makeDynamicSyncTable<K extends string, L extends string, ParamDe
     },
   });
   const getName = wrapMetadataFunction(getNameDef);
-  const getSchema = wrapMetadataFunction(getSchemaDef);
+  const getSchema = wrapGetSchema(wrapMetadataFunction(getSchemaDef));
```
jonathan-codaio, line 1521: It looks like we call `makeSyncTable()` right below this, which also does the wrapping. So seems like it's getting double-wrapped? The logic ensures the output will still be correct, but wonder if we should avoid having the wrapper called multiple times? Could just not wrap here and leave a comment that it happens later in `makeSyncTable()`.

api.ts
```diff
@@ -970,11 +1001,11 @@ export type MetadataFormulaMetadata = Omit<MetadataFormula, 'execute'>;
 /**
  * A JavaScript function that can implement a {@link MetadataFormulaDef}.
  */
-export type MetadataFunction = <K extends string, L extends string>(
+export type MetadataFunction = (
   context: ExecutionContext,
   search: string,
   formulaContext?: MetadataContext,
-) => Promise<MetadataFormulaResultType | MetadataFormulaResultType[] | ArraySchema | ObjectSchema<K, L>>;
+) => Promise<MetadataFormulaResultType | MetadataFormulaResultType[] | ArraySchema | ObjectSchema<any, any>>;
```
jonathan-codaio, line 1008: How come this was needed?

testing/execution.ts
```diff
@@ -163,10 +167,26 @@ export async function executeFormulaOrSyncFromCLI({
...
-    const formulaSpecification: SyncFormulaSpecification | StandardFormulaSpecification = {
-      type: syncFormula ? FormulaType.Sync : FormulaType.Standard,
-      formulaName,
-    };
+
+    let formulaSpecification:
+      | SyncFormulaSpecification
+      | StandardFormulaSpecification
+      | SyncMetadataFormulaSpecification;
+    if (getSchema) {
```
jonathan-codaio, line 175: Per the above comment, may just want to split this part out to a separate PR so we can iterate on executing metadata formulas more generally.

helpers/object_utils.ts
huayang-codaio: yeah, can do that. was just copying this method over from coda repo. I'll probably tweak the language a bit before moving them to the jsdoc comment.

test/packs/fake_v2.ts
huayang-codaio: let me see if I can add a unittest instead.

cli/coda.ts
ekoleda-codaio: Ya, the ability to test more of a Pack from the CLI is something I know I've wanted. +1 to making this flexible, but also understand the scope creep aspect.

huayang-codaio: I reverted these changes from this PR. let's make that a separate change to support `--metadataFormula`

jonathan-codaio (approved):

PR #1882, opened 2022-04-05
Add validation rule preventing usage of varargParameters for sync tables

alexd-codaio:

Since the UI does not currently support varargs in this context, just throw an error until we can fix it.

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -56,6 +56,7 @@ import type {StringEmbedSchema} from '../schema';
 import type {StringPackFormula} from '../api';
 import type {StringTimeSchema} from '../schema';
 import type {SyncFormula} from '../api';
+import type {SyncTable} from '../api';
 import type {SyncTableDef} from '../api';
 import type {SystemAuthenticationTypes} from '../types';
 import {Type} from '../api_types';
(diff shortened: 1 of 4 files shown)


testing/upload_validation.ts
```diff
@@ -885,7 +886,19 @@ const genericDynamicSyncTableSchema = zodCompleteObject<
   getSchema: formulaMetadataSchema,
 }).strict();
 
-const syncTableSchema = z.union([genericDynamicSyncTableSchema, genericSyncTableSchema]);
+const syncTableSchema = z.union([genericDynamicSyncTableSchema, genericSyncTableSchema])
+  .superRefine((data, context) => {
+    const syncTable = data as SyncTable;
+
+    if (syncTable.getter.varargParameters) {
```
jonathan-codaio, line 893: Just confirming that we don't need to check if this is an empty list here. Not sure if our definition wrappers spit out empty lists for fields that are omitted in the hand-authored definition.

jonathan-codaio (approved):

testing/upload_validation.ts
alexd-codaio: Done.

PR #1884, opened 2022-04-07
identityName will now be required for all sync tables, including dynamic ones

patrick-codaio:
*** This is the north star, but first we will merge https://github.com/coda/packs-sdk/pull/1915 to be live during a migration period ***

With this, even dynamic schemas in sync tables shouldn't have to worry about generating the `identity` object themselves. When this SDK change goes live in Coda, it will be accompanied by code to inject the appropriate identity into every dynamically generated schema.

`identityName` will now fail a build if it conflicts with the `identity.name` value on a sync table's schema. This will break some existing pack builds, but we think it's very valuable in minimizing accidental mistakes.

To apply `identityName` to dynamic schemas, we need to store the `identityName` in the persisted sync table definition so it can be used by the code that generates those dynamic schemas.

See the downstream changes in these PRs:
- https://github.com/coda/coda/pull/70218
- https://github.com/coda/packs/pull/3055


patrick-codaio:
@ekoleda-codaio https://golinks.io/bug/19739 has context about identityName vs identity.name (and I just saw Jonathan's comment at the bottom that I should throw validation errors when both are set - I'll update that soon). The other half is that the identityName shortcut never worked on dynamic schemas at all, so I'm supporting that. I can zoom today if that would help?

api.ts
```diff
@@ -1285,8 +1287,12 @@ export interface SyncTableOptions<
    * For example, if your identity name was 'Product', another formula or sync could return
    * shell objects that reference rows in this table, so long as they contain the id
    * of the object, and the schema is declared as `{identity: {name: 'Products'}}`.
+   *
+   * The only time you don't need to specify this is if your schema already sets it's own
+   * `identity.name`. Technically, this value is only the default identity name for your table's
+   * schema(s).
    */
-  identityName: string;
+  identityName?: string;
```
jonathan-codaio, line 1295: Per discussion, sounds like we'll leave this required. I'll wait to review further until you've updated!

patrick-codaio:
@jonathan-codaio @ekoleda-codaio @huayang-codaio I've updated this PR based on this morning's chat. I think I'll do another pass on the tests later today, but this should be good enough to start review on.

api.ts
```diff
@@ -161,6 +161,8 @@ export interface SyncTableDef<
   description?: string;
   /** See {@link SyncTableOptions.schema} */
   schema: SchemaT;
+  /** See {@link SyncTableOptions.identityName} */
+  identityName: string;
```
huayang-codaio, line 165: I wonder if there's a space to add a documentation indicating it's only used by dynamic sync tables.

huayang-codaio (commented):
makes sense to me.

api.ts
```diff
@@ -1404,10 +1414,16 @@ export function makeSyncTable<
 }: SyncTableOptions<K, L, ParamDefsT, SchemaDefT>): SyncTableDef<K, L, ParamDefsT, SchemaT> {
   const {getSchema: getSchemaDef, entityName} = dynamicOptions;
   const {execute: wrappedExecute, ...definition} = maybeRewriteConnectionForFormula(formula, connectionRequirement);
-  if (schemaDef.identity) {
-    schemaDef.identity = {...schemaDef.identity, name: identityName || schemaDef.identity.name};
-  } else if (identityName) {
-    schemaDef.identity = {name: identityName};
+  // We don't fail on a missing identityName because the legacy functions don't set it.
+  if (identityName) {
```
jonathan-codaio, line 1418: I'm not sure why I get why all the below logic is newly conditional on this? Naively I would think that we just want to add the validation about the names matching.

testing/upload_validation.ts
```diff
@@ -873,6 +873,7 @@ const genericSyncTableSchema = zodCompleteObject<GenericSyncTableDef & {isDynami
   // it in the validator like this helps zod flag it in the way we need.
   isDynamic: zodDiscriminant(false).optional(),
   getSchema: formulaMetadataSchema.optional(),
+  identityName: z.string().optional(),
```
jonathan-codaio, line 876: How come it's optional here?

testing/upload_validation.ts
```diff
@@ -931,7 +933,29 @@ const unrefinedPackVersionMetadataSchema = zodCompleteObject<PackVersionMetadata
...
+          context.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: `Every non-dynamic sync table must specify an identity for its schema. "${tableDef.name}" has neither an identityName nor an identity.name on its schema.`,
+          });
+        }
+        if (tableDef.identityName && tableDef.schema.identity?.name) {
+          if (tableDef.identityName !== tableDef.schema.identity.name) {
+            context.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: `Sync table "${tableDef.name}" defines an identityName that conflicts with its schema's identity.name.`,
```
jonathan-codaio, line 953: Can we include the conflicting names in the error message?

jonathan-codaio (commented):
Before merging this, can we queue up the work on the `coda` side to fully inject the identity into dynamic sync tables so that that's ready to go live, and when we merge this PR and subsequently make an npm version for it, we can announce those two together?

I'm just worried about merging this and having a while between when the injection changes are live, blocking us from making an npm release if we needed to. We could make a release with this as it is, but makers would hit the breaking change, but not get the benefit of the identity injection that this enables.

Also should add a CHANGELOG entry for this.

api.ts
patrick-codaio: The old code was already a no-op if identityName was falsy, I just thought this would be easier to read.

jonathan-codaio: Ah, that's subtle, I see it now, yeah you're right.

testing/upload_validation.ts
huayang-codaio: oh do we intend to make this a breaking change? probably a good time to do this I guess?

huayang-codaio: oh never mind. saw the other comment.

patrick-codaio: Oops, forgot to update that, good catch.

patrick-codaio:
@jonathan-codaio I'll prep PRs in both `coda` and `packs` before merging, yup! The rest of the comments are addressed now.

jonathan-codaio (approved):

huayang-codaio:
Just merged https://github.com/coda/packs-sdk/pull/1886 you may need to rebase.

patrick-codaio:
@jonathan-codaio @huayang-codaio I'm leaving this PR here to be used or referenced after our migration period, but the new https://github.com/coda/packs-sdk/pull/1915 is the code I intend submit for the migration period.

PR #1885, opened 2022-04-07
Document incremental OAuth

ekoleda-codaio:
Also update the Todoist sample to take advantage of it.


docs/guides/advanced/authentication.md
```diff
@@ -305,7 +305,38 @@ However if the API provider deviates too far from the OAuth 2.0 specification it
     Coda doesn't currently support the older 1.0 or 1.0a versions of the OAuth specification. If you would like to connect to an API that only supports these versions of the standard please [contact support][support] so that we can continue to gauge interest.
 
 
-<!-- TODO: Progressive OAuth -->
+#### Incremental OAuth
+
+If your Pack includes a lot of features you may likewise need to request a lot of scopes during authentication, which can scare away new users. An alternative approach is to use incremental authorization, where you start out only requesting a small initial set of scopes and request more as the user starts using features that require them.
```
jonathan-codaio, line 310: Maybe also worth a section on the other core use case, which is that of a pack that's already live that needs more scopes in a new version. You can simply include the new scopes in the new version, and as described below, if the initial request fails, the user is prompted to reauth and try again.

docs/guides/advanced/authentication.md
```diff
@@ -305,7 +305,38 @@ However if the API provider deviates too far from the OAuth 2.0 specification it
...
+```
+
+When the Pack above is installed the user will only be required to grant access to the `read` scope. However, when they try to use the `UpdateItem` action formula they will then be prompted grant additional access to the `write` scope. This prompt is displayed as a pop-up dialog at the bottom of the doc:
+
+<img src="../../../images/auth_oauth_incremental.png" srcset="../../../images/auth_oauth_incremental_2x.png 2x" class="screenshot" alt="Prompting the user for additional permissions">
+
+When the user signs in again they will be prompted to approve the additional scopes, after which they will be able to use the formula successfully.
+
+!!! info "401 response required"
+    Even if a formula specifies `extraOAuthScopes`, Coda will first try to run the formula using the existing credentials. The user will only be prompted to approve additional scopes if the formula fails with a 401 "Unauthorized" response from the API. If the API you are connecting to returns a different response code on failure you won't be able to take advantage of the incremental authorization feature.
```
jonathan-codaio, line 339: FWIW the Figma API incorrectly returns 403 and so our pack hacks the responses so they are treated as 401s, otherwise OAuth refreshes would not work for Figma (bonkers that this is how their API works). So it is possible, but not sure if we want to talk about that.

https://github.com/coda/packs/blob/main/packs/figma/main/formulas.ts#L97

I guess if it comes up enough we can expose an SDK helper function that abstracts this away.

jonathan-codaio (commented):
FYI @patrick-codaio if there's anything you want to add

docs/guides/advanced/authentication.md
patrick-codaio: +1 - my bet is this will be more commonly relevant. How about having a paragraph above this like:

As your pack grows you may find that you need to request more OAuth scopes than you initially did when your existing users connected. Coda allows new scopes to be added to Pack OAuth settings in a non-breaking way: we don't prompt the user to re-authorize until they try to use a Pack feature that fails. Once that happens, we notice that the connection the user was using was created with a stale list of OAuth scopes and we prompt them to re-authenticate it to get your new scopes. 

This progressive expansion of OAuth scopes can also work on a per-formula basis... all of your stuff

docs/guides/advanced/authentication.md
```diff
@@ -305,7 +305,38 @@ However if the API provider deviates too far from the OAuth 2.0 specification it
...
+pack.addFormula({
+  name: "UpdateItem",
+  // ...
+  isAction: true,
+  extraOAuthScopes: ["update"],
+  // ...
+});
+```
+
+When the Pack above is installed the user will only be required to grant access to the `read` scope. However, when they try to use the `UpdateItem` action formula they will then be prompted grant additional access to the `write` scope. This prompt is displayed as a pop-up dialog at the bottom of the doc:
```
patrick-codaio, line 332: prompted *to* grant

docs/guides/advanced/authentication.md
patrick-codaio: Technically, today we don't pay attention to the status code at all - any and all errors from a formula execution are clobbered if we notice there are missing scopes. We have this bug to change that: https://golinks.io/bug/19751 - maybe we should bump it up to a P2?

patrick-codaio (commented):
There's also a small issue where some services aren't completely compliant to the oauth spec and we can't support incremental scopes on them. I don't know if that deserves mention here, though... off the top of my head I think we noticed that zoom can't do it.

docs/guides/advanced/authentication.md
ekoleda-codaio: I love that language @patrick-codaio! Added almost verbatim.

docs/guides/advanced/authentication.md
ekoleda-codaio: Oh, good to know. Until we fix that bug I've just removed this warning block.

docs/guides/advanced/authentication.md
ekoleda-codaio: Good catch, done.

jonathan-codaio (approved):

patrick-codaio (approved):

PR #1886, opened 2022-04-07
add hideNewColumnsByDefault option to sync table

huayang-codaio:
PTAL @jonathan-codaio @patrick-codaio @coda/packs 

Two changes in this PR, 

1. add an optional placeholder `schema` to dynamic sync tables. this is used by Jira Pack where it shows a placeholder grid of selected columns before syncing issues. 

screenshot (testing dynamic grid using a placeholder schema):

image

it's worth pointing out only the featured columns will be rendered. non-featured columns are not just absent from canvas but also missing from the column list. This is probably not a big deal for a placeholder schema so I am not fixing it. 

```
  schema: coda.makeObjectSchema({
    properties: {
      name: { type: coda.ValueType.String },
      extraColumn: { type: coda.ValueType.String },
      anotherExtraColumn: { type: coda.ValueType.String },
    },
    id: "name",
    primary: "name",
    featured: ['name', 'anotherExtraColumn'],
    identity: {
      name: "Test",
    },
  }),
```

image

2. add `doNotAddNewSyncColumns` to allow pack maker set a default value of the `ADD NEWLY FOUND COLUMNS` option. This is important for Jira and similar packs with hundreds of columns, where otherwise will populates all columns into the canvas and overwhelm the user. 

Note that:
- only the featured columns in placeholder schema will show up in the final sync with `doNotAddNewSyncColumns=true`. 
- `featured` property of `getSchema` doesn't do anything here, probably by design?

Before: 

image

After:

image


api.ts
```diff
@@ -1372,6 +1376,16 @@ export interface DynamicSyncTableOptions<
...
+  /**
+   * If the sync table should add newly found columns to the canvas. By setting this to true (user can
+   * still change this later), newly found columns will be hidden by default. The hidden columns can
+   * be revealed manually.
+   */
+  doNotAddNewSyncColumns?: boolean;
+  /**
+   * Optional placeholder schema before the dynamic schema is retrieved.
+   */
+  schema?: SchemaT;
```
huayang-codaio, line 1388: @ekoleda-codaio please advise.

api.ts
```diff
@@ -1372,6 +1376,16 @@ export interface DynamicSyncTableOptions<
    * this sync table (including autocomplete formulas).
    */
   connectionRequirement?: ConnectionRequirement;
+  /**
+   * If the sync table should add newly found columns to the canvas. By setting this to true (user can
```
jonathan-codaio, line 1380: Yeah, tricky to explain. Maybe:

```
If true, when subsequent syncs discover new schema fields, these fields will not automatically added as new columns on the table. The user can still manually add columns for these new fields. This only applies to tables that use dynamic schemas.

When tables with dynamic schemas are synced, the {@link getSchema} formula is run each time, which may return a schema that is different than that from the last sync. The default behavior is that any schema fields that are new in this sync are automatically added as new columns, so they are apparent to the user. However, in rare cases when schemas change frequently, this can cause the number of columns to grow quickly and become overwhelming. Setting this value to true leaves the columns unchanged and puts the choice of what columns to display into the hands of the user.
```

api.ts
jonathan-codaio: Maybe `defaultSchema` or `placeholderSchema`?

api.ts
```diff
@@ -1372,6 +1376,24 @@ export interface DynamicSyncTableOptions<
    * this sync table (including autocomplete formulas).
    */
   connectionRequirement?: ConnectionRequirement;
+  /**
+   * If true, when subsequent syncs discover new schema fields, these fields will not automatically
```
patrick-codaio, line 1380: you are missing a "be" at the end of the line

api.ts
```diff
@@ -1372,6 +1376,24 @@ export interface DynamicSyncTableOptions<
...
+   *
+   * When tables with dynamic schemas are synced, the {@link getSchema} formula is run each time,
+   * which may return a schema that is different than that from the last sync. The default behavior
+   * is that any schema fields that are new in this sync are automatically added as new columns,
+   * so they are apparent to the user. However, in rare cases when schemas change frequently,
+   * this can cause the number of columns to grow quickly and become overwhelming. Setting this
+   * value to true leaves the columns unchanged and puts the choice of what columns to display
+   * into the hands of the user.
+   */
+  doNotAddNewSyncColumns?: boolean;
```
patrick-codaio, line 1392: Would it be crazy to make this `doNotAddNewSyncColumnsByDefault` for accuracy, since users will be able to override this value?

api.ts
```diff
@@ -1507,18 +1537,21 @@ export function makeDynamicSyncTable<K extends string, L extends string, ParamDe
...
-    },
-  });
+  const placeholderSchema: any =
+    placeholderSchemaInput ||
+    // default placeholder only shows a column of id, which will be replaced later by the dynamic schema.
+    makeObjectSchema({
+      type: ValueType.Object,
+      id: 'id',
+      primary: 'id',
+      identity: {name},
```
patrick-codaio, line 1550: We're going to have a merge conflict here, FWIW

api.ts
huayang-codaio: good catch!

api.ts
```diff
@@ -1372,6 +1376,24 @@ export interface DynamicSyncTableOptions<
...
+   * which may return a schema that is different than that from the last sync. The default behavior
+   * is that any schema fields that are new in this sync are automatically added as new columns,
+   * so they are apparent to the user. However, in rare cases when schemas change frequently,
+   * this can cause the number of columns to grow quickly and become overwhelming. Setting this
+   * value to true leaves the columns unchanged and puts the choice of what columns to display
+   * into the hands of the user.
+   */
+  doNotAddNewSyncColumnsByDefault?: boolean;
+  /**
+   * Optional placeholder schema before the dynamic schema is retrieved.
```
ekoleda-codaio, line 1394: What's the behavior if the user doesn't supply the placeholder schema? What's the advantage of supplying one?

api.ts
```diff
@@ -1372,6 +1376,24 @@ export interface DynamicSyncTableOptions<
    * this sync table (including autocomplete formulas).
    */
   connectionRequirement?: ConnectionRequirement;
+  /**
+   * If true, when subsequent syncs discover new schema fields, these fields will not automatically be
```
ekoleda-codaio, line 1380: Should we call these schema "properties" instead of "fields" for consistency?

api.ts
```diff
@@ -1372,6 +1376,24 @@ export interface DynamicSyncTableOptions<
...
+   *
+   * When tables with dynamic schemas are synced, the {@link getSchema} formula is run each time,
+   * which may return a schema that is different than that from the last sync. The default behavior
+   * is that any schema fields that are new in this sync are automatically added as new columns,
+   * so they are apparent to the user. However, in rare cases when schemas change frequently,
+   * this can cause the number of columns to grow quickly and become overwhelming. Setting this
+   * value to true leaves the columns unchanged and puts the choice of what columns to display
+   * into the hands of the user.
+   */
+  doNotAddNewSyncColumnsByDefault?: boolean;
```
ekoleda-codaio, line 1392: This field name is a bit verbose, and the `doNot` prefix is uncommon. What about something like:

```ts
disableNewColumnsOnSync: true
```

api.ts
huayang-codaio: If the sync table has any parameter, we won't trigger the sync right away when it's added to the canvas. At this moment, we'll still render a placeholder table based on if there's any placeholder schema given. 

For example, Jira passes in a placeholder schema of 6 columns (issue, reporter, start date, etc) that would render in this phase. Without a placeholder schema, the temporary table would just have one column with the object name. 

In addition, if `doNotAddNewSyncColumnsByDefault`/`disableNewColumnsOnSync` is set to true, we would only add the placeholder schema columns to the canvas by default. (hmm I think I need to add this to document as well)

api.ts
```diff
@@ -1250,6 +1252,8 @@ export interface DynamicOptions {
   getSchema?: MetadataFormulaDef;
   /** See {@link DynamicSyncTableOptions.entityName} */
   entityName?: string;
+  /** See {@link DynamicSyncTableOptions.disableNewColumnsOnSyncByDefault} */
+  disableNewColumnsOnSyncByDefault?: boolean;
```
ekoleda-codaio, line 1256: Nitpick: I know @patrick-codaio recommended the `ByDefault` suffix in another comment, but I don't think it's necessary now. `OnSync` clarifies that this behavior only kicks in during the sync, and even though a user can manually add new columns that's not happening during the sync, but afterwards. I'd vote for dropping `ByDefault` for brevity, but I don't want to bike shed this too much :-)

huayang-codaio: either way works for me. let me update.

patrick-codaio: I also don't want to bike shed, I just want to make sure we're talking about the same thing - I wasn't worried about the pack maker thinking manual column adding would be disabled, I was worried the pack maker might think this end user toggle would be disabled:
Screen Shot 2022-04-08 at 11 03 45 AM

If you still want to keep the "ByDefault" off, no biggie.

jonathan-codaio (approved):
Agree with Patrick's thought that maybe some makers could get confused about whether it disables the setting, but I think maybe that's best handled by the existing documentation rather than a longer property name.

api.ts
ekoleda-codaio: Ah, I didn't realize that this is additionally controllable by the user. So this sync table field only sets the default toggle position for newly created tables?

huayang-codaio: yes. it only sets the default value.

ekoleda-codaio: Ah, now I understand the logic. I'm fine with either the current name (shorter) or what you had previous (more accurate).

ekoleda-codaio (approved):

PR #1887, opened 2022-04-07
Set one networkDomain in auth metadata

dweitzman-codaio:
The auth networkDomain can be inferred for packs one only one overall networkDomain

This change is backwards incompatible for packs with multiple domains:
they'll need to explicitly choose which domain gets auth in order to
upload new versions in the future.


testing/upload_validation.ts
```diff
@@ -1114,4 +1115,36 @@ const legacyPackMetadataSchema = validateFormulas(
         "Specify the domain that your pack makes http requests to using `networkDomains: ['example.com']`",
       path: ['networkDomains'],
     },
-  );
+  )
+  .superRefine((data, context) => {
```
jonathan-codaio, line 1119: NIt, is `data` typed as `any` here? I think it's safe to cast it at this point in the validation flow, doing so may help catch future refactoring issues.

testing/upload_validation.ts
```diff
@@ -1114,4 +1115,36 @@ const legacyPackMetadataSchema = validateFormulas(
...
+    if (!data.defaultAuthentication) {
+      return;
+    }
+    // Pack has multiple network domains and uses authentication.
+    if (data.defaultAuthentication.requiresEndpointUrl || data.defaultAuthentication.type === AuthenticationType.None) {
+      // We're ok if there's a user-supplied endpoint domain if auth is just being disabled explicitly.
```
jonathan-codaio, line 1125: I think there's a missing "or".

helpers/metadata.ts
```diff
@@ -96,13 +99,22 @@ function compileSyncTable(syncTable: GenericSyncTable): PackSyncTable {
...
   if (authentication.type === AuthenticationType.None || authentication.type === AuthenticationType.Various) {
     return authentication;
   }
+
+  // User credentials must be explicitly pinned to exactly one domain so they can be invalidated if the pack
+  // owner changes the network domain. For simplicity and backwards compatibility, the authentication networkDomain
+  // can be omitted if there's exactly one call to setNetworkDomain elsewhere in the pack: we can safely assume
+  // authentication wants to be pinned to that same domain.
+  if (!authentication.networkDomain && networkDomains && networkDomains.length === 1) {
+    authentication.networkDomain = networkDomains[0];
```
jonathan-codaio, line 116: Just wondering about the value of populating this, if the `coda` code is always going to tolerate this being missing for single-domain packs. It raises the expectation that this will always be populated, but a year from now there could still be old builds where it isn't. Maybe that's fine and it's better to have most packs have this while still noting that old ones may not?

jonathan-codaio (commented):
Yep cool.

helpers/metadata.ts
dweitzman-codaio: For the moment the option that feels the most appealing to me is:
- Remove this logic in the SDK, with the intent that `coda` will support omitted auth networkDomain forever
- Centralize metadata normalization/transformation immediately after it's read in `coda`. Maybe put the version normalization function here in `packs-sdk` and have it return a type where the auth `networkDomain` is non-optional?
- Maybe start recording some kind of a version with the metadata?

Others options might be:
- Keep this logic with the hope that it makes a backfill simpler someday in the future
- Remove this logic and let the `coda` repo deal with the uncertainty about which SDK built some metadata in either a centralized or decentralized way

huayang-codaio (approved):
lgtm probably add a test for `upload_validations.ts`

huayang-codaio:
oh and it's a breaking change to oauth packs with multiple domains. @jonathan-codaio maybe we should release a version first and then make all the breaking changes to the next release?

testing/upload_validation.ts
dweitzman-codaio: Added a cast

testing/upload_validation.ts
dweitzman-codaio: Simplified this comment because switching from `any` to `PackVersionMetadata` required me to check for `None` earlier anyway.

helpers/metadata.ts
huayang-codaio: I am leaning towards `Remove this logic in the SDK, with the intent that we'll support existing metadata with omitted auth networkDomain forever` unless we want to backfill the existing external metadata (doable but I don't feel worth for this change)

jonathan-codaio: Yeah I would be inclined to remove in the SDK and support existing metadata forever.

FWIW we do record the SDK version in the DB with every pack upload, which is meant to give us an indication of what feature set/structure was available to the pack at the time it was built. So it wouldn't be unreasonable to keep this logic here in the SDK and have the `coda` side only care about handling an undefined networkDomain if `sdkVersion < x.y.z`.

jonathan-codaio:
> oh and it's a breaking change to oauth packs with multiple domains. @jonathan-codaio maybe we should release a version first and then make all the breaking changes to the next release?

It looks like the only pending change is also a breaking one. Patrick is also soon to merge a breaking change, and then there's all the potential breaking changes I enumerated in https://staging.coda.io/d/Packs-Platform-IA-go-packs_d36WK4zgrYx/SDK-Breaking-Changes_suWuJ#SDK-Breaking-Changes_tuBp0

I'm tempted to say we work fast to implement the breaking changes in the doc (just the parts needed for makers to start their migrations) so that in about a week we can create one new release that includes all of our planned breaking changes and announce them at once.

helpers/metadata.ts
dweitzman-codaio: Removed the metadata.ts changes

jonathan-codaio (approved):

dweitzman-codaio:
> lgtm probably add a test for `upload_validations.ts`

Added unit tests

test/upload_validation_test.ts
```diff
@@ -1689,10 +1689,13 @@ describe('Pack metadata Validation', () => {
           type: AuthenticationType.CodaApiHeaderBearerToken,
           deferConnectionSetup: true,
           shouldAutoAuthSetup: true,
+          networkDomain: 'other.domain',
         },
         networkDomains: ['coda.io', 'other.domain'],
       });
       const err = await validateJsonAndAssertFails(metadata);
+      // TODO(dweitzman): Get rid of this error when credential pinning is launched
```
dweitzman-codaio, line 1697: Also added a TODO to this unit test. We can theoretically relax the CodaApiHeaderBearerToken validation to only look at the auth `networkDomain` once the enforcement of auth `networkDomain` is live

dweitzman-codaio:
Before merging this, two things to figure out:

- Should this PR bump the SDK version?
- Any concerns with the new `networkDomain` having `string` type, knowing that someday it may need to generalize to `string|string[]`?

jonathan-codaio:
> Before merging this, two things to figure out:
> 
> * Should this PR bump the SDK version?
> * Any concerns with the new `networkDomain` having `string` type, knowing that someday it may need to generalize to `string|string[]`?

No ned to bump the SDK version anymore, new changes are now considered "unrelease" until we explicitly create an npm release, which will assign a new version. I believe this is now handled by a GitHub Actions workflow, though we've only released this way once so far.

Per the comment thread, `string` for now is good and moving to `string | string[]` if needed in the future seems reasonable.

PR #1888, opened 2022-04-08
skip marshaling deep-nested value instead of throwing

huayang-codaio:
https://codaproject.slack.com/archives/C02K6EABVMK/p1649435284318259

turns out there's a legit case that Pack may return deep nested object. 

@jonathan-codaio @alan-codaio @coda/packs

### DIFF
--- a/bundles/thunk_bundle.js
+++ b/bundles/thunk_bundle.js
@@ -4525,7 +4525,7 @@ module.exports = (() => {
   }
   function processValue(val, depth = 0) {
     if (depth >= MaxTraverseDepth) {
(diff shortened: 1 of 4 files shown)


alan-codaio (approved):

jonathan-codaio:
I want to understand why Jira is returning such a weird value first before proceeding with this.  I think it was really useful for now that the existing code caught this.

jonathan-codaio:
Let's not merge this, it was great that it caught the Jira pack bug :)

jonathan-codaio (approved):
Nm, I was corrected that this affects fetcher responses too.

PR #1889, opened 2022-04-08
Add new names for id, primary, and featured for clarity.

jonathan-codaio:
One of our breaking changes from the doc.

I was afraid of how big a change like this might be, I started scoping it and just started making sample code changes as part of that, and it wound up being not quite so horrible so I just did the full change. With the use of ObjectSchemaHelper it helps make the migration period more robust, and the `coda` side changes make use of it too. So it really just becomes an exercise in making sure that all references to the old fields go through the helper.


ekoleda-codaio:
Would you want to replace the references to these fields in the docs or samples, or do that in a follow-on PR?

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 # Changelog
+
 This changelog keeps track of all changes to the packs-sdk. We follow conventions from [keepachangelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Unreleased
+
 - **Breaking Change** Added a validation rule that prevents the usage of varargsParameters for sync table getters
-which are not currently supported in the UI.
+  which are not currently supported in the UI.
+- **Future Breaking Change** The properties `id`, `primary`, and `featured` of object schemas will be renamed to `idProperty`, `primaryProperty`, and `featuredProperties`, respectively, to better clarify that their values refer to property names within the schema. During a migration period, you may use either name; the original names are now marked as deprecated and will be removed at a future date, no earlier than May 1, 2022.
```
patrick-codaio, line 9: I still feel like we can improve on `primaryProperty`... Its only purpose is to set what the chip shows and what the display column is in a sync table, right? I think `displayProperty` might be better.

bundles/thunk_bundle.js
```diff
@@ -4293,6 +4293,9 @@ module.exports = (() => {
   // helpers/ensure.ts
   init_buffer_shim();
 
+  // helpers/migration.ts
+  init_buffer_shim();
```
patrick-codaio, line 4297: this looks like a duplicate of the line above it... hopefully it's idempotent

jonathan-codaio:
@ekoleda-codaio VS Code is not respecting the lint exceptions for the documentation example files. When I edit those, it applies the normal lint rules, e.g. single quotes instead of double-quotes. I can get around this by just running `make lint-fix` after I edit the files, but the auto-fix does not fix lines that are > 80 characters.

How are you editing these files in general? Did you get VS Code to respect the alternate lint rules, or are you just using a different editor for these files?

CHANGELOG.md
ekoleda-codaio: Ya, I think `displayProperty` would be an improvement.

patrick-codaio: cc @alicia-codaio who is thinking of future directions pack objects could be "displayed" - any concerns about this change?

alicia-codaio: hm. `displayProperty` seems okay but maybe too vague for me to feel super confident about futureproofing. based on those two main areas (what's visible on the chip, and the primary column on the sync table) what about something like `titleProperty`, where it feels more definitively singular (whereas you might "display" more than one property)?

patrick-codaio:
@jonathan-codaio I was turning off auto-format while editing those files

ekoleda-codaio:
> VS Code is not respecting the lint exceptions

I don't have that issue, but I'm not sure how exactly. I may not have the same set of extensions installed or changed a setting, but I can't recall.

CHANGELOG.md
jonathan-codaio: @ekoleda-codaio thoughts on `titleProperty` vs `displayProperty`? I will defer to you on this weighty decision 😄 Both are reasonable options to me.

bundles/thunk_bundle.js
huayang-codaio: yeah, esbuild injects this whenever a new import is added. it ensures the buffer shim is available. but the generated code does seem stupid because the module has been imported before..

schema.ts
```diff
@@ -756,18 +757,22 @@ export interface ObjectSchemaDefinition<K extends string, L extends string> exte
   type: ValueType.Object;
   /** Definintion of the key-value pairs in this object. */
   properties: ObjectSchemaProperties<K | L>;
+  /** @deprecated Use {@link idProperty} */
```
huayang-codaio, line 760: this `link` didn't work for me. I never figured out how to link a property within the same class.

CHANGELOG.md
ekoleda-codaio: I named the section of the docs ["Display value"](https://coda.io/packs/build/latest/guides/advanced/schemas/#display-value), so I'm biased towards `displayProperty`. To me "title" implies something more formal, and for some schemas the displayed value could be as simple as number. If I get the tie breaker, then put me down for `displayProperty`.

schema.ts
jonathan-codaio: Will look at this.

CHANGELOG.md
jonathan-codaio: `displayProperty` is probably the most intuitive. Dunno about future-proofing against hypotheticals, but if we think we might allow multiple properties to feed into that, doesn't seem inconceivable we'd add `displayProperties` at that point, or change the type also allow an array.

schema.ts
jonathan-codaio: Hmm, this does seem to work for me as-is in the staged documentation? I think I did read that the link resolution strategy does try to resolve to properties on the same object first.

schema.ts
```diff
@@ -1016,7 +1016,7 @@ export type InferrableTypes = boolean | number | string | object | boolean[] | n
  * inputs, it may be useful to us this helper to sniff the return value and generate a basic
  * inferred schema from it.
  *
- * This utility does NOT attempt to determine {@link id} or {@link primary} attributes for
+ * This utility does NOT attempt to determine {@link idProperty} or {@link displayProperty} attributes for
```
huayang-codaio, line 1019: the old values were ambiguous. https://coda.io/packs/build/latest/reference/sdk/functions/generateSchema/

it's more precise to prefix with a classname.

schema.ts
huayang-codaio: right it seems to work here. i am a bit confused. I'll see if I hit the issue next time. sorry for the false alarm.

jonathan-codaio (commented):
Let me know if any final thoughts on this, otherwise feel free to approve and I'll update the `coda` side!

huayang-codaio (approved):

PR #1890, opened 2022-04-08
Added 'force' option to LinkSchema

hewitt-codaio:
Some embedded URLs will only render on the client, for example if user login is required.  For canvas embeds, our `Embed()` formula has a boolean `force` parameter that renders the embed on the client instead of the server.  However, there was no equivalent for Link columns.

Thus, we are adding a "force" property to the Link column config that triggers the same behavior.  This PR adds support to the packs-sdk for the new property.

IMPORTANT:
* The first commit contains the actual source changes.
* The second commit is the `make build` output.

Note:  Here is a [link to the corresponding Coda PR](https://github.com/coda/coda/pull/70334).

force-property

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This changelog keeps track of all changes to the packs-sdk. We follow convention
 ### Unreleased
 - **Breaking Change** Added a validation rule that prevents the usage of varargsParameters for sync table getters
 which are not currently supported in the UI.
+- Added "force" option to `LinkSchema` for `LinkDisplayType.Embed`.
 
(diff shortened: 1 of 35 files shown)


jonathan-codaio (approved):

patrick-codaio:
Should we be sure not to release this SDK change before https://github.com/coda/coda/pull/70334 is live?

jonathan-codaio:
> Should we be sure not to release this SDK change before [coda/coda#70334](https://github.com/coda/coda/pull/70334) is live?

Yep. But we're not planning to release for a while because of the breaking changes flowing in, so should be nbd.

PR #1891, opened 2022-04-11
Run "make build"

dweitzman-codaio:
I think my branch for https://github.com/coda/packs-sdk/pull/1887 may have been old enough that it wasn't using the latest compilation settings?...


jonathan-codaio (commented):
LGTM, but I think my latest PR https://github.com/coda/packs-sdk/pull/1889 will include the fix for this as well, so we could also just fix forward with that and then I wouldn't have to rebase again.

dweitzman-codaio:
> LGTM, but I think my latest PR #1889 will include the fix for this as well, so we could also just fix forward with that and then I wouldn't have to rebase again.

Works for me. I'll close this PR

PR #1901, opened 2022-04-11
Handling a bunch of dependabots in 1 PR

patrick-codaio:


package.json
```diff
@@ -76,11 +76,11 @@
     "@types/url-parse": "^1.4.8",
     "@types/uuid": "^8.3.4",
     "@types/xml2js": "^0.4.9",
-    "@types/yargs": "^17.0.8",
-    "@typescript-eslint/eslint-plugin": "5.13.0",
+    "@types/yargs": "^17.0.10",
+    "@typescript-eslint/eslint-plugin": "5.19.0",
     "@typescript-eslint/experimental-utils": "5.17.0",
```
jonathan-codaio, line 81: Probably want this one too? Generally all 3 typescript-eslint ones move in lockstep.

patrick-codaio: Oh, right. That's strange that dependabot didn't suggest that one, it is indeed at 5.19 now - https://www.npmjs.com/package/@typescript-eslint/experimental-utils

Fixed!

PR #1904, opened 2022-04-12
make array parameter sparse

huayang-codaio:
https://staging.coda.io/d/Packs-Platform-IA-go-packs_d36WK4zgrYx/SDK-Breaking-Changes_suWuJ#SDK-Breaking-Changes_tuBp0/r1&view=modal

This PR makes array parameter sparse. It is not completed yet and only has sparse **number** array implemented (other types support will be added later in this PR). 

TODO: 
- [x] other array types.
- [x] change log. 

A few considerations:
- the sparse array behavior should not affect existing pack versions. it can be done by either 1) check packs-sdk version in coda, or 2) add an array type option (i.e. `allowEmpty`) that's only set for the newly complied metadata. 
- if we should all arrays sparse (e.g. schema arrays, suggested value arrays) or only parameter array sparse. I think the impact is too much to make the general sdk `Array` type sparse. This PR would only make parameter arrays sparse. 
- TS should reflect the nullable array parameters.

PTAL @jonathan-codaio @patrick-codaio @coda/packs


huayang-codaio:
@jonathan-codaio btw. maybe this one isn't a breaking change. if the input was sparse, the pack formula will show error anyway. this PR would only open up the formulas for more potential executions rather than breaking existing executions.

api_types.ts
```diff
@@ -32,6 +32,12 @@ export interface ArrayType<T extends Type> {
   type: 'array';
   /** The type of the items in this array. */
   items: T;
+  /** If true, this array will accept empty values as `undefined` for non-string types. */
```
jonathan-codaio, line 35: Interesting, is the idea that for strings we don't have to support sparse arrays since we'll coerce to the empty string? So that means StringArray, HtmlArray, ImageArray, FileArray don't need to be sparse? And boolean could similarly coerce to false. So that just means it's probably just NumberArray and DateArray that require declaring params as sparse?

I don't object, but the inconsistency between array types is a bit odd and I wonder if developers will have a hard time with it. As usual, mostly deferring to @ekoleda-codaio about whether consistency is better or avoiding sparseness when not necessary is better.

huayang-codaio: oh I think HtmlArray, ImageArray and FileArray would still be sparse. (just think about an empty column)

I exempted non-string value because an empty string column still still a valid string. while empty image/file column should undefined.

jonathan-codaio: Ok yeah that seems reasonable.

I guess it's still a little weird why the same blank cell results in undefined for HtmlArray but `''` for a StringArray, but maybe coercion isn't that hard for people to grasp. Even if we typed StringArray to be sparse, I guess CFL is already coercing blanks to strings so we'd never see `undefined` in practice?

@ekoleda-codaio WDYT, should StringArray be a special snowflake that doesn't need `undefined`, or should it be consistent with all the rest?

huayang-codaio: thinking more about this I guess all array types can be undefined. essentially if the input value is not coerce'able for the given type it would be passed in as undefined. 

so if we pass an image column into a string array (assuming coerce doesn't work in this case), I guess it should be undefined?

huayang-codaio: but empty text column should be just `''` for string array I think. it makes sense to most cases I think.

jonathan-codaio: That makes sense that it depends on the coercability of the source data type 👍

huayang-codaio: I'll proceed with this direction, and finish up all the remaining data types.

ekoleda-codaio: I think ideally we would make all array types work the same. Having an empty cell result in an empty string for `StringArray` but undefined for `NumberArray` is unexpected, and exposes an implementation detail of Coda. That said, if it's too hard to change I think we could document the inconsistency any be fine.

jonathan-codaio:
Are both these ready for re-review? Just let me know.

jonathan-codaio (commented):
Code looks good.

Can we also create 2 helper, and export them in index.ts:

`condenseItems()` or some such name, that takes an array param and returns the same array with `undefined` values removed, for when a maker wants to implement something like `sum` and not error on unspecified values.

`requireAllItemsDefined()` or some such name, which throws `new UserVisibleError()` with a nice message if some array param item is `undefined`, so makers can trivially enforce the prior behavior if they don't want to accept sparse inputs. This should probably also be a type guard (return type is something like `: arr is T[]`) so TS does the right thing too.

I wonder about rolling this out though, it's an instantly-breaking change, both for makers and us in the `packs` repo. It's not hard to deal with, you can simply call `requireAllParametersDefined(params)` as the first line of every array formula, but it kind of muddies our story about there being a transition period for all of our breaking changes.

Should we introduce things like `ParameterType.SparseStringArray` while leaving `ParameterType.StringArray` with the non-sparse type, so that people can transition on their own time? When we're ready to stop supporting non-sparse arrays, we could just make `ParameterType.StringArray` mean the same thing as `ParameterType.SparseStringArray`.

WDYT @ekoleda-codaio ?

huayang-codaio:
@jonathan-codaio this one is probably ready to be reviewed. The other one got into many corner cases that I am fixing right now. 

I can add those two helper methods. but I wonder if it's easier to simply add an option to make a parameter array non-sparse.

jonathan-codaio:
> @jonathan-codaio this one is probably ready to be reviewed. The other one got into many corner cases that I am fixing right now.
> 
> I can add those two helper methods. but I wonder if it's easier to simply add an option to make a parameter array non-sparse.
> 
> edit: oh, it probably aligns with the `ParameterType.SparseStringArray` idea, which is one way to do it.

Yeah, based on the `coda` PR, it seems like it's not a big deal to support sparse and non-sparse arrays forever, I wonder if we just wind up doing that instead @ekoleda-codaio ? We have to do it if we want to have a transition period, but it's only a few extra lines of code to maintain I think.

ekoleda-codaio:
If we were starting from scratch I'd probably say we should only have a sparse option and have developers code around it. I think the utility of a dedicated "dense" option isn't that high, when you can enforce dense-ness in code if you really need it. That said, if creating a separate Sparse version of our parameter types is the best way to transition, then I think it's fine to keep both around for the foreseeable future and see how the usage pans out.

huayang-codaio:
I updated this PR to 
- add new parameter type `SparseStringArray` and etc. 
- make this PR a non-breaking change (we can also make it a breaking change by defaulting array parameters to sparse. It feels unnecessary to me). 
- mark some legacy parameter helper functions as deprecated.

jonathan-codaio (approved):

PR #1905, opened 2022-04-13
Add file as a supported param type

gary-codaio:
Adds file and file array as supported parameter types. Together with https://github.com/coda/coda/pull/70481


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This changelog keeps track of all changes to the packs-sdk. We follow convention
 - **Future Breaking Change** The properties `id`, `primary`, and `featured` of object schemas will be renamed to `idProperty`, `displayProperty`, and `featuredProperties`, respectively, to better clarify that their values refer to property names within the schema. During a migration period, you may use either name; the original names are now marked as deprecated and will be removed at a future date, no earlier than May 1, 2022.
 - Added "force" option to `LinkSchema` for `LinkDisplayType.Embed`.
 - **Breaking Change** Removed dead code related to `defaultConnectionType`. This was an unused feature and should not have been previously referenced.
+- Added support for files as parameters. Previously, files could be parameters by using the Image parameter type, but an error would show in the doc.
```
spencer-codaio, line 12: would specifically mention the type here (i.e. `ParameterType.File` like done with `LinkDisplayType.Embed`

schema.ts
```diff
@@ -87,6 +87,19 @@ export enum ValueHintType {
...
    */
   Currency = 'currency',
+  /**
+   * Indicates to interpret and render the value as a file. The provided value should be a URL that
+   * points to a file. Coda will hotlink to the file when rendering it a doc.
+   *
+   * Using {@link FileAttachment} is recommended instead, so that the file is always accessible
+   * and won't appear as broken if the source file is later deleted.
+  */
+  FileReference = 'file',
```
spencer-codaio, line 97: hmm im actually not sure if we need to support a reference for files. For images, this makes sense because you can directly embed an image but idk how that would work for files generally cc @jonathan-codaio

schema.ts
```diff
@@ -143,11 +156,6 @@ export enum ValueHintType {
    * ```
    */
   Reference = 'reference',
-  /**
-   * Indicates to interpret and render a value as a file attachment. The provided value should be a URL
-   * pointing to a file of a Coda-supported type. Coda will ingest the file and host it from Coda infrastructure.
-   */
-  Attachment = 'attachment',
```
spencer-codaio, line 150: hmm this will be a breaking change i believe. is that intentional? maybe relates to whether we need FileReference and if not we should stick to `Attachment`

gary-codaio: Ah yup this would be breaking, I should have called this out for discussion. If we do want `FileReference` it seemed confusing to me that we had something that was just called `Attachment`, so figured it would be good to rename this in conjunction.

CHANGELOG.md
jonathan-codaio: Yeah, +1. Would also label this as a future breaking change, that eventually ParameterType.Image will only accept actual image values from the doc.

api_types.ts
```diff
@@ -129,6 +136,10 @@ export enum ParameterType {
    * Indicates a parameter that is a Coda image. The pack is passed an image URL.
    */
   Image = 'image',
+  /**
+   * Indicates a file that is a Coda file. The pack is passed a file URL.
```
jonathan-codaio, line 140: Nit, "Indicates a _parameter_ that is...`

schema.ts
jonathan-codaio: Yeah, we don't support file references, that's unique to images (and maybe will go away eventually too).

schema.ts
jonathan-codaio: Yeah, I wouldn't rename this. `Attachment` already means `FileAttachment` and we already have support for this as a return value, so I wouldn't want to touch it.

jonathan-codaio (commented):
Nice!

FYI @huayang-codaio who is working on making array params sparse, as this introduces a new array param type, might be a merge conflict lurking.

huayang-codaio:
thanks for the heads up. the packs-sdk change is straightforward to my task. should be easy merging.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This changelog keeps track of all changes to the packs-sdk. We follow convention
 - **Future Breaking Change** The properties `id`, `primary`, and `featured` of object schemas will be renamed to `idProperty`, `displayProperty`, and `featuredProperties`, respectively, to better clarify that their values refer to property names within the schema. During a migration period, you may use either name; the original names are now marked as deprecated and will be removed at a future date, no earlier than May 1, 2022.
 - Added "force" option to `LinkSchema` for `LinkDisplayType.Embed`.
 - **Breaking Change** Removed dead code related to `defaultConnectionType`. This was an unused feature and should not have been previously referenced.
+- Added support for files as parameters with `ParameterType.File`. Previously, files could be used as parameters by using the `ParameterType.Image` parameter type, but an error would show in the formula builder.
```
jonathan-codaio, line 12: Would label this as `**Future Breaking Change**` and say at the end, that in the future we may enforce that error, requiring that only images can be passed when ParameterType.Image is specified.

jonathan-codaio (approved):
🎉

CHANGELOG.md
gary-codaio: Will do!

schema.ts
```diff
@@ -146,8 +146,8 @@ export enum ValueHintType {
   /**
    * Indicates to interpret and render a value as a file attachment. The provided value should be a URL
    * pointing to a file of a Coda-supported type. Coda will ingest the file and host it from Coda infrastructure.
-   */
-  Attachment = 'attachment',
+  */
+   Attachment = 'attachment',
```
spencer-codaio, line 150: nit: looks like an extra space here?

spencer-codaio (approved):
nice!!

PR #1907, opened 2022-04-13
Rename SetEndpoint.getOptionsFormula to getOptions. Allow both during migration.

jonathan-codaio:
The last change was a little cleaner because we were renaming a optional fields (like `primary`), this is a required field (on an optional object), I don't love it but it seems like we have to make both new and old optional but validate at upload time that at least one is present.


huayang-codaio (approved):

testing/upload_validation.ts
```diff
@@ -279,8 +279,12 @@ const setEndpointPostSetupValidator = zodCompleteObject<SetEndpoint>({
...
   description: z.string(),
   // TODO(jonathan): Remove this from the metadata object, only needs to be present in the full bundle.
-  getOptionsFormula: z.unknown(),
-});
+  getOptions: z.unknown().optional(),
+  getOptionsFormula: z.unknown().optional(),
+}).refine(
+  data => data.getOptions || data.getOptionsFormula,
+  'Either getOptions or getOptionsFormula must be specified.',
+);
```
huayang-codaio, line 287: 👍

PR #1912, opened 2022-04-15
support downloadsAs parameter for file storage

spencer-codaio:
handles https://golinks.io/bug/22077

implemented in https://github.com/coda/coda/pull/70548

@ekoleda-codaio @jonathan-codaio @coda/packs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This changelog keeps track of all changes to the packs-sdk. We follow convention
 ### Added
 
 - Added "force" option to `LinkSchema` for `LinkDisplayType.Embed`.
-- Added "downloadsAs" option to `storeBlob` and `storeUrl` in `TemporaryBlobStorage` to specify the file name that it should download as when opening the temporary url.
+- Added "downloadFilename" option to `storeBlob` and `storeUrl` in `TemporaryBlobStorage` to specify the file name that it should download as when opening the temporary url.
 
 ### Changed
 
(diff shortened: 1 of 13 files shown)


ekoleda-codaio:
👍 Interface LGTM

huayang-codaio (approved):

api_types.ts
```diff
@@ -513,16 +513,20 @@ export interface TemporaryBlobStorage {
    *
    * The URL expires after 15 minutes by default, but you may pass a custom expiry, however
    * Coda reserves the right to ignore long expirations.
+   *
+   * You can control what the file downloads as if opened in a browser using the `downloadAs` parameter.
```
huayang-codaio, line 517: can you update the documentation to be more specific about filename? I thought it was file type in the first place.

spencer-codaio: ah yes i should update the changelog too

huayang-codaio: oh, may simply make `file` to `file name`

`what the file downloads as` I was thinking maybe you want to control what a binary blob is downloaded as jpeg or png or what file type. lol

spencer-codaio: ahh gotcha ok ill make this more clear

jonathan-codaio: @ekoleda-codaio Hmm, is `downloadAs` actually clearer than `filename`? I feel like the latter is much more intuitive. If we go with `downloadAs`, it seems like it needs more of a qualifer, because I agree with Huayang it is ambiguous what download property this refers to.

ekoleda-codaio: As you mentioned in the bug comments, it doesn't actually set the filename portion of the temporary URL, so just `filename` would be misleading. I think keeping `download` somewhere in the property name is important, but I think adding `filename` makes it clearer what the value should be. `downloadFilename` seems like a good option to me. (I think "filename" is more common than "file name" these days, and it's what the Node `fs` library uses.)

spencer-codaio: changed to `downloadFilename

huayang-codaio (approved):

CHANGELOG.md
```diff
@@ -2,16 +2,25 @@
 
 This changelog keeps track of all changes to the packs-sdk. We follow conventions from [keepachangelog](https://keepachangelog.com/en/1.0.0/).
 
-### Unreleased
+## [Unreleased]
+
+### Added
+
+- Added "force" option to `LinkSchema` for `LinkDisplayType.Embed`.
+- Added "downloadsAs" option to `storeBlob` and `storeUrl` in `TemporaryBlobStorage` to specify the file name that it should download as when opening the temporary url.
```
jonathan-codaio, line 10: Nit, update here too.

jonathan-codaio (approved):

CHANGELOG.md
spencer-codaio: oops good catch

PR #1914, opened 2022-04-15
Don't implicitly map id/featured/primary to new names during normalization, so we can warn users.

jonathan-codaio:
Also, export our zod error mapping utility so we can reuse it on the `coda` side.

In my first pass at this, `normalizeSchema()` kind of magically remapped id/featured/primary to their new names, which immediately made them future-compatible, but that's actually probably not desirable if we want to warn people to update their source code. Whenever we later rip out the old fields from the SDK, their code would break at that much later date. Better to catch it early.


huayang-codaio (approved):

ekoleda-codaio:
@jonathan-codaio - What is the net effect of this for the Pack maker? No difference to the deprecation effects, it just shows up differently to our server?

patrick-codaio (approved):

jonathan-codaio:
> @jonathan-codaio - What is the net effect of this for the Pack maker? No difference to the deprecation effects, it just shows up differently to our server?

Before this PR, a maker who used old field names would have them transformed automatically to the new field names without their knowledge (which I did not intend), and so would bypass any opportunity we had to warn them about deprecating those field names. But if we eventually remove the old field names entirely from the SDK, they would stop getting the auto-conversion and their code would break at that much later date, without ever having gotten a warning.

The net effect of this PR is to create the intuitive behavior you'd effect from deprecating fields--they don't get automagically mapped to new fields and makers get prompted to change the fields themselves at their own time.

PR #1916, opened 2022-04-18
Bump the SDK version number so that yarn won't resolve packs-sdk-for-web-editor the same commit hash accidentally.

jonathan-codaio:
PTAL @huayang-codaio @patrick-codaio @dweitzman-codaio @coda/packs


huayang-codaio (approved):
oh, interesting.

patrick-codaio (approved):

PR #1918, opened 2022-04-19
Include SDK version in metadata and condition auth.networkDomain validation on it

dweitzman-codaio:
This is to support making backwards incompatible changes in a way that gives a grace period for people on older SDKs before they need to upgrade

A follow-up at some point should be server-side validation to make sure the uploading SDK version isn't "too old", however we define that. This would be the forcing function to make people eventually upgrade their SDK


builder.ts
```diff
@@ -74,6 +75,12 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
    * See {@link PackVersionDefinition.version}.
    */
   version?: string;
+
+  /**
+   * See {@link PackVersionDefinition.sdkVersion}.
+   */
+  sdkVersion?: string;
```
jonathan-codaio, line 82: Could we maybe do something like have `validatePackVersionMetadata()` take the sdkVersion as a second argument and inject it into into the `metadata` value before calling `safeParse()`? (Then remove it before returning it.)

I'd rather that the SDK version remain an out-of-band piece of data that the maker doesn't even have the opportunity to specify themselves, it should always be inferred by the tooling.

helpers/sdk_version.ts
```diff
@@ -0,0 +1,8 @@
+import {ensureExists} from './ensure';
+import semver from 'semver';
+import {version} from '../package.json';
```
jonathan-codaio, line 3: I think we initially did this much simpler and more correct approach for getting the version from package.json, but it wreaked havoc with our CI build-is-up-to-date checks for dependabot PRs. See the PR comments here for some more details: https://github.com/coda/packs-sdk/pull/831

So we probably want to go back to other approach.

builder.ts
dweitzman-codaio: Updated so `validatePackVersionMetadata()` takes an extra argument and consults a list called `packMetadataSchemaBySdkVersion` that can modify the Zod schema conditionally based on the SDK version. Seemed a little cleaner for the moment than mutating the `metadata` temporarily, but I don't have strong feelings either way

helpers/sdk_version.ts
dweitzman-codaio: Reverted this code while removing `sdkVersion` from the metadata itself

testing/upload_validation.ts
```diff
@@ -95,10 +96,26 @@ export class PackMetadataValidationError extends Error {
...
+export async function validatePackVersionMetadata(
+  metadata: Record<string, any>,
+  sdkVersion: string | undefined,
+): Promise<PackVersionMetadata> {
+  let combinedSchema: z.ZodType<Partial<PackVersionMetadata>> = legacyPackMetadataSchema;
+
+  // Server-side validation may be running a different SDK version than the pack maker
+  // is using, so some breaking changes to metadata validation can be set up to only
+  // take effect before or after an SDK version bump.
+  if (sdkVersion) {
```
huayang-codaio, line 108: I guess sdkVersion is only absent for some very old pack versions?

testing/upload_validation.ts
```diff
@@ -1194,43 +1211,52 @@ const legacyPackMetadataSchema = validateFormulas(
...
       path: ['networkDomains'],
     },
-  )
-  .superRefine((untypedData, context) => {
+  );
+const packMetadataSchemaBySdkVersion = [
+  {
     // Check that packs with multiple network domains explicitly choose which domain gets auth.
+    // This is a backward-incompatible validation that takes effect in any pack release after 0.9.0.
+    versionRange: '>0.9.0',
```
huayang-codaio, line 1219: is this >0.9.0 or >=0.9.0? just double check. I don't know which versions they were added.

huayang-codaio (approved):
lgtm.

this approach works. I was thinking about plugging sdkVersion into zod's context object but that requires a patch to the package. This approach is definitely easier.

testing/upload_validation.ts
dweitzman-codaio: Looks like sdkVersion started being included in https://github.com/coda/packs-sdk/pull/970 in May 2021

huayang-codaio: yeah, feel like we can just assume sdkVersion is always available.

testing/upload_validation.ts
dweitzman-codaio: The code was changed while the SDK version was still 0.9.0 without bumping the SDK version, so I used `>0.9.0` to say, "whatever the next version is"

Turns out the next release was `0.10.0`, so there would only be a `0.9.1` release if we decided we wanted to maintain the `0.9` branch do a bugfix or security patch release

testing/upload_validation.ts
huayang-codaio: or actually, I am not sure if we have any packs that old. maybe we do. the current way is good. is it possible to add some log here?

testing/upload_validation.ts
```diff
@@ -1194,43 +1211,52 @@ const legacyPackMetadataSchema = validateFormulas(
         "Specify the domain that your pack makes http requests to using `networkDomains: ['example.com']`",
       path: ['networkDomains'],
     },
-  )
-  .superRefine((untypedData, context) => {
+  );
+const packMetadataSchemaBySdkVersion = [
```
jonathan-codaio, line 1215: Mind giving this a static type?

Right now if somebody adds a new entry to the list and gets the interface wrong, TS will just happily infer this as the union of whatever it was given and we'll fail at runtime instead.

jonathan-codaio (approved):
Nice, this is quite clean.

Just so I understand, once we e.g. require all new uploads with >0.9.0 we would just move these checks into the base schema, right?

testing/upload_validation.ts
dweitzman-codaio: Not sure that here in the SDK wouldn't be the best place to log, but it'd be simple enough to logging or metrics directly from server code before calling into `validatePackVersionMetadata`.

(My preferred approach to tracking the use of deprecated features would probably be a workflow that scans all packs daily and gathers any deprecation warnings or other findings that might impact changes we're hoping to make)

huayang-codaio: > Not sure that here in the SDK wouldn't be the best place to log, but it'd be simple enough to logging or metrics directly from server code before calling into validatePackVersionMetadata.

makes sense. 

> (My preferred approach to tracking the use of deprecated features would probably be a workflow that scans all packs daily and gathers any deprecation warnings or other findings that might impact changes we're hoping to make)

I still feel like it's easier to just query athena if we make all external metadata queryable. but a workflow idea definitely works.

huayang-codaio (approved):

testing/upload_validation.ts
dweitzman-codaio: Done

dweitzman-codaio:
> Nice, this is quite clean.
> 
> Just so I understand, once we e.g. require all new uploads with >0.9.0 we would just move these checks into the base schema, right?

Yes, exactly

test/upload_validation_test.ts
```diff
@@ -40,12 +40,15 @@ import {validateSyncTableSchema} from '../testing/upload_validation';
 import {validateVariousAuthenticationMetadata} from '../testing/upload_validation';
 
 describe('Pack metadata Validation', () => {
-  async function validateJson(obj: Record<string, any>) {
-    return validatePackVersionMetadata(obj);
+  async function validateJson(obj: Record<string, any>, sdkVersion?: string) {
+    const packageJson = await import('../package.json');
```
jonathan-codaio, line 44: I think this is what's tripping up the CI. One of the build tools is too smart for its own good and as far as I can tell sniffs the dynamic imports for string literals, and includes the imported literal paths in the compiled output in `dist`.

The other places where we dynamically import this (intentionally) don't use string literals in part because of this. If you change this line to something like `await import('' + '../package.json');` (and then rebuild) it should avoid the issue. 😱

PR #1920, opened 2022-04-20
Check for deprecated field names in object schemas. Add a warning mode for validation that applies all schema extensions.

jonathan-codaio:
Will update things on the `coda` side in my pending change to call this to generate warnings.

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -46,6 +46,7 @@ import {PostSetupType} from '../types';
 import type {QueryParamTokenAuthentication} from '../types';
 import {ScaleIconSet} from '../schema';
 import type {ScaleSchema} from '..';
+import type {Schema} from '..';
 import type {SetEndpoint} from '../types';
 import {SimpleStringHintValueTypes} from '../schema';
 import type {SimpleStringSchema} from '../schema';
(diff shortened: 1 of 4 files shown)


testing/upload_validation.ts
```diff
@@ -107,7 +111,7 @@ export async function validatePackVersionMetadata(
   // take effect before or after an SDK version bump.
   if (sdkVersion) {
     for (const {versionRange, schemaExtend} of packMetadataSchemaBySdkVersion) {
-      if (semver.satisfies(sdkVersion, versionRange)) {
+      if (warningMode || semver.satisfies(sdkVersion, versionRange)) {
```
dweitzman-codaio, line 114: As a thought experiment, I'm wondering if there could be a hypothetical future where we're temporarily supporting two SDK versions during an upgrade grace period and we'd want to add a validation that only impacts the older one and not the newer one

Accepting "warningMode" as a boolean would still work as an API here, but the implementation might change a little so that `warningMode` is like using the maximum theoretical semantic version

For now, at least, this logic seems fine

dweitzman-codaio (approved):
Seems to get the job done

testing/upload_validation.ts
jonathan-codaio: Yep totally, similar scenarios have occurred to me too. I hurt my head quite a bit trying to arrive at this general approach, so figured we'd start with the simplest thing and then can address specific shortcomings or new use cases as we need to.

PR #1922, opened 2022-04-21
Show deprecation warnings in coda validate.

jonathan-codaio:
To finish this off, we should next read the server-generated deprecation warnings that come back in the API response from `coda upload` and render those too. But this is helpful to allow people to (eventually) check most things for themselves locally. The major issue though is that if they're on an SDK version before this is supported, the ability to check for deprecation warnings locally doesn't exist, and so they would only know they're using deprecated stuff when `coda upload` tells them.

### DIFF
--- a/cli/validate.ts
+++ b/cli/validate.ts
@@ -13,9 +13,10 @@ import {validatePackVersionMetadata} from '../testing/upload_validation';
 
 interface ValidateArgs {
   manifestFile: string;
+  checkDeprecationWarnings: boolean;
 }
 
-export async function handleValidate({manifestFile}: ArgumentsCamelCase<ValidateArgs>) {
+export async function handleValidate({manifestFile, checkDeprecationWarnings}: ArgumentsCamelCase<ValidateArgs>) {
   const fullManifestPath = makeManifestFullPath(manifestFile);
   const {bundlePath} = await compilePackBundle({manifestPath: fullManifestPath, minify: false});
   const manifest = await importManifest<PackVersionDefinition>(bundlePath);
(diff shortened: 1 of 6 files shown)


cli/validate.ts
```diff
@@ -26,10 +27,13 @@ export async function handleValidate({manifestFile}: ArgumentsCamelCase<Validate
...
 
   const metadata = compilePackMetadata(manifest);
-  return validateMetadata(metadata);
+  return validateMetadata(metadata, {checkDeprecationWarnings});
 }
 
-export async function validateMetadata(metadata: PackVersionMetadata) {
+export async function validateMetadata(
+  metadata: PackVersionMetadata,
+  {checkDeprecationWarnings = true}: {checkDeprecationWarnings?: boolean} = {},
```
jonathan-codaio, line 35: Adding the ability to turn this off when called directly, since the `packs` repo will likely have thousands of such warnings before we go fix everything.

setup-env.sh
```diff
@@ -0,0 +1,3 @@
+BASEDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+export PATH=${BASEDIR}/build/node/bin:${BASEDIR}/node_modules/.bin:${PATH}
```
jonathan-codaio, line 3: Just copying this from the `packs` repo, allows us to run `ts-node` unqualified when trying to test the CLI locally. (In the `packs` repo it allows us to run `coda` unqualified.)

dweitzman-codaio: Maybe put that as a comment in the code for future reference?

dweitzman-codaio (approved):

setup-env.sh
jonathan-codaio: Done

PR #1923, opened 2022-04-22
Validate more deprecated fields.

jonathan-codaio:
I think this covers the rest of the breaking changes described in the changelog that can be validated statically in this way.

@patrick-codaio I had wanted to validate that dynamic sync table are specifying an `identityName`, but because we validate the post-compiled metadata, and the code applies the sync table's name as a fallback identity name right now, we can't distinguish the cases and so can't flag anything for the maker here. Ideally we could tell a maker all the breaking changes they will face in the future, but I don't see how here. I think what we'll want to remember to do, is when we eliminate the fallback and require identityName, we should do so with a clear error message that tells people to add `identityName` and to set it to the value of `name` to avoid breaking an existing table that might have already been added to docs.

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -61,6 +61,7 @@ import type {SyncTable} from '../api';
 import type {SyncTableDef} from '../api';
 import type {SystemAuthenticationTypes} from '../types';
 import {Type} from '../api_types';
+import type {UnionType} from '../api_types';
 import type {ValidationError} from './types';
 import {ValueHintType} from '../schema';
 import {ValueType} from '../schema';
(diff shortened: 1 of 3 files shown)


testing/upload_validation.ts
```diff
@@ -1326,6 +1354,17 @@ function validateObjectSchemaDeprecatedFields(
     pathPrefix,
     context,
   });
+
+  if (schema.identity?.attribution) {
```
dweitzman-codaio, line 1358: Doesn't seem that critical, but I noticed that the unit tests don't cover this warning

dweitzman-codaio (approved):

testing/upload_validation.ts
jonathan-codaio: Ah I had a feeling I forgot a case, thanks!

PR #1925, opened 2022-04-25
Show server-generated deprecation warnings at the end of coda upload.

jonathan-codaio:
Should be the last piece of the puzzle. Makes the CLI show the same output as the web editor when uploading a new pack version if the server determines there are deprecation warnings. In theory this is the same output as what `validateMetadata()` with `checkDeprecationWarnings: true` would return locally, but we should consider the server authoritative in case we add server-side only checks.

All the changes are in `upload.ts`. The rest is just bringing the API codegen up to date to get the new type def that includes deprecation warnings on the upload response.

PTAL @dweitzman-codaio @patrick-codaio @coda/packs

### DIFF
--- a/cli/upload.ts
+++ b/cli/upload.ts
@@ -171,7 +171,7 @@ export async function handleUpload({
     }
 
     const {deprecationWarnings} = uploadCompleteResponse;
-    if (deprecationWarnings) {
+    if (deprecationWarnings?.length) {
       print(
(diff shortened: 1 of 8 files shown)


dweitzman-codaio (approved):
Nice

jonathan-codaio:
Oops forgot to include screenshot

Screen Shot 2022-04-25 at 1 22 04 PM

PR #1927, opened 2022-04-26
Don't hard fail on breaking changes until future version 1.0.0.

jonathan-codaio:
Our intent is that our next SDK version will expose the new fields and allow for a transition period, and 2 releases from now is when we would actually start enforcing them, so we need our semver filters to reflect that.

PTAL @dweitzman-codaio @huayang-codaio @alexd-codaio @coda/packs

### DIFF
--- a/dist/testing/upload_validation.js
+++ b/dist/testing/upload_validation.js
@@ -1021,8 +1021,8 @@ const legacyPackMetadataSchema = validateFormulas(unrefinedPackVersionMetadataSc
 const packMetadataSchemaBySdkVersion = [
     {
         // Check that packs with multiple network domains explicitly choose which domain gets auth.
-        // This is a backward-incompatible validation that takes effect in any pack release after 0.9.0.
-        versionRange: '>0.9.0',
+        // This is a backward-incompatible validation that takes effect in any pack release after 1.0.0.
+        versionRange: '>=1.0.0',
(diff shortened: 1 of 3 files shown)


spencer-codaio:
how does this work with our next release if we don't want to enforce the breaking changes for 2 more releaes? Is the thinking that we'd bump this number after the next release to our future "breaking changes" one?

jonathan-codaio:
> how does this work with our next release if we don't want to enforce the breaking changes for 2 more releaes? Is the thinking that we'd bump this number after the next release to our future "breaking changes" one?

The next release (intended to be 0.10, but there was some wonkiness on the `coda` side, so it'll likely be 0.11) will expose all the new properties and features, so folks can start using them, but not require them. The logic to show deprecation warnings will apply every schema extension in `packMetadataSchemaBySdkVersion` regardless of SDK version and generate warnings. So when people upgrade to 0.11 they'll be able to see warnings, and switch to new fields at their leisure, but not get any hard errors. Only when they upgrade to 1.0 will we enforce these as errors.

dweitzman-codaio (approved):
I'll need to make sure I update my comment in the other PR to say that networkDomain is required in >=1.0.0 rather than 0.10.0

huayang-codaio (approved):

PR #1928, opened 2022-04-26
mark a few more methods as deprecated

huayang-codaio:
saw these when I was here. @coda/packs

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -261,8 +261,6 @@ export type GenericDynamicSyncTable = DynamicSyncTableDef<any, any, ParamDefs, a
 export type SyncTable = GenericSyncTable | GenericDynamicSyncTable;
 
 /**
- * @deprecated
- *
  * Helper to determine if an error is considered user-visible and can be shown in the UI.
  * See {@link UserVisibleError}.
  * @param error Any error object.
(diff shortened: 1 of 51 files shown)


api.ts
```diff
@@ -261,6 +261,8 @@ export type GenericDynamicSyncTable = DynamicSyncTableDef<any, any, ParamDefs, a
 export type SyncTable = GenericSyncTable | GenericDynamicSyncTable;
 
 /**
+ * @deprecated
```
jonathan-codaio, line 264: How come this one?

huayang-codaio: this is export from `legacy_exports`. maybe it's a mistake?

jonathan-codaio: Ah ok. I would just remove `@deprecated` but you can leave otherwise leave it alone for now.

I think what probably happened is that everything that used to be in `index.ts` that we seemed to only use for implementation purposes I moved to legacy_exports. I think I was trying to avoid importing anything that wasn't in index.ts or legacy_exports.ts so we could be disciplined about what the interface of the SDK was. We kind of gave up on that for decent reasons and so if it was today I would just remove this from legacy_exports and import it directly from `/dist/api`, but no need to bother right now.

jonathan-codaio (approved):

PR #1929, opened 2022-04-27
Normalize schema property

alexd-codaio:


huayang-codaio (approved):

jonathan-codaio (approved):
Ah ☹️ I should see if we can add a simple test that all the properties are coming through in practice.

PR #1930, opened 2022-04-28
Add migration guide, and update the rest of the docs

ekoleda-codaio:
Staged: [Migration guide](https://coda-packs-sdk--1930.com.readthedocs.build/en/1930/support/migration/)
Also updated the rest of the documentation to reflect the post-migration state of the SDK. The `.ts` samples were previously fixed, but not the prose or inline samples.


docs/support/migration.md
```diff
@@ -0,0 +1,207 @@
...
+
+
+## Version 0.11.0
+
+In preparation for a public launch, we intentionally introduced a number of breaking changes to make the SDK easier to use and understand. We hope to avoid breaking changes on this scale in the future.
+
+
+### Rename schema fields `id`, `primary`, and `featured`
+
+__Affects__: Packs that define a [schema][schemas].<br>
```
jonathan-codaio, line 17: I think technically only affects object schemas.

docs/support/migration.md
```diff
@@ -0,0 +1,207 @@
+---
+title: Migration
+---
+
+# Migrating between SDK versions
+
+While the Packs team strives to maintain backwards compatibility between SDK versions, there may occasionally be times where you need to change your code to work with a new SDK version. This page will include instructions on how to perform the migration. A detailed list of all changes in the SDK can be found in the [Changelog][changelog].
+
+
+## Version 0.11.0
```
jonathan-codaio, line 10: Probably fine but may be worth extra emphasis that they must have upgraded to version 0.11.0 in order to be able to attempt migration. (And we should provide the command for upgrading if they use the CLI.)

jonathan-codaio (approved):
So good!

Probably also worth mentioning that once they are on 0.11.0, they can use `coda validate` to generate deprecation warnings about the things they should fix. I think we will produce warnings about everything except the `identityName` change, which isn't currently detectable because we put the sync table name as a fallback currently.

The web editor will also start showing the same deprecation warnings when we bump it to 0.11.0 and flip a flag to show them.

docs/support/migration.md
```diff
@@ -0,0 +1,207 @@
...
+
+
+### Add `identityName` to dynamic sync tables
+
+__Affects__: Packs that include a dynamic sync table.<br>
+__Action Required__: Add new code.
+
+Like regular sync tables, dynamic sync tables now require the `identityName` field to be set. This will be used along with the dynamic URL to set the identity of the table. You no longer need to set the `identity` field of the schema generated in the `getSchema` function, as it will be constructed for you automatically.
+
+To ensure that existing dynamic sync tables in your user's docs continue to work we strongly recommend you set this value to be the same as the name of the table.
```
patrick-codaio, line 51: Can we be more explicit? We could add "If you don't, references to your dynamic sync table may break and row updates from action return values may stop working."?

documentation/examples/cats/full.ts
```diff
@@ -121,10 +121,10 @@ const CatSchema = coda.makeObjectSchema({
...
       codaType: coda.ValueHintType.DateTime,
     },
-    id: { type: coda.ValueType.String },
+    catId: { type: coda.ValueType.String },
   },
   displayProperty: "image",
-  idProperty: "id",
+  idProperty: "catId",
   featuredProperties: ["tags"],
   identity: {
```
patrick-codaio, line 129: Looks like we can delete this?

ekoleda-codaio: This was me trying to fix something unrelated in the same PR :-) I can remove for focus.

documentation/examples/cats/sync_table.ts
```diff
@@ -16,10 +16,10 @@ const CatSchema = coda.makeObjectSchema({
       type: coda.ValueType.String,
       codaType: coda.ValueHintType.DateTime,
     },
-    id: { type: coda.ValueType.String },
+    catId: { type: coda.ValueType.String },
   },
   displayProperty: "image",
-  idProperty: "id",
+  idProperty: "catId",
   featuredProperties: ["tags"],
```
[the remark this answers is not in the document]
ekoleda-codaio: Ditto.

docs/support/migration.md
ekoleda-codaio: Done. Added some new sections to the permanent Migration guide, and moved the version-specific stuff to a sub-page.

docs/guides/advanced/schemas.md
```diff
@@ -406,7 +403,7 @@ let PersonReferenceSchema =
     coda.makeReferenceSchemaFromObjectSchema(PersonSchema);
 ```
 
-In some instances you may have to create the reference schema manually however, like when you want a row to be able to reference other rows in the same table. A reference schema is an object schema with the `codaType` field set to `Reference`. It must specify both an `id` and `primary` property, and those properties must be marked as `required`. It must also have an `identity` set, with the name matching that of the schema used by the target sync table.
+In some instances you may have to create the reference schema manually however, like when you want a row to be able to reference other rows in the same table. A reference schema is an object schema with the `codaType` field set to `Reference`. It must specify both an `idProperty` and `displayProperty`, and those properties must be marked as `required`. It must also have an `identity` set, with the name matching that of the schema used by the target sync table.
```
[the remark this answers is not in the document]
ekoleda-codaio: Done

docs/guides/blocks/sync-tables/index.md
```diff
@@ -128,7 +128,7 @@ In some areas of the UI, such as the settings pane, the name will be rewritten t
     Changing the name of a sync table will break any existing docs that use it. When creating your Pack select your names carefully.
 
 
-## Identity
+## Identity {: #identity}
 
 Every sync table is required to specify an [`identityName`][identityName], which is a unique identifier for the type of item being synced in that table. By convention this is usually the singular form of the name of the item being synced. For instance, a "Tasks" table would use an identity name of "Task". This identity name is used to reference the table when [creating row references](#references) or [updating rows with actions](#actions).
```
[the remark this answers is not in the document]
ekoleda-codaio: I've tried to keep the regular sync tables guide free of dynamic sync table details, to avoid overloading the reader. I'll take an AI to add this detail to the dynamic sync tables guide, as I have another update I want to do there outside this migration work.

patrick-codaio (approved):

PR #1959, opened 2022-05-02
Remove transformBody from lambda

alexd-codaio:
@jonathan-codaio @spencer-codaio @coda/packs PTAL

Looks good? As it turns out, I don't think it's a huge deal if we run `transformBody` more than once (it will just do nothing), but we need to run it at least once. :D


alexd-codaio:
Huge failwhale on the tests... digging in.

handler_templates.ts
```diff
@@ -354,7 +354,5 @@ export function generateObjectResponseHandler<T extends Schema>(
     if (!finalSchema) {
       return projectedBody;
     }
-
-    return transformBody(projectedBody, finalSchema);
```
huayang-codaio, line 358: do you just want to return `projectedBody` regardless of `finalSchema` here?

huayang-codaio (commented):
lgtm but be aware that 0.11.0 has been cut. this might be 0.11.0

handler_templates.ts
alexd-codaio: Yes. Done.

huayang-codaio (approved):
cool. just realized we are cutting this to 0.12.0 from https://github.com/coda/coda/pull/71238

huayang-codaio:
oh man, there are tons of test cases to update. :)

alexd-codaio:
Fixed the tests.

@jonathan-codaio @spencer-codaio  - One artifact of this change is that it will/could break tests that assume that `transformBody` is supposed to be called by the lambda glue. Not sure if we're worried about that or not.

test/execution_test.ts
```diff
@@ -207,7 +207,7 @@ describe('Execution', () => {
               }),
             },
             execute: async ([valid]) => {
-              return valid ? {foo: 'blah'} : {foo: 123};
+              return valid ? {Foo: 'blah'} : {Foo: 123};
```
jonathan-codaio, line 210: How come many of these are changing to use normalized-style keys?

jonathan-codaio (commented):
Hmm, yeah this is going to break all the tests in `packs`, `packs-examples`, and any test a third-party developer has written, since they all assert based on normalized output.

Mostly I think that's a good thing, it was really confusing before that the output you see in a unittest or in the CLI is normalized when you as the developer don't really care. It will be much more intuitive that when testing with only the SDK the output is literally what you return from your `execute` implementation.

But it's disruptive, and we've gone to great lengths to package up our current round of breaking changes into a big announcement this week including a migration guide, and we can't immediately follow it up with more breaking changes.

I wonder if we want to add an option to `ExecuteOptions` in `executeFormulaFromPackDef` and `executeSyncFormulaFromPackDef` that toggles whether or not the output is normalized, and default it to true to preserve backwards compatibility, and then for 1.0.0 maybe we invert the default?

CC @ekoleda-codaio

test/execution_test.ts
alexd-codaio: The tests relied on the fact that we did normalization of the object output?

ekoleda-codaio:
It seems like a good change to make, and it would be nice if we could work this into our existing breaking moment. Given this only affects the CLI we don't _need_ to provide a toggle for it, since CLI users can choose when to upgrade SDK versions.

test/execution_test.ts
huayang-codaio: I guess the argument is that we should probably change the expected output, instead of changing the `execute` function for the tests.

alexd-codaio:
OK. Updated this to retain the behavior within tests and a flag for disabling it. Added changelog entry.

PTAL @jonathan-codaio @ekoleda-codaio

jonathan-codaio (commented):
Mind just patching in this commit hash to `packs` and making sure `make test` still passes?

test/validation_test.ts
```diff
@@ -85,7 +85,7 @@ describe('Property validation in objects', () => {
     examples: [],
     parameters: [makeBooleanParameter('boolParam', 'Pass in a boolean (malformed is ok)')],
     execute: async ([boolParam]) => {
-      return {bool: boolParam};
+      return {Bool: boolParam};
```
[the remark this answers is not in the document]
alexd-codaio: Well, the schema is also being normalized, so `bool: {type: ValueType.Bool}` is actually `Bool: {type: ValueType.Bool}` in the tests...

I can revert these changes if you'd like, but thought it was easier to keep them around since it's going to have to be changed anyway.

alexd-codaio:
@jonathan-codaio Tests pass with that commit BTW.

image

test/validation_test.ts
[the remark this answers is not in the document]
alexd-codaio: As discussed, I reverted most of the test changes here and will let someone clean it up once we remove this flag. 

I retained the VM ones, since I didn't think it was necessary to add the normalization logic since these didn't appear in the user-visible test framework(?).

jonathan-codaio (approved):

PR #1962, opened 2022-05-03
Use cross-fetch instead of isomorphic-fetch

oleg-codaio:
https://www.npmjs.com/package/isomorphic-fetch doesn't really seem maintained anymore - moving to an alternative, `cross-fetch`, as suggested in the README.

PTAL @coda/packs


jonathan-codaio (commented):
Is this a fully drop-in replacement for isomorphic-fetch? Any code that does error handling is reliant on the quirks of the given library.

oleg-codaio:
There could be subtle differences, but isomorphic-fetch isn't maintained and takes a transitive dependency on an early version of node-fetch that's vulnerable to CVE-2022-0235: https://nvd.nist.gov/vuln/detail/CVE-2022-0235

jonathan-codaio:
Would love if somebody on @coda/packs has a few minutes to check out this branch and test the CLI with it. Should be fine to test against prod or staging, it's only the client changes that we would care about. In particular want to check that things like `coda upload` work as written when there are errors such that the various API calls return non-200 responses, making sure the error handling in the API client and CLI still work as intended.

jonathan-codaio:
> Would love if somebody on @coda/packs has a few minutes to check out this branch and test the CLI with it. Should be fine to test against prod or staging, it's only the client changes that we would care about. In particular want to check that things like `coda upload` work as written when there are errors such that the various API calls return non-200 responses, making sure the error handling in the API client and CLI still work as intended.

@alan-codaio would you mind testing out the CLI commands using this branch tomorrow to make sure there aren't any behavior changes, particularly around error handling?

As a reminder, after running `source setup-env.sh` in the repo you can do `ts-node cli/coda.ts upload ...` and run any of the CLI commands using code right in the branch.

alan-codaio:
Screen Shot 2022-05-05 at 10 07 58 PM

Works fine for me 👍

jonathan-codaio (approved):

PR #1965, opened 2022-05-03
Allow sync tables to reuse schemas

patrick-codaio:
https://golinks.io/bug/23002


requirements.txt
```diff
@@ -1,3 +1,10 @@
+
+            Warning: The lock flag -r/--requirements will be deprecated in a future version
```
dweitzman-codaio, line 2: Curious how if this warning was added manually or by some kind of tooling. Does it not need to be inside a `#`-style comment?

test/upload_validation_test.ts
```diff
@@ -771,6 +771,60 @@ describe('Pack metadata Validation', () => {
...
+        const metadata = createFakePack({
+          syncTables: [table1, table2],
+        });
+        const validatedMetadata = await validateJson(metadata);
+
+        const {schema: schema1} = validatedMetadata.syncTables[0];
+        const {schema: schema2} = validatedMetadata.syncTables[1];
+        assert.ok(schema1);
+        assert.ok(schema2);
+        // assert.deepEqual(schema.identity, {name: 'SomeIdentity'});
```
dweitzman-codaio, line 825: There's a commented-out line here. Did you intend to have this assert on the identity of the return values?

requirements.txt
ekoleda-codaio: This file is generated here:

https://github.com/coda/packs-sdk/blob/c3a47e4bfc19ac0fe9bb2da11f02fd31fcb6d7c6/Makefile#L42

Readthedocs requires it (it doesn't support Pipfile). Looks like this does [break the Readthedocs build](https://readthedocs.com/projects/coda-packs-sdk/builds/934991/).

jonathan-codaio (approved):

requirements.txt
patrick-codaio: Oops, sorry about that, removed for now. I'll take a stab in a different PR.

test/upload_validation_test.ts
patrick-codaio: Thank you for reminding me that I never finished the test!!

requirements.txt
patrick-codaio: https://github.com/coda/packs-sdk/pull/1972

PR #1971, opened 2022-05-04
Fix a super minor typo in the migration guide

dweitzman-codaio:
s/net/set/ in "For affected Packs, net the networkDomain field..."

Also removed a newline in the changelog that wraps a line early

All the other changes were automatically made by my format-on-save settings
in vscode. Using "**" instead of "__" for bold text does seem more common
in the rest of the codebase, but also happy to revert those changes (I really
just wanted to fix the typo).


ekoleda-codaio (approved):
Perfect, thanks for the fix!

PR #1972, opened 2022-05-04
Fix pipenv warning

patrick-codaio:
As [seen here](https://github.com/coda/packs-sdk/pull/1965#discussion_r864022046) pipenv does not want us to use `lock` anymore.


Makefile
```diff
@@ -39,7 +39,7 @@ _bootstrap-python-requirements:
 	# Ensure requirements.txt (used by Read the Docs) is in sync.
 	# Keep this separate from _bootstrap-python since the output changes based
 	# on the version of pipenv installed.
-	${PIPENV} lock -r > requirements.txt
+	${PIPENV} requirements > requirements.txt
```
dweitzman-codaio, line 42: Could be nice to make this something in the spirit of `echo "# These requirements were autogenerated by 'make boostrap'" && ${PIPENV} requirements > requirements.txt` so the file still has a note about being auto-generated

dweitzman-codaio (approved):

requirements.txt
```diff
@@ -1,34 +1,27 @@
-#
-# These requirements were autogenerated by pipenv
-# To regenerate from the project's Pipfile, run:
-#
-#    pipenv lock --requirements
-#
-
 -i https://pypi.org/simple
-bracex==2.2.1; python_version >= '3.6'
```
jonathan-codaio, line 9: Does the loss of these `; python_version >= '3.6'` clauses matter? I suspect not.

patrick-codaio: It didn't change the output of `make bs && make build`, so I guess not.

Makefile
patrick-codaio: Done!

PR #1976, opened 2022-05-05
Allow multiple auth domains

dweitzman-codaio:
While uncommon, some APIs do seem to require the same credentials to be used across multiple domains.

An example would be the Google Drive API, which serves thumbnail images off a different domain than file metadata but still requires thumbnail requests to send credentials.

The change here is to modify the type of the authentication `networkDomain` from `string` to `string | string[]` and update `upload_validation.ts` accordingly.

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -308,6 +308,13 @@ const setEndpointPostSetupValidator = zodCompleteObject<SetEndpoint>({
   'Either getOptions or getOptionsFormula must be specified.',
 );
 
+const singleAuthDomainSchema = z
+  .string()
+  .nonempty()
+  .refine(domain => domain.indexOf(' ') < 0, {
+    message: 'The `networkDomain` in setUserAuthentication() cannot contain spaces. Use an array for multiple domains.',
+  });
+
 const baseAuthenticationValidators = {
   // TODO(jonathan): Remove these after fixing/exporting types for Authentication metadata, as they're only present
   // in the full bundle, not the metadata.
@@ -318,7 +325,7 @@ const baseAuthenticationValidators = {
   endpointDomain: z.string().optional(),
   // The items are technically a discriminated union type but that union currently only has one member.
   postSetup: z.array(setEndpointPostSetupValidator).optional(),
-  networkDomain: z.union([z.string().nonempty(), z.string().nonempty().array().nonempty()]).optional(),
+  networkDomain: z.union([singleAuthDomainSchema, z.array(singleAuthDomainSchema).nonempty()]).optional(),
 };
 
 const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny> = {
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -318,7 +318,7 @@ const baseAuthenticationValidators = {
   endpointDomain: z.string().optional(),
   // The items are technically a discriminated union type but that union currently only has one member.
   postSetup: z.array(setEndpointPostSetupValidator).optional(),
-  networkDomain: z.string().optional(),
+  networkDomain: z.union([z.string().nonempty(), z.string().nonempty().array().nonempty()]).optional(),
 };
 
 const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny> = {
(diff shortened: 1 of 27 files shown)


testing/upload_validation.ts
```diff
@@ -318,7 +318,7 @@ const baseAuthenticationValidators = {
   endpointDomain: z.string().optional(),
   // The items are technically a discriminated union type but that union currently only has one member.
   postSetup: z.array(setEndpointPostSetupValidator).optional(),
-  networkDomain: z.string().optional(),
+  networkDomain: z.union([z.string().nonempty(), z.string().nonempty().array().nonempty()]).optional(),
```
huayang-codaio, line 321: I thought we usually do `z.array(z.string().nonempty()).nonempty()`. but as long as it works.

huayang-codaio (approved):

types.ts
```diff
@@ -253,10 +253,12 @@ export interface BaseAuthentication {
...
 
   /**
-   * Which domain should get auth credentials, when a pack is configured with multiple domains.
+   * Which domain(s) should get auth credentials, when a pack is configured with multiple domains.
    * Packs configured with only one domain or with requiredsEndpointUrl set to true can omit this.
+   *
+   * Using multiple authenticated network domains is uncommon and requires Coda approval.
    */
-  networkDomain?: string;
+  networkDomain?: string | string[];
```
huayang-codaio, line 261: btw. is anyone (external) actually using this? if not we can probably just make a breaking change? @jonathan-codaio

dweitzman-codaio: I suspect it's only the unreleased Box pack which needs multi-domain auth and would require changes anyway, but potentially anyone with multiple domains who switched to SDK 0.11.0 

Not sure if there's any quick and easy way to search for packs with this in the metadata, but we could scan s3

testing/upload_validation.ts
```diff
@@ -1268,13 +1275,16 @@ const packMetadataSchemaBySdkVersion: SchemaExtension[] = [
...
         // headers.
-        if (!data.networkDomains?.includes(data.defaultAuthentication.networkDomain)) {
-          context.addIssue({
-            code: z.ZodIssueCode.custom,
-            path: ['defaultAuthentication.networkDomain'],
-            message: 'The `networkDomain` in setUserAuthentication() must match a previously declared network domain.',
-          });
-          return;
+        for (const authNetworkDomain of authNetworkDomains) {
+          if (!data.networkDomains?.includes(authNetworkDomain)) {
```
huayang-codaio, line 1279: can you also validate that the network domain has no space characters? I just saw that you use that to encode the list.

types.ts
huayang-codaio: not a big deal though. just calling out in case if it's easy.

dweitzman-codaio: Oh, actually validation is currently for SDK versions `>=1.0.0`, so probably the Box pack is the only one that even noticed this existed -- before we added the version-based verification. Seems likely that that pack + first party packs are the only ones using this field

dweitzman-codaio: Personally I'm comfortable making this a breaking change since it affects essentially nobody who doesn't have to update their config anyway

The change would presumably be to rename the field to `networkDomains` with an "s" and make type always an array `string[]`? This would only affect people using multiple network domains, since packs with a single domain can omit the field entirely

@jonathan-codaio and @ekoleda-codaio, thoughts?

testing/upload_validation.ts
dweitzman-codaio: Done

types.ts
jonathan-codaio: Probably wouldn't affect anybody, but on principle considering we just announced the new field yesterday, we shouldn't go break it right away.

I actually think the non-breaking `networkDomain?: string | string[];` might be the best interface anyway--for the vast majority of developers they will either omit this or provide a single domain as a string. It's only for really really rare cases that it needs to be an array, so it would kind of be the tail wagging the dog if we designed the interface around that.

What we're learning from our makers is that the conceptually correct interface isn't really as helpful as one that kinds them towards common behaviors.

jonathan-codaio (commented):
Mind updating the CHANGELOG to include a new [Unreleased] entry with this?

types.ts
ekoleda-codaio: +1 to Jonathan's points.

testing/upload_validation.ts
dweitzman-codaio: Done

testing/upload_validation.ts
dweitzman-codaio: > Mind updating the CHANGELOG to include a new [Unreleased] entry with this?

Done

PR #1979, opened 2022-05-06
Document sparse array parameters

ekoleda-codaio:
- Documented sparse array parameters
- Created a sample for regular array parameters and sparse array parameters
- Added existing samples into the parameter docs, in collapsable sections.
- Added existing samples into the parameters sample category.


docs/guides/basics/parameters/index.md
```diff
@@ -131,10 +146,30 @@ Images and files that the user uploaded to the doc will be hosted on the `codaho
...
+
+??? example "Example: Longest string formula"
+    ```ts
+    --8<-- "examples/parameter/longest.ts"
+    ```
+
+
+### Table column
+
+Passing a table column into an array parameter can be error prone, because if that column is empty for some rows then the formula will fail to run. To accept a list of values that may include empty values there use the sparse variant of the array parameter (`SparseStringArray`, `SparseNumberArray`, etc). Blank cells will be representd as `null`, and you'll need to make sure your code can handle those values.
```
huayang-codaio, line 167: maybe `there use` -> `there, use`? ignore if it's just my reading..

huayang-codaio (approved):
thank you

docs/guides/basics/parameters/index.md
ekoleda-codaio: Good catch, some leftover word. Removed, and made other fixes / improvements.

PR #1988, opened 2022-05-10
update docs to mention you must be on latest release to publish or templatize

gary-codaio:
ptal @ekoleda-codaio


ekoleda-codaio (approved):
Excellent! Thanks so much for contributing.

PR #1992, opened 2022-05-13
Undo bad auto-formatting

ekoleda-codaio:
The file `authentication.md` was touched in PR #1907 and it looks like an auto-formatter went awry. Our documentation is using some plugins that introduce non-standard markdown syntax that broken when reformatted.

I used the approach [documented here](https://stackoverflow.com/a/7196615) to revert the change to this file.


jonathan-codaio (approved):
Oof, sorry about that!

PR #1993, opened 2022-05-13
Add digram for continuations

ekoleda-codaio:
I've noticed a few developers being confused by the continuation flow, tried adding a diagram to clear it up. The layout isn't 100% perfect, but Mermaid doesn't provide much control there.

Staged here: https://coda-packs-sdk--1993.com.readthedocs.build/en/1993/guides/blocks/sync-tables/#longrunning


jonathan-codaio (approved):
"A picture is worth 100 lines of documentation" so some such

PR #1995, opened 2022-05-16
Add test to catch markdown auto-formatting problems.

ekoleda-codaio:


test/doc_formatting_test.ts
```diff
@@ -0,0 +1,45 @@
...
+const MarkdownPattern = path.join(DocsRoot, '**/*.md');
+
+interface MarkdownFile {
+  path: string;
+  content: string;
+}
+
+async function getMarkdownFileContent(): Promise<MarkdownFile[]> {
+  const files = glob.sync(MarkdownPattern, {});
+  const buffers = await Promise.all(files.map(file => fs.promises.readFile(file)));
```
jonathan-codaio, line 21: Nit, is this actually better than just doing `fs.readFileSync()` and doing it all synchronously?

jonathan-codaio (approved):

test/doc_formatting_test.ts
ekoleda-codaio: It is 33% faster, but that's only 30ms 😄  Changed it to sync for simplicity.

PR #1996, opened 2022-05-16
Allow empty auth network domain with multiple networks

dweitzman-codaio:
Now that we have multi-domain credential pinning, we can just infer all
networks as auth networks if nothing explicit is specified. Using multiple
auth networks would still be blocked on the server-side without manual
approval from Coda.

### DIFF
--- a/dist/testing/upload_validation.js
+++ b/dist/testing/upload_validation.js
@@ -1061,26 +1061,18 @@ const packMetadataSchemaBySdkVersion = [
                 else if (data.defaultAuthentication.networkDomain) {
                     authNetworkDomains = [data.defaultAuthentication.networkDomain];
                 }
-                if (!(authNetworkDomains === null || authNetworkDomains === void 0 ? void 0 : authNetworkDomains.length)) {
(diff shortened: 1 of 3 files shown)


dweitzman-codaio:
Chatted with Huayang a bit. Another possibility would be keep the requirement that pack authors explicitly list auth domains when there are multiple domains, but make it conditional on SDK version.

That'd encourage new pack authors to be explicit about what domains get auth while allowing existing packs to keep functioning until they upload a new version

huayang-codaio:
oh haha, apparently we commented on different PRs. :P

jonathan-codaio:
Ah, yeah I think it would be better if you had to explicitly specify your auth domain(s) when using multiple network domains. It would be exceedingly rare to use multiple auth domains even when using multiple network domains, so I don't think we should default to that rare and less safe case.

Which existing packs are affected, is it just Dropbox? I'm fine if you want to add a hack so the existing version doesn't break, otherwise if all it requires is re-releasing the pack let's just do that, we want to re-release all of our packs on >= 0.11.0 anyway.

PR #1999, opened 2022-05-18
Add special OAuth parameters that probably only Slack needs

patrick-codaio:
The new [Slack OAuth API](https://api.slack.com/authentication/migration) has a couple changes that our current SDK can't handle. We could hack these features into the Coda code base, but that would not allow other developers to make their own Slack Pack.

First, Slack Apps cannot request scopes that are beyond what the Slack App's profile is updated to. This is a problem because in SlackOauth world, `scope` means bot scopes, and `user_scope` is what we want to use. We had already supported `user_scope` as an `additionalParam`, but now Slack will refuse to work unless we additionally *remove* the `scope` param.

Second, the `access_token` in the OAuth response from Slack is now the bot scope, and the *user* token that we want is now nested in a sub-field (`authed_user`).

I cannot test these changes with the CLI for real because Slack no longer allows non-https redirect URLs, but I did test them end-to-end in dev by [creating a new Slack Dev pack](https://github.com/coda/packs/pull/3103) and [verifying it works](https://github.com/coda/coda/pull/72001). 

If we like this approach, I can add unit tests (it's not trivial because `oauth_server.ts` is only tested by `auth_test.ts` currently, which treats it like a black box).


types.ts
```diff
@@ -433,6 +433,20 @@ export interface OAuth2Authentication extends BaseAuthentication {
...
    */
   useProofKeyForCodeExchange?: boolean;
+
+  /**
+   * In rare cases, OAuth providers may want the permission scopes in a different query parameter
+   * than `scope`. You can use the `additionalParams` above to accomplish that. In even rarer
+   * cases, some of those OAuth providers sometimes want you to also *not* send the normal scope
+   * param (which this parameter enables).
+   */
+  suppressScopeParam?: boolean;
```
jonathan-codaio, line 443: Would it work to make this `scopeParamName` instead and have it override what we use for the scope param? For Slack, we then set `scopeParamName: 'user_scope'` and then remove the hack where we currently pass the scopes as additionalParams.

jonathan-codaio:
Surfacing for @ekoleda-codaio for visibility as well, though he's out sick.

types.ts
```diff
@@ -433,6 +433,20 @@ export interface OAuth2Authentication extends BaseAuthentication {
    * See https://datatracker.ietf.org/doc/html/rfc7636 for more details.
    */
   useProofKeyForCodeExchange?: boolean;
+
+  /**
```
dweitzman-codaio, line 437: Thinking aloud... I wonder if it'd make sense at some point to have a way to clearly tag certain new options as "experimental" in the SDK so users know that those options are new and their behavior could change in future SDK releases.

If we had a way to mark experimental options, seems like an unusual corner case with only one known 1st-party-pack usage might be a good use case for it.

jonathan-codaio: Yeah, would be nice. Possibly we could use the `@internal` or `@category` options that are available in TypeDoc?

https://typedoc.org/guides/doccomments/

patrick-codaio: I like this idea, but I feel like it deserves more thought beyond the urgent need of avoiding slack downtime, so I'd like to defer.

testing/oauth_server.ts
```diff
@@ -90,19 +90,24 @@ export function launchOAuthServerFlow({
...
+    const responseBody = await oauthResponse.json();
+    const tokenContainer = nestedResponseKey ? responseBody[nestedResponseKey] : responseBody;
+    const {access_token: accessToken, refresh_token: refreshToken, ...data} = tokenContainer;
 
     return {accessToken, refreshToken, data};
   };
   const serverContainer = new OAuthServerContainer(callback, afterTokenExchange, port);
 
-  const authorizationUri = withQueryParams(authorizationUrl, {
-    scope,
```
jonathan-codaio, line 100: Nit, if you wanted to still do this inline, I think you could do `[scopeParamName || 'scope']: scope`.

jonathan-codaio (approved):

types.ts
patrick-codaio: Great idea! Done, and tested with `coda auth` on an updated https://github.com/coda/packs/pull/3103

testing/oauth_server.ts
patrick-codaio: I tried this, but something very strange happens:

Screen Shot 2022-05-18 at 2 15 43 PM

Screen Shot 2022-05-18 at 2 15 52 PM

So I'm reverting to the verbose version.

jonathan-codaio: Oh this is just specific to the Slack pack though because there's still `additionalParams: {user_scope: '...'}` in that manifest, right?

The only difference is the ordering--in the verbose one you're writing user_scope after additionalParams, in the terse one you're writing user_scope before additionalParams. Maybe you could just move `[scopeParamName || 'scope']: scope,` to the bottom if you want it to take precedence over `additionalParams`.

patrick-codaio: OHHHH yes I need to remove `additionalParams` explicitly, DOH

PR #2001, opened 2022-05-18
Fix validation of auth network domains

dweitzman-codaio:
Auth network domains are optional until a future SDK release, but if someone
chooses to set them now we still want to ensure the value they choose matches
a pack network domain

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -70,6 +70,7 @@ import type {VariousSupportedAuthenticationTypes} from '../types';
 import type {WebBasicAuthentication} from '../types';
 import {assertCondition} from '../helpers/ensure';
 import {isArray} from '../schema';
+import {isDefined} from '../helpers/object_utils';
 import {isNil} from '../helpers/object_utils';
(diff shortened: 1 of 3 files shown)


testing/upload_validation.ts
```diff
@@ -1232,13 +1233,58 @@ const legacyPackMetadataSchema = validateFormulas(
         "Specify the domain that your pack makes http requests to using `networkDomains: ['example.com']`",
       path: ['networkDomains'],
     },
-  );
+  )
+  .superRefine((untypedData, context) => {
```
dweitzman-codaio, line 1237: This is enforced regardless of SDK version

testing/upload_validation.ts
```diff
@@ -1248,26 +1294,14 @@ const packMetadataSchemaBySdkVersion: SchemaExtension[] = [
       return schema.superRefine((untypedData, context) => {
```
dweitzman-codaio, line 1294: This is only enforced starting at SDK 1.0.0

jonathan-codaio (approved):

PR #2002, opened 2022-05-19
check formula conn requirement before defaulting for sync tables

spencer-codaio:
fixing up some handling of `connectionRequirement` input in both formula and sync table def https://codaproject.slack.com/archives/C01835XEBSS/p1652313021825569

@jonathan-codaio @coda/packs


spencer-codaio:
> Hmm, say a bit more about what was breaking here? Introducing a `connectionRequirement` property on the top-level sync table def was meant to be a convenience but otherwise work the same. Are the autocomplete formulas not being rewritten properly or something?
> 
> Even if desirable, this is a breaking change, so this would likely cause some fallout. I think we'd rather deprecate providing connectionRequirement on the getter.



> Hmm, say a bit more about what was breaking here? Introducing a `connectionRequirement` property on the top-level sync table def was meant to be a convenience but otherwise work the same. Are the autocomplete formulas not being rewritten properly or something?
> 
> Even if desirable, this is a breaking change, so this would likely cause some fallout. I think we'd rather deprecate providing connectionRequirement on the getter.

Oh hmm.. good call on breaking change. The bug here is that if you put a `connectionRequirement` in the formula parameter and not in the top-level sync table configuration, it will not get read properly. Instead it defaults to the "defaultAuthentication" one which is by default `Required`. I think it is especially problematic because our templates seed the `connectionRequirement` in the formula parameter rather than the sync table, so if you ever just change that to something that's not the default Required it doesn't get passed through

jonathan-codaio:
> > Hmm, say a bit more about what was breaking here? Introducing a `connectionRequirement` property on the top-level sync table def was meant to be a convenience but otherwise work the same. Are the autocomplete formulas not being rewritten properly or something?
> > Even if desirable, this is a breaking change, so this would likely cause some fallout. I think we'd rather deprecate providing connectionRequirement on the getter.
> 
> > Hmm, say a bit more about what was breaking here? Introducing a `connectionRequirement` property on the top-level sync table def was meant to be a convenience but otherwise work the same. Are the autocomplete formulas not being rewritten properly or something?
> > Even if desirable, this is a breaking change, so this would likely cause some fallout. I think we'd rather deprecate providing connectionRequirement on the getter.
> 
> Oh hmm.. good call on breaking change. The bug here is that if you put a `connectionRequirement` in the formula parameter and not in the top-level sync table configuration, it will not get read properly. Instead it defaults to the "defaultAuthentication" one which is by default `Required`. I think it is especially problematic because our templates seed the `connectionRequirement` in the formula parameter rather than the sync table, so if you ever just change that to something that's not the default Required it doesn't get passed through

Sorry what do you mean by "templates" here--our examples and slash menu snippets? We should definitely update anything like that.

spencer-codaio:
> Sorry what do you mean by "templates" here--our examples and slash menu snippets? We should definitely update anything like that.

yeah just looked and i think I was using an old example, but updating some leftover ones. By deprecate providing it on the getter, do you mean deprecate providing it with the formula when it's nested in a sync table or deprecate providing it on the top-level sync table?

jonathan-codaio:
> > Sorry what do you mean by "templates" here--our examples and slash menu snippets? We should definitely update anything like that.
> 
> yeah just looked and i think I was using an old example, but updating some leftover ones. By deprecate providing it on the getter, do you mean deprecate providing it with the formula when it's nested in a sync table or deprecate providing it on the top-level sync table?

Yeah the former. We should only support it on the top-level of a sync table def now, and should just stop letting you do it nested in the formula. As always, the blocker is usually that our first party packs use legacy syntax and just need to be updated, though maybe this happened at this point.

PR #2004, opened 2022-05-23
Allow multi-network CodaApiHeaderBearerToken with credential pinning

dweitzman-codaio:
There's server-side enforcement now that credentials aren't sent to domains that aren't declared to have auth.

That makes is safe to use `CodaApiHeaderBearerToken` in a multi-network pack as long as only coda domains are listed as receiving auth credentials.


testing/upload_validation.ts
```diff
@@ -1060,16 +1060,34 @@ function validateFormulas(schema: z.ZodObject<any>) {
...
-            (domain: string) => !['coda.io', 'codahosted.io', 'localhost', 'calc-grpc-proxy'].includes(domain),
-          )
-        ),
+      untypedMetadata => {
+        const metadata = untypedMetadata as PackVersionMetadata;
+        if (metadata.defaultAuthentication?.type !== AuthenticationType.CodaApiHeaderBearerToken) {
+          return true;
+        }
+
+        const codaDomains = ['coda.io', 'codahosted.io', 'localhost', 'calc-grpc-proxy'];
```
dweitzman-codaio, line 1069: Seems like maybe `codahosted.io` and `calc-grpc-proxy` should be removed from this list of coda domains, since they don't need auth tokens?...

jonathan-codaio: Yeah, I don't recall why they made it into this particular validation, @huayang-codaio do you recall?

jonathan-codaio (approved):

testing/upload_validation.ts
dweitzman-codaio: I'm not sure what `localhost` is here for either, but I removed `codahosted.io` and `calc-grpc-proxy` from the list at least because I don't see what either of those domains would do with a coda bearer token

testing/upload_validation.ts
```diff
@@ -1069,7 +1069,7 @@ function validateFormulas(schema: z.ZodObject<any>) {
           return true;
         }
 
-        const codaDomains = ['coda.io', 'codahosted.io', 'localhost', 'calc-grpc-proxy'];
```
jonathan-codaio: Ok I remember a bit more now. It's not really about the fact that `codahosted.io` needs to use an auth token, it's that we don't want to generate an error if you include that. If you use an `Image` or `File` param type, it'll come through as a url on codahosted.io that you can download. You're free to declare codahosted.io as a network domain alongside coda.io and it's not supposed to count as an error, those two together are just what defines "connecting to Coda".

Now, I think we made it so that you don't have to declare codahosted.io at all anymore so it's just part of the system, so these days you don't even need to declare it. But if you do it shouldn't be an error.

And then I think `localhost` and `calc-grpc-proxy` are declared alongside `coda.io` by the Coda Debug pack which I guess needs to use Coda API token auth, and I think this carveout was just so that pack doesn't get an error during upload. I'm fine however we want to handle the carveout but we should avoid breaking that pack.

dweitzman-codaio: Seems like removing `codahosted.io` should be ok here since only authenticated domains are being checked now. You could still declare a `codahosted.io` network as long as it's not an authenticated network

The coda debug pack is already updated to only use `coda.io` as an authenticated domain: https://github.com/coda/packs/blob/main/packs/coda/debug/manifest.ts

testing/upload_validation.ts
```diff
@@ -1063,16 +1063,34 @@ function validateFormulas(schema: z.ZodObject<any>) {
...
+        const hasNonCodaAuthDomain = authDomains.some((domain: string) => !codaDomains.includes(domain));
+
+        // A non-coda auth domain is always an issue.
+        return !hasNonCodaAuthDomain;
+      },
       {
-        message: 'CodaApiHeaderBearerToken can only be used for coda.io domains',
-        path: ['networkDomains'],
+        message:
+          'CodaApiHeaderBearerToken can only be used for coda.io domains. Restrict `defaultAuthentication.networkDomain` to coda.io or codahosted.io',
```
jonathan-codaio, line 1092: Does codahosted.io need to be in this message?

dweitzman-codaio: Fixed

jonathan-codaio (approved):

dweitzman-codaio:
Do we care that this will force anyone using coda bearer token and multiple domains to upgrade to the 0.12.0 SDK?

If we care about that, I could modify the PR to keep the old check for old SDK versions and only do this check on SDK releases higher than 0.12.0

dweitzman-codaio:
> Do we care that this will force anyone using coda bearer token and multiple domains to upgrade to the 0.12.0 SDK?
> 
> If we care about that, I could modify the PR to keep the old check for old SDK versions and only do this check on SDK releases higher than 0.12.0

My vote is that we probably don't care. I think all the impacted packs would be first party or written by Codans

jonathan-codaio:
I don't think we care. More carrots and sticks to upgrade to modern versions seems reasonable at this point 😄

testing/upload_validation.ts
huayang-codaio: I think we were sending creds to `codahosted.io` before the change to always allow the CDN domain. 

`calc-grpc-proxy` and `localhost` were used by the debug pack (which uses the coda api bearer token auth). I am not sure if it's still using these two domains though (maybe not as from some earlier changes). There's a separate check to ensure that these local domains are only used by this pack.

dweitzman-codaio: The debug pack is already updated so that `coda.io` is its only auth domain

PR #2005, opened 2022-05-24
Add SDK support for boolean Toggle rendering for cross-doc fidelity.

jonathan-codaio:
PTAL @patrick-codaio @coda/packs CC @adam-codaio @evanbrooks-codaio @coda/api 

Evan, we can do this part for you for status bars, will probably take a little bit of think for how best to represent those in the SDK.


jonathan-codaio:
@ekoleda-codaio too

testing/upload_validation.ts
```diff
@@ -543,6 +544,7 @@ const booleanPackFormulaSchema = zodCompleteObject<Omit<BooleanPackFormula<any>,
   schema: zodCompleteObject<BooleanSchema>({
     type: zodDiscriminant(ValueType.Boolean),
     description: z.string().optional(),
+    displayType: z.nativeEnum(BooleanDisplayType),
```
patrick-codaio, line 547: nativeEnum is optional by default? surprising

patrick-codaio (approved):

testing/upload_validation.ts
jonathan-codaio: Oops it's not, fixing!

ekoleda-codaio:
I would have expected this to be implemented as a `codaType` / `ValueHintType` like `Slider` and `Scale`. Was that approach considered?

jonathan-codaio:
> I would have expected this to be implemented as a `codaType` / `ValueHintType` like `Slider` and `Scale`. Was that approach considered?

Yeah, I did consider that. There's a slight inconsistency between this case and those cases. In those cases, if you don't provide a hint type at all, the value renders as a number. When you provide a hint type, you're saying, don't render as a raw number, render as a UI element, and oh by the way there's some extra schema properties that you can use to control how that UI element renders.

For booleans, there's no equivalent of "just render as a number". The existing behavior is to render as a UI element, namely a checkbox. So it's like a type hint was already applied, and all that's left are modifiers for how the UI element should be rendered.

It doesn't seem right that if we had 9 ways to render a boolean we'd have 9 ValueHintTypes--it seems like we'd have one ValueHintType and 9 `displayType`s. In this case, a hint of "render as a checkbox-like UI element" is already implied.

ekoleda-codaio:
While I see that there is a difference in the default behavior for booleans, that seems more like legacy than a fundamental difference. Checkbox/Toggle feels a lot like Slider/Scale IMO. I'd be fine with a world where Slider vs Scale is controlled by `displayType` instead of `codaType`, but if that ship has sailed than I think it's better to be consistent with this new option.

jonathan-codaio:
> While I see that there is a difference in the default behavior for booleans, that seems more like legacy than a fundamental difference. Checkbox/Toggle feels a lot like Slider/Scale IMO. I'd be fine with a world where Slider vs Scale is controlled by `displayType` instead of `codaType`, but if that ship has sailed than I think it's better to be consistent with this new option.

Ok, trying it as `ValueHintType.Toggle`. I suppose if a toggle eventually has sub-properties (say, a color), then it would be better to have a ToggleSchema where `codaType: ValueHintType.Toggle` is used as a discriminant like other schemas, then a schema using `displayType: BooleanDisplayType.Toggle` as a discriminant.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Added
 
 - Added rarely-needed OAuth options: `scopeParamName` and `nestedResponseKey`.
+- Added `displayType` to `BooleanSchema`, allowing boolean values to be rendered as toggles in tables.
```
ekoleda-codaio, line 10: I think this needs to be updated for the change to ValueHint.

schema.ts
```diff
@@ -200,12 +207,26 @@ interface BaseSchema {
   description?: string;
 }
 
+/**
+ * Enumeration of display types supported by schemas that use {@link ValueType.Boolean}.
+ *
+ * These affect how a boolean value is rendered in tables.
+ */
+export enum BooleanDisplayType {
```
ekoleda-codaio, line 215: Should this be deleted now?

ekoleda-codaio (changes requested):

schema.ts
jonathan-codaio: Whoops yep deleted.

CHANGELOG.md
jonathan-codaio: Done.

huayang-codaio:
wow. just found this impressive slack thread. thanks for digging into this.

jonathan-codaio:
I also looked into supporting `codaType: ValueHintType.Toggle` on formulas that return booleans like we do for strings and numbers:

https://github.com/coda/packs-sdk/commit/edbd6d64caf5981ea9f401dad42776f18cc0c0fd

But it doesn't work because I don't think we currently those kinds of schema hints on formulas/formats, only sync tables. We respect type hints and schemas that affect how structured values are parsed, e.g. a number getting parsed as a currency with a given currency code and precision, because that happens at the value level.

So since this doesn't currently work I was going to leave this code out for now, but I could also include it to future-proof things.

ekoleda-codaio (approved):

PR #2006, opened 2022-05-24
Use node-fetch for "coda execute"

dweitzman-codaio:
This is to keep the behavior of "coda execute" similar to the current behavior in production.

Test strategy:

Made a pack formula that queries httpbin.org/deflate. The current SDK returns compressed garbage from `coda execute`

With these changes, `coda execute` returns:

```
{
  deflated: true,
  headers: {
    Accept: '*/*',
    'Accept-Encoding': 'gzip,deflate',
    Host: 'httpbin.org',
    'User-Agent': 'Coda-Test-Server-Fetcher',
    'X-Amzn-Trace-Id': 'Root=1-628d2498-299a40d15e7de4f3152b0ed9'
  },
  method: 'GET',
  origin: '136.25.111.210'
}
```

That's a successful deflate, and it also shows the latest version of the "Accept"
header correctly as `*/*`.


jonathan-codaio (commented):
Nice!

Would add an entry to the CHANGELOG.

jonathan-codaio (approved):

PR #2008, opened 2022-05-25
Update getSchema documentation

ekoleda-codaio:
- Document that `getSchema` can now access parameter values.
- Now that bug/20975 is fixed, remove the `Array` schema wrapper around the schema.


docs/guides/basics/parameters/autocomplete.md
```diff
@@ -118,6 +118,11 @@ coda.makeParameter({
     },
     ```
 
+!!! bug "All parameters passed as string"
```
jonathan-codaio, line 121: After https://github.com/coda/coda/pull/72397 this should no longer be true 😄 , if you want to hold off on this one.

jonathan-codaio (approved):

PR #2011, opened 2022-05-26
Accept "disableAuthentication" on TemporaryBlobStore.storeUrl

dweitzman-codaio:
Packs may need to fetch images from unauthenticated domains


api_types.ts
```diff
@@ -599,7 +599,11 @@ export interface TemporaryBlobStorage {
    * If the `downloadFilename` parameter is specified, the data will be interpreted as a file (`attachment` content
    * disposition) that will be downloaded when accessed as the file name provided.
    */
-  storeUrl(url: string, opts?: {expiryMs?: number; downloadFilename?: string}): Promise<string>;
+  storeUrl(
+    url: string,
+    opts?: {expiryMs?: number; downloadFilename?: string},
+    fetchOpts?: Pick<FetchRequest, 'disableAuthentication'>,
```
dweitzman-codaio, line 605: Could potentially allow anything from `FetchRequest`, but for now `disableAuthentication` was all I needed.

It might be ideal if the first argument here was just a fetch request but it wouldn't be backwards compatible

If we imagine a future where the `fetch` replaces the `url` then it might make sense to name it `fetch` instead of `fetchOpts`

ekoleda-codaio:
API surface LGTM.

api_types.ts
jonathan-codaio: Great, yeah this is forward-compatible with adding more options, so seems fine to start here.

jonathan-codaio (approved):

jonathan-codaio:
Oops forgot have you update the CHANGELOG.

PR #2013, opened 2022-05-30
Typo ("schip" -> "chip")

loucadufault:


docs/reference/sdk/interfaces/ObjectSchemaDefinition.md
```diff
@@ -42,7 +42,7 @@ A hint for how Coda should interpret and render this object value.
 
 For example, an object can represent a person (user) in a Coda doc, with properties for the
 email address of the person and their name. Using `ValueHintType.Person` tells Coda to
-render such a value as an @-reference to that person, rather than a basic object schip.
+render such a value as an @-reference to that person, rather than a basic object chip.
```
jonathan-codaio, line 45: So this is actually an auto-generated file, to fix this you'd make the change here:

https://github.com/coda/packs-sdk/blob/main/schema.ts#L788

and then run `make build` to generate everything downstream (which assumes you've run `make bootstrap` first to install all dependencies).

If either of those commands fail, just let us know and we can make and build the fix for you.

ekoleda-codaio:
This PR seems to be stalled, so fixing in #2053 instead.

PR #2014, opened 2022-05-31
Update description of 'featuredProperties' to be more comprehensive.

nmuegge:
As discussed with @ekoleda-codaio in the [forum](https://community.coda.io/t/how-to-include-nested-properties-is-featuredproperties/31029/2), there is some misunderstanding about how featuredProperties can and can't be used and the workarounds existing for these cases. This additional description should clear up some of that - it would've for me 👍

### DIFF
--- a/schema.ts
+++ b/schema.ts
@@ -796,7 +796,7 @@ export interface ObjectSchemaDefinition<K extends string, L extends string> exte
    * columns are created for each of the featured properties. The user can easily add additional
    * columns for any other properties, as desired. All featured properties need to be top-level.
    * If you can't or don't want to change the received data format, consider changing the
-   * received object after fetching and before returning and assigning it to the scheme.
+   * received object after fetching and before returning and assigning it to the schema.
    *
    * This distinction exists for cases where a sync table may include dozens of properties,
    * which would create a very wide table that is difficult to use. Featuring properties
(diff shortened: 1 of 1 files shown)


jonathan-codaio (commented):
Thanks for this!

You'll need to generate the downstream documentation files by running `make build` and then committing those. As a one-time setup step (and then periodically if dependencies change) you'll need to run `make bootstrap` first.

But, we've never had people outside of Coda try this so you might encounter some errors in those commands. If you do, just shout back and we can build and commit the generated files for you.

schema.ts
```diff
@@ -794,7 +794,9 @@ export interface ObjectSchemaDefinition<K extends string, L extends string> exte
    * A list of property names from within {@link properties} for the "featured" properties
    * of this object, used in sync tables. When a sync table is first added to a document,
    * columns are created for each of the featured properties. The user can easily add additional
-   * columns for any other properties, as desired.
+   * columns for any other properties, as desired. All featured properties need to be top-level.
+   * If you can't or don't want to change the received data format, consider changing the
+   * received object after fetching and before returning and assigning it to the scheme.
```
ekoleda-codaio, line 799: "scheme" => "schema".

ekoleda-codaio (changes requested):
Thanks for the contribution!

schema.ts
nmuegge: Just updated it in commit d3cf1be4bd03e0a5ec1e4193676a862b114cf313 👍

nmuegge:
Awesome! Thanks!

PR #2015, opened 2022-05-31
Add changelog entry for new storeUrl "fetchOpts" parameter

dweitzman-codaio:
Left this off of https://github.com/coda/packs-sdk/pull/2011


huayang-codaio (approved):

jonathan-codaio (approved):
Forgot to think of this when you first added this, but technically `temporaryBlobStorage.storeUrl()` only exists to help the cases where you _need_ authentication. If you don't need authentication, you should just return the url you already have, and we'll ingest it outside of the pack.

So stepping back, do we actually need this functionality, or can the pack in question here just return the unauthenticated url as-is?

PR #2038, opened 2022-06-06
Add API samples

ekoleda-codaio:
This PR creates a new code sample category focused on the APIs being connected to. Some are repurposed from the existing authentication samples, but many new ones were added, pulled from the list of most requested Packs. Adding this in preparation for the Packathon.


documentation/examples/apis/asana.ts
```diff
@@ -0,0 +1,26 @@
...
+    let response = await context.fetcher.fetch({
+      method: "GET",
+      url: "https://app.asana.com/api/1.0/users/me",
+    });
+    let user = response.body.data;
+    return user.name;
+  },
+});
+
+// Allow the pack to make requests to the Asana.
```
jonathan-codaio, line 25: Nit, "to Asana."

jonathan-codaio (approved):
Wow, what a great resource.

documentation/examples/apis/asana.ts
ekoleda-codaio: Done

PR #2043, opened 2022-06-07
Require identityName on all sync tables

patrick-codaio:
This is targeted to be released in SDK release 1.0.


jonathan-codaio (approved):

testing/upload_validation.ts
```diff
@@ -932,8 +932,7 @@ const baseSyncTableSchema = {
   getter: syncFormulaSchema,
   entityName: z.string().optional(),
   defaultAddDynamicColumns: z.boolean().optional(),
-  // TODO(patrick): After migration period, this will become required.
-  identityName: z.string().optional(),
+  identityName: z.string(),
```
dweitzman, line 935: Should this change be conditioned on SDK version? If not, it's a breaking change even for people who haven't upgraded their SDK.

patrick-codaio: I guess I thought we were ok with that. @jonathan-codaio did I misunderstand the plan?

jonathan-codaio: Ah shoot, I missed this. I was just thinking about this taking effect within the SDK, which would have meant only affecting subsequent releases starting with 1.0.0.

But because this same logic runs within the server on uploads for every SDK version, as David points out it will indeed break for every SDK version. So we need to fix this.

patrick-codaio: Ok, I'll work on that today.

PR #2047, opened 2022-06-08
Add samples index page

ekoleda-codaio:
Adds an index page for our samples, making it easier to link to all of the sample instead of a specific category. Uses a Python  macro to automatically generate the table from the navigation, so it remains in sync. Added a description and optional icon for each sample page to improve the UX.

Staged: https://coda-packs-sdk--2047.com.readthedocs.build/en/2047/samples/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 dist/** linguist-generated
 docs/embedded-snippets/** linguist-generated
 docs/reference/sdk/** linguist-generated
-docs/samples/** linguist-generated
+docs/samples/*/** linguist-generated
 documentation/generated/** linguist-generated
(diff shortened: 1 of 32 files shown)


jonathan-codaio (commented):
Nice. Where is the Python macro specified? Couldn't quite see in the diff where this is specified.

ekoleda-codaio:
Whoops, GitHub thought it was an auto-generated file and collapsed it by default in the diff. The file in question [is here](https://github.com/coda/packs-sdk/pull/2047/files#diff-f32a7d2e6a232a50369df8c8ba3ffd5fede6afe02df88ea5ece244ac4c0e6718), and I've adjusted the collapsing so it's visible now.

jonathan-codaio (approved):

PR #2048, opened 2022-06-09
Fix NPM update instructions and adjust nav

ekoleda-codaio:
`npm update @codahq/packs-sdk` doesn't bump the major version (or minor version for pre-1.0.0 releases), which means it will never work for upgrading during a breaking change. Changed the instructions to instead recommend using `npm install` at the target version in the migration page.

While I was there, noticed how it was hard to tell visually that the "Migration" header in the nav was actually a page with content. Disabled the feature that folded index pages into headers, and renamed those index pages "Overview" in the nav.

Before / After:
Screen Shot 2022-06-09 at 10 24 34 AMScreen Shot 2022-06-09 at 10 24 39 AM


jonathan-codaio (approved):
I guess the issue is that `npm update` abides by whatever your semver constraint is in your package.json? So I guess what it does depends on how your semver got populated, does `npm install --save` update package.json with a highly constrained semver?

https://docs.npmjs.com/cli/v8/commands/npm-update

Separately, and we've gone back and forth on this, the `packs-examples` repo describes using `yarn` which is inconsistent with other directions we give, so we should unify those.

ekoleda-codaio:
> I guess the issue is that `npm update` abides by whatever your semver constraint is in your package.json?

Yes, and that there is no semver way to say "update to the latest version, even if it's breaking." `npm install` adds a caret dependency, which allows for minor version changes but not major ones. Pre-1.0 though, the minor version is considered breaking, so `^0.10.0` means `0.10.*`.

Good catch on `yarn`, let me think about what to do and send a PR.

PR #2055, opened 2022-06-15
Convenience method for action return schemas

patrick-codaio:


schema.ts
```diff
@@ -1256,3 +1256,15 @@ export function makeReferenceSchemaFromObjectSchema(
...
+
+/**
+ * Convenience for defining the result schema for an action. The identity enables Coda to
+ * update the corresponding sync table row, if it exists.
+ * You could add the identity directly, but that would make the schema less re-usable.
+ */
+export function withIdentity(schema: GenericObjectSchema, identityName: string): GenericObjectSchema {
+  return makeObjectSchema({
+    ...schema,
+    identity: {name: ensureExists(identityName)},
```
jonathan-codaio, line 1268: Do you want ensureNonEmptyString() here? As-is I think this is a no-op.

schema.ts
```diff
@@ -1256,3 +1256,15 @@ export function makeReferenceSchemaFromObjectSchema(
...
 }
+
+/**
+ * Convenience for defining the result schema for an action. The identity enables Coda to
+ * update the corresponding sync table row, if it exists.
+ * You could add the identity directly, but that would make the schema less re-usable.
+ */
+export function withIdentity(schema: GenericObjectSchema, identityName: string): GenericObjectSchema {
+  return makeObjectSchema({
+    ...schema,
```
jonathan-codaio, line 1267: Do we want to deep copy the schema, in case the maker mutates it?

patrick-codaio: Oh, duh. Thanks! Done.

schema.ts
patrick-codaio: Yup, done!

jonathan-codaio (approved):
Would consider adding a CHANGELOG entry too.

PR #2056, opened 2022-06-15
Allow object schemas to track a pack ID without an identity

patrick-codaio:
Here's how I plan to use it in the `coda` repo - https://github.com/coda/coda/pull/73515


schema.ts
```diff
@@ -837,11 +837,17 @@ export type ObjectSchemaDefinitionType<
...
+// worry about identity when things outside of a sync table should interact with it, like references or
+// action return values.
 /** @hidden */
 export interface ObjectSchema<K extends string, L extends string> extends ObjectSchemaDefinition<K, L> {
   identity?: Identity;
+
+  /**
+   * Pack makers should never need to interact with this, it's just present for Coda's internal plumbing.
+   */
+  __packId?: number;
```
spencer-codaio, line 852: is it intentional that you put two `_`s for this

spencer-codaio (approved):
overall lgtm! but would want @jonathan-codaio to take a look

schema.ts
patrick-codaio: Yeah, but I could be convinced to change it if folks feel strongly.

testing/upload_validation.ts
```diff
@@ -832,6 +832,7 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
     attribution: attributionSchema,
     properties: z.record(objectPropertyUnionSchema),
     includeUnknownProperties: z.boolean().optional(),
+    __packId: z.number().optional(),
```
jonathan-codaio, line 835: Is there a way to validate that this is NOT provided? I think that should work--we don't want anybody ever actually specifying this in a pack definition, it will only ever be injected by us after they upload and we validate the schema.

schema.ts
```diff
@@ -839,11 +839,17 @@ export type ObjectSchemaDefinitionType<
 > = ObjectSchemaType<T>;
 
 // This is basically the same as an ObjectSchemaDefinition but includes an Identity that may be injected
-// at upload time (for static schemas) or at runtime (for dynamic schemas). This is the type that is
-// used on the `coda` side for implementations, but should not need to be used by pack makers.
```
jonathan-codaio, line 843: Should we keep the comment line about this being used only in the `coda` implementation and that this isn't a type relevant to pack makers?

schema.ts
spencer-codaio: i just haven't seen that before haha is that a practice?

schema.ts
patrick-codaio: Reverted! I got confused about the 2 different `identity` fields before.

schema.ts
patrick-codaio: A bit, yeah. Like, a programmer might make code that uses a single underscore, so if you're writing code for a programmer to use, you add another layer to avoid potential collisions.

If it's good enough for javascript, it's good enough for me:
Screen Shot 2022-06-15 at 1 08 50 PM

For example, we used it when customizing our json serialization: `var HACK_UNDEFINED_JSON_VALUE = '__CODA_UNDEFINED__';`

testing/upload_validation.ts
jonathan-codaio: Ok. We may want to throw on the `coda` side if we detect this just to be safe--we'll be overwriting it, so maybe we don't have to throw, but might be best to warn the maker and just ensure it's not being abused.

TS won't allow somebody to provide this field in non-legacy functions, but if a maker isn't using TS or abusing TS they can still upload whatever they want, and indeed we got a 500 error in the diff checker that seems to have resulted from somebody bypassing TS and us not accounting for that.

jonathan-codaio (approved):

testing/upload_validation.ts
patrick-codaio: Awesome, I'll add that to https://github.com/coda/coda/pull/73515

PR #2063, opened 2022-06-21
Update typescript to 4.7; update eslint

oleg-codaio:
See https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/

PTAL @coda/packs

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "sinon": "14.0.0",
     "source-map": "^0.8.0-beta.0",
     "stacktrace-parser": "^0.1.10",
+    "typescript": "^4.7.4",
     "uglify-js": "^3.15.5",
     "url-parse": "^1.5.10",
     "uuid": "^8.3.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -44,7 +44,6 @@
     "sinon": "14.0.0",
     "source-map": "^0.8.0-beta.0",
     "stacktrace-parser": "^0.1.10",
-    "typescript": "^4.6.4",
```
jonathan-codaio, line 47: I'd be a little hesitant about moving this to devDependencies. It seems innocent and correct, but many of the development-related packages here are in dependencies because the SDK itself does development and compilation. I can't recall specifically if `typescript` is one of the ones that has to remain here, but my intuition is that this is going to break something downstream, so I'd recommend keeping it where it is.

oleg-codaio: Fair enough, will move back.

jonathan-codaio (approved):
Thanks!

PR #2065, opened 2022-06-21
Slightly improve error message for references when "displayProperty" has a typo

dweitzman-codaio:
This is a fix for o/bug/23491

Could potentially make the error message even more accurate by saying the field doesn't exist. I'm open
to opinions on whether that's worth it.

Old error message:

```
TypeError: Cannot read property 'required' of undefined
    at checkSchemaPropertyIsRequired (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/schema.js:497:53)
    at validateObjectSchema (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/schema.js:479:9)
    at makeObjectSchema2 (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/schema.js:464:5)
    at Object.makeReferenceSchemaFromObjectSchema2 (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/schema.js:559:12)
    at Object.<anonymous> (../../../../../../private/var/folders/Users/DavidWeitzman/src/my-test-pack/schemas.ts:27:32)

    <more lines>
```

New error message:

```
Error: displayProperty set to undefined field "Name"
    at assertCondition (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/helpers/ensure.js:82:15)
    at checkSchemaPropertyIsRequired (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/schema.js:497:18)
    at validateObjectSchema (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/schema.js:479:9)
    at makeObjectSchema2 (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/schema.js:464:5)
    at Object.makeReferenceSchemaFromObjectSchema2 (../../../../../../private/var/folders/Users/DavidWeitzman/src/packs-sdk/dist/schema.js:560:12)
    at Object.<anonymous> (../../../../../../private/var/folders/Users/DavidWeitzman/src/my-test-pack/schemas.ts:27:32)
```


schema.ts
```diff
@@ -1186,7 +1186,7 @@ function checkSchemaPropertyIsRequired<K extends string, L extends string, T ext
 ) {
   const {properties, codaType} = schema;
   assertCondition(
-    properties[field].required,
+    properties[field]?.required,
```
jonathan-codaio, line 1189: Should we add a new check right above it that `properties[field]` is actually defined? Seems easy enough, to be able to accurate stately what the issue is in that case.

dweitzman-codaio: Updated the error to `Error: displayProperty set to undefined field "Name"`

One unfortunate thing is that it'll say `displayProperty` or `idProperty` even if the user was using the deprecated `primary` or `id` parameters, but my first attempt to distinguish between those cases didn't work and I think it's probably not worth spending more time on. Eventually `id` and `primary` will be unsupported anyway

jonathan-codaio: Yeah I don't think it's worth spending more time on. Onward!

PR #2067, opened 2022-06-21
Make identityName optional for pre-1.0 packs

patrick-codaio:
Context: https://github.com/coda/packs-sdk/pull/2043/files#r900703433

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -69,6 +69,7 @@ import {ValueType} from '../schema';
 import type {VariousAuthentication} from '../types';
 import type {VariousSupportedAuthenticationTypes} from '../types';
 import type {WebBasicAuthentication} from '../types';
+import {ZodParsedType} from 'zod';
 import {assertCondition} from '../helpers/ensure';
 import {isArray} from '../schema';
 import {isDefined} from '../helpers/object_utils';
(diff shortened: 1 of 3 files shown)


testing/upload_validation.ts
```diff
@@ -1367,6 +1369,24 @@ const packMetadataSchemaBySdkVersion: SchemaExtension[] = [
...
+
+          // A blank string will fail the existing `min(1)` requirement in the zod schema, so
+          // we only need to additionally make sure there is a value present.
+          if (typeof syncTable.identityName !== 'string') {
+            let receivedType: ZodParsedType = ZodParsedType.unknown;
+            if (syncTable.identityName === undefined) {
+              receivedType = ZodParsedType.undefined;
+            } else if (syncTable.identityName === null) {
+              receivedType = ZodParsedType.null;
+            }
```
jonathan-codaio, line 1381: Hmm, I guess it's fine to give you the extra info that you passed undefined or null, but I think `expected` and `received` are optional, so this could all maybe be simplified to this I think?

```typescript
if (!syncTable.identityName) {
  context.addIssue({
    code: z.ZodIssueCode.invalid_type,
    path: ['syncTables', i, 'identityName'],
    message: 'An identityName is required on all sync tables',
  });
}
```

jonathan-codaio: Ah I see, we've been using `ZodIssueCode.custom` for all of these similar custom validations, which seems fine here too if you want.

jonathan-codaio (approved):

PR #2069, opened 2022-06-22
throw error if number params are NaN or boolean params are not true or false when coerced

gary-codaio:
For https://coda-final.oslash.com/o/bug/23223 (request for validation when passing in incorrect data type)

Added these errors and also updated code to reflect that we don't allow object params. Here's also a more detailed rundown of why we only want to throw for numbers and booleans:
- the change to booleans could be left as is, but since we're not treating this as truthy as we only accept "true" as true, figured we should limit the possible inputs
- for a date, we could throw for something that fails, but I think it's possible in a Coda doc to pass in any string to be turned into a date and what would that would be passed around as would be something representing an invalid date which is what happens currently for this cli simulated execution
- change to an number is straightforward, we should just check if it's a number
- for an object, unreachable, this changes the code path to throw 
- for all other possible types, anything that can typed in as an input for the cli should work so should not check anything/nothing to check

tested locally:
image


testing/coercion.ts
```diff
@@ -38,21 +39,25 @@ function coerceParamValue(paramDef: ParamDef<any>, paramValue: any): any {
...
   }
-  return coerceParam(paramDef.type, paramValue);
+  return coerceParam(paramDef.type, paramDef.name, paramValue);
 }
 
-function coerceParam(type: Type, value: any): any {
+function coerceParam(type: ParamType, name: string, value: any): any {
   switch (type) {
     case Type.boolean:
+      if((value || '').toLowerCase() !== 'true' && (value || '').toLowerCase() !== 'false') {
```
jonathan-codaio, line 50: Nit, since we do it three times, let's just make a local var of `(value || '').toLowerCase()`

jonathan-codaio (approved):
Would just note the improvements in the CHANGELOG.

For dates, makes sense, but can you just confirm what happens if you try to pass a garbage string in a doc to a pack expecting a Date parameter? In the CLI, we're getting a Date object with value `Invalid date`, tbh I'm not sure what happens in a real doc.

gary-codaio:
@jonathan-codaio good point on checking the date in a doc. We actually do validate that the param is not an invalid date, so I'll add a check for that in the cli as well and also add tot he changelog 
image

PR #2070, opened 2022-06-22
Target Node 14 in esbuild config

ekoleda-codaio:
Ensures that the generated bundle is compatible with the Packs runtime. Should fix [this issue](https://community.coda.io/t/sync-tables-formulas-fail/31692?u=eric_koleda), where Packs fail with:

> Unexpected token '||='

### DIFF
--- a/testing/compile.ts
+++ b/testing/compile.ts
@@ -156,6 +156,7 @@ async function buildWithES({
     globalName: format === 'iife' ? 'module.exports' : undefined,
 
     platform: 'node',
+    target: 'node14',
 
     inject: getInjections(buildOptions),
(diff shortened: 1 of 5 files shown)


testing/compile.ts
```diff
@@ -156,6 +156,7 @@ async function buildWithES({
     globalName: format === 'iife' ? 'module.exports' : undefined,
 
     platform: 'node',
+    target: 'node14',
```
jonathan-codaio, line 159: Mind adding a comment that this is here because Lambda runs on Node 14?

jonathan-codaio (approved):
Would add a CHANGELOG entry too.

huayang-codaio:
thank you for fixing this.

PR #2071, opened 2022-06-22
add default allowed get domains

gary-codaio:
For https://coda-final.oslash.com/o/bug/23665

Verified this works with
```import * as coda from "@codahq/packs-sdk";

// Regular expression that matches Coda-hosted images.
const HostedImageUrlRegex = new RegExp("^https://(?:[^/]*\.)?codahosted.io/.*");

// Formula that calculates the file size of an image.
pack.addFormula({
  name: "FileSize",
  description: "Gets the file size of an image, in bytes.",
  parameters: [
    coda.makeParameter({
      type: coda.ParameterType.Image,
      name: "image",
      description:
        "The image to operate on. Not compatible with Image URL columns.",
    }),
  ],
  resultType: coda.ValueType.Number,
  execute: async function ([imageUrl], context) {
    // Throw an error if the image isn't Coda-hosted. Image URL columns can
    // contain images on any domain, but by default Packs can only access image
    // attachments hosted on codahosted.io.
    if (!imageUrl.match(HostedImageUrlRegex)) {
      throw new coda.UserVisibleError("Not compatible with Image URL columns.");
    }
    // Fetch the image content.
    let response = await context.fetcher.fetch({
      method: "GET",
      url: imageUrl,
      isBinaryResponse: true, // Required when fetching binary content.
    });
    // The binary content of the response is returned as a Node.js Buffer.
    // See: https://nodejs.org/api/buffer.html
    let buffer = response.body as Buffer;
    // Return the length, in bytes.
    return buffer.length;
  },
});
```
image

Not sure there's a good way to keep this in sync with the runtime config on the coda side

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 - Added validation that `networkDomain` does not include slashes since it's a domain, not a path.
 - Changed Pack compilation to explicitly target Node version 14, to ensure compatibility with the Packs runtime.
 - Added parameter type validation for `execute` command.
+- Added several implicitly-allowed domains including codahosted.io to the `execute` command. 
 
 ## [1.0.0] - 2022-06-16
 
--- a/testing/constants.ts
+++ b/testing/constants.ts
@@ -1,3 +1,12 @@
 export enum HttpStatusCode {
   Unauthorized = 401,
 }
+
+export const DEFAULT_ALLOWED_GET_DOMAINS = [
+  'coda-us-west-2-prod-blobs-upload.s3.us-west-2.amazonaws.com',
+  'coda-us-west-2-staging-blobs-upload.s3.us-west-2.amazonaws.com',
+  'coda-us-west-2-head-blobs-upload.s3.us-west-2.amazonaws.com',
+  'coda-us-west-2-adhoc-blobs-upload.s3.us-west-2.amazonaws.com',
+  'coda-us-west-2-dev-blobs-upload.s3.us-west-2.amazonaws.com',
+  'codahosted.io',
+]
--- a/testing/constants.ts
+++ b/testing/constants.ts
@@ -2,11 +2,8 @@ export enum HttpStatusCode {
   Unauthorized = 401,
 }
 
-export const DEFAULT_ALLOWED_GET_DOMAINS = [
-  'coda-us-west-2-prod-blobs-upload.s3.us-west-2.amazonaws.com',
-  'coda-us-west-2-staging-blobs-upload.s3.us-west-2.amazonaws.com',
-  'coda-us-west-2-head-blobs-upload.s3.us-west-2.amazonaws.com',
-  'coda-us-west-2-adhoc-blobs-upload.s3.us-west-2.amazonaws.com',
-  'coda-us-west-2-dev-blobs-upload.s3.us-west-2.amazonaws.com',
-  'codahosted.io',
+// Note that this should be kept in sync with the "PacksAllowedPublicDomains" runtime config 
+export const DEFAULT_ALLOWED_GET_DOMAINS_REGEXES = [
+  /^coda-us-west-2-\w+-blobs-upload\.s3\.us-west-2\.amazonaws\.com$/,
+  /^codahosted\.io$/,
 ]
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 - Added validation that `networkDomain` does not include slashes since it's a domain, not a path.
 - Changed Pack compilation to explicitly target Node version 14, to ensure compatibility with the Packs runtime.
 - Added parameter type validation for `execute` command.
+- Added several implicitly-allowed domains including codahosted.io to the fetcher used by the CLI.
```
dweitzman-codaio, line 10: Might be more clear to say "to the `execute` command" rather than "to the fetcher used by the CLI"

dweitzman-codaio (approved):

testing/constants.ts
```diff
@@ -1,3 +1,12 @@
 export enum HttpStatusCode {
   Unauthorized = 401,
 }
+
+export const DEFAULT_ALLOWED_GET_DOMAINS = [
```
spencer-codaio, line 5: can you leave a note here to keep it in sync with the runtime config we use to actually gate it?

testing/constants.ts
```diff
@@ -1,3 +1,12 @@
...
+  'coda-us-west-2-prod-blobs-upload.s3.us-west-2.amazonaws.com',
+  'coda-us-west-2-staging-blobs-upload.s3.us-west-2.amazonaws.com',
+  'coda-us-west-2-head-blobs-upload.s3.us-west-2.amazonaws.com',
+  'coda-us-west-2-adhoc-blobs-upload.s3.us-west-2.amazonaws.com',
+  'coda-us-west-2-dev-blobs-upload.s3.us-west-2.amazonaws.com',
```
spencer-codaio, line 10: nit: might be nice to make this an array of regexes to handle all of these with one line

jonathan-codaio: Yeah, can we maybe do a regex check here? I think I'd rather not advertise all of these domains in this public repo.

gary-codaio: Since our docs only say codahosted, should we only allow codahosted for the cli?
image

testing/constants.ts
```diff
@@ -1,3 +1,9 @@
 export enum HttpStatusCode {
   Unauthorized = 401,
 }
+
+// Note that this should be kept in sync with the "PacksAllowedPublicDomains" runtime config 
+export const DEFAULT_ALLOWED_GET_DOMAINS_REGEXES = [
+  /coda-us-west-2-[prod|staging|head|adhoc|dev]-blobs-upload.s3.us-west-2.amazonaws.com$/,
```
jonathan-codaio, line 7: Nit, let's just make the env part `\w+` so we don't advertise all the other environments.

testing/constants.ts
```diff
@@ -1,3 +1,9 @@
...
+  /codahosted.io$/,
```
jonathan-codaio, line 8: Nit, you'll want to escape all of the dots so they only match literal dots and not everything.

testing/constants.ts
gary-codaio: changed to regexes, let me know if we want to do this or just allow the "prod" env 

also tested that it works
image

testing/constants.ts
gary-codaio: ah see this, will do

gary-codaio:
Also @ekoleda-codaio somehow telepathically knew this was happening and updated the docs in https://github.com/coda/packs-sdk/pull/2072

testing/fetcher.ts
```diff
@@ -561,12 +563,13 @@ export class AuthenticatingFetcher implements Fetcher {
...
 
-  private _validateHost(url: string): void {
+  private _validateHost(url: string, method: FetchMethodType): void {
     const parsed = new URL(url);
     const host = parsed.host.toLowerCase();
     const allowedDomains = this._networkDomains || [];
     if (
-      !allowedDomains.map(domain => domain.toLowerCase()).some(domain => host === domain || host.endsWith(`.${domain}`))
+      !allowedDomains.map(domain => domain.toLowerCase()).some(domain => host === domain || host.endsWith(`.${domain}`)) &&
+      !(method === 'GET' ? DEFAULT_ALLOWED_GET_DOMAINS_REGEXES : []).some(domain => host.toLowerCase().match(domain))
```
jonathan-codaio, line 572: Optional nit, `.test()` returns a boolean so if you're only interested in whether a regex maches, that's preferable.

jonathan-codaio (approved):

PR #2074, opened 2022-06-22
Prevent pack makers from using reserved column names

patrick-codaio:
Fix for https://coda-final.oslash.com/o/bug/23804

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -744,6 +744,8 @@ function isValidObjectId(component: string): boolean {
   return Base64ObjectRegex.test(component);
 }
 
+const SystemColumnNames = ['id', 'value', 'synced', 'connection'];
+
 enum ExemptionType {
   IdentityName = 'IdentityName',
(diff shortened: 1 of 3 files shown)


testing/upload_validation.ts
```diff
@@ -932,7 +934,11 @@ const baseSyncTableSchema = {
   entityName: z.string().optional(),
   defaultAddDynamicColumns: z.boolean().optional(),
   // TODO(patrick): Make identityName non-optional after SDK v1.0.0 is required
-  identityName: z.string().min(1).optional(),
+  identityName: z
+    .string()
+    .min(1)
+    .optional()
+    .refine(val => !val || !SystemColumnNames.includes(val), 'Invalid identityName'),
```
jonathan-codaio, line 941: Maybe a slightly longer and friendlier error message, like "This property name is reserved for internal use by Coda and can't be used as an identity name, sorry!"

patrick-codaio: Done, thanks!

jonathan-codaio (approved):

PR #2076, opened 2022-06-27
Add MissingScopesError for pack makers to explicitly trigger progressive oauth flow

patrick-codaio:
This should be launched ASAP after the accompanying change in coda repo stops Coda defaulting to forcing progressive OAuth: https://github.com/coda/coda/pull/74035


jonathan-codaio (approved):
Does the simulated fetcher need to handle this error at all? Can't remember if we have missing scopes retry logic in the CLI fetcher.

ekoleda-codaio (approved):

patrick-codaio:
@jonathan-codaio The CLI `coda auth` command takes an optional `--extra_oauth_scopes='first second third'` param - it's up to the pack maker to use that to manually refresh the scopes. The formula execution does try to refresh the access token on 401, but that's the only special casing. I don't think this change needs an update to all that.

PR #2113, opened 2022-07-12
added new schema for images with outline and border properties

mphillips-codaio:
## Summary

> Added `ImageSchema`, for use with images. Allows packs to set two flags on a formula returning an image: if the image should be rendered with or without an outline, and whether to turn off the rounded corners. If the outline flag `ImageOutline` is not set on a schema, the default is `Solid`, and the image will be rendered with an outline. If the corners flag `ImageCornerStyle` is not set, the default is `Rounded`, and the image will be rendered with rounded corners.

> Created to give pack makers the option to remove special image styling (rounded corners and outline) created by Beautiful Images story, specifically the Latex pack, as the formulas rendered are sometimes cut off by the outline and rounding.

>  See example below of `Latex::Math('e^{\pi}')`
Screen Shot 2022-07-14 at 4 43 45 PM


schema.ts
```diff
@@ -568,6 +568,26 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
   timeFormat?: string;
 }
 
+/**
+ * A schema representing a return value or object property that is provided as a string,
+ * which Coda should interpret as an image.
+ */
+ export interface StringImageSchema extends BaseStringSchema<
+  ValueHintType.ImageReference | ValueHintType.ImageAttachment> {
+  /** Instructs Coda to render this value as a Image with proper beautifulImage properties. */
```
gleb-codaio, line 577: I wouldn't mention the beautifulImage here, as this documentation is public. Just say "Instructs Coda to render this value as an Image"?

schema.ts
```diff
@@ -568,6 +568,26 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
...
+  /**
+   * 
+   * Boolean specifying whether or not to add outline to rendered images. Defaults to true.
+   */
+  outline?: boolean;
+  /**
+   *
+   * Boolean specifying whether or not to add rounded corners to rendered images. Defaults to true.
+   */
+  round?: boolean;
```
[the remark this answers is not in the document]
mphillips-codaio: should I do one with 6, 8, or 10px for rounded corners?

schema.ts
```diff
@@ -568,6 +568,31 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
   timeFormat?: string;
 }
 
+/**
+ * Enabling options that can be used with a {@link ImageSchema}.
+ */
+ export enum ImageState {
```
gleb-codaio, line 574: It shouldn't be the same enum for the two properties, they just happen to have the same values right now. Make it two enums, named Outline and RoundedCorners.

schema.ts
```diff
@@ -568,6 +568,31 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
...
+/**
+ * A schema representing a return value or object property that is provided as a string,
+ * which Coda should interpret as an image.
+ */
+ export interface ImageSchema extends BaseStringSchema<
+  ValueHintType.ImageReference | ValueHintType.ImageAttachment> {
+  /** Instructs Coda to render this value as an Image. */
+  codaType: ValueHintType.ImageReference | ValueHintType.ImageAttachment;
+
+  // Boolean specifying whether or not to add outline to rendered images. Defaults to true.
```
gleb-codaio, line 590: It's not a boolean anymore.
Also by single-line I meant surround it with `/** text */` so that it is still a jsdoc comment, not a simple comment.

mphillips-codaio: ahh I see I was trying to figure out the proper comment styling

schema.ts
```diff
@@ -568,6 +568,40 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
...
+  Disabled = 'disabled',
+  /** Image rounded corners are applied. */
+  Default = 'default',
+}
+
+/**
+ * A schema representing a return value or object property that is provided as a string,
+ * which Coda should interpret as an image.
+ */
+ export interface ImageSchema extends BaseStringSchema<
```
jonathan-codaio, line 595: Now that you have an explicit schema for these hint types, I think you'll want to remove them from here: https://github.com/coda/packs-sdk/blob/main/schema.ts#L623

jonathan-codaio (commented):
@ekoleda-codaio should review this too.

Do you have a project writeup explaining the product goals here so we can make sure this adequately models what you're trying to achieve? And anything we add to the SDK we have to support forever, so additions here need to be carefully considered.

schema.ts
```diff
@@ -568,6 +568,40 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
...
+  ValueHintType.ImageReference | ValueHintType.ImageAttachment> {
+  /** Instructs Coda to render this value as an Image. */
+  codaType: ValueHintType.ImageReference | ValueHintType.ImageAttachment;
+  /** Outline type specifying whether or not to add outline to rendered images. Defaults to true. */
```
ekoleda-codaio, line 599: > Defaults to true.

This isn't a boolean field, so not quite accurate. Either change to an enum, or perhaps just leave unspecified, since it's likely to change in the product without being updated here.

Ditto below.

schema.ts
```diff
@@ -568,6 +568,40 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
...
 }
 
+/**
+ * State of outline on images that can be used with a {@link ImageSchema}.
+ */
+export enum Outline {
+  /** Image outline is disabled. */
+  Disabled = 'disabled',
+  /** Image outline is applied. */
+  Default = 'default',
```
ekoleda-codaio, line 578: I don't think `Default` is a good name for this option. What if in the future we change the default in the product to disabled? Did developers intend to use the default regardless of what it was, or intend to turn outlines on?

What we do for other enums is have only explicit options and interpret no option set as "use the default". For example, the [`CurrencySchema`](https://github.com/coda/packs-sdk/blob/e9c08785a83d0073f82787a33d1a8d2f94910c78/schema.ts#L349) has a a `format` field that accepts an enum value. The [`CurrencyFormat`](https://github.com/coda/packs-sdk/blob/e9c08785a83d0073f82787a33d1a8d2f94910c78/schema.ts#L314) enum only has concrete options. If the developer doesn't specify which format to use, then they get the product default.

I suggest changing this to `Solid` or something that concretely defines this option.

schema.ts
```diff
@@ -568,6 +568,40 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
...
+ * which Coda should interpret as an image.
+ */
+ export interface ImageSchema extends BaseStringSchema<
+  ValueHintType.ImageReference | ValueHintType.ImageAttachment> {
+  /** Instructs Coda to render this value as an Image. */
+  codaType: ValueHintType.ImageReference | ValueHintType.ImageAttachment;
+  /** Outline type specifying whether or not to add outline to rendered images. Defaults to true. */
+  outline?: Outline;
+  /** roundedCorners type specifying whether or not to add rounded corners to rendered images. Defaults to true. */
+  roundedCorners?: RoundedCorners;
```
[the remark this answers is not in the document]
gleb-codaio: Yep, exactly.

schema.ts
mphillips-codaio: > This feels a bit oddly specific. I might call this `ImageCornerStyle` and have values of `Default` and `Rounded`, which leaves room for more styles in the future.

Based on this, would it seem reasonable to set something like `Standard` as without the styling for both properties, then `Rounded` and `Outlined` as the other enum values for now?

schema.ts
mphillips-codaio: the imageProperty in the object schema requires that the property contains one of those hints (https://github.com/coda/packs-sdk/blob/main/schema.ts#L878) so a lot of errors were brought up around that. Do you have any suggests for how can I can resolve that?

jonathan-codaio: Oh, yeah it's because you have to implement a new validator for your schema. This should do it, you can copy this into this PR:

https://github.com/coda/packs-sdk/commit/86c3dcfdcd56b02ee8f6d2218fffafb15822d564#diff-8f115faeb93eef8da8c710afe7051c558275f28ab6015e1a6bd21f2f7a599f51

jonathan-codaio: Mind including these tests too, to make sure the new fields validate? https://github.com/coda/packs-sdk/commit/fc01a240e5a0f3c1a6ff94ba073cc4826801e776#diff-e51cad99b49dae3d3ff0b09db2a20be1dfbef9e67950e077b326088856860551

gleb-codaio:
> Do you have a project writeup explaining the product goals here so we can make sure this adequately models what you're trying to achieve? And anything we add to the SDK we have to support forever, so additions here need to be carefully considered.

I wrote up the context and what we want to achieve here: https://staging.coda.io/d/Canvas-IA-o-canvasia_doZuOLrJWJ_/Pack-images_suE9R#_luyBS

schema.ts
```diff
@@ -568,6 +568,40 @@ export interface StringDateTimeSchema extends BaseStringSchema<ValueHintType.Dat
...
+
+/**
+ * A schema representing a return value or object property that is provided as a string,
+ * which Coda should interpret as an image.
+ */
+ export interface ImageSchema extends BaseStringSchema<
+  ValueHintType.ImageReference | ValueHintType.ImageAttachment> {
+  /** Instructs Coda to render this value as an Image. */
+  codaType: ValueHintType.ImageReference | ValueHintType.ImageAttachment;
+  /** ImageOutline type specifying style of outline on images. */
```
jonathan-codaio, line 599: Can you document here what the default behavior is if left unspecified?

jonathan-codaio (commented):
Oh sorry, and one more thing,  after rebasing (since a new SDK version was just released this morning) can you document the new functionality in CHANGELOG.md by adding a `# Not yet released` section at the top and a bullet point describing the new fields? Just like https://github.com/coda/packs-sdk/commit/81ee4c21525b105c2ba873f1b49ecec5cfc06531#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

jonathan-codaio (commented):
LGTM

@ekoleda-codaio any final feedback?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## Not yet released
 
+- Added `ImageSchema`, for use with images. Allows packs to set two flags on a formula returning an image: if the image should be rendered with or without an outline, and whether to turn off the rounded corners. If the outline flag `ImageOutline` is not set on a schema, the default is `Solid`, and the image will be rendered with an outline. If the corners flag `ImageCornerStyle` is not set, the default is `Rounded`, and the image will be rendered with rounded corners.
+
+## Not yet released
```
ekoleda-codaio, line 9: I don't think we should have two "Not yet released headers". I think the stuff below was just released, probably just need to sync with main.

ekoleda-codaio:
One small comment about the changelog, but SDK change LGTM.

ekoleda-codaio:
LGTM on the SDK design and documentation, but leaving it to @jonathan-codaio to give the final PR approval.

jonathan-codaio (approved):
Nicely done!

PR #2114, opened 2022-07-12
Unit test for max memory used to marshal/unmarshal values for isolated-vm

dweitzman-codaio:
The current implementation of marshalValue() is extremely memory hungry.
This makes it impossible to return larger files like multi-megabyte images
from the fetcher without exceeding memory limits.

This unit test can be useful as we try to slim down the marshal/unmarshal
memory usage.


test/thunk_test.ts
```diff
@@ -47,4 +47,41 @@ describe('Thunk', () => {
 
     assert.isTrue(result !== 'Object');
   });
+
+  it('marshaling buffers uses too much memory', async () => {
```
huayang-codaio, line 51: maybe add a note that this should be fixed.

huayang-codaio (approved):

test/thunk_test.ts
dweitzman-codaio: Done

PR #2115, opened 2022-07-13
Support MkDocs Material Insiders edition

ekoleda-codaio:
As a sponsor of the MkDocs Material project, we get access to their [Insiders edition](https://squidfunk.github.io/mkdocs-material/insiders/), which provides early access to new features. This edition is distributed via a private GitHub repo, with sponsors able to get access through designated GitHub accounts. Pipenv / pip can install from this repo, provided an API token is included in the URL.

The Makefile was adjusted so that if the GitHub token is found in an environment variable:
- It installs the Insiders edition of the library.
- Uses an Insiders version of the mkdocs.yml file with additional Insiders-only plugins enabled.

Many of these patterns come from the [Installation instructions](https://squidfunk.github.io/mkdocs-material/insiders/getting-started/) for MkDocs Material Insiders.

This change should be a no-op for other engineers and the Circle CI pipeline. A follow-on effort will involve determining which account to use to generate the GitHub token and then setting the appropriate environment variable in CircleCI.

I chose this "upgrade" approach because I wanted to reduce friction for internal engineers (they don't need to setup the environment variable) and make it possible for external contributors to build and preview the docs locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,20 +189,29 @@ build-mkdocs:
 # For example, if you wanted to force upload (to skip the existing directory check), you can run
 # make publish-docs-<env> FLAGS=--forceUpload
 .PHONY: publish-docs-adhoc
-publish-docs-adhoc:
+publish-docs-adhoc: validate-mkdocs-insiders
 	(cd ${ROOTDIR}; ./node_modules/.bin/ts-node documentation/scripts/documentation_publisher.ts push adhoc ${FLAGS})
 
 .PHONY: publish-docs-head
-publish-docs-head:
+publish-docs-head: validate-mkdocs-insiders
 	(cd ${ROOTDIR}; ./node_modules/.bin/ts-node documentation/scripts/documentation_publisher.ts push head ${FLAGS})
 
 .PHONY: publish-docs-staging
-publish-docs-staging:
+publish-docs-staging: validate-mkdocs-insiders
 	(cd ${ROOTDIR}; ./node_modules/.bin/ts-node documentation/scripts/documentation_publisher.ts push staging ${FLAGS})
 
 .PHONY: publish-docs-prod
-publish-docs-prod:
+publish-docs-prod: validate-mkdocs-insiders
 	(cd ${ROOTDIR}; ./node_modules/.bin/ts-node documentation/scripts/documentation_publisher.ts push prod ${FLAGS})
+
+# Verify that the MkDocs Material Insiders GitHub token is present in the environment.
+# Used to verify that the published docs are using the desired version of the library.
+.PHONY: validate-mkdocs-insiders
+validate-mkdocs-insiders:
+	if [[ -z "${MKDOCS_INSIDERS_TOKEN}" ]]; then \
+		echo "Not using the MkDocs Material Insiders library."; \
+		exit 1; \
+	fi
 
 ###############################################################################
 


mkdocs.yml
```diff
@@ -45,25 +45,27 @@ markdown_extensions:
...
-          # 2022-01-14: Reorganize guides and tutorials.
-          "guides/advanced/autocomplete.md": "guides/basics/parameters/autocomplete.md"
-          "guides/advanced/dynamic-sync-tables.md": "guides/blocks/sync-tables/dynamic.md"
-          "guides/advanced/cli.md": "guides/development/cli.md"
-          "guides/advanced/libraries.md": "guides/development/libraries.md"
-          "guides/advanced/testing.md": "guides/development/testing.md"
-          "guides/troubleshooting.md": "guides/development/troubleshooting.md"
-          "get-started/cli.md": "tutorials/get-started/cli.md"
-          "get-started/web.md": "tutorials/get-started/web.md"
+  # Client-side search.
```
ekoleda-codaio, line 48: FYI: The plugins content was changed from a list to a hash, in order to support the inheritance correctly, as mentioned at the bottom of [this page](https://squidfunk.github.io/mkdocs-material/insiders/getting-started/#built-in-plugins).

Makefile
```diff
@@ -194,17 +204,6 @@ publish-docs-staging:
 publish-docs-prod:
 	(cd ${ROOTDIR}; ./node_modules/.bin/ts-node documentation/scripts/documentation_publisher.ts push prod ${FLAGS})
 
-.PHONY: publish-docs-gh-pages
```
ekoleda-codaio, line 197: FYI: Deleted this old GitHub Pages publishing target, since we stopped using it long ago.

spencer-codaio (approved):

jonathan-codaio (commented):
Hmm can you say a bit more? If I have a token and you don't or the CI doesn't, or the CI has a token and I don't, doesn't that mean we'll get different build outputs? How do we get consistent reproducible builds if the build step is environment-dependent?

ekoleda-codaio:
> I don't see how it's possible to have builds that differ--in order to merge you need a developer's build to match what the CI does.

Running `make build` doesn't actually build the docs, and the resulting HTML is not checked in. So the fact that it differs based on environment variables wouldn't trip the check we have for un-checked-in diffs.

>What kind of features are in the Insiders edition right now?

All new features start in Insiders, and only graduate to the open source version when the project hits a certain funding goal. This is the [current list](https://squidfunk.github.io/mkdocs-material/insiders/#whats-in-for-me). 

> Are any of particular interest?

The one I am currently the most interested in is [Social Cards](https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/). It will images from the metadata of the page, which show up when you share the URL on social. For example:

image

Our docs currently use a static placeholder image, which isn't attractive or informative:

image

jonathan-codaio:
Ah, ok, I didn't realize the html wasn't getting checked in. So it's only `make publish-docs-{env}` that does this?

I'm more ok with this then. I would error if a token is not found though, we shouldn't be able to accidentally publish docs that are a regression. We should just be committing to using Insiders.

ekoleda-codaio:
> So it's only `make publish-docs-{env}` that does this?

Ya, exactly.

> I would error if a token is not found though, we shouldn't be able to accidentally publish docs that are a regression.

Good call, done.

> I'm more ok with this then.

It's certainly a tradeoff either way. Happy to defer to your judgement on how to balance these competing interests.

Makefile
```diff
@@ -13,6 +13,12 @@ PIPENV := PYTHONPATH=${ROOTDIR} PIPENV_IGNORE_VIRTUALENVS=1 pipenv
 
 DOC_GIT_REVISION ?= main
 
+MKDOCS_INSIDERS_VERSION="8.3.9-insiders-4.20.0"
+ifeq "$(MKDOCS_INSIDERS_TOKEN)" ""
+    MKDOCS_CONFIG_FILE:=mkdocs.yml
+else
+    MKDOCS_CONFIG_FILE:=mkdocs.insiders.yml
```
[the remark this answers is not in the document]
ekoleda-codaio: In addition to publishing the docs there are two other uses of mkdocs:

1. `make build-mkdocs`, which is run as part of `make docs` (and therefore `make build`) to verify that the docs are still build-able. The results aren't checked in.
1. `make view-docs`, to preview the docs locally.

My thought was that neither of these need to operate with full fidelity, so it's OK if they use the Open Source version, but that would require having two `mkdocs.yml` config files.

Remove non-Insiders support from those as well would mean that everyone needs a token to build or preview.

ekoleda-codaio:
I was always a little queasy about my approach in this PR, and your concerns certainly amplify that. I think I'll shelve this for now since the need it's pressing.

PR #2116, opened 2022-07-13
More memory-efficient marshal/unmarshal for isolated-vm

dweitzman-codaio:
The previous implementation relied on JSON.stringify() which is crazily memory-inefficient for multi-megabyte objects like a fetched image.

This new implementation passes data into isolated-vm using javascript's structuredClone() algorithm instead.

See the comment at the top of runtime/common/marshalling/index.ts for an overview of how the transformation to a clone-safe object works.

Other notes:
- The unit tests for marshal/unmarshal now pass the data through isolated-vm to make sure the output format actually works.
- More isolated-vm calls need to set "copy: true" now, since the marshal/unmarshal output isn't just a primitive string anymore
- Set up a memory regression test in thunk_test.ts
- Some types that didn't work before work now, like RegExp
- Some types that previously silently did nothing raise errors now, like a Promise.


dweitzman-codaio:
This doesn't work yet on a real pack. I'll take this out of draft mode when it does

huayang-codaio (commented):
this looks much easier

runtime/common/marshaling/index.ts
```diff
@@ -1,106 +1,154 @@
...
   }
 
-  if (Array.isArray(val)) {
-    return val.map(item => processValue(item, depth + 1));
+  if (val instanceof Buffer || global.Buffer?.isBuffer(val)) {
+    // Theoretically it should be possible to pass an array buffer
+    // through structured copy with some transfer options, but it's
+    // simpler to just encode it as a string.
+    postTransforms.push({
+      type: 'Buffer',
```
chris-codaio, line 63: Same here

runtime/common/marshaling/index.ts
```diff
@@ -1,106 +1,154 @@
...
-  const serializedValue = serialize(val);
-
-  if (typeof serializedValue === 'object') {
-    let objectValue = serializedValue;
-    if ('toJSON' in serializedValue && typeof serializedValue.toJSON === 'function') {
-      objectValue = serializedValue.toJSON();
+  if (Array.isArray(val)) {
+    const maybeModifiedArray: any[] = [];
+    let someItemHadModifications = false;
+    val.forEach((item, index) => {
```
chris-codaio, line 72: Super nit (feel free to ignore): prefer `for` loop over `forEach` in modern js due to performance difference.

runtime/common/marshaling/index.ts
```diff
@@ -1,106 +1,154 @@
...
+    postTransforms,
+    [MarshalingInjectedKeys.CodaMarshaler]: CodaMarshalerType.Object,
+  };
+  return result;
 }
 
-export function unmarshalValue(marshaledValue: string | undefined): any {
-  if (marshaledValue === undefined) {
-    return marshaledValue;
+function applyTransform(input: any, path: string[], fn: (encoded: any) => any): any {
```
huayang-codaio, line 144: I think this approach works. but I slightly lean towards an in-place marshaling where you can easily apply the same traverse function with a marshaler / unmarshaler function. 

but I do see the postTransforms could be more efficient in most cases.

runtime/common/marshaling/index.ts
```diff
@@ -1,106 +1,154 @@
...
+  if (!val) {
+    return {val, hasModifications: false};
   }
 
-  if (val === null) {
-    return val;
+  const maybeError = marshalError(val);
+  if (maybeError) {
+    postTransforms.push({
+      type: 'Error',
```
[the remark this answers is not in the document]
dweitzman-codaio: Done

runtime/common/marshaling/index.ts
dweitzman-codaio: Done

runtime/common/marshaling/index.ts
dweitzman-codaio: Done

runtime/common/marshaling/index.ts
```diff
@@ -1,106 +1,154 @@
...
-  // Instead of passing a replacer to `JSON.stringify`, we chose to preprocess the value before
-  // passing it to `JSON.stringify`. The reason is that `JSON.stringify` may call the object toJSON
-  // method before calling the replacer. In many cases, that means the replacer can't tell if the
-  // input has been processed by toJSON or not. For example, Date.prototype.toJSON simply returns
-  // a date string. The replacer can't identify if the initial value is a Date or a date string.
-  //
-  // processValue is trying to mimic the object processing of JSON but the behavior may not be
-  // identical. It will only serve the purpose of our internal marshaling use case.
-  return JSON.stringify(processValue(val));
+export function marshalValue(val: any): any {
```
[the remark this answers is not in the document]
dweitzman-codaio: Done

huayang-codaio (commented):
lgtm!

runtime/common/marshaling/index.ts
huayang-codaio: let me add another note on this comment. I think the `postTransforms` is more efficient here while I was considering complicated cases like nested marshaling (e.g. Error has a property of Buffer). When the non-copyable types get more complicated, in-place marshaling would be more adaptive while you need to carefully control the `postTransforms` ordering and paths (in case unmarshaling changes the hierarchy of a nested object).

dweitzman-codaio: I don't think the old implementation recursively supported buffers within errors either

dweitzman-codaio: Hypothetically if we want to supported buffers in errors, we could recursively call marsalValue and the post transforms would be embedded deep within the output instead of merged into the top level

huayang-codaio: I just verified that buffer in error works in the old form. It's a real bug we found before and had to apply a fix. would be a bit surprised if it no longer works. 

You may be able to make it work in this PR by not directly returning when it detects an error, and keep serializing its sub fields. I was just worried if the path in `postTransforms` is going to be precise.

dweitzman-codaio: I think I can actually fix all the issues with errors, but still trying to get the code working. Looks like `v8.serialize()` and `v8.deserialize()` exist and are compatible with structure clone, so one thing I plan to do is encode errors as `new Error(Buffer.from(v8.serialize(marshalValue(...)), "base64"))`

Also planning to make errors recursively marshal any spare arguments so errors-within-errors work properly

huayang-codaio: I am not sure if I follow what you plan to do. 

on the other hand, it looks like v8.serialize/deserialize doesn't seem to work for custom errors. 

```
> class NewError extends Error {}
undefined
> v8.deserialize(v8.serialize(new NewError()))
Error
    at REPL14:1:29
    at Script.runInThisContext (node:vm:129:12)
    at REPLServer.defaultEval (node:repl:562:29)
    at bound (node:domain:421:15)
    at REPLServer.runBound [as eval] (node:domain:432:12)
    at REPLServer.onLine (node:repl:889:10)
    at REPLServer.emit (node:events:402:35)
    at REPLServer.emit (node:domain:475:12)
    at REPLServer.Interface._onLine (node:readline:487:10)
    at REPLServer.Interface._line (node:readline:864:8)
> v8.deserialize(v8.serialize(new NewError())) instanceof NewError
false
```

It looks like the same behavior of how isolated-vm passes error objects around. we need a special error marshaling to keep all non-standard error information.

dweitzman-codaio: `v8.serialize` isn't for simplifying the logic, it's for fixing the problem with `wrapError` since isolated-vm errors only seem to be able to bubble up a string and this new `marshalValue` implementation wasn't returning a string

If you look at the unit tests currently in place, they show that `wrapError`/`unwrapError` are losing type information. I should be able to fix that with `serialize`/`deserialize`

huayang-codaio: oh you are right. just noticed that.

dist/bundles/thunk_bundle.js
```diff
@@ -7,10 +7,17 @@ module.exports = (() => {
   var __getOwnPropNames = Object.getOwnPropertyNames;
   var __getProtoOf = Object.getPrototypeOf;
   var __hasOwnProp = Object.prototype.hasOwnProperty;
+  var __require = /* @__PURE__ */ ((x) => typeof require !== "undefined" ? require : typeof Proxy !== "undefined" ? new Proxy(x, {
+    get: (a, b) => (typeof require !== "undefined" ? require : a)[b]
+  }) : x)(function(x) {
+    if (typeof require !== "undefined")
+      return require.apply(this, arguments);
+    throw new Error('Dynamic require of "' + x + '" is not supported');
```
dweitzman-codaio, line 15: Hmm. I'm trying to figure if if there's a way I can use`v8.serialize` in SDK code that doesn't end up triggering some kind of `Dynamic require of ... is not supported` error

Makefile
```diff
@@ -100,6 +100,7 @@ compile-thunk:
 		--inject:${ROOTDIR}/testing/injections/buffer_shim.js \
 		--format=iife \
 		--global-name=module.exports \
+		--platform=node \
```
huayang-codaio: does this make a difference?

dist/bundles/thunk_bundle.js
huayang-codaio: is v8 not available in isolated-vm?

huayang-codaio: oh so isolated-vm doesn't have v8 in its runtime. interesting even if it mentions v8 thousands times in its readme..

```
> context.evalSync('v8.serialize')
Uncaught ReferenceError: v8 is not defined
    at <isolated-vm>:1:1
    at (<isolated-vm boundary>)
    at REPL8:1:9
```

huayang-codaio: it only has language primitives. I guess you don't want to compile a full v8 module into isolated-vm issue. maybe the only option is to inject a function that calls back to the native v8 module.

runtime/common/marshaling/index.ts
```diff
@@ -113,12 +184,12 @@ export function wrapError(err: Error): Error {
   //     'to actually fetch from the remote API.';
   // }
 
-  return new Error(marshalValue(err));
+  return new Error(v8.serialize(marshalValue(err)).toString('base64'));
```
huayang-codaio, line 187: in the worst case, just using JSON for this case. if an error object is 2MB, we are screwed. lol

dweitzman-codaio:
I've made all the tests pass using yet another layer of weird hackery. The main files of interest are `runtime/common/marshaling/serializer.ts` which flips between the in-lambda vs out-of-lambda serialize implementations and `runtime/bootstrap/index.ts` which has a new `injectSerializer` to help set up the isolated-vm environment

dweitzman-codaio:
Holy moly... I don't want to count my chickens before they hatch, but on a local demo it looks like this epic hack might actually be working! And as the unit tests demonstrate, types recursively nested within errors should work correctly.

The adhoc demo linked from https://github.com/coda/coda/pull/75049 is still updating with the new code

runtime/bootstrap/index.ts
```diff
@@ -80,7 +80,7 @@ export async function injectAsyncFunction(
        });
      };`,
     [stub],
-    {arguments: {reference: true}},
+    {arguments: {reference: true}, result: {copy: true}},
```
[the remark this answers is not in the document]
dweitzman-codaio: Actually I think these changes are unnecessary: the argument here is the injected function, and the closure doesn't actually return anything, it just defines a function. The `apply` call above does have `copy: true` set for both the argument and result, and I think that's where it matters.

runtime/common/marshaling/index.ts
```diff
@@ -1,106 +1,154 @@
...
-
-  return val;
-}
-
-function processValue(val: any, depth: number = 0): any {
+// pathPrefix can be temporarily modified, but needs to be restored to its original value
+// before returning.
+//
+// "hasModifications" is to avoid trying to copy objects that don't need to be copied.
+// Only objects containing a buffer or an error should need to be copied.
```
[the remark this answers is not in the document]
dweitzman-codaio: I added a comment below explaining why the copy is being avoided. Basically, we want to make sure we don't convert something like a `Date` into a object with the equivalent fields

An alternative implementation could be to have an allow list of types that we know structured copied can handle (dates, regexps, etc.) and immediately return those without trying to recurse. The benefit would be better performance but the downside would be that this code would gain more knowledge about how structured clone works and we'd need a long list of checks for supported types.

dweitzman-codaio:
Also happy to hop on a call whenever and talk through all the changes and the path of trial and error that led to the current state

huayang-codaio (commented):
with playing the branch, I have two observations. 
- it doesn't look like Buffer needs to be serialized now. I am not sure what happened. But after I commented out `if (val instanceof Buffer || global.Buffer?.isBuffer(val)) {` etc in marshaling/index.ts. it seems like Buffer can still be passed as arguments and results. I'll play more with it. 
- the injection approach doesn't seem to work for custom errors. with a further think I wonder if it is because we are creating a loop of serialization. the fact of passing `Error` to serialize (which goes across isolated-vm boundary) already peels off the custom error information from it.

dweitzman-codaio:
To record some discussion Huayang and I had offline: we'll probably retain both the old and new marshal implementations for now so calc service can send both of those formats as input to pack runtime code.

runtime/common/marshaling/marshal_errors.ts
```diff
@@ -37,7 +37,7 @@ function getErrorClassType(err: Error): ErrorClassType {
   return ErrorClassType.Other;
 }
 
-export function marshalError(err: any): object | undefined {
+export function marshalError(err: any, marshalFn: (val: any) => any): object | undefined {
```
huayang-codaio, line 40: I suppose `marshalValue` is the only possible value of `marshalFn`. Do we still need this parameter?

runtime/common/marshaling/marshal_errors.ts
```diff
@@ -76,7 +83,10 @@ function getErrorClass(errorClassType: ErrorClassType, name: string): ErrorConst
   return errorClasses.find(cls => cls.name === name) || Error;
 }
 
-export function unmarshalError(val: {[key: string]: any | undefined}): Error | undefined {
```
huayang-codaio, line 79: ditto

test/marshaling_test.ts
```diff
@@ -27,35 +47,65 @@ describe('Marshaling', () => {
...
     const testObjects: any = [
       [null, undefined, 0, false, NaN, Infinity, {undefined: 1, null: undefined}],
       {Array: [], Boolean: false, new: {_: null}, function: undefined, NaN: 1},
       [{undefined: [{false: [{true: [{null: 0}]}]}]}],
+      ['foo', {1: Buffer.from('bar')}, [[Buffer.from('baz')]]],
     ];
     testObjects.forEach((x: any) => assert.deepEqual(transform(x), x));
   });
 
   it('does not throw error for unhandled objects', () => {
```
huayang-codaio, line 94: description seems outdated.

huayang-codaio (approved):

runtime/common/marshaling/marshal_errors.ts
dweitzman-codaio: This was just a lazy way to get around circular dependencies, since marshalValue depends on marshalError.

I'll just move the contents of marshal_errors.js into index.js for simpler code

huayang-codaio: cool. thanks

test/marshaling_test.ts
dweitzman-codaio: Fixed

chris-codaio:
Did we have anyone review this from a security PoV?  Would be great to look over this for isolated-vm breakouts via object ref.

dweitzman-codaio:
> Did we have anyone review this from a security PoV? Would be great to look over this for isolated-vm breakouts via object ref.

Who might be a good person to review from a security perspective? Definitely happy to have more eyes on this.

dweitzman-codaio:
For anyone following along abc reviewing from a security perspective: as far as I understand, it's not possible to pass an actual object into or out of isolated-vm.

Setting "reference" means you pass a remote pointer to an object. In the case of a function you can invoke the function, but all arguments and results still need to be copied or transferable to cross the vm border

chris-codaio:
Perhaps @bala-codaio or @az-codaio would be good reviewers?  In particular, the "Security" section in the `isolated-vm` repo is a good foundation on what to look for - particularly around passing in references to `isolated-vm` objects or any v8/node object that can be used directly or indirectly to affect memory or timing.  https://github.com/laverdet/isolated-vm

az-codaio:
Chatted about this with @dweitzman-codaio. This sounds fine to me from a security perspective.

Some notes:

The new functions being injected [serialize, deserialize](https://github.com/coda/packs-sdk/pull/2116/files#diff-50a5e070285bcea931e1dcba7424c8da6bf255485306b181e7f904c679c2e365R179) have arguments being passed in via reference. Normally, this could pose a problem as the reference could be used to gain access to data out of the isolate, but that's mitigated here by using a copy (i.e. a [structured clone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)).

David also clarified that:
> `$0.applySync()` is the way of asking `isolated-vm` to execute a function from the outside world, with the specified parameters for how to transfer arguments and outputs across the vm boundary

Note: One point I was confused by was why `evalClosure()` doesn't also specify `{copy: true}` for the return value, but that function is only used to assign the `serializeFn` itself to `codaInternal.serializer.serialize` so that `marshalValue` can later get access it. (And it sounds like non-primitives being returned would throw an error anyways.)

More generally, it looks like we're using an existing pattern of injecting these types of functions into the isolate. I'm not terribly familiar with the Packs architecture, but reading through the comments, it looks like we've been careful not to expose references to `Reference`, `ExternalCopy`, etc to an isolate running untrusted packs code. Probably worth reviewing at some point in the future, however.

PR #2120, opened 2022-07-20
Updated ImageSchema exports in index

mphillips-codaio:
Updated exports for `ImageOutline` and `ImageCornerStyle` to export as enum instead of types. Used in object wrappers for pack images.

Built on top of [added ImageSchema](https://github.com/coda/packs-sdk/pull/2113) PR.


jonathan-codaio (approved):
Oops, sorry I should have caught this initially.

PR #2122, opened 2022-07-20
Add NumericDurationSchema for hinting numbers as durations

alexd-codaio:

Useful in many contexts, and seems like a small oversight.


jonathan-codaio (commented):
I think you need to add a new Zod validator to upload_validation.ts that handles the new schema.

Probably need to add a new entry to the `numberPropertySchema` union that is a clone of `durationPropertySchema` but with type Number.

alexd-codaio:
> I think you need to add a new Zod validator to upload_validation.ts that handles the new schema.
> 
> Probably need to add a new entry to the `numberPropertySchema` union that is a clone of `durationPropertySchema` but with type Number.

Done.

jonathan-codaio (approved):

PR #2127, opened 2022-07-25
Validate the authorizationUrl and tokenUrl are actually URLs

dweitzman-codaio:
This is just earlier feedback to a pack developer if there's something wrong with the provided authorizationUrl or tokenUrl

As a side benefit, once all packs are running at the next SDK release or higher we can simplify validation rules involving the domains of these URLs on the server side because parse errors could be server errors instead of a user-facing request error.

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -380,8 +380,8 @@ const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny> 
   }),
   [AuthenticationType.OAuth2]: zodCompleteStrictObject<OAuth2Authentication>({
     type: zodDiscriminant(AuthenticationType.OAuth2),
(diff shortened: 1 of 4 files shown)


testing/upload_validation.ts
```diff
@@ -380,8 +380,8 @@ const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny>
   }),
   [AuthenticationType.OAuth2]: zodCompleteStrictObject<OAuth2Authentication>({
     type: zodDiscriminant(AuthenticationType.OAuth2),
-    authorizationUrl: z.string(),
-    tokenUrl: z.string(),
+    authorizationUrl: z.string().url(),
```
jonathan-codaio, line 383: TIL

jonathan-codaio (approved):

PR #2128, opened 2022-07-28
Fix endless loop

ekoleda-codaio:
Identified when trying to help [a user in the community](https://community.coda.io/t/testing-dynamic-sync-formulas/32585), `newRealFetcherSyncExecutionContext` was calling itself instead of `newRealFetcherExecutionContext`, resulting in an endless loop.

### DIFF
--- a/dist/testing/execution.js
+++ b/dist/testing/execution.js
@@ -298,7 +298,7 @@ function newRealFetcherExecutionContext(packDef, manifestPath) {
 }
 exports.newRealFetcherExecutionContext = newRealFetcherExecutionContext;
 function newRealFetcherSyncExecutionContext(packDef, manifestPath) {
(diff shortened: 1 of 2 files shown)


jonathan-codaio (approved):
Oops 😬

spencer-codaio:
sounds like we probably want to release the new version w this?

ekoleda-codaio:
I [just recommended](https://community.coda.io/t/testing-dynamic-sync-formulas/32585/3) they install from GitHub for now, so perhaps not urgent. That said, it would be nice if that wasn't needed.

jonathan-codaio:
Some discussion here about the logistics of the next release: https://codaproject.slack.com/archives/C01835XEBSS/p1659022661141509

PR #2129, opened 2022-07-28
remove extra properties from helper

spencer-codaio:
* remove extra properties from helper so you grab from the schema itself
* allow object in title


jonathan-codaio (approved):
Nice, thanks.

PR #2130, opened 2022-07-28
Add length validation to various fields that will be indexed

cqian-codaio:
Adds length validation to:
- Building block names and descriptions
- Total building block count
- Column format matchers length
- Network domain url length and column format matcher regex length
PTAL @gary-codaio @jonathan-codaio @huayang-codaio

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -94,6 +94,7 @@ export const PACKS_VALID_COLUMN_FORMAT_MATCHER_REGEX = /^\/(.*)\/([a-z]+)?$/;
 export const BUILDING_BLOCK_COUNT_LIMIT = 500;
 export const BUILDING_BLOCK_NAME_CHARACTER_LIMIT = 50;
 export const BUILDING_BLOCK_DESCRIPTION_CHARACTER_LIMIT = 500;
+export const COLUMN_MATCHERS_CHARACTER_LIMIT = 100;
 export const COLUMN_MATCHERS_COUNT_LIMIT = 10;
 export const NETWORK_DOMAIN_CHARACTER_LIMIT = 100;
 
@@ -1087,7 +1088,8 @@ const formatMetadataSchema = zodCompleteObject<PackFormatMetadata>({
   hasNoConnection: z.boolean().optional(),
   instructions: z.string().optional(),
   placeholder: z.string().optional(),
-  matchers: z.array(z.string().refine(validateFormatMatcher)).max(COLUMN_MATCHERS_COUNT_LIMIT),
+  matchers: z.array(
+    z.string().max(COLUMN_MATCHERS_CHARACTER_LIMIT).refine(validateFormatMatcher)).max(COLUMN_MATCHERS_COUNT_LIMIT),
 });
 
 const syncFormulaSchema = zodCompleteObject<Omit<SyncFormula<any, any, ParamDefs, ObjectSchema<any, any>>, 'execute'>>({
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -91,12 +91,15 @@ import * as z from 'zod';
  * a real RegExp object.
  */
 export const PACKS_VALID_COLUMN_FORMAT_MATCHER_REGEX = /^\/(.*)\/([a-z]+)?$/;
-export const BUILDING_BLOCK_COUNT_LIMIT = 500;
-export const BUILDING_BLOCK_NAME_CHARACTER_LIMIT = 50;
-export const BUILDING_BLOCK_DESCRIPTION_CHARACTER_LIMIT = 500;
-export const COLUMN_MATCHERS_CHARACTER_LIMIT = 100;
-export const COLUMN_MATCHERS_COUNT_LIMIT = 10;
-export const NETWORK_DOMAIN_CHARACTER_LIMIT = 100;
+
+export const Limits = {
+  BuildingBlockCountPerType: 100,
+  BuildingBlockName: 50,
+  BuildingBlockDescription: 100,
+  ColumnMatcherRegex: 100,
+  NumColumnMatchers: 10,
+  NetworkDomainUrl: 253,
+}
 
 enum CustomErrorCode {
   NonMatchingDiscriminant = 'nonMatchingDiscriminant',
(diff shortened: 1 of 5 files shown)


cqian-codaio:
Some things left out were parameters names/descriptions and examples, which I'm leaning towards not indexing. From looking at various pack metadata out there these seem too noisy to attempt to match on.

testing/upload_validation.ts
```diff
@@ -91,6 +91,12 @@ import * as z from 'zod';
  * a real RegExp object.
  */
 export const PACKS_VALID_COLUMN_FORMAT_MATCHER_REGEX = /^\/(.*)\/([a-z]+)?$/;
+export const BUILDING_BLOCK_COUNT_LIMIT = 500;
+export const BUILDING_BLOCK_NAME_CHARACTER_LIMIT = 50;
+export const BUILDING_BLOCK_DESCRIPTION_CHARACTER_LIMIT = 500;
+export const COLUMN_MATCHERS_CHARACTER_LIMIT = 100;
+export const COLUMN_MATCHERS_COUNT_LIMIT = 10;
+export const NETWORK_DOMAIN_CHARACTER_LIMIT = 100;
```
cqian-codaio, line 100: I came up with random values for these. Open to discussing and changing them

testing/upload_validation.ts
```diff
@@ -91,6 +91,12 @@ import * as z from 'zod';
  * a real RegExp object.
  */
 export const PACKS_VALID_COLUMN_FORMAT_MATCHER_REGEX = /^\/(.*)\/([a-z]+)?$/;
+export const BUILDING_BLOCK_COUNT_LIMIT = 500;
```
jonathan-codaio, line 94: Nit, oof, I don't love that we were using deprecated style for constant variable names here, rather than being consistent with the unlovely style, mind just updating them all to use UpperCamelCase?

testing/upload_validation.ts
```diff
@@ -91,6 +91,12 @@ import * as z from 'zod';
...
+export const BUILDING_BLOCK_NAME_CHARACTER_LIMIT = 50;
```
jonathan-codaio, line 95: You could maybe group these together like

```typescript
const Limits = {
  BuildingBlockDescription: 500,
  ColumnMatcherRegex: 100,
  NumColumnMatchers: 10,
};
```

Better namespaced, and makes your imports cleaner too.

testing/upload_validation.ts
jonathan-codaio: Apparently domains are officially limited to 253 characters, maybe we could just go with that?

testing/upload_validation.ts
```diff
@@ -91,6 +91,12 @@ import * as z from 'zod';
...
+export const BUILDING_BLOCK_DESCRIPTION_CHARACTER_LIMIT = 500;
+export const COLUMN_MATCHERS_CHARACTER_LIMIT = 100;
+export const COLUMN_MATCHERS_COUNT_LIMIT = 10;
```
jonathan-codaio, line 98: Is this per column format? Might want to update the naming to reflect that.

testing/upload_validation.ts
```diff
@@ -1269,6 +1278,16 @@ function validateFormulas(schema: z.ZodObject<any>) {
         path: ['defaultAuthentication.networkDomain'],
       },
     )
+    .superRefine((data, context) => {
+      const buildingBlockCount = (data.syncTables?.length as number || 0) + (data.formulas?.length as number ?? 0) +
```
jonathan-codaio, line 1282: I think I'd just simplify this and say you can have at most 100 formulas, 100 sync tables, and 100 column formats. Little easier to understand, document, and implement.

jonathan-codaio (commented):
Nice.

Mind going ahead and validating parameter names and descriptions too? Agree we don't need to index them, but good to do to avoid this as an abuse vector, and prevent weird UI issues from long values. Fine to use the same limits for building block names and descriptions.

testing/upload_validation.ts
cqian-codaio: Added. I was wondering what places I had to document this in? I added something to the changelog. Should I add something to the SDK docs as well (maybe the troubleshooting section?)

cqian-codaio:
- Renamed limit variables
- Made building block count limit per building block type
- Added parameter name and description limits

testing/upload_validation.ts
```diff
@@ -92,6 +92,15 @@ import * as z from 'zod';
  */
 export const PACKS_VALID_COLUMN_FORMAT_MATCHER_REGEX = /^\/(.*)\/([a-z]+)?$/;
 
+export const Limits = {
+  BuildingBlockCountPerType: 100,
+  BuildingBlockName: 50,
+  BuildingBlockDescription: 100,
+  ColumnMatcherRegex: 100,
+  NumColumnMatchers: 10,
```
jonathan-codaio, line 100: Nit, `NumColumnMatchersPerFormat`.

testing/upload_validation.ts
jonathan-codaio: When @ekoleda-codaio is back he can advise on the best place in the documentation to mention these.

jonathan-codaio (approved):

gary-codaio (approved):

testing/upload_validation.ts
```diff
@@ -92,6 +92,15 @@ import * as z from 'zod';
  */
 export const PACKS_VALID_COLUMN_FORMAT_MATCHER_REGEX = /^\/(.*)\/([a-z]+)?$/;
 
+export const Limits = {
+  BuildingBlockCountPerType: 100,
+  BuildingBlockName: 50,
+  BuildingBlockDescription: 100,
+  ColumnMatcherRegex: 100,
```
ekoleda-codaio, line 99: As discussed over chat, these character limits feel a little low. The regex one in particular, given that regular expressions are known to get quite verbose sometimes.

testing/upload_validation.ts
ekoleda-codaio: As discussed in chat, my advice is to actually not document the limits for now. As long as the error message is clear, it will be just as informative and guaranteed to be up-to-date.

testing/upload_validation.ts
cqian-codaio: Yeah especially if we're allowing 253 characters in the network domain
I think we can set this to 300 or something? Most domains won't be 253 characters long, but just in case it is this gives a bit of leeway

testing/upload_validation.ts
```diff
@@ -92,6 +92,15 @@ import * as z from 'zod';
  */
 export const PACKS_VALID_COLUMN_FORMAT_MATCHER_REGEX = /^\/(.*)\/([a-z]+)?$/;
 
+export const Limits = {
+  BuildingBlockCountPerType: 100,
+  BuildingBlockName: 50,
+  BuildingBlockDescription: 250,
```
huayang-codaio, line 98: this feels too small for description. maybe we should not limit the description field but can trim the description before indexing.

cqian-codaio: It sounds like less but it's just the description of a particular building block, not the pack. Usually these are 1 sentence long but 250 allows for multiple sentences.

huayang-codaio: 250 could be the standard we enforce. if there's a situation when the pack maker needs more, they will report a bug. 

comparing to the other fields though, I just don't see this particular description length requirement as critical as others.

PR #2131, opened 2022-07-29
fix build options and allow the serializer release to build with coda repo webpack

huayang-codaio:
The trigger of this PR is that the latest packs-sdk doesn't build in the coda repo because of the new serializer module (using v8 or coda internal) doesn't build with webpack. Though the module isn't used in the browser at all, we'll need to define a stub to allow compiler to process. 

With further investigation I found a few more interesting stuff.
- the `dist/bundle.js` lib is loaded in both browser and isolated-vm. so far it works fine with a single bundle (alternative is to build one bundle for each targeting env). I'll leave it as is but modify the serializer to work with both environmentss. 
- our thunk and pack bundles are running in isolated-vm and bootstrap bundle is running in the lambda handler (node 14). The isolated-vm environment is not exactly node. I am switching to target `es2020` which is the closest one I found.
- a few other changes I'll comment inline.

## Tests

coda build: 

see https://github.com/coda/coda/pull/75966

image

CLI

```
-> % npx ts-node cli/coda.ts execute test/packs/fake_v2.ts StatusCode 404
Unrecogonized error {"name":"StatusCodeError","statusCode":404,"body":"","error":"","options":{"url":"https://httpbin.org/status/404","method":"GET"},"response":{"body":"","headers":{"access-control-allow-credentials":"true","access-control-allow-origin":"*","connection":"keep-alive","content-length":"0","content-type":"text/html; charset=utf-8","date":"Fri, 29 Jul 2022 18:15:06 GMT","server":"gunicorn/19.9.0"}}}
```

PTAL @dweitzman-codaio @jonathan-codaio


Makefile
```diff
@@ -99,12 +99,12 @@ compile-thunk:
...
-		--define:process.env.IS_THUNK=true \
+		--define:process.env.IN_ISOLATED_VM_OR_BROWSER=true \
 		--global-name=module.exports \
-		--target=node14;
+		--target=es2020;
 
-.PHONY: compile
-compile:
+.PHONY: compile-ts
+compile-ts:
```
huayang-codaio, line 107: pulling this out so that it can be used in local debugging without building the documents.

testing/compile.ts
```diff
@@ -156,13 +157,17 @@ async function buildWithES({
...
+    // logical assignment
+    target: 'ES2020',
 
     inject: getInjections(buildOptions),
     minify: false, // don't minify here since browserify doesn't minify anyway.
     sourcemap: 'both',
+
+    // The pack bundle is always targeting the isolated-vm environment.
+    define: {'process.env.IN_ISOLATED_VM_OR_BROWSER': 'true'},
+    external: ['codaInternal'], // for serializer
```
huayang-codaio, line 170: this allows the serializer module to work in CLI.

testing/node_fetcher.ts
```diff
@@ -178,12 +159,6 @@ export async function nodeFetcher(options: BaseFetcherOptions): Promise<FetcherF
     body: resultBody,
   };
 
-  if (simple) {
-    if (!response.ok) {
-      throw new StatusCodeError(response.status, resultBody, options, fullResponse);
-    }
-  }
```
huayang-codaio, line 185: node fetcher doesn't throw with status code by itself. this is done by `handleFetcherStatusError`. 

while it's not a big deal, this particular implementation uses a separate instance of `StatusCodeError` that confuses error handling. I figured it's easier to just remove the whole thing.

Makefile
dweitzman-codaio: Nice!

test/packs/fake_v2.ts
```diff
@@ -178,3 +179,30 @@ pack.addFormula({
...
+pack.addFormula({
+  name: 'StatusCode',
+  description: '',
+  parameters: [
+    coda.makeParameter({
+      type: coda.ParameterType.Number,
+      name: 'status',
+      description: '',
+    }),
+    // Add more parameters here and in the array below.
```
dweitzman-codaio, line 192: I think this comment is just leftover template code?

testing/node_fetcher.ts
dweitzman-codaio: This code is similar to what's in production, so if a change is needed here then a corresponding change may be needed in `coda`

bundles/thunk_bundle.js
```diff
@@ -4809,8 +4809,18 @@ module.exports = (() => {
   var serialize;
   var deserialize;
   if (true) {
-    serialize = codaInternal.serializer.serialize;
-    deserialize = codaInternal.serializer.deserialize;
+    serialize = (value) => {
+      if ("codaInternal" in global) {
```
huayang-codaio, line 4813: this actually doesn't work in the iframe. hold on..

dweitzman-codaio (approved):

testing/node_fetcher.ts
huayang-codaio: The node-fetch in production did throw but the fetcher service doesn't. it will catch the status code error and return a regular fetcher response. https://github.com/kr-project/experimental/blob/8a9d698d9af8e89822b7a8b51981bf3ad146be60/modules/server/fetcher-server/private/fetcher_service.ts#L352

testing/compile.ts
huayang-codaio: false alarm.. didn't refresh the webpack.

bundles/thunk_bundle.js
huayang-codaio: false alarm.

jonathan-codaio (commented):
What's the specific change that caused things to break?

huayang-codaio:
@jonathan-codaio https://github.com/coda/packs-sdk/pull/2116/files#diff-a483b6451dab123519bcc8e8bdc21493a9bfe3ea70fb10351338a88144fbc30e

This file.

jonathan-codaio:
> @jonathan-codaio https://github.com/coda/packs-sdk/pull/2116/files#diff-a483b6451dab123519bcc8e8bdc21493a9bfe3ea70fb10351338a88144fbc30e
> 
> This file.

Ah, I thought this is already live on the `coda` side? Is it not completely broken?

I trust you guys on these changes, no need to block on me, just trying to understand better.

huayang-codaio:
@jonathan-codaio yeah it's confusing. it took me a while a figure out the build chain. The short answer is that `packs-sdk-for-web-editor` is still on old version.

The coda repo uses @codahq/packs-sdk for backend services and browser bundling. On the other hand, I was expecting `packs-sdk-for-web-editor` to be only used in moncao syntax but it turns out that we are also using `packs-sdk-for-web-editor` in the Pack Studio (pack code) bundling (https://github.com/coda/coda/blob/2376268dbd961f8db8cbd2eef811b8adc0ffec69/modules/browser-website/private/postload/private/packs/contexts/pack_build_context.tsx#L220). 

This browser-generated pack code bundle is then loaded in the browser (to extract the pack metadata) and in isolated vm (formula execution). 

`packs-sdk-for-web-editor` will be bundled with webpack as well. That will loads the code in browser.

jonathan-codaio:
Ah yeah, we do compilation and metadata extraction on the client so we don't evaluate untrusted code on the server during that process. So all of that needs to be tied to the web editor sdk version.

Are there any parts of the new marshaling that are part of that client-side build, or is all the marshaling stuff part of the bundling that we do on the server side? Just curious if the new marshaling stuff isn't fully "launched" if there's a compile element to it that won't happen until the web-editor version is bumped.

huayang-codaio:
The marshaling code isn't executed in the browser. It's only to make the compilers (esbuild / webpack) happy.

The issues that require this PR to fix are:
- though not being used, `packs-sdk-for-web-editor` is an input of our webpack bundles. It's neither isolated-vm nor node. `serializer.ts` needs to a stub (i.e. throw not-implemented error) to make the compiler happy. otherwise it complains about missing symbols or missing modules. 
- The `dist/bundle.js` file will be running in the browser and in isolated-vm and the `serializer.ts` needs to accommodate.

PR #2132, opened 2022-07-29
add instanceof helpers

huayang-codaio:
The issue: `error instanceof coda.StatusCodeError` doesn't work in the packs code bundle. It's working well in all the other places (e.g. thunk bundle, lambda, or coad service code).

The root reason is that we don't do marshaling/unmarshaling between the thunk bundle and the pack bundle. Unfortunately they use different instances of the error class. 

After almost a day of exploration, I can't figure out a way to solve the problem without making a breaking change (of new / existing pack bundles). So I figured maybe we should just go the easiest solution: avoid using instanceof towards error classes in the packs bundle code.

Here's s 6 min walkthrough for more details. 
https://www.loom.com/share/da1c796ca3ae4381ade0278221cf9787

@jonathan-codaio @dweitzman-codaio


jonathan-codaio (approved):
Yeah, last year I dealt with a similar issue on the `coda` side and basically did the same thing.

PR #2156, opened 2022-08-02
Make console.log marshaling more forgiving

dweitzman-codaio:
An unwanted side effect of marshaling changes in https://github.com/coda/packs-sdk/pull/2116 was making console.log more rigid about what types it accepts.  People will want console.log to succeed with less information rather than fail because the type can't be represented.

This implementation calls `util.format` within isolated-vm rather than trying to marshal objects across the vm boundary and waiting for the `util.format` call later in the logger implementation

o/bug/24444


huayang-codaio:
For logging only, I was thinking about maybe simply do return `inspect` or `JSON.stringify` for unrecognized types. It's tricky to identify and handle all the unrecognized types. WDTY?

huayang-codaio:
I am not against to try this version with the up-coming release though. just evaluating the potential risks.

dweitzman-codaio:
> For logging only, I was thinking about maybe simply do return `inspect` or `JSON.stringify` for unrecognized types. It's tricky to identify and handle all the unrecognized types. WDTY?

Seems like `JSON.stringify(() => return 1)` returns `undefined` and `JSON.stringify(Promise.resolve(1))` returns `{}`, so I suspect the text "<function>" and "<promise>" is probably more helpful to people in practice

dweitzman-codaio:
I guess `inspect` would have more promise. The values could be large, but maybe we don't care

dweitzman-codaio:
Switched to using "inspect"

huayang-codaio:
yeah, JSON.stringify isn't working on the promise/function cases that you handled in the PR. just suggesting to have some general fallback to unrecognized types (for logging).

huayang-codaio:
and yeah inspect probably works better.

dweitzman-codaio:
It's not that easy to know if a type can or can't be cloned. https://github.com/ungap/structured-clone/blob/main/cjs/serialize.js shows one approach where you use `Object.prototype.toString.call(value)` and then break apart the returned string to see what it reveals about the true type of an object

Something like that may be possible, but it feels kind of awkward and complex so I was thinking of doing something simple here that would work for most people most of the time

runtime/common/marshaling/index.ts
```diff
@@ -125,15 +133,22 @@ function fixUncopyableTypes(
     }
   }
 
+  if (forLogging) {
+    if (typeof val === 'function' || val instanceof Promise) {
+      return {val: inspect(val), hasModifications: true};
```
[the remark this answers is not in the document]
huayang-codaio: after inspecting our code, the path for `console.log` seems really simple. we can simply turn the entire input into a string here without worrying for unmarshaling into the original object. eventually the fetcher proxy simply calls `util.format(*args)`

runtime/common/marshaling/index.ts
```diff
@@ -158,16 +173,30 @@ function isMarshaledValue(val: any): boolean {
...
 }
 
+// The marshaling for console.log is willing to be more lossy/forgiving than the marshaling
+// for other functions since you usually want console.log to print _something_ rather than
+// throw an error.
+//
+// Currently only functions and promises are detected and turned into strings. Other unsupported
+// types could still raise an error.
+export function marshalValueForLogging(val: any): MarshaledValue {
+  return marshalValueInternal(val, true);
```
huayang-codaio, line 193: assuming `util.format` is browserified. maybe this can simply be `return format(val)`;

dweitzman-codaio:
Changed the implementation to call `util.format` within isolated-vm so a wider variety of types are supported out of the box.

test/marshaling_test.ts
```diff
@@ -192,4 +197,14 @@ describe('Marshaling', () => {
     assert.throws(() => unmarshalValue(1), 'Not a marshaled value: 1');
     assert.throws(() => unmarshalValue({foo: 'bar'}), 'Not a marshaled value: {"foo":"bar"}');
   });
+
+  it('marshals values for logging', () => {
```
huayang-codaio, line 201: 👍

test/thunk_test.ts
```diff
@@ -126,4 +128,39 @@ describe('Thunk', () => {
 
     assert.equal(result, true);
   });
+
+  it('log marshaling works', async () => {
```
huayang-codaio, line 132: 👍

Makefile
```diff
@@ -100,6 +100,7 @@ compile-thunk:
 		--inject:${ROOTDIR}/testing/injections/buffer_shim.js \
 		--format=iife \
 		--define:process.env.IN_ISOLATED_VM_OR_BROWSER=true \
+		--define:process.env.NODE_DEBUG=false \
```
huayang-codaio, line 103: can you add a note in code why this is added?

huayang-codaio (approved):
thank you!

test/marshaling_test.ts
```diff
@@ -192,4 +197,14 @@ describe('Marshaling', () => {
...
+    assert.deepEqual(transformForLogging(['one string']), ['one string']);
+    assert.deepEqual(transformForLogging(['two', 'strings']), ['two strings']);
+    assert.deepEqual(transformForLogging(['a %s c', 'b']), ['a b c']);
+    assert.deepEqual(transformForLogging(['%o vs %j', {a: 'b'}, {a: 'b'}]), [`{ a: 'b' } vs {"a":"b"}`]);
```
dweitzman-codaio, line 205: Just an FYI, the "%o" here won't actually work for user packs because it seems like it's not supported by the pure-js version of `util.format` that the esbuild thunk uses. The `%j` does work, though

dweitzman-codaio: Added comments in the code about this and a thunk_test to demonstrate it

Makefile
dweitzman-codaio: Added a comment mentioning that util.format depends on debuglog which depends on NODE_DEBUG, with a link to the source code at https://github.com/nodejs/node/blob/6b055f385744d2ca71c19d46a0ec3bcfc51f5cd3/lib/internal/util/debuglog.js#L21

huayang-codaio: ah, got it. that makes sense.

test/marshaling_test.ts
huayang-codaio: good to know. thanks for calling this out.

PR #2159, opened 2022-08-02
bump node-fetch to 3.2.10 and fix timeout

gary-codaio:


gary-codaio:
Coda node fetch not bumped, this simulates that

PR #2170, opened 2022-08-09
add source to release handler

gary-codaio:
Mark release from the cli as such

### DIFF
--- a/cli/release.ts
+++ b/cli/release.ts
@@ -1,6 +1,7 @@
 import type {ArgumentsCamelCase} from 'yargs';
 import type {BasicPackDefinition} from '..';
 import type {PackVersionDefinition} from '..';
+import { PublicApiPackSource } from '../helpers/external-api/v1';
 import {build} from './build';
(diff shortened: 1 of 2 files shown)


jonathan-codaio (commented):
Is this just for logging? We currently don't support this on the endpoint, so to actually do this you'd have to add in this field in the openapi spec and add handling on the `coda` side. `v1.ts` is actually just copied from the codegen SDK that's generated from the openapi spec rather than something that gets hand-edited here.

Up to you if you want to do all that work for logging this field, I'm fine skipping it--you can probably get a bit for logging just by inspecting the userAuth object in the endpoint and seeing if you can determine whether it was cookie auth or if it had an API token, the latter indicating this came from the CLI or the API directly.

gary-codaio:
@jonathan-codaio yeah this is just for logging and I was in the midst of working through all the changes on the coda side. Given how much that is starting to balloon, I'll look at checking the cookie vs token auth

PR #2171, opened 2022-08-11
Fix small typo in schemas.md

zack-codaio:
fix typo: coda.ValeType.Number -> coda.ValueType.Number


ekoleda-codaio (approved):
Thanks for the fix!

PR #2173, opened 2022-08-16
Allow flexible yes/no responses in prompts

ekoleda-codaio:
A developer noticed that [the documentation](https://coda.io/packs/build/latest/tutorials/get-started/cli/#register-an-api-token) said to enter `y` but that only `yes` was accepted. I noticed that the codebase was using a mix of `yes` and `y`, without the ability to convert between the two.

I updated the `promptForInput` helper function to include a special `yesOrNo` mode that used the `yn` NPM library to handle all the variations. I then updated the codebase to use the this new mode wherever a yes/no question was being asked.

I also added tests for this function, using the newly added `proxyquire` NPM library, as I couldn't get `sinon` to mock a lop-level export of a library.

### DIFF
--- a/testing/helpers.ts
+++ b/testing/helpers.ts
@@ -6,6 +6,7 @@ import * as readlineSync from 'readline-sync';
 import {translateErrorStackFromVM} from '../runtime/common/source_map';
 import {unwrapError} from '../runtime/common/marshaling';
 import {wrapError} from '../runtime/common/marshaling';
+import yn from 'yn';
 
 export function getManifestFromModule(module: any): PackVersionDefinition {
   if (!module.manifest && !module.pack) {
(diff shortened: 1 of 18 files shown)


testing/helpers.ts
```diff
@@ -6,6 +6,7 @@ import * as readlineSync from 'readline-sync';
 import {translateErrorStackFromVM} from '../runtime/common/source_map';
 import {unwrapError} from '../runtime/common/marshaling';
 import {wrapError} from '../runtime/common/marshaling';
+import yn from 'yn';
```
jonathan-codaio, line 9: How much do we care about using this vs rolling our own simple heuristic? Based on yarn.lock, this is pulling in many new packages, just for some simple string checks, not sure if it's worth it.

jonathan-codaio (commented):
Great! Glad to see this standardized finally.

testing/helpers.ts
ekoleda-codaio: I don't think `yn` has an dependencies:

https://github.com/sindresorhus/yn/blob/main/package.json

I think the ones you are seeing in `yarn.lock` are from `proxyquire`.

jonathan-codaio (approved):
Oh got it, I missed that.

PR #2174, opened 2022-08-17
Add chai and mocha imports to testing code snippets

ekoleda-codaio:
The assert one seems required at the very last.


jonathan-codaio (approved):
Ah yeah, these are a bit foreign to us because our repos have some magic settings that add these as globals, but this would be more typical.

PR #2175, opened 2022-08-23
tweak words to let people get help for oauth url domain mismatch

huayang-codaio:
small tweaks to the language to make it support the related case where oauth uses different domains for authorization and token url. 

@dweitzman-codaio @ekoleda-codaio @jonathan-codaio


ekoleda-codaio (approved):
Thanks for adding to the docs!

PR #2176, opened 2022-08-29
Add helper `getEffectivePropertyKeysFromSchema` to sync formulas

huayang-codaio:
If the doc user has picked certain properties using the new column picker, how does the pack sync formula know which columns were picked? 

The updated schema will be passed in with the sync execution context as in `context.sync.schema`. To get the list correctly, the pack maker basically needs to do `const originalKeys = Object.entries(context.sync.schema.items.properties).map([key, property] => property.fromKey || key);` to get the list of picked properties. 

Passing `selectedOriginalKeys` in here would make it easier for pack makers. 

However, the challenge is that we may be supporting picking properties in a nested structure. (e.g. picking properties in issues->project, where issues is the top level object). A simple array won't be enough to represent the options. 

A few alternatives:

1. passing as a dict from schema identity to the list of keys picked. this may not work if the top level object has two fields of the same object type. (e.g. issue.creator and issue.assignee can be both of schema `User`)
2. passing a nested structure like `{fieldA: {}, fieldB: {}, fieldC: {subField1: {}, subField2: {}}, fieldD: {}}`

I figured either options above are pretty complicated to parse and won't give the pack maker much convenience than they parsing the `context.sync.schema` object directly. So in this PR I am only adding a simple `string[]` field that can be used by most simple pack implementations which don't care about nested fields. I'd love to hear what you think about this: do we need this convenient field at all or should we make it more complicated? 

@jonathan-codaio @alan-codaio @dweitzman-codaio


jonathan-codaio (commented):
Yeah, this is pretty tricky. Since the existing `schema` field is already complete and accurate, I think I would actually instead explore adding utility methods that allow the maker to extract the fields they want from the schema.

So you could just export a method that just takes the schema and returns the top-level fromKeys, and possibly another that returns the top-level property names (rather than the fromKeys) and if useful, yet another that returns some kind of nesting. It's only one extra line of code for the maker, but it puts them in control of deciding which fields are most actionable for them.

(A similar approach, structured differently, is changing `context.sync` into an object that has methods on it in addition to data fields, so that you can call methods to get at this data rather than using utility methods.)

Would love to hear @ekoleda-codaio 's thoughts on this as always, though he's on vacation this week.

huayang-codaio:
With discussions of https://staging.coda.io/d/Packs-Platform-IA-o-packs_d36WK4zgrYx/Nested-properties-in-the-column-picker_suAJG#_lub1p, I don't think it's necessary to support a full nested property selection at the point but we should still make the API flexible. Let me change to a helper approach instead.

jonathan-codaio (approved):
Don't forget to update the CHANGELOG.

helpers/schema.ts
```diff
@@ -0,0 +1,20 @@
+import type {Schema} from '../schema';
+import {ValueType} from '../schema';
+
+/**
+ * A helper to extract properties fromKeys from a schema object. This is mostly useful
+ * in processing the context.sync.schema in a sync formula, where the schema would only
+ * include a subset of properties which were manually selected by the Pack user.
+ */
+export function getSelectedFromKeysFromSchema(schema: Schema): string[] | undefined {
```
ekoleda-codaio, line 9: I think there is room to improve the name of this helper function. "Selected" isn't accurate here, since it works for any schema, even one that hasn't be altered by the user. Also "FromKeys" isn't completely accurate, since it returns a value for properties that don't have `fromKey` set.

What about `getPropertyKeysFromSchema`?

ekoleda-codaio: SGTM.

huayang-codaio:
we are close to ship column picker. I am coming back to merge this now and update the jira pack to leverage the user picked columns to optimize its data fetching. cc @patrick-codaio as we just chatted.

helpers/schema.ts
jonathan-codaio: Did you want to rename this as Eric suggested?

jonathan-codaio (commented):
@chris-codaio This is the helper I was referring to the other day. Given a schema, it returns the list of key names before we mangled them (if we mangled a name we keep the original name in the fromKey property).

Hoping the eventual 2-way sync handling and this use case (filtering a sync implementation to only the user-selected fields) can share an implementation, as both do the same thing conceptually: mapping property names back into the maker's vocabulary.

helpers/schema.ts
huayang-codaio: just realized that I have not renamed it yet.

jonathan-codaio (approved):

PR #2201, opened 2022-09-09
Reorganize left nav

ekoleda-codaio:
[Preview](https://coda-packs-sdk--2201.com.readthedocs.build/en/2201/guides/)

1.  For expandable sections, get rid of the "Overview" terminology ("Sync tables > Overview" becomes "Sync tables" > "Sync tables"). I've observed some people use the documentation recently, and they seem to avoid the Overview pages. My thought is that they assume the page contains only introductory fluff, when in fact it contains the meat of the topic. The duplicative naming isn't great, but I think it's tolerable.

    Before & After:
    Before After

1. Move "Authentication" and "Fetching remote data" into the "Basics" section. The vast majority of Packs use these features, and while not simple, they should probably be considered standard topics. Also "Basics" was looking somewhat empty compared the ever expanding "Advanced" section.

    Before & After:
    Before After

1.  Create a new "Distribution" grouping, moving "Sharing and Publishing" to it and adding "Selling Packs". Both are hosted in the Help center.

    Before & After:
    Before After


jonathan-codaio (approved):
If I saw two things labeled "Sync tables" I'd probably assume the first was only an expander and click the second one still, so I'm not sure this properly directs people to the first link still 😄 If we can come up with a distinct label for each page I think that'd be ideal.

jonathan-codaio:
> The first "Sync tables" is only expander, and I want people to click the second "Sync tables" since that's the actual content, so I think we're OK? The issue before was that I think people saw "Overview" and thought "I don't need an overview, I need the details!".

Ohh, got it.

PR #2203, opened 2022-09-12
Add some SDK fidelity for progress bars.

jonathan-codaio:
To address https://staging.coda.io/d/_d-GJF-DmEUK#Bugs_tuq-W/r25097&view=modal

Turns out we've had progress bars for several months but never support in cross-doc. I added some TypeScript checks on the `coda` side 5 months ago that this wouldn't happen again, but there was a human race condition between when I added these checks and when the feature was being built.

The full progress bar column format includes color + gradient specifications here, so this doesn't yet give us full fidelity. I figure I don't want to commit to a bunch of color-related stuff in the SDK before we think about it more plus it's a bunch of boilerplate work to do, seems good enough for now to at least support this level of fidelity.


requirements.txt
```diff
@@ -1,28 +1,28 @@
 # Autogenerated by 'make bs'
 -i https://pypi.org/simple
-bracex==2.2.1
-click==8.1.2
+bracex==2.2.1; python_version >= '3.6'
```
jonathan-codaio, line 3: These diffs come along when you run `make bs` on linux. I don't know if this file is just stale or if this is a conflict with Mac.

huayang-codaio (approved):
didn't realize we have a progress bar either. lol

requirements.txt
ekoleda-codaio: I think it was just stale. I think this new format used to not be compatible with ReadTheDocs, but seems to work now, so go for it.

ekoleda-codaio:
Similar to https://github.com/coda/packs-sdk/pull/2005#issuecomment-1136346270, I like the mental model of Column Type -> `codaType`. Having to pick `Slider` to create a progress bar isn't very obvious, and sort of exposes our implementation details. I know it makes the SDK code a little more complicated, but I'd prefer `ProgressBar` be it's own `codaType`.

huayang-codaio:
yeah, it's very convincing that I won't realize `ProgressBar` is a subtype of `Slider`.

jonathan-codaio:
Ok giving this a shot. Really hoping the implementation on the `coda` side is not a pain, was intending for this to be a quick fix and not a project 🤞

huayang-codaio (approved):

jonathan-codaio:
@ekoleda-codaio does this look ok now?

ekoleda-codaio:
Yes! Thanks for going the extra mile on this.

ekoleda-codaio (approved):

PR #2204, opened 2022-09-13
Make isolated-vm an optional dependency.

jonathan-codaio:
Just creates a wrapper for accessing ivm functionality which does a try/catch on the require. Most functionality just returns early or skips tests if the package is not available. As a Codan you would still need to have it installed in order to build the runtime, hoping that would hard fail if it wasn't, but haven't tested it.

The stated behavior of npm on optionalDependencies is that it will try to install them but just skip past if they fail, which is what we want. Weirdly, I was testing on my M2 Mac which could not install the SDK as it currently exists due to `isolated-vm` issues, but with this fix, it not only installed fine, but `isolated-vm` was actually installed after all! Like, it skipped past sub-dependencies of isolated-vm or something. I don't understand, but it made it work great on the M2.


package.json
```diff
@@ -50,8 +49,12 @@
     "uuid": "^8.3.2",
     "xml2js": "^0.4.23",
     "yargs": "^17.5.1",
+    "yn": "^3.1.1",
```
jonathan-codaio, line 52: This is used at runtime but wasn't in package.json, I think we were getting it transitively from isolated-vm. The latest version of this package is 5.0.0, but that version uses a different module style that is not compatible with our code. So pegged this to the same version we were already using. When dependabot tries to update this, hopefully something will fail so we don't accidentally upgrade it...

huayang-codaio, line 52: is this a missing required package?

huayang-codaio (approved):
works for me. I'd expect some error message to remind the user of the failed dependency installation so that they would know what happened.

package.json
huayang-codaio: just saw your comment.

jonathan-codaio: Yep, see https://github.com/coda/packs-sdk/pull/2204#discussion_r970138214

PR #2207, opened 2022-09-15
Update MkDocs Material, generate social cards, and improve nav

ekoleda-codaio:
[Preview](https://coda-packs-sdk--2207.com.readthedocs.build/en/2207/). [Social cards preview](https://socialsharepreview.com/?url=https://coda-packs-sdk--2207.com.readthedocs.build/en/2207/).

1.  Updates the mkdocs-material dependency to get access to newly released features.
    - The new version introduced a bug on our homepage, where we were previously hiding the `<h1>` tag. Decided to stop hiding that tag and moved up a line of text to make the layout look better.

1.  Enabled the MkDocs Material "social" plugin, to generate unique images for each page to use in social sharing link preview. web
    - Removed previous social meta tags, since the plugin now generates these instead.
    - Added flag to disable generating cards when running a local preview server, since it greatly slows down incremental builds.
    - Updated the CircleCI config to install the image processing libraries needed by the social plugin.
    - Discovered that the plugin relies on `site_url` being set correctly in `mkdocs.yaml`, which wasn't happening correctly for builds done by Read the Docs. Created a helper script that preprocesses the file to make it work with Read the Docs, and added a `.readthedocs.yaml` file to run the script before the docs are built.
    
1.  Enabled the `navigation.indexes` feature, which allows `index.md` files to be flattened into the parent zippy item.
    -  This helps address some [previous feedback](https://github.com/coda/packs-sdk/pull/2201#pullrequestreview-1102416515) that the duplicate nav items wasn't ideal.
        Before & After
        Screen Shot 2022-09-15 at 1 39 26 PM Screen Shot 2022-09-15 at 1 39 08 PM

    -  Moved the main Overview page from `index.md` to `overview.md` since it shouldn't be collapsed.


ekoleda-codaio:
I haven't asked the brand team yet. This change replaces the previous static image, which was a generic banner:

image

I think the brand team could create a better image then what this plugin generates, but it would be static across the entire site. The advantage of this plugin is that it generates a unique image for each page, using the title and description metadata for the page. This is similar to what GitHub does, for example, for each repo:

image

There's a tradeoff between visual appeal and link-specific, and there is perhaps a debate to be had there. I think this is enough of an improvement to move forward as-is, but I'll add a TODO to sync with the brand team.

PR #2208, opened 2022-09-15
Fix zod validation for slider min/max/step that are formulas.

jonathan-codaio:
Part of the fix for https://staging.coda.io/d/_d-GJF-DmEUK#Bugs_tuq-W/r25095&view=modal. Sliders schemas can have formulas for these attributes, our types already supported that but zod did not. So if you try to cross-doc a column that uses formulas for these, you'd hit a zod validation error in the cross-doc workflow.

I don't think we have any other examples of passing CFL in the SDK yet, so I don't think this is functionality we should advertise, but if we're just blindly passing through values from cross-doc, seems ok.

This doesn't fully fix the issue though, because there's a bug in the API, and the actual cell values for sliders that use formulas are structured values and the API is returning them as serialized JSON strings rather than numbers (and those JSON values don't even have the cell value! just the calculated min/max/step values) so these do not round-trip properly.

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -613,12 +613,14 @@ const scalePropertySchema = zodCompleteStrictObject<ScaleSchema & ObjectSchemaPr
   ...basePropertyValidators,
 });
 
+const optionalStringOrNumber = z.union([z.number(), z.string()]).optional();
+
 const sliderPropertySchema = zodCompleteStrictObject<SliderSchema & ObjectSchemaProperty>({
   type: zodDiscriminant(ValueType.Number),
   codaType: zodDiscriminant(ValueHintType.Slider),
-  maximum: z.number().optional(),
-  minimum: z.number().optional(),
-  step: z.number().optional(),
+  maximum: optionalStringOrNumber,
+  minimum: optionalStringOrNumber,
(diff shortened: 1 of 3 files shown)


jonathan-codaio:
PTAL @huayang-codaio @coda/packs

testing/upload_validation.ts
```diff
@@ -613,22 +613,24 @@ const scalePropertySchema = zodCompleteStrictObject<ScaleSchema & ObjectSchemaPr
...
 
+const optionalStringOrNumber = z.union([z.number(), z.string()]).optional();
+
 const sliderPropertySchema = zodCompleteStrictObject<SliderSchema & ObjectSchemaProperty>({
   type: zodDiscriminant(ValueType.Number),
   codaType: zodDiscriminant(ValueHintType.Slider),
-  maximum: z.number().optional(),
-  minimum: z.number().optional(),
-  step: z.number().optional(),
+  maximum: optionalStringOrNumber,
```
huayang-codaio, line 621: do you want to validate that the only allowed string is `[Max]`?

jonathan-codaio: Oh it can be any CFL formula, the test was just an example.

huayang-codaio (approved):

PR #2210, opened 2022-09-17
params can be undefined

balupton:
Currently there is an edge case with `optional: true` paramaters where they can be undefined without typescript catching it.

This PR leans on the safer side and marks that all params can be undefined, which inspires pack authors to set default values to prevent undefined exceptions.

See discussion at: https://community.coda.io/t/incorrect-types-for-optional-array-params/33848/5?u=balupton

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -339,7 +339,7 @@ type TypeOfMap<T extends UnionType> = T extends Type
  * the parameter defintion for that formula.
  */
 export type ParamValues<ParamDefsT extends ParamDefs> = {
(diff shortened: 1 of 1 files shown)


balupton:
As to whether it is undefined or null, that depends on what the code is actually doing. I'm not sure which is which. If it is null, then doing `TypeOfMap<T>?` is sufficient.

jonathan-codaio:
Ok, I think this will do the trick: https://github.com/coda/packs-sdk/pull/2223

Getting TypeScript to cooperate here was...painful.

ekoleda-codaio:
Fixed by #2223.

PR #2211, opened 2022-09-20
Correct typo in embeds.md

jleverenz:


jonathan-codaio:
Thanks for this and sorry for the delay! We had a team offsite.

LGTM, we just have a bit of a tooling issue where our continuous integration pipeline doesn't run for people outside of Coda, which makes it hard to merge open-source contributions so they don't pass our build/test checks. We're looking into a good solution there, but in the meantime, I've put up a versions of this PR that will pass the CI that I can then merge.

ekoleda-codaio:
This was fixed in #2222. Thanks again for the PR!

PR #2223, opened 2022-09-23
When a parameter is declared as optional, include undefined in its inferred type in execute().

jonathan-codaio:
PTAL @ekoleda-codaio @huayang-codaio @coda/packs


api_types.ts
```diff
@@ -339,7 +343,11 @@ type TypeOfMap<T extends UnionType> = T extends Type
  * the parameter defintion for that formula.
  */
 export type ParamValues<ParamDefsT extends ParamDefs> = {
-  [K in keyof ParamDefsT]: ParamDefsT[K] extends ParamDef<infer T> ? TypeOfMap<T> : never;
+  [K in keyof ParamDefsT]: ParamDefsT[K] extends OptionalParamDef<infer S>
+    ? TypeOfMap<S> | undefined
```
jonathan-codaio, line 347: This is the heart of the change. The rest of it is getting TypeScript to get enough type inference from the return value of `makeParameter()` such that this check works--that it understands a parameter definition could possibly extend OptionalParamDef.

api.ts
```diff
@@ -372,6 +372,34 @@ export type ParameterOptions<T extends ParameterType> = Omit<ParamDef<ParameterT
     : undefined;
 };
 
+type UnionParameterOptions =
```
jonathan-codaio, line 375: I did not want to do this horrendous union, but when I tried do this with generics, TypeScript would not properly infer the return value of `makeParameter()`:

```typescript
makeParameter<T extends ParameterType, O extends ParameterOptions<T>>(options: O): DeriveStuffFromO<O> 
```

The output of this was not a parameter bound to a specific value of ParameterType as I think it should be, but rather a parameter bound to the non-specific `ParameterType` enum class, which breaks all specific inference downstream.

huayang-codaio: it seems to work for me removing `UnionParameterOptions`. just pushed a commit for demonstration. can you take a look if that makes sense?

api.ts
```diff
@@ -372,6 +372,14 @@ export type ParameterOptions<T extends ParameterType> = Omit<ParamDef<ParameterT
     : undefined;
 };
 
+type ParamDefFromOptionsUnion<T extends ParameterType, O extends ParameterOptions<T>> = Omit<
+  O,
+  'type' | 'autcomplete'
+> & {
+  type: O extends ParameterOptions<infer S> ? ParameterTypeMap[S] : never;
```
huayang-codaio, line 379: i honestly have no idea why this works...

maybe because [`OptionalParamDef<infer S>`](https://github.com/coda/packs-sdk/blob/4f3b641575b031e1620e1951b75db48e7ee6ecd5/api_types.ts#L346) is also using inferring S?

huayang-codaio: image
image

huayang-codaio (approved):

jonathan-codaio:
Merging, WCGW...

balupton:
Good work. Would be keen to see a video breaking down how this works, as it seems you have to step into the realm of sorcery.

jonathan-codaio:
> Good work. Would be keen to see a video breaking down how this works, as it seems you have to step into the realm of sorcery.

Yeah, the heart of it is to give a name to a param with `optional: true` (`OptionalParamDef` is now a named type for that) that can be checked for in the part of the code you were working with before, to differentiate this case from a non-optional param.

The rest of it is applying more generics to `makeParameter` so that its return value becomes very specific, so that TypeScript can infer the output of `makeParameter({optional: true})` as being compatible with `OptionalParamDef`. When the return value is statically declared as `ParamDef`, TypeScript will simply never trigger the conditional for `OptionalParamDef`.

PR #2229, opened 2022-09-27
validate building blocks of the same type have unique names

gary-codaio:
Sync tables already checked this, added it for formulas and formats too.

Tested on dev after updating the pack-sdk version we used
image


requirements.txt
```diff
@@ -1,44 +1,44 @@
 # Autogenerated by 'make bs'
 -i https://pypi.org/simple
-bracex==2.3.post1; python_version >= '3.7'
-cairocffi==1.3.0; python_version >= '3.7'
+bracex==2.3.post1 ; python_version >= '3.7'
```
jonathan-codaio, line 3: Do you use devbox or Mac? I get these same whitespace diffs (on devbox), been wondering if it's platform-dependent.

testing/upload_validation.ts
```diff
@@ -1202,8 +1202,34 @@ const unrefinedPackVersionMetadataSchema = zodCompleteObject<PackVersionMetadata
...
-  formulas: z.array(formulaMetadataSchema).max(Limits.BuildingBlockCountPerType).optional().default([]),
-  formats: z.array(formatMetadataSchema).max(Limits.BuildingBlockCountPerType).optional().default([]),
+  formulas: z
+    .array(formulaMetadataSchema)
+    .max(Limits.BuildingBlockCountPerType)
+    .optional()
+    .default([])
+    .superRefine((data, context) => {
+      const formulaNames = data.map(formulaDef => formulaDef.name);
+      for (const dupe of getNonUniqueElements(formulaNames)) {
```
jonathan-codaio, line 1212: Ah this is my bad, but can we make getNonUniqueElements() case-insensitive? The formula language will treat formulas the same regardless of case, so we should do the same here.

jonathan-codaio (commented):
Great, thank you! Mind adding a test for this in upload_validation_test.ts? Should be boilerplate for the sync table name case that we can just copy.

requirements.txt
gary-codaio: I'm on mac

testing/upload_validation.ts
```diff
@@ -192,7 +192,11 @@ export function validateSyncTableSchema(schema: any): ArraySchema<ObjectSchema<a
 function getNonUniqueElements<T>(items: T[]): T[] {
```
jonathan-codaio, line 192: I got too cute with the typing here, I think we only use it for strings, so it's also fine to just type this as `item: string[]` or do `<T extends string>` as the generic (handles strings and enums).

testing/upload_validation.ts
gary-codaio: Had to do a questionable bit of type casting to make it case insensitive. Would be worth checking if that's what you had in mind

testing/upload_validation.ts
```diff
@@ -192,7 +192,11 @@ export function validateSyncTableSchema(schema: any): ArraySchema<ObjectSchema<a
...
   const set = new Set<T>();
   const nonUnique: T[] = [];
-  for (const item of items) {
+  for (let item of items) {
+    // make this case insensitive for strings
+    if (typeof item === 'string') {
+      item = item.toUpperCase() as T
+    }
     if (set.has(item)) {
       nonUnique.push(item);
```
jonathan-codaio, line 201: Super nit, this will put the normalized name in the error message; the case I think we think will be most common is repeating e.g. `'Foo'` twice by accident, would be best if the original name was returned as-is. Maybe:

```typescript
for (const item of items) {
  const normalized = item.toUpperCase();
  if (set.has(normalized)) {
    nonUnique.push(item);
  }
  set.add(normalized);
}
```

jonathan-codaio (approved):
🎉

PR #2233, opened 2022-09-28
enable property path projection

spencer-codaio:
This enables handling for Pack Makers to project out properties from nested objects in an object for any of our `...property` properties on the object schema. 

For example, if you had an object schema of the following:

```
const user = makeObjectSchema({
  type: ValueType.Object,
  displayProperty: 'login',
  properties: {
    login: {type: ValueType.String},
    avatar: {type: ValueType.String, fromKey: 'avatar_url', codaType: ValueHintType.ImageReference},
  },

export const simplePullRequestSchema = makeObjectSchema({
  type: ValueType.Object,
  displayProperty: 'title',
  idProperty: 'url',
  properties: {
    title: {type: ValueType.String},
    author: {
      ...user,
      fromKey: 'user',
    },
    url: {type: ValueType.String, fromKey: 'html_url', codaType: ValueHintType.Url},
  },
```

Before, you wouldn't be able to access any of the properties of `author` natively and would have to manually add a duplicate property at the top-level object in order to use it for these properties. With this change you'll be able to write the following to extract the author's avatar:

```
export const simplePullRequestSchema = makeObjectSchema({
    ...originalProperties
    imageProperty: 'author.avatar',
}
```

internal doc link: https://staging.coda.io/d/Pack-Previews-links-embeds-and-objects-go-pack-previews_dp4gmVCYbYv/SDK-Object-Property-Projection_sui2n?utm_source=slack&utm_content=comment_notification#_lu08n

This uses the `json-path` library which matches the pathing logic we support in the document. 

**Some known limitations**
* we only support `.` syntax for accessing object properties. This means that the projection logic will not work for properties that have periods in the name. 
* `[]` accessing will always be interpreted as array access rather than being interchangeable for object access

This currently only enables it for the new properties used for pack cards but will follow-up with adding capability for existing `...property` values after that works

@coda/packs


spencer-codaio:
I should probably add some documentation in comments but not sure where to put it.. open to ideas or if this should just be documented in the actual SDK documentation? cc @ekoleda-codaio

schema.ts
```diff
@@ -839,9 +839,18 @@ export interface Identity extends IdentityDefinition {
   packId: number;
 }
 
+export interface ObjectProperty {
+  label: string;
+  value: string;
+}
+
+export type PropertyType<K extends string = string> = K | string | ObjectProperty;
```
jonathan-codaio, line 847: Nit, PropertyIdentifier may be a little clearer. The properties themselves are the contents of the `properties` dict in the schema def, and those property defs have a `type` field; this type has nothing to do with those values and that field, so we probably want to distinguish the naming.

Similar, ObjectProperty might be better as `PropertyIdentifierSpec` or something.

schema.ts
```diff
@@ -1371,11 +1425,11 @@ export function normalizeSchema<T extends Schema>(schema: T): T {
       description: schema.description,
       attribution: schema.attribution,
       includeUnknownProperties: schema.includeUnknownProperties,
-      titleProperty: titleProperty ? normalizeSchemaKey(titleProperty) : undefined,
-      subtitleProperties: subtitleProperties ? subtitleProperties.map(normalizeSchemaKey) : undefined,
-      imageProperty: imageProperty ? normalizeSchemaKey(imageProperty) : undefined,
-      descriptionProperty: descriptionProperty ? normalizeSchemaKey(descriptionProperty) : undefined,
-      linkProperty: linkProperty ? normalizeSchemaKey(linkProperty) : undefined,
+      titleProperty: titleProperty ? tryNormalizeSchemaPropertyType(titleProperty) : undefined,
```
jonathan-codaio, line 1428: So just confirming that we're omitting `idProperty` and `featuredProperties` from allowing paths? I think for consistency it would probably make people's intuition if these supported it, but it requires code changes and may break cross-doc, TypeForm and other dynamic schema situations where the featured columns could resemble paths, so I guess safer this way. @ekoleda-codaio WDYT?

testing/upload_validation.ts
```diff
@@ -933,138 +942,122 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
-          if (!(schema.titleProperty in schema.properties)) {
+      function validateProperty(
+        propertyKey: keyof typeof schema,
+        isValidSchema: (schema: Schema & ObjectSchemaProperty) => boolean,
+        invalidSchemaMessage: string,
+      ) {
+        if (schema[propertyKey]) {
+          const propertyValue =
+            typeof schema[propertyKey] === 'string'
+              ? (schema[propertyKey] as string)
```
jonathan-codaio, line 953: Hmm, surprised these casts are needed, would expect TS to properly narrow these types. It may work without the casts if you store `schema[propertyKey]` into a local var and check that instead.

testing/upload_validation.ts
```diff
@@ -933,138 +942,122 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
+            })?.[0];
+          }
+
+          if (!propertySchema) {
             context.addIssue({
               code: z.ZodIssueCode.custom,
-              path: ['titleProperty'],
-              message: `The "titleProperty" field name "${schema.titleProperty}" does not exist in the "properties" object.`,
+              path: [propertyKey],
+              message: `The "${propertyKey}" field name "${propertyValue}" does not exist in the "properties" object.`,
```
jonathan-codaio, line 972: Nit, we probably want to update these messages when we interpreted the value as a path to indicate that it was a path. 'The "titleProperty" path did not resolve property in the schema.' or something.

schema.ts
spencer-codaio: Yeah called out in the PR but I left a todo to go enable this but after starting down that path realized there was a lot more functionality required to enable that, so didn't want to muddle the waters currently

spencer-codaio: The way the pathing logic is implemented I don't think it should break those, since it will always look for an exact match first. It'll just be wonky if you try to extend it to do anything complicated (which I don't imagine people will do for dynamic schemas given there should inherently be no guarantees on what properties are present?)

testing/upload_validation.ts
spencer-codaio: ah good call

huayang-codaio (commented):
great.

ekoleda-codaio:
> I should probably add some documentation in comments but not sure where to put it.. open to ideas or if this should just be documented in the actual SDK documentation? cc @ekoleda-codaio

I think we could add a section at the bottom of the [Schemas guide](https://coda.io/packs/build/latest/guides/advanced/schemas/) that talks about property projection. That said, if you just adjust the TypeScript comments that will flow into the reference docs which will be a good start.

schema.ts
spencer-codaio: mmm yeah good call, updated

testing/upload_validation.ts
```diff
@@ -947,11 +947,13 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
-        if (schema[propertyKey]) {
+        const propertyValueRaw = schema[propertyKey];
+        if (propertyValueRaw) {
           const propertyValue =
-            typeof schema[propertyKey] === 'string'
-              ? (schema[propertyKey] as string)
-              : (schema[propertyKey] as ObjectProperty).value;
+            typeof propertyValueRaw === 'string'
+              ? propertyValueRaw
+              : (propertyValueRaw as PropertyIdentifierDetails)?.property;
```
jonathan-codaio, line 955: The type safety here could be a bit better, which will fix the need for the cast. (Trying to avoid casting except when really necessary, like when zod gives you an `any` value or something.) As the number of contributors to the SDK grows, people won't have as much context to know where in the code to proactively look to fix stuff and we'll need to rely on TS and tests to catch potential edge cases, so casting too much will mask a class of bugs.

If we declare an intermediate type like:

```typescript
type SchemaPathProperties = Pick<GenericObjectSchema, 'titleProperty' | 'linkProperty' | 'imageProperty' | 'descriptionProperty'>;
```

Then you can type `propertyKey` as `propertyKey: keyof SchemaPathProperties`, and now TS knows that `propertyValueRaw` is `PropertyIdentifier<string>` and the cast isn't necessary.

spencer-codaio: ah very good call, will add that

schema.ts
```diff
@@ -839,9 +839,36 @@ export interface Identity extends IdentityDefinition {
...
+ * This is useful for specifying a label for a property reference that uses a json path, where the
+ * label of the underlying property might not be descriptive enough at the top-level object.
+ */
+export interface PropertyIdentifierDetails {
+  label: string;
+  property: string;
+}
+
+/**
+ * An identifier for an object schema property that comprises of either an exact property match with the top-level
```
jonathan-codaio, line 853: Nit, typo in "comprises of"

schema.ts
```diff
@@ -1321,9 +1348,62 @@ function checkSchemaPropertyIsRequired<K extends string, L extends string, T ext
...
+ * and "[]" as accessing array items.
+ */
 export function normalizeSchemaKey(key: string): string {
-  // Colons cause problems in our formula handling.
-  return pascalcase(key).replace(/:/g, '_');
+  // Try splitting by . to handle json paths.
+  return (
+    key
+      .split('.')
+      // Colons cause problems in our formula handling.
```
jonathan-codaio, line 1360: Nit, should this comment move back down above line 1371

schema.ts
```diff
@@ -1321,9 +1348,62 @@ function checkSchemaPropertyIsRequired<K extends string, L extends string, T ext
...
+ */
+function normalizeSchemaPropertyIdentifier(key: PropertyIdentifier): PropertyIdentifier {
+  if (typeof key === 'string') {
+    return normalizeSchemaKey(key);
+  }
+
+  const {label, property: value} = key;
+  return {
+    property: normalizeSchemaKey(value),
+    label: normalizeSchemaKey(label),
```
jonathan-codaio, line 1388: Should this just get regular pascal casing? This isn't a path so it probably shouldn't be subject to the carveout for dots and braces right?

schema.ts
```diff
@@ -1321,9 +1348,62 @@ function checkSchemaPropertyIsRequired<K extends string, L extends string, T ext
   );
 }
 
+/**
+ * Normalizes a schema property key into PascalCase. This interprets "."s as accessing object properties
+ * and "[]" as accessing array items.
+ */
 export function normalizeSchemaKey(key: string): string {
-  // Colons cause problems in our formula handling.
-  return pascalcase(key).replace(/:/g, '_');
```
jonathan-codaio, line 1326: What happens if you were cross-docing a column called, say, "Foo [bar]", will that break as soon as this gets checked in? I think before it would get normalized to "FooBar", but now it will remain "Foo [bar]" and so I think both it will get a new column name when resyncing, and it may also just cause the model to barf in some way. Seems like we should test this first.

Assuming it breaks something, we may want to make the detection of path-like strings much more conservative here, and use a regex to only consider strings that are syntactically valid paths, without whitespace and whose brackets only contain numbers. Even still we may break people who use periods in cross-doc column names.

jonathan-codaio:
I'm getting more nervous about random dynamic schema cases with e.g. Jira, Salesforce, and TypeForm that we might break if we're normalizing schema keys wrong by detecting them as paths. I wonder if we should just do the safer and only-slightly-more-inelegant form `myKey: {path: 'foo.bar'}`.

In addition to causing annoyances if we create duplicate columns by normalizing incorrectly across subsequent syncs, since we're normalizing the idProperty with this same function, if the id property name accidentally changes as a result of this, as we've see in the past we'll have data loss.

spencer-codaio:
> I'm getting more nervous about random dynamic schema cases with e.g. Jira, Salesforce, and TypeForm that we might break if we're normalizing schema keys wrong by detecting them as paths. I wonder if we should just do the safer and only-slightly-more-inelegant form `myKey: {path: 'foo.bar'}`.
> 
> In addition to causing annoyances if we create duplicate columns by normalizing incorrectly across subsequent syncs, since we're normalizing the idProperty with this same function, if the id property name accidentally changes as a result of this, as we've see in the past we'll have data loss.

Hmm nothing should be actually changing here though. It's not changing how the underlying key is represented in the `properties` object, and if it tries to do an exact match before trying to do any pathing logic, all of the existing behavior should be kept. The phasing where the existing `...property` objects aren't implementing this yet also gives me room to test it with an actual dynamic schema table and see how it reacts to make extra sure

schema.ts
spencer-codaio: yeah let me add a bunch more tests for these edge cases. I think in the logic on the coda side it should always look for an exact object key match before it even tries to do any pathing

jonathan-codaio:
> > I'm getting more nervous about random dynamic schema cases with e.g. Jira, Salesforce, and TypeForm that we might break if we're normalizing schema keys wrong by detecting them as paths. I wonder if we should just do the safer and only-slightly-more-inelegant form `myKey: {path: 'foo.bar'}`.
> > In addition to causing annoyances if we create duplicate columns by normalizing incorrectly across subsequent syncs, since we're normalizing the idProperty with this same function, if the id property name accidentally changes as a result of this, as we've see in the past we'll have data loss.
> 
> Hmm nothing should be actually changing here though. It's not changing how the underlying key is represented in the `properties` object, and if it tries to do an exact match before trying to do any pathing logic, all of the existing behavior should be kept. The phasing where the existing `...property` objects aren't implementing this yet also gives me room to test it with an actual dynamic schema table and see how it reacts to make extra sure

Hmm I'm not sure I follow--the whole schema is getting rewritten with new normalization logic here, right?

https://github.com/coda/packs-sdk/blob/sc-path-property-projection/schema.ts#L1432

spencer-codaio:
@jonathan-codaio okay i reworked the logic a bit here so that the properties in `properties` will never be affected by the new schema pathing logic. This means that the actual normalization logic for properties will not change, so no actual properties should change from this work.

The pathing normalization logic only applies to the property references (`titleProperty`, etc.). In these property references, the normalization process optimistically tries to do an exact match with the top-level properties (to allow accessing any properties directly under the same normalization scheme) and then if no match is found, it will do the normalization according to pathing logic. This should handle doing this safely for things like `idProperty` and others that I haven't added the logic to so that it doesn't inadvertently break on dynamic schema columns with weird names

This all happens before the validation step which then will interpret whatever is found in the same way (first for a direct match and then for a pathing match)

schema.ts
jonathan-codaio: Nit ^^

schema.ts
```diff
@@ -1321,11 +1348,79 @@ function checkSchemaPropertyIsRequired<K extends string, L extends string, T ext
...
+  normalizedProperties: ObjectSchemaProperties,
+): PropertyIdentifier {
+  if (typeof key === 'string') {
+    return normalizeSchemaKeyPath(key, normalizedProperties);
+  }
+
+  const {label, property: value} = key;
+  return {
+    property: normalizeSchemaKeyPath(value, normalizedProperties),
+    label: normalizeSchemaKey(label),
```
jonathan-codaio, line 1403: Is it intentional that we want to normalize this, given that it's just a display label? Before, everything did double-duty as a schema property and a display label, but now we actually have something that's just a display label.

We may decide we want to normalize it if people are using weird casing and we want cards to look consistent. But maybe that's something we could do at render time vs losing the original info at compile time.

testing/upload_validation.ts
```diff
@@ -933,138 +942,125 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
+          if (
+            'codaType' in subtitlePropertySchema &&
+            subtitlePropertySchema.codaType &&
+            [
+              ValueHintType.ImageAttachment,
+              ValueHintType.Attachment,
+              ValueHintType.ImageReference,
+              ValueHintType.Embed,
+              ValueHintType.Scale,
+              ValueHintType.Scale,
```
jonathan-codaio, line 1053: These got duplicated, not sure if this should be removed or if the second one was supposed to be a different hint type.

testing/upload_validation.ts
```diff
@@ -933,138 +942,125 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
+            ].includes(subtitlePropertySchema.codaType)
+          ) {
             context.addIssue({
               code: z.ZodIssueCode.custom,
-              path: ['linkProperty'],
-              message: `The "linkProperty" field name "${schema.linkProperty}" must refer to a "ValueType.String" property with a "ValueHintType.Url" "codaType".`,
+              path: ['subtitleProperties', i],
+              message: `The "subtitleProperties" field name "${f}" must refer to a value that does not have a codaType corresponding to one of ImageAttachment, Attachment, ImageReference, Embed, Scale, or Scale.`,
```
jonathan-codaio, line 1059: Same dupe here in the error message.

testing/upload_validation.ts
```diff
@@ -933,138 +942,125 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
+          const subtitlePropertySchema = schema.properties[f];
+          if (
+            'codaType' in subtitlePropertySchema &&
+            subtitlePropertySchema.codaType &&
+            [
+              ValueHintType.ImageAttachment,
+              ValueHintType.Attachment,
+              ValueHintType.ImageReference,
+              ValueHintType.Embed,
+              ValueHintType.Scale,
```
jonathan-codaio, line 1052: We should convert this to an exhaustive`switch` statement over the string hint types, so anybody adding a new hint type is prompted to decide if it's appropriate for subtitle properties or not. It seems like the deciding factor is whether it's actually free text vs a url or structured value?

jonathan-codaio (commented):
Cool, yeah much safer now with the extra checks.

schema.ts
spencer-codaio: oh wait what is the typo?

jonathan-codaio: "is comprised of" is how I would read it

schema.ts
spencer-codaio: yeah ill get rid of normalization for now -- was just worried about it appearing inconsistent since we already do some smart "change from pascal case to sentence case" logic at render time

testing/upload_validation.ts
spencer-codaio: good call done

testing/upload_validation.ts
```diff
@@ -933,137 +943,141 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
+          `must refer to a "ValueType.String" property with a "ValueHintType.Url" "codaType".`,
+        );
+      };
+
+      validateTitleProperty();
+      validateLinkProperty();
+      validateImageProperty();
+      validateSnippetProperty();
+      (schema.subtitleProperties || []).forEach((f, i) => {
+        if (typeof f === 'string') {
```
jonathan-codaio, line 1034: Is this missing a validation case for when f is a PropertyIdentifierDetails?

testing/upload_validation.ts
```diff
@@ -933,137 +943,141 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
+          switch (subtitlePropertySchema.codaType) {
+            case ValueHintType.ImageAttachment:
+            case ValueHintType.Attachment:
+            case ValueHintType.ImageReference:
+            case ValueHintType.Embed:
+            case ValueHintType.Scale:
+              context.addIssue({
+                code: z.ZodIssueCode.custom,
+                path: ['subtitleProperties', i],
+                message: `The "subtitleProperties" field name "${f}" must refer to a value that does not have a codaType corresponding to one of ImageAttachment, Attachment, ImageReference, Embed, or Scale.`,
```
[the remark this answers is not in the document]
spencer-codaio: these values would render as the actual image / asset which is why disallowing those url-like hints

testing/upload_validation.ts
```diff
@@ -933,137 +943,141 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
+            case ValueHintType.Html:
+            case ValueHintType.Markdown:
+            case ValueHintType.Percent:
+            case ValueHintType.Person:
+            case ValueHintType.ProgressBar:
+            case ValueHintType.Reference:
+            case ValueHintType.Slider:
+            case ValueHintType.Toggle:
+            case ValueHintType.Time:
+            case ValueHintType.Url:
```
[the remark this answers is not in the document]
spencer-codaio: this feels fine to me compared to the others because its a text value

testing/upload_validation.ts
```diff
@@ -933,137 +943,141 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
-            path: ['subtitleProperties', i],
-            message: `The "subtitleProperties" field name "${f}" must refer to a value that does not have a codaType corresponding to one of ImageAttachment, Attachment, ImageReference, Embed, Scale, or Scale.`,
-          });
-          return;
+          switch (subtitlePropertySchema.codaType) {
+            case ValueHintType.ImageAttachment:
+            case ValueHintType.Attachment:
+            case ValueHintType.ImageReference:
+            case ValueHintType.Embed:
+            case ValueHintType.Scale:
```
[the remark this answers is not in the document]
spencer-codaio: yeah i was mostly wondering what progress and slider would look like and then fine-tune this since i haven't gotten a chance to test (and this was the existing behavior) 

toggle is another in-between where it's a bit richer than just text but also feels more useful than leaving it raw as a boolean of true or false.

I'd guess we probably don't want to start with any rich representations of data like this but wanted to show the team first

testing/upload_validation.ts
spencer-codaio: oops good catch i forgot to refactor this one. done

testing/upload_validation.ts
```diff
@@ -988,6 +990,18 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
             return;
           }
         }
+
+        const propertyValueRaw = schema[propertyKey];
+        if (propertyValueRaw) {
+          if (Array.isArray(propertyValueRaw)) {
+            propertyValueRaw.forEach(propertyIdentifier => {
+              validatePropertyIdentifier(propertyIdentifier);
```
jonathan-codaio, line 998: Sorry for another nit 😦 I think this is making the error paths for these incorrect, since the index in the array is not take into account. The old code correctly had `path: ['subtitleProperties', i],`.

`validateJsonAndAssertFails` gives you back the error objects so you can make assertions on the path and message.

jonathan-codaio (approved):

jonathan-codaio:
Test would be good for making sure the path is right for a bad subtitleProperty (I think there's already a test, just need to an assertion for the path)

PR #2235, opened 2022-09-29
make release notes required from CLI

huayang-codaio:
Enforce CLI release to come with an initial release notes to best support the breaking changes story. 

PTAL @jonathan-codaio @patrick-codaio @alan-codaio  

cc @erickoledadevrel 

```
-> % npx ts-node cli/coda.ts release test/packs/fake_v2.ts 
coda.ts release <manifestFile> [packVersion]

Set the Pack version that is installable for users. You may specify a specific
version, or omit a version to use the version currently in the manifest file.
The version must always be higher than that of any previous release.

Options:
      --version  Show version number                                   [boolean]
      --help     Show help                                             [boolean]
  -n, --notes    Notes about the contents of this Pack release
                                                             [string] [required]

Missing required argument: notes
Please provide release notes, which will be shown to Pack users to understand the release.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ## Unreleased
 
 - Added helper method `getEffectivePropertyKeysFromSchema`. This method can be used in the sync table formulas to retrieve the user manually selected property keys by `getEffectivePropertyKeysFromSchema(context.sync.schema)`.
+- Now `--notes` is a required option in `coda release` command.
 
 ## [1.1.0] - 2022-09-27
 
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -225,6 +225,7 @@ if (require.main === module) {
           string: true,
           alias: 'n',
           describe: 'Notes about the contents of this Pack release',
+          demandOption: 'Please provide release notes, which will be shown to Pack users to understand the release.',
         },
         codaApiEndpoint: {
           string: true,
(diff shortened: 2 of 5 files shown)


ekoleda-codaio:
Making them required feels like it adds unnecessary friction for those that are building for their internal team, etc. If we make release notes editable, then I think we don't need to be so forceful as to always require them.

huayang-codaio:
If we don't require it, it would likely to be always ignored. Even codans would ignore release notes. 

As a result the release history page observed by the pack user will be completely empty.

I understand that internal tool pack may not need a release note (I doubt though as the maker would have no idea what they have released, just like I have no idea what we have released for 1pp), most pack makers are for public packs and they should attach a release note. 

At the very minimum, a release with breaking changes should be forced to add notes. 

Also curious what others think.

jonathan-codaio (approved):
Yeah, I think we should try this. It's a pretty gentle requirement, at the moment we don't care what you put in them, but it at least provides a moment where you should consider writing down what actually changed. Will be great to see this for 1PP, and I agree if we didn't require it even for 1PP they wouldn't get written.

If we're really concerned that this is too much of a burden for internal packs (but even there I think users of the pack will get tremendous value from these) we could allow bypassing notes with a flag, but still keep the default behavior of prompting you.

ekoleda-codaio:
> most pack makers are for public packs

FWIW, I don't think this is true today (or at least not overwhelmingly the case). This is not a one way door so I'm fine proceeding, but I think we shouldn't forget about the significance of the internal Pack use case.

huayang-codaio:
sure. I agree that "public" isn't the correct thing here. 

What I meant to say is that even for internal packs, you'd only create a release when you are sharing the pack to someone else (otherwise you just create a build, which doesn't require a note). And in that case, we should prompt them to enter a release note so that whoever using the pack (not the pack maker) would have some information of what happened.

PR #2258, opened 2022-10-04
RFC: Rich description support for formula/table/parameter descriptions

alexd-codaio:

Adds support for rich-text within formula / table / parameter descriptions.

Initially, intend to support bold, italics, and links. Could extend in the future to new formatting as needed. These are implemented as a new structured type (`DescriptionToken` and friends). When you call `makeFormula`/`makeParameter`/`makeSyncTable`, we will take the `description` string and turn it into a `DescriptionToken[]`

**BREAKING CHANGES**

* Our internal interfaces have been updated to this model along with the upload validation. This will require us to update the `coda` repo. I suspect we'll also need to have CLI clients bump their version when the SDK bump hits prod? 
* Pack authors will author these descriptions using standard Markdown syntax. If any pack authors use Markdown control characters in their descriptions, once we release this SDK we will interpret those characters as Markdown.


api.ts
```diff
@@ -193,7 +199,7 @@ export interface SyncTableDef<
   /** See {@link SyncTableOptions.name} */
   name: string;
   /** See {@link SyncTableOptions.description} */
-  description?: string;
+  description?: DescriptionToken[];
```
jonathan-codaio, line 202: As you noted, this and the other fields will probably have to stay `string | DescriptionToken[]` forever, because existing packs will have this as `string`. Even if we disallow old packs at runtime at some point, if we are ever iterating over older pack versions in some workflow, we'd want the typing to be accurate to their contents.

testing/upload_validation.ts
```diff
@@ -498,7 +541,7 @@ const paramDefValidator = zodCompleteObject<ParamDef<any>>({
         message: 'Object parameters are not currently supported.',
       },
     ),
-  description: z.string().max(Limits.BuildingBlockDescription),
+  description: descriptionTokens,
```
jonathan-codaio, line 544: Similar throughout, have to do a union of string and this new type to support down-version SDK clients.

If you want to support that current SDK versions are passing tokens,  can add that as an add-on check within `packMetadataSchemaBySdkVersion` in this file.

api.ts
alexd-codaio: Done.

testing/upload_validation.ts
alexd-codaio: Done. Added tests as well.

alexd-codaio:
Updated with the new behavior and tests. PTAL @jonathan-codaio

jonathan-codaio (approved):

PR #2259, opened 2022-10-04
Update docs bases on vararg and array structured builder changes

ekoleda-codaio:
- Mention partial support for varargs in structured builders.
- Mention limitations reading previous parameter values in structured builders for array and vararg parameters.
- Add note to the sample code as well.
- Use the abbreviations extensions + snippets extension to add a tooltip on all mention of "structured builders".

image


docs/guides/basics/parameters/autocomplete.md
```diff
@@ -119,6 +119,13 @@ coda.makeParameter({
     },
     ```
 
+!!! warning "Manual refresh required for `StringArray` parameters"
+    When using `StringArray` parameters in the structured builders, the autocomplete values are only fetched once when the parameter UI is loaded. The autocomplete options can be manually refreshed by the user with a menu item, but this limitation doesn't make them a great fit for use cases where the options depend on the values of previous parameters. <!-- o/bug/25402 -->
```
jonathan-codaio, line 123: FYI @alexd-codaio, just making sure these are accurate.

jonathan-codaio (approved):
Do we us the term "structured builder" externally elsewhere? I guess we need a way to refer to builders, just curious if this is adopted elsewhere.

ekoleda-codaio:
> Do we us the term "structured builder" externally elsewhere? I guess we need a way to refer to builders, just curious if this is adopted elsewhere.

Not in the SDK docs previously. The Help Center uses the term "builder" fairly often, but usually in a more specific context like "filter builder". I'm conflicted about introducing a new term, but found it too award to write "action configuration and sync table settings" every time I needed to refer to it.

ekoleda-codaio:
Based on feedback from @alicia-codaio I've replaced "structured builders" with "builder UIs".

PR #2261, opened 2022-10-06
Fix silly validation checks using old property names.

jonathan-codaio:
Chris bumped into this earlier, couldn't complete a build because this validation was checking for an old field name.

PTAL @huayang-codaio @patrick-codaio CC @chris-codaio


huayang-codaio (approved):
Eh. How did we live with this for until now.

jonathan-codaio:
Yeah, pretty surprised. Guess only first-party packs are using these...

PR #2264, opened 2022-10-07
Reorganize samples to improve VSCode experience

ekoleda-codaio:
VSCode didn't know to apply `tsconfig.samples.json` to the code samples, so it showed a lot of incorrect TypeScript errors. This change moves all of the user-facing TypeScript code (examples and snippets) into a common directory, which has a nested `tsconfig.json` file. The main `tsconfig.json` file uses the `rootDirs` property to mark this as a subproject. This is applying the guidance in this [Stack Overflow answer](https://stackoverflow.com/a/56542874).

This PR introduces no changes to anything in `dist` or `documentation/generated`, so I think it should be a no-op for the Coda binary. Include paths in the markdown files were updated, as were the paths in `documentation_config.ts`.


jonathan-codaio (commented):
LGTM, one question though, is the output file of the embedded snippets in the same place? There's `coda` code for the web editor that pulls in an output JSON file that powers the side panel content, just making sure that doesn't need to change.

ekoleda-codaio:
> is the output file of the embedded snippets in the same place?

Yes, the documentation compiler script reads the data from different paths, but writes it to the same location. This PR introduces no changes to those files.

jonathan-codaio (approved):

PR #2265, opened 2022-10-07
Validation for names of parameters and sync table names

patrick-codaio:


test/upload_validation_test.ts
```diff
@@ -849,6 +862,41 @@ describe('Pack metadata Validation', () => {
         assert.isTrue(childSchema.required);
       });
 
+      it('invalid sync table name', async () => {
+        const syncTable = makeSyncTable({
+          name: 'Sync@Table',
```
patrick-codaio, line 867: spaces (only) are blocked in `makeSyncTable`, so it's less code to test a different char

test/upload_validation_test.ts
```diff
@@ -2507,6 +2555,32 @@ describe('Pack metadata Validation', () => {
         ]);
       });
 
+      it('column format names are allowed to have funky characters', async () => {
```
patrick-codaio, line 2558: Should they? I think they're just shown as a label to a user, so... maybe this is intentional?

huayang-codaio: I am also curious. @jonathan-codaio @alexd-codaio

huayang-codaio (approved):
thanks!

testing/upload_validation.ts
```diff
@@ -485,7 +494,10 @@ const variousSupportedAuthenticationValidators = Object.entries(defaultAuthentic
 const primitiveUnion = z.union([z.number(), z.string(), z.boolean(), z.date()]);
 
 const paramDefValidator = zodCompleteObject<ParamDef<any>>({
-  name: z.string().max(Limits.BuildingBlockName),
+  name: z
+    .string()
+    .max(Limits.BuildingBlockName)
+    .regex(regexFormulaName, 'Parameter names can only contain alphanumeric characters and underscores.'),
```
jonathan-codaio, line 500: I guess if we're trying to make sure the tokenizer doesn't choke on these, should use the exact same regex that the tokenizer users:

https://github.com/coda/coda/blob/main/modules/common/formula/tokens.ts#L248

(The above is copied from this file anyway, it was just the subset of the values relevant to formula names.)

test/upload_validation_test.ts
jonathan-codaio: Yeah looks like the format name doesn't matter, the formula catalog gets a formula called `Format${formulaName}` where the suffix is the _formula_ that the format references, not the format's name itself. So this seems ok, if confusingly permissive compared to everything else.

jonathan-codaio (commented):
Would add a note to the changelog that we're tightening up validation here, inevitably this will break somebody trying to upload a new version of an existing pack.

testing/upload_validation.ts
patrick-codaio: The only difference I see is that this upload_validation code doesn't allow an underscore as the first character. Am I misreading this? Is there still a tightening needed?

jonathan-codaio: Ah, are you saying we're stricter than the tokenizer is here, so it should be fine?

If so I agree but I think we should be more explicit about it. Could maybe assign `const regexParameterName = regexParameterName` and leave a note that this is currently the same as the tokenizer's restrictions except stricter because we don't allow leading underscores.

It seems like we're going to have bugs similar to the one you found if parameter and formula name validation ever diverges from what the tokenizer allows, so best to be really careful and explicit to minimize the likelihood of that happening again.

patrick-codaio: Ok, done!

PR #2267, opened 2022-10-08
Update core.autocompleteSearchObjects.md to correct inaccuracy

loucadufault:
See https://github.com/coda/packs-sdk/blob/main/api.ts#L1253-L1254

`search` query and `displayKey` prop of searched objects are both made lowercase, making the search case-insensitive (or non case-sensitive, feel free to rename).

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -1214,7 +1214,7 @@ export function simpleAutocomplete<T extends AutocompleteParameterTypes>(
  * filtering to only those that match a search string, and converting the matching
  * objects into the format needed for autocomplete results.
  *
- * A case-sensitive search is performed over each object's `displayKey` property.
+ * A case-insensitive search is performed over each object's `displayKey` property.
  *
  * A common pattern for implementing autocomplete for a formula pattern is to
  * make a request to an API endpoint that returns a list of all entities,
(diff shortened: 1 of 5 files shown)


ekoleda-codaio:
Thanks for the fix @loucadufault. The reference docs are auto-generated from the TypeScript source code, so you'll need to change the text here:

https://github.com/coda/packs-sdk/blob/63827d09dc4d6efc91d842150be8b4ff9927669d/api.ts#L1217

Then run `make bs` to setup your local env and `make build` to rebuild the docs and other artifacts. It's a bit of a pain of make small fixes to the reference docs, so happy to do it on your behalf if you'd rather.

ekoleda-codaio:
There is more information in the Contributing guide:

https://coda.io/packs/build/latest/support/contributing/#documentation-changes

loucadufault:
Thanks for the resources @ekoleda-codaio. I have made the changes and rebuilt the doc artifacts with `make docs`. 

One note is that the build failed (but still made changes to 4/5 of the files ultimately changed) with:
```
Please config your git origin to git@github.com:coda/packs-sdk.git
make: *** [typedoc] Error 1
```

when my repo was configured to point to my forked patch repo for the PR. Might be worth documenting to change the remote when cloning from a fork before building (and then change the remote back to fork url in order to push changes).

ekoleda-codaio:
> Might be worth documenting to change the remote when cloning from a fork before building (and then change the remote back to fork url in order to push changes).

Ah, good catch @loucadufault, let me see if I can simplify that.

ekoleda-codaio (approved):

PR #2270, opened 2022-10-12
[TwoWaySync] Add mutability/update properties

chris-codaio:
PTAL @coda/packs


dragonghy:
there was a thread about if we should merge the experimental changes into main or use a separate branch. so I guess we landed with the decision to merge into main?

huayang-codaio:
there was a thread about if we should merge the experimental changes into main or use a separate branch. so I guess we decided to use main branch?

api.ts
```diff
@@ -211,6 +211,8 @@ export interface SyncTableDef<
   entityName?: string;
   /** See {@link DynamicOptions.defaultAddDynamicColumns} */
   defaultAddDynamicColumns?: boolean;
+  /** See {@link SyncTableOptions.maxUpdateBatchSize} */
```
huayang-codaio, line 214: maybe add `/** @hidden */` to these new fields.

huayang-codaio (approved):

api.ts
chris-codaio: Oh, cool.  Will add that to all of the definitions.

api.ts
```diff
@@ -1406,6 +1409,12 @@ export interface SyncTableOptions<
    * sync tables that have a dynamic schema.
    */
   dynamicOptions?: DynamicOptions;
+  /**
+   * If the pack supports object updates, the maximum number of objects that will be sent to the pack
```
jonathan-codaio, line 1413: Super nit, technically it's if the _table_ supports object updates. Other stuff in the pack isn't relevant.

api.ts
```diff
@@ -211,6 +211,9 @@ export interface SyncTableDef<
   entityName?: string;
   /** See {@link DynamicOptions.defaultAddDynamicColumns} */
   defaultAddDynamicColumns?: boolean;
+  /** See {@link SyncTableOptions.maxUpdateBatchSize} */
+  /** @hidden */
+  maxUpdateBatchSize?: number;
```
jonathan-codaio, line 216: If the goal is to put `executeUpdate` next to `execute`, that would be in `SyncFormula`/`SyncFormulaDef`. If the goal is to put this field next to `executeUpdate`, then it would go there too, not here on `SyncTableDef`.

huayang-codaio: yeah I had the question as well in the first place. It does look like `maxUpdateBathSize` should be a property of the sync table but not to a particular formula so this change makes sense. 

and as it is to the `executeUpdate`, it becomes an interesting question as if makes sense for it to live in the `getter` in the longer term. and if we decided `getter` should be `formula` then it would make sense to have `maxUpdateBathSize` there as well.

jonathan-codaio: Yeah I think the getter is the right place to put this stuff, it's just named poorly at the moment. Not sure the feasibility of renaming though, but at least it's an internal field name even if the misnomer sticks around.

chris-codaio: In our discussion, we had decided to put this metadata next to other table fields, not within the formula.  Happy to change that if it makes sense, though.  Just tell me where I should put it.

```
 makeDynamicSyncTable({
    name: 'Tasks',
    identityName: 'Tasks',
    connectionRequirement: ConnectionRequirement.Required,
    listDynamicUrls: handlers.listDynamicTaskUrls,
    getName: handlers.getDynamicTaskName,
    getDisplayUrl: handlers.getDynamicTaskDisplayUrl,
    getSchema: handlers.getDynamicTaskSchema,
    maxUpdateBatchSize: 15, // defaults to 1, min 1, max 1000?
    ...
```

chris-codaio: Clarified over slack that we indeed want to move this into the formula.  Done.

PR #2271, opened 2022-10-13
Improve reference doc generation

ekoleda-codaio:
As per #2267, external contributions to the reference docs have two friction points:
- Contributors don't realize they need to change the comments in the TS definitions, not in the markdown file directly.
- Contributors have to change their git origin temporarily.

This PR attempts to improve both by:
- Adding a note at the top of every generated file.
- Removing the link to the source file when generating the checked-in documentation. The links are kept when generating the docs during the publishing flow.


jonathan-codaio (commented):
Ah interesting, so the documentation that's checked in isn't (and I guess hasn't been) what will be pushed? The publish step builds it from scratch and pushes that? So is checking in the built documentation just so that we can see diffs?

ekoleda-codaio:
> So is checking in the built documentation just so that we can see diffs?

Ya, I think checking in the generated docs serves two purposes:
- Understanding the impact of interface changes on the generated docs (though this is somewhat diminished by hiding the diffs by default in GitHub)
- Being able to preview the documentation changes in a PR using ReadTheDocs (which doesn't have the ability to run a build script).

jonathan-codaio (approved):

PR #2272, opened 2022-10-13
Explicitly export various internal interfaces

alexd-codaio:

Prepares the SDK to be migrated to an ES module. Explicitly export a number of internal properties via an `internal` namespace.


index.ts
```diff
@@ -129,9 +129,6 @@ export {getQueryParams} from './helpers/url';
 export {join as joinUrl} from './helpers/url';
 export {withQueryParams} from './helpers/url';
 
-// Schema helpers.
-export {getEffectivePropertyKeysFromSchema} from './helpers/schema';
```
jonathan-codaio, line 133: @huayang-codaio I think you meant this one to be exposed to makers, so it should probably stay here, right?

helpers/legacy_marshal/index.ts
```diff
@@ -102,7 +102,6 @@ export function legacyMarshalValue(val: any): string | undefined {
   // identical. It will only serve the purpose of our internal marshaling use case.
   return JSON.stringify(processValue(val));
 }
-
```
jonathan-codaio, line 105: Super nit, don't think this file needed to change.

internal.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export {normalizeSchema} from './schema';
+export {normalizeSchemaKey} from './schema';
+export {normalizePropertyValuePathIntoSchemaPath} from './schema';
+
+// Fetcher helpers
+export {transformBody} from './handler_templates';
+
+
+export * as legacyImports from './legacy_exports';
```
jonathan-codaio, line 41: How bad is the refactoring in the `packs` repo? Hopefully it will work there to just change the imports and not have to destructure all the legacy bits?

jonathan-codaio (commented):
Seems largely ok. Adding @ekoleda-codaio for any thoughts/concerns as well.

index.ts
huayang-codaio: yeah this intentionally stays here.

PR #2273, opened 2022-10-13
remove ensure helpers from packs-sdk exports

huayang-codaio:
[WIP]

just a quick idea while I was looking at https://github.com/coda/packs-sdk/pull/2272 

not sure if it's just me but it feels boring that in the coda repo whenever I use `ensureExists` VS will pop up an option from packs-sdk and I had to manually pick the other one..

if there's no strong objection I'll remove these helpers in the next release. 

PTAL @coda/packs @erickoledadevrel

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -135,12 +135,6 @@ export {getEffectivePropertyKeysFromSchema} from './helpers/schema';
 // SVG constants.
 export {SvgConstants} from './helpers/svg';
 
-// General Utilities
-export {assertCondition} from './helpers/ensure';
-export {ensureExists} from './helpers/ensure';
-export {ensureNonEmptyString} from './helpers/ensure';
-export {ensureUnreachable} from './helpers/ensure';
-
 // Object Schemas
 export type {ArraySchema} from './schema';
 export type {AttributionNode} from './schema';


jonathan-codaio (commented):
We use these a lot in the `packs` repo so we'd have to change all the imports there. I think the intent is they are exposed as helpers for pack makers.

huayang-codaio:
we can easily replace the usages in the packs repo. 

> I think the intent is they are exposed as helpers for pack makers.

yeah I understand that was the intention. but honestly there are many helpers we can export to the pack makers but are not necessarily related to the packs. If this is helpful enough to people I believe there will be another npm package for it.

ekoleda-codaio:
FWIW they aren't advertised in the documentation, except in the auto-generated reference docs. I don't know what the uptake is among Pack developers, but it's likely low. That said, I think we should avoid breaking changes unless required. Is there a "hidden" tag we could apply to these functions to get VSCode to not suggest them?

jonathan-codaio:
> FWIW they aren't advertised in the documentation, except in the auto-generated reference docs. I don't know what the uptake is among Pack developers, but it's likely low. That said, I think we should avoid breaking changes unless required. Is there a "hidden" tag we could apply to these functions to get VSCode to not suggest them?

Yeah, I think there's some way to demote or ban things from Intellisense, I would go that direction.

For `ensureExists` sure maybe it's a little contrived to provide that as a utility for developers, but `withQueryParams` is extremely useful, and that one will conflict on the `coda` side too, so would rather solve this in a way that doesn't constrain what we provide in the SDK.

huayang-codaio:
okay. maybe there's a way for VsCode to resolve this better. let's try that direction first.

PR #2277, opened 2022-10-13
[TwoWaySync] Add key unmapping functions

chris-codaio:
Due to the key mangling we do for results coming out of sync table syncs, we need to reverse that back to what the pack expects when sending in updates to be processed.

PTAL @coda/packs @jonathan-codaio @huayang-codaio @ekoleda-codaio


handler_templates.ts
```diff
@@ -332,6 +332,58 @@ export function transformBody(body: any, schema: Schema | undefined): any {
...
+function unmapKeys(obj: {[key: string]: any}, schema?: Schema): object {
+  if (!(schema && isObject(schema))) {
+    return obj;
+  }
+
+  const {properties} = schema;
+
+  // Look at the properties of the schema and invert any keys if present.
+  const remappedKeys: Map<string, string> = new Map();
+  for (const key in properties) {
```
jonathan-codaio, line 344: Nit, does `Object.keys(properties)` already restrict to only own-properties? If used throughout, could eliminate the need for hasOwnProperty() checks.

jonathan-codaio (commented):
LGTM, but mind adding some tests? Wish we had a baseline of tests for the existing key mapping since that always trips people up.

chris-codaio:
> Is there an expectation that the update function takes in identical input to what the output function emitted? That's not the case for every pack (e.g. Jira).

Yes, the new function to be implemented will take in input identical to what the output function emitted (subject to column selection on the Coda side).

alexd-codaio:
> > Is there an expectation that the update function takes in identical input to what the output function emitted? That's not the case for every pack (e.g. Jira).
> 
> Yes, the new function to be implemented will take in input identical to what the output function emitted (subject to column selection on the Coda side).

Maybe I'm missing something, but I don't see how / where you can reverse the transformation of the pack value that happens in the `ObjectWrapper` code where we take the schema and fully transform it?

I guess there's a plan to have an `ModelWrapperInverser` that undoes the transformations that happen through that?

handler_templates.ts
chris-codaio: Yeah, it restricts to own enumerable properties.  Probably not a bad idea, but I'd prefer to be consistent with code in the existing `mapKeys`.  Perhaps we come back and fix them both up.

chris-codaio:
Added a basic set of unit tests.  PTAL @jonathan-codaio

jonathan-codaio (approved):

PR #2278, opened 2022-10-13
[TwoWaySync] Add executeUpdate method

chris-codaio:
PTAL @jonathan-codaio @huayang-codaio @ekoleda-codaio 
CC @coda/packs


api.ts
```diff
@@ -1592,6 +1631,7 @@ export function makeSyncTable<
       ...definition,
       cacheTtlSecs: 0,
       execute,
+      executeUpdate: executeUpdate as any,
```
chris-codaio, line 1634: I could use some help on removing this `any` cast.  The `SchemaDefT` and `SchemaT` types are fighting given input vs. output types on the new method; unclear to me what the right fix is.

api.ts
```diff
@@ -736,6 +746,19 @@ export interface SyncFormulaDef<
...
+   *
+   * This function takes in parameters, updated sync table objects, and a sync context, 
+   * and is responsible for pushing those updated objects to the external system then returning
+   * the new state of each object.
+   */
+  /** @hidden */
+   executeUpdate?(
+    params: ParamValues<ParamDefsT>,
+    updates: Array<SyncUpdate<K, L, SchemaT>>,
+    context: SyncExecutionContext,
```
jonathan-codaio, line 760: We may want a dedicated type for this context. The current type mostly applies here but it has `continuation` which is confusing, but at least it's optional.

api.ts
```diff
@@ -1583,6 +1610,18 @@ export function makeSyncTable<
       continuation,
     } as SyncFormulaResult<K, L, SchemaT>;
   };
+  const executeUpdate = wrappedExecuteUpdate ? async function execUpdate(
+    params: ParamValues<ParamDefsT>,
+    updates: Array<SyncUpdate<K, L, SchemaDefT>>,
+    context: SyncExecutionContext,
+  ) {
```
jonathan-codaio, line 1617: If you add an explicit return type here, does that help with the `any`?

helpers/metadata.ts
```diff
@@ -73,7 +73,7 @@ function compileSyncTable(syncTable: GenericSyncTable): PackSyncTable {
   }
 
   const {getter, ...rest} = syncTable;
-  const {execute, ...getterRest} = getter;
+  const {execute, executeUpdate, ...getterRest} = getter;
```
jonathan-codaio, line 76: So here we're losing any notion in the metadata that this table supports updates. So we'll need to add a new bit to the compiled metadata for `supportsUpdates?: boolean;` or something and set it when `executeUpdate` exists.

Right now `getter` is typed as `PackFormulaMetadata` so we'll need some new sync-specific metadata type.

api.ts
chris-codaio: Unfortunately, no:
```
Type 'ObjectSchemaDefinitionType<K, L, SchemaT>' is not assignable to type 'Partial<{ [K in keyof SchemaWithNoFromKey<SchemaDefT>]: SchemaType<SchemaWithNoFromKey<SchemaDefT>[K]>; }>'
```

api.ts
chris-codaio: Added - did you have something like this in mind?

helpers/metadata.ts
chris-codaio: I added something, but it feels a bit clunky.  Let me know if you have a better pattern for this.

jonathan-codaio: > Probably not enough to go looking for the actual function in a bunch of places where we want to expose this in the UX.

Yeah, you literally can't find it in the UI (or really anywhere outside of lambda) otherwise, since the existence of the functions are stripped out entirely.

jonathan-codaio: Hmm what seems clunky about it? This seems direct and reasonable.

jonathan-codaio (approved):
LGTM, I'm going to play around with trying to fix the `any` cast right now though, up to you if you want to wait for me or not.

chris-codaio:
> LGTM, I'm going to play around with trying to fix the `any` cast right now though, up to you if you want to wait for me or not.

Thanks!  I really appreciate the help!

jonathan-codaio:
Sigh, this is really weird. The error in the chain right above the leaf error you pasted is this:

```
          Types of property 'previousValue' are incompatible.
            Type 'ObjectSchemaDefinitionType<K, L, SchemaT>' is not assignable to type 'ObjectSchemaDefinitionType<K, L, SchemaDefT>'.
```

It's just saying Foo is not assignable to Foo, which does not make sense. So the only thing I can think of is that maybe the generics are not aligning, and K, L, and SchemaDefT are actually referring to different things in each case. But I don't think that's actually happening, so I'm confused.

Also weird, is that if you don't try to do the wrapping and just pass through wrappedExecuteUpdate as-is, you get the same error. So it's not even the intermediate function wrapper that's causing it, it's busted from the start, which smells a little closer to a TS bug.

The only potential improvement I was able to make was to narrow the cast so instead of `any` we're casting it to the function type we actually expect. Better for type-safety, but (further) harms the readability of the sync table definition, so still a judgement call.

https://github.com/coda/packs-sdk/commit/cc5059f8f6b9c35a15c92b475830514e0dce814b

chris-codaio:
> Sigh, this is really weird. The error in the chain right above the leaf error you pasted is this:
> 
> ```
>           Types of property 'previousValue' are incompatible.
>             Type 'ObjectSchemaDefinitionType<K, L, SchemaT>' is not assignable to type 'ObjectSchemaDefinitionType<K, L, SchemaDefT>'.
> ```
> 
> It's just saying Foo is not assignable to Foo, which does not make sense. So the only thing I can think of is that maybe the generics are not aligning, and K, L, and SchemaDefT are actually referring to different things in each case. But I don't think that's actually happening, so I'm confused.
> 
> Also weird, is that if you don't try to do the wrapping and just pass through wrappedExecuteUpdate as-is, you get the same error. So it's not even the intermediate function wrapper that's causing it, it's busted from the start, which smells a little closer to a TS bug.
> 
> The only potential improvement I was able to make was to narrow the cast so instead of `any` we're casting it to the function type we actually expect. Better for type-safety, but (further) harms the readability of the sync table definition, so still a judgement call.
> 
> [cc5059f](https://github.com/coda/packs-sdk/commit/cc5059f8f6b9c35a15c92b475830514e0dce814b)

I was thinking the problem was related to the difference between `SchemaDefT` and `SchemaT` given the presence of the Identity stuff; unclear if that's really the case and how we should think about that identity piece before and after this method call.

chris-codaio:
I'll go ahead and proceed for now, but would really like to fix this at some point.  May require revisiting the duality of types.

jonathan-codaio:
> > Sigh, this is really weird. The error in the chain right above the leaf error you pasted is this:
> > ```
> >           Types of property 'previousValue' are incompatible.
> >             Type 'ObjectSchemaDefinitionType<K, L, SchemaT>' is not assignable to type 'ObjectSchemaDefinitionType<K, L, SchemaDefT>'.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It's just saying Foo is not assignable to Foo, which does not make sense. So the only thing I can think of is that maybe the generics are not aligning, and K, L, and SchemaDefT are actually referring to different things in each case. But I don't think that's actually happening, so I'm confused.
> > Also weird, is that if you don't try to do the wrapping and just pass through wrappedExecuteUpdate as-is, you get the same error. So it's not even the intermediate function wrapper that's causing it, it's busted from the start, which smells a little closer to a TS bug.
> > The only potential improvement I was able to make was to narrow the cast so instead of `any` we're casting it to the function type we actually expect. Better for type-safety, but (further) harms the readability of the sync table definition, so still a judgement call.
> > [cc5059f](https://github.com/coda/packs-sdk/commit/cc5059f8f6b9c35a15c92b475830514e0dce814b)
> 
> I was thinking the problem was related to the difference between `SchemaDefT` and `SchemaT` given the presence of the Identity stuff; unclear if that's really the case and how we should think about that identity piece before and after this method call.

Oh yep I totally missed that part of the error message. Yeah, that's the issue, and it might indicate an actual problem we'll have to confront.

PR #2282, opened 2022-10-18
Docs bug fixes

ekoleda-codaio:


jonathan-codaio (commented):
Ah, I didn't realize this was the  case for forms. If you don't pass in string values, do you get an error? Is this fields-are-strings behavior standard per an http spec?

ekoleda-codaio:
Ya, if you get a type error if you pass in non-string input:

image

We define it explicitly as string-string pairs:

https://github.com/coda/packs-sdk/blob/6f69d388b8c39aea64efca9e32a5fc36608c9957/api_types.ts#L502

The spec doesn't doesn't really have types, as I understand it, which I guess means it's all strings. I think we most likely enforce string-string pairs because that's what the proto uses, and maybe because we want developers to control the stringification rather than have us guess at it.

jonathan-codaio (approved):
Makes sense. Ultimately this is encoded as foo=bar&blah=baz so yeah it's all strings in the end.

PR #2284, opened 2022-10-19
add value export to sdk

spencer-codaio:
per https://staging.coda.io/d/Pack-Previews-links-embeds-and-objects-go-pack-previews_dp4gmVCYbYv/10-18-Property-Label-Logic_suOJU#_luDop


schema.ts
```diff
@@ -854,6 +854,8 @@ export interface Identity extends IdentityDefinition {
 export interface PropertyIdentifierDetails {
   /**
    * An optional label for the property. This will be used in locations where the label appears with the property.
+   *
+   * If set to '', the label will be omitted.
```
patrick-codaio, line 858: Or not present, right?

patrick-codaio (approved):

schema.ts
spencer-codaio: for now the default is that labels always show though we have ongoing discussions on whether that should be true

spencer-codaio: will update this if we actively change that decision

PR #2285, opened 2022-10-19
[TwoWaySync] Execution / thunk path changes

chris-codaio:
PTAL @jonathan-codaio @huayang-codaio 
CC @coda/packs

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -755,7 +755,7 @@ export interface SyncFormulaDef<
    * the new state of each object.
    */
   /** @hidden */
-   executeUpdate?(
+  executeUpdate?(
     params: ParamValues<ParamDefsT>,
     updates: Array<SyncUpdate<K, L, SchemaT>>,
     context: UpdateSyncExecutionContext,
--- a/runtime/thunk/thunk.ts
+++ b/runtime/thunk/thunk.ts
@@ -18,7 +18,9 @@ import {PostSetupType} from '../../types';
 import {StatusCodeError} from '../../api';
 import type {SyncExecutionContext} from '../../api_types';
 import type {SyncFormulaSpecification} from '../types';
+import type {SyncUpdate} from '../../api';
 import type {TypedPackFormula} from '../../api';
+import type {UpdateSyncExecutionContext} from '../../api_types';
 import {findFormula} from '../common/helpers';
 import {findSyncFormula} from '../common/helpers';
 import {isDynamicSyncTable} from '../../api';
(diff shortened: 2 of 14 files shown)


runtime/thunk/thunk.ts
```diff
@@ -43,25 +45,29 @@ export async function findAndExecutePackFunction<T extends FormulaSpecification>
   manifest: BasicPackDefinition,
   executionContext: ExecutionContext | SyncExecutionContext,
   shouldWrapError: boolean = true,
+  updates?: Array<SyncUpdate<any, any, any>>,
```
chris-codaio, line 48: Not married to this approach, but it seemed safest from a backwards/forwards compatibility point of view.

api.ts
```diff
@@ -755,7 +755,7 @@ export interface SyncFormulaDef<
    * the new state of each object.
    */
   /** @hidden */
-   executeUpdate?(
+  executeUpdate?(
```
chris-codaio, line 758: I noticed an extra space crept in here, so just cleaning that up while I'm at it.

runtime/thunk/thunk.ts
```diff
@@ -198,6 +212,13 @@ function findParentFormula(
...
+export function ensureExists<T>(value: T | null | undefined, message?: string): T {
+  if (typeof value === 'undefined' || value === null) {
+    throw new Error(message || `Expected value for ${String(value)}`);
+  }
+  return value;
+}
+
 export function ensureSwitchUnreachable(value: never): never {
   throw new Error(`Unreachable code hit with value ${String(value)}`);
 }
```
huayang-codaio, line 224: these methods probably don't need to be exported. 

I recall this method was here instead being imported to avoid dependencies in the old days. it's probably okay to simply import them now?

chris-codaio: Yeah, don't need to export, you're right.  I haven't kept up to date on what dependencies are safe w.r.t. the thunk bundle, etc.

runtime/thunk/thunk.ts
huayang-codaio: I think updating the thunk file is probably safe. the runtime files are
- pack bundle, uploaded by the maker.
- the lambda bootstrap bundle and the thunk lambda which are both injected by the upload api. these two are the ones you are changing. they should always match each other.

runtime/thunk/thunk.ts
huayang-codaio: yeah I was not sure but I saw it's importing from '../../helpers/migration' now.

huayang-codaio (approved):
lgtm overall.

chris-codaio:
> lgtm overall.

Cool! Thanks for the quick review!

runtime/thunk/thunk.ts
jonathan-codaio: The interface of `executeThunk` needs to remain backwards-compatible, but `findAndExecutePackFunction` is only called internally, so it seems like that interface can change as we want? Seems like we maybe want start cleaning it up and pass a param dict.

jonathan-codaio (commented):
Nice

chris-codaio:
I'm doing another cleanup pass around this now that I'm trying to get the end-to-end tests working.  Will come back and add the param dict since it'll be a lot cleaner.

PR #2286, opened 2022-10-20
[TwoWaySync] Tweak bootstrap for easier use

chris-codaio:
* Cleanup thunk to take a hash instead of ordered params
* Export a generic version of the update payload type

PTAL @jonathan-codaio @huayang-codaio 
CC @coda/packs

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -720,6 +720,12 @@ export interface SyncUpdate<K extends string, L extends string, SchemaT extends 
   updatedFields: string[];
 }
 
+/**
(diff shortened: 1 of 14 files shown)


jonathan-codaio (approved):
Ah great! Looks like this resolved a bug that the updates were being passed as the wrong positional arg before.

chris-codaio:
> Ah great! Looks like this resolved a bug that the updates were being passed as the wrong positional arg before.

Yes indeed.

huayang-codaio (commented):
looks better!

PR #2288, opened 2022-10-20
Bring CHANGELOG inline with spec and add linter

ekoleda-codaio:
Although [keepachangelog.com](https://keepachangelog.com) doesn't include a formal spec, tools have been built around it's recommended format ([example](https://github.com/oscarotero/keep-a-changelog)), resulting in an unofficial one. Our CHANGELOG wasn't parseable by those tools, since we didn't always follow the conventions.

This PR updates our CHANGELOG to bring it inline with the spec:
- All changes in bulleted lists.
- All changes grouped under h3's that indicate the type of change.
- All versions have a title (the part after the dash)
- Added version URLs (for versions that exist in GitHub)

Also adds a linter to make sure we stay in compliance, and adjusts the existing plugin to generate correct links.

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -11,7 +11,7 @@
       "addUnreleased": true,
       "addVersionUrl": true,
       "versionUrlFormats": {
-        "repositoryUrl": "https://github.com/coda"
+        "repositoryUrl": "https://github.com/coda/packs-sdk"
       }
     }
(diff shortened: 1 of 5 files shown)


spencer-codaio (approved):
amazing! thanks for adding this. this will be a huge help

PR #2316, opened 2022-11-02
add setup_remote_docker

huayang-codaio:
seems like we need this to run `docker run`.


jonathan-codaio (approved):
Weird, dunno why this broke

PR #2317, opened 2022-11-02
Add a bit to fetcher requests for ignoreRedirects.

jonathan-codaio:
PTAL @huayang-codaio @ekoleda-codaio @packs

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -526,6 +526,14 @@ export interface FetchRequest {
    * wish to make an unauthenticated supporting request as part of a formula implementation.
    */
   disableAuthentication?: boolean;
+  /**
+   * Whether to follow the redirects indicated by an HTTP 301 or 302 response
+   * and fetch the URL indicated by the redirect. If false, requests returning
+   * an HTTP 301 or 302 response will be returned as-is, and you may inspect the
+   * `Location` header of the response to observe the indicated redirect URL.
+   * Defaults to true if unset.
+   */
+  followRedirects?: boolean;
 }
 
 /**
--- a/api_types.ts
+++ b/api_types.ts
@@ -527,13 +527,10 @@ export interface FetchRequest {
    */
   disableAuthentication?: boolean;
   /**
-   * Whether to follow the redirects indicated by an HTTP 301 or 302 response
-   * and fetch the URL indicated by the redirect. If false, requests returning
-   * an HTTP 301 or 302 response will be returned as-is, and you may inspect the
-   * `Location` header of the response to observe the indicated redirect URL.
-   * Defaults to true if unset.
+   * If true, will immediatey return a response when encountering an HTTP 301
+   * You may inspect the `Location` header of the response to observe the indicated redirect URL.
    */
-  followRedirects?: boolean;
+  ignoreRedirects?: boolean;
 }
 
 /**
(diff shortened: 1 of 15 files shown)


api_types.ts
```diff
@@ -526,6 +526,14 @@ export interface FetchRequest {
...
    */
   disableAuthentication?: boolean;
+  /**
+   * Whether to follow the redirects indicated by an HTTP 301 or 302 response
+   * and fetch the URL indicated by the redirect. If false, requests returning
+   * an HTTP 301 or 302 response will be returned as-is, and you may inspect the
+   * `Location` header of the response to observe the indicated redirect URL.
+   * Defaults to true if unset.
+   */
+  followRedirects?: boolean;
```
jonathan-codaio, line 536: This is the basically the heart of the change.

I chose to call this `followRedirects` instead of the setting our node-fetcher uses (probably derived from request-promise) called `followRedirect` because I think it reads better and is more accurate, since there can be a chain of redirects in one request and it will follow all of them.

bundles/thunk_bundle.js
```diff
@@ -7001,7 +7010,7 @@ module.exports = (() => {
   }
   __name(handleError, "handleError");
   function handleFetcherStatusError(fetchResult, fetchRequest) {
-    if (fetchResult.status >= 300) {
+    if (fetchResult.status >= 400) {
```
jonathan-codaio, line 7013: This used to mimic request-promise behavior and throw for anything > 300. This seems undesirable, a user is likely expecting to get a response to expect and not an exception (which they can still inspect to get the same data, but seems gross). This never came up before because the SDK would never let you get 301/302 responses until now.

runtime/thunk/thunk.ts
```diff
@@ -47,10 +47,12 @@ interface FindAndExecutionPackFunctionArgs<T> {
 /**
  * The thunk entrypoint - the first code that runs inside the v8 isolate once control is passed over.
  */
-export async function findAndExecutePackFunction<T extends FormulaSpecification>(
-  {shouldWrapError = true, ...args}:
-  {shouldWrapError: boolean} & FindAndExecutionPackFunctionArgs<T>,
-): Promise<T extends SyncFormulaSpecification ? GenericSyncFormulaResult : PackFormulaResult> {
+export async function findAndExecutePackFunction<T extends FormulaSpecification>({
```
jonathan-codaio, line 50: Looks like `prettier` doing its thing.

testing/node_fetcher.ts
```diff
@@ -23,6 +21,14 @@ export interface BaseFetcherOptions {
   qs?: {[key: string]: any};
 
   followRedirect?: boolean;
+  /**
```
jonathan-codaio, line 24: This file is just copying over the latest from `coda` for the simulator, with a few minor changes to remove unneeded functionality. More comments on the `coda` PR about why these changes were necessary.

ekoleda-codaio:
SDK surface and documentation LGTM.

api_types.ts
chris-codaio, line 536: Would changing this to `ignoreRedirects` help with both the double negatives we need to deal with on the coda side as well as make the default value of `undefined` === "don't ignore redirects" be the expected default behavior?

jonathan-codaio: Sure, I like it.

huayang-codaio (approved):
lgtm. agreed that ignoreRedirect would be cleaner here.

PR #2333, opened 2022-11-11
Fixed named anchor tags

vedran:
## Motivation

I noticed that the named anchor tags on this page were not the right case, and named anchor links are case sensitive. This meant that clicking on any of these links would lead to the top of the destination page, rather than the correct spot.

## Changes

I changed the case to match the working links I found and manually tested the ones I modified.


ekoleda-codaio:
Thanks for the fix @vedran! It looks like some other stuff changed out from under you, causing the build to fail. Would you mind rebasing your fork on the latest main from this repo and see if that solves the issue?

ekoleda-codaio:
Hey @vedran - I've just gone ahead and done that myself in #2346. I'll close this out for now, but thanks again for taking the time!

PR #2344, opened 2022-11-16
Bump pip:pillow to version 9.3.0

tommyyan-codaio:
Resolve dependabot alerts:
* https://github.com/coda/packs-sdk/security/dependabot/23 ([CVE-2022-45199](https://github.com/advisories/GHSA-q4mp-jvh2-76fj))
* https://github.com/coda/packs-sdk/security/dependabot/24 ([CVE-2022-45199](https://github.com/advisories/GHSA-q4mp-jvh2-76fj))

@coda/infosec @jonathan-codaio

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,7 +13,7 @@ mkdocs-awesome-pages-plugin = "*"
 mkdocs-macros-plugin = "*"
 mkdocs-redirects = "*"
 mkdocs-simple-hooks = "*"
-pillow = "*"
+pillow = "==9.3.0"
 cairosvg = "*"
 
 [dev-packages]
--- a/Pipfile
+++ b/Pipfile
@@ -7,6 +7,8 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+cairosvg = "*"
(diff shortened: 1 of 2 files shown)


jonathan-codaio (commented):
The CI is failing with `ModuleNotFoundError: No module named 'importlib_metadata'` in an `mkdocs` import. I wonder if this means that our version of `mkdocs` is not compatible with this version of `pillow`, if e.g. a transitive dependency was removed. Might need to see if there is a newer version of `mkdocs` available?

ekoleda-codaio:
It looks like MkDocs has a direct dependency on `importlib_metadata`:

https://github.com/mkdocs/mkdocs/blob/56b235a8ad43f2300d17f87e6fa4de7a3d764397/pyproject.toml#LL43-L43C57

Not sure why it got pruned from `Pipfile.lock`.

jonathan-codaio:
> Hoping that explicitly pulling in `importlib_metadata` will fix it. Should it go under dev packages instead?

I guess putting it in the same section as mkdocs is logical, so seems fine where you have it.

PR #2348, opened 2022-11-23
Relax URL validation to accept data: URLs (not just http/https)

jtsay-codaio:
## Summary
Relax URL validation to accept `data:` URLs (not just http/https)

We already use `data:` urls in ImageAttachment fields in the Google Drive pack – this change makes it so validation passes when testing locally using `coda execute`.

## Testing
- Updated validation unit tests
- Tested this locally with `coda execute` on the google drive pack. Verified that with the change, we no longer get a URL validation error.

### DIFF
--- a/testing/validation.ts
+++ b/testing/validation.ts
@@ -219,12 +219,12 @@ function tryParseDateTimeString(result: unknown, schema: BaseStringSchema) {
 
 function tryParseUrl(result: unknown, schema: BaseStringSchema) {
   const invalidUrlError = {
-    message: `Property with codaType "${schema.codaType}" must be a valid HTTP(S) url, but got "${result}".`,
+    message: `Property with codaType "${schema.codaType}" must be a valid HTTP(S) or data url, but got "${result}".`,
   };
   try {
     const url = urlParse(result as string);
 
-    if (!(url.protocol === 'http:' || url.protocol === 'https:')) {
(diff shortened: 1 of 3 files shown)


testing/validation.ts
```diff
@@ -219,12 +219,12 @@ function tryParseDateTimeString(result: unknown, schema: BaseStringSchema) {
...
 function tryParseUrl(result: unknown, schema: BaseStringSchema) {
   const invalidUrlError = {
-    message: `Property with codaType "${schema.codaType}" must be a valid HTTP(S) url, but got "${result}".`,
+    message: `Property with codaType "${schema.codaType}" must be a valid HTTP(S) or data url, but got "${result}".`,
   };
   try {
     const url = urlParse(result as string);
 
-    if (!(url.protocol === 'http:' || url.protocol === 'https:')) {
+    if (!(url.protocol === 'http:' || url.protocol === 'https:' || url.protocol === 'data:')) {
```
jonathan-codaio, line 227: Totally unrelated, but while you're here...we dropped support for non-https urls a little while back, we might as well flag them too here. Mind removing the 'http:' check and updating the error message?

jonathan-codaio (approved):
Wow, thanks! I put it on my backlog to fix this, didn't expect you to have to do it yourself but very much appreciate the contributions here.

testing/validation.ts
jtsay-codaio: Yep, done!

PR #2349, opened 2022-11-28
Add OAuth2 config options for credentials location

ekoleda-codaio:
Our current OAuth2 token exchange logic first tries to pass the client ID and secret in the request body, and if that fails retries it in an Authorization basic header. Some APIs (Airtable for example) require the Authorization header, but invalidate the authorization code after the first failed attempt in the body.

This PR adds a new option to the OAuth2 config that allows the developer to explicitly specify which method to use. A corresponding change will need to be made to the Coda server to honor this setting.


huayang-codaio:
lgtm. let me send out a quick change reflecting this sdk option.

jonathan-codaio (approved):
👍

huayang-codaio (approved):
implemented with https://github.com/coda/coda/pull/82557

ekoleda-codaio:
> if some oauth provider adopts difference practices for exchange v.s. refresh

I haven't seen that before in the wild, so fingers crossed it never happens 🤞

testing/upload_validation.ts
```diff
@@ -424,6 +424,7 @@ const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny>
     pkceChallengeMethod: z.enum(['plain', 'S256']).optional(),
     scopeParamName: z.string().optional(),
     nestedResponseKey: z.string().optional(),
+    credentialsLocation: z.string().optional(),
```
jonathan-codaio, line 427: Oops should have caught this earlier, this should be `z.nativeEnum(TokenExchangeCredentialsLocation).optional()`

ekoleda-codaio: Fixed in https://github.com/coda/packs-sdk/pull/2354

PR #2350, opened 2022-11-29
don't import our crypto shim if crypto is supported natively

alan-codaio:
For environments using Node version >= 19.0.0, running any Coda CLI command would result in the following error:
```
TypeError: Cannot set property crypto of #<Object> which has only a getter
    at ../../../packs/node_modules/@codahq/packs-sdk/dist/testing/injections/crypto_shim.js (/private/var/folders/50/_v53nyjd0vb2r94bzn5nbcqm0000gp/T/coda-packs-4ea73dfd-3391-4b5e-82f7-c91f11182216nGo4p5/bundle.js:148:21)
    at __init (/private/var/folders/50/_v53nyjd0vb2r94bzn5nbcqm0000gp/T/coda-packs-4ea73dfd-3391-4b5e-82f7-c91f11182216nGo4p5/bundle.js:15:58)
    at ../../../../../../private/var/folders/50/_v53nyjd0vb2r94bzn5nbcqm0000gp/T/coda-packs-4ea73dfd-3391-4b5e-82f7-c91f11182216nGo4p5/browserify-bundle.js (/private/var/folders/50/_v53nyjd0vb2r94bzn5nbcqm0000gp/T/coda-packs-4ea73dfd-3391-4b5e-82f7-c91f11182216nGo4p5/bundle.js:155:7)
    at __require2 (/private/var/folders/50/_v53nyjd0vb2r94bzn5nbcqm0000gp/T/coda-packs-4ea73dfd-3391-4b5e-82f7-c91f11182216nGo4p5/bundle.js:18:52)
    at /private/var/folders/50/_v53nyjd0vb2r94bzn5nbcqm0000gp/T/coda-packs-4ea73dfd-3391-4b5e-82f7-c91f11182216nGo4p5/bundle.js:8420:10
    at Object.<anonymous> (/private/var/folders/50/_v53nyjd0vb2r94bzn5nbcqm0000gp/T/coda-packs-4ea73dfd-3391-4b5e-82f7-c91f11182216nGo4p5/bundle.js:8421:3)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
    at Module._load (node:internal/modules/cjs/loader:878:12)
```

[...]

PR #2351, opened 2022-11-30
enforce node version to be >=14 and < 19

huayang-codaio:
when we play with https://github.com/coda/packs-sdk/pull/2350, it turns out that isolated-vm doesn't support node 19 at all.

also I was worried that skipping crypto shim would make the bundle erroring out in isolated-vm. 

let's enforce the node version to be 14 and <19. Ideally we may want to ping it to 14 but I guess any version <19 works as far as we know. 

PTAL @alan-codaio @jonathan-codaio @patrick-codaio 

## Testing

```
huayang@Huayangs-MacBook-Pro [16:25:43] [~/code/packs-sdk] [hg-enforce-node-version]
-> % node -v 
v19.2.0
huayang@Huayangs-MacBook-Pro [16:27:49] [~/code/packs-sdk] [hg-enforce-node-version]
-> % make bs 
yarn config v1.22.19
warning ../package.json: No license field
success Set "cache-folder" to "/Users/huayang/.yarncache".
✨  Done in 0.05s.
yarn install v1.22.19
warning ../package.json: No license field
[1/5] 🔍  Validating package.json...
error @codahq/packs-sdk@1.2.1: The engine "node" is incompatible with this module. Expected version ">=14.17.x <19". Got "19.2.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
make[1]: *** [_bootstrap-node] Error 1
make: *** [bootstrap] Error 2
```

CLI check. 
```
huayang@Huayangs-MacBook-Pro [16:30:54] [~/code/packs-sdk] [hg-enforce-node-version *]
-> % node -v 
v19.2.0
huayang@Huayangs-MacBook-Pro [16:30:55] [~/code/packs-sdk] [hg-enforce-node-version *]
-> % npx ts-node cli/coda.ts execute test/packs/fake_v2.ts Random --no-vm
Unsupported node verison 19.2.0
```

### DIFF
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -17,8 +17,15 @@ import {handleSetOption} from './set_option';
 import {handleUpload} from './upload';
 import {handleValidate} from './validate';
 import {handleWhoami} from './whoami';
+import {printAndExit} from '../testing/helpers';
+import semver from 'semver';
 import {tryGetIvm} from '../testing/ivm_wrapper';
 import yargs from 'yargs';
+
+const version = typeof process !== 'undefined' ? process.version.substring(1) : ''; // strip out first char from v19.x.x
+if (!semver.valid(version) || semver.compare(version, '19.0.0') >= 0 || semver.compare(version, '14.17')) {
+  printAndExit(`Unsupported node verison ${version}`);
+}
 
 if (require.main === module) {
   // eslint-disable-next-line @typescript-eslint/no-unused-expressions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Added
 
 - Added an option `credentialsLocation` to `OAuth2Authentication`, to specify how the client ID and secret should be passed during the token exchange. The default `Automatic` should be sufficient for most OAuth2 providers.
+- Enforce Node version to be >=14.17 and <19 from the CLI.
 
 ## [1.2.1] - 2022-11-15
 
--- a/dist/cli/coda.js
+++ b/dist/cli/coda.js
@@ -21,8 +21,14 @@ const set_option_1 = require("./set_option");
 const upload_1 = require("./upload");
 const validate_1 = require("./validate");
 const whoami_1 = require("./whoami");
+const helpers_1 = require("../testing/helpers");
+const semver_1 = __importDefault(require("semver"));
 const ivm_wrapper_1 = require("../testing/ivm_wrapper");
 const yargs_1 = __importDefault(require("yargs"));
+const version = typeof process !== 'undefined' ? process.version.substring(1) : ''; // strip out first char from v19.x.x
+if (!semver_1.default.valid(version) || semver_1.default.compare(version, '19.0.0') >= 0 || semver_1.default.compare(version, '14.17')) {
+    (0, helpers_1.printAndExit)(`Unsupported node verison ${version}`);
+}
 if (require.main === module) {
     // eslint-disable-next-line @typescript-eslint/no-unused-expressions
     void yargs_1.default
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.2.1",
   "license": "MIT",
   "engines": {
-    "node": ">=14.17.x"
+    "node": ">=14.17.x <19"
   },
   "main": "dist/index.js",
   "bin": {
--- a/test/packs/fake_v2.ts
+++ b/test/packs/fake_v2.ts
@@ -223,3 +223,14 @@ pack.addFormula({
   },
   cacheTtlSecs: 0,
 });
+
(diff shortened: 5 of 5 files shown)


huayang-codaio:
oh wait this doesn't work. it only enforces the node version of packs-sdk but not the node version of the pack maker. 

guess we need to check the node version in the CLI.

cli/coda.ts
```diff
@@ -17,9 +17,16 @@ import {handleSetOption} from './set_option';
...
 import {handleWhoami} from './whoami';
+import {printAndExit} from '../testing/helpers';
+import semver from 'semver';
 import {tryGetIvm} from '../testing/ivm_wrapper';
 import yargs from 'yargs';
 
+const version = typeof process !== 'undefined' ? process.version.substring(1) : ''; // strip out first char from v19.x.x
+if (!semver.valid(version) || semver.compare(version, '19.0.0') >= 0 || semver.compare(version, '14.17')) {
+  printAndExit(`Unsupported node verison ${version}`);
+}
```
huayang-codaio, line 28: cc @erickoledadevrel

alan-codaio (approved):

ekoleda-codaio:
Given that `isolated-vm` is an optional dependency, should we really block the entire CLI based incompatibility with it?

jonathan-codaio:
I guess the critical question is, ignoring whether `coda execute` works or not, if you build and upload a pack using node 19, will it fail to run for real in our sandbox? I think Huayang is suggesting that it will indeed fail currently.

ekoleda-codaio:
> I think Huayang is suggesting that it will indeed fail currently.

I got the sense that we are worried it might, but haven't confirmed that. And if so, why would we block users from using Node 19 vs just fixing the underlying issue, whatever that is.

My worry is that this temporary measure might be around for some time, if there isn't a clear end state that we are waiting for.

huayang-codaio:
Okay I did more research and here's what I learned. 

The pack bundle (if using crypto) needs the crypto shim to run in isolated-vm or lambda, regardless it's build with node 19 or an older version. Otherwise it throws for missing function. So the shim needs to be applied for the bundle regardless of node version until isolated-vm supports crypto natively. 

The interesting fact of node 19 is that isolated-vm doesn't build with it. So `coda execute` will always run in the node19 environment (which has a native crypto library in node 19) instead of isolated-vm. In this execution environment, the shim should/could not be applied. 

When we build the pack bundle, the shim is applied by `esbuild` and `browserify`. The compiling context has no `process.version` and no `crypto`. So that the compiled bundle will always apply the crypto shim, regardless of the node version. 

With the observations above, it will look "fine" if we only skip applying the shim if `global.crypto?.getRandomValues` is present (which is only the case of `coda execute ... --no-vm` with node 19). For the other cases, the shim is always applied. 

The downside with this approach is that the pack maker using node 19 will always execute one bundle (w/o crypto shim) and upload a different one (with crypto shim). Ideally this should be avoided but a citizen developer may not care? idk. @erickoledadevrel what do you think?

huayang-codaio:
also I want to point out that node 19 isn't a LTS yet. The next LTS version will be node 20. 

To Eric's point of the timeline of this temporary support, I think it will be until isolated-vm is supported with the node version. 

Anyways, after the learnings, I figured it might be fine to support node 19 knowing that the subtle difference in the bundles that pack makers execute v.s upload. as long as isolated-vm is supported, this will be fixable.

ekoleda-codaio:
> The downside with this approach is that the pack maker using node 19 will always execute one bundle (w/o crypto shim) and upload a different one (with crypto shim). Ideally this should be avoided but a citizen developer may not care? idk. @erickoledadevrel what do you think?

My hunch is that now that `isolated-vm` is optional, and because it has so many dependencies, most developers are failing to install it and executing their Pack locally in Node instead of the VM without even knowing. For many reasons this can give a false sense of compatibility that will be revealed when they upload, and my guess is that the crypto shim will be minor compared to other differences between Node and the VM.

> To Eric's point of the timeline of this temporary support, I think it will be until isolated-vm is supported with the node version.

I imagine they will add support at some point, at which point the above problem will be mitigated on it's own. I think fully blocking the CLI until then because of potential differences between Node crypto and the shim would be overkill.

huayang-codaio:
cool. let's do the optional shim approach then. @alan-codaio

```
if (!global.crypto?.getRandomValues) {
  global.crypto = crypto;
}
```

ekoleda-codaio:
> cool. let's do the optional shim approach then. @alan-codaio

@huayang-codaio @alan-codaio was this ever put into place? I'm still talking to makers running into this error:

```Cannot set property crypto of #<Object> which has only a getter```

huayang-codaio:
hah, no. let me send a PR out.

huayang-codaio:
oh @alan-codaio  has that PR but it was never merged. https://github.com/coda/packs-sdk/pull/2350 better merge that one.

PR #2353, opened 2022-11-30
Validate sample Packs and fix errors

ekoleda-codaio:
Added a build target to the Make file that validates the Packs with the production server. This caught a few errors, which are also fixed in this PR. (Determined the Dropbox samples no longer need multiple domains defined.)

Added this new step to the post-commit build step, since I want it to run regularly and alert us, but don't want to block unrelated SDK PRs if new server validation has broken existing samples.


Makefile
```diff
@@ -192,6 +192,13 @@ compile-documentation-scripts:
 compile-samples:
 	${ROOTDIR}/node_modules/.bin/tsc --project ./documentation/samples/tsconfig.json
 
+.PHONY: validate-samples
+validate-samples:
```
jonathan-codaio, line 196: FWIW here's the similar command we use in the `packs` repo to run this step:

https://github.com/coda/packs/blob/main/Makefile#L123

Up to you if you think it's worth it to use `parallel` here vs just looping.

Makefile
```diff
@@ -192,6 +192,13 @@ compile-documentation-scripts:
...
+	find documentation/samples/packs -name "*.ts" | \
+	parallel -k \
+	'echo "Validating {}"; node -r ts-node/register --no-deprecation cli/coda.ts validate {}'
```
jonathan-codaio, line 199: Is it `--no-checkDeprecationWarnings`?

jonathan-codaio (commented):
Great idea.

This doesn't hit a server though, right? `coda validate` basically just calls the zod validation in upload_validation.ts. So I think it would be fine/commit to make this block merges.

Makefile
jonathan-codaio, line 199: Also running cli/coda.ts I think is kinda slow because it has to compile everything, dunno if it's possible to run the pre-built version in `dist/` for this.

ekoleda-codaio: I can't find any reference to that. But the docs do include `--no-deprecation`: https://nodejs.org/api/cli.html#--no-deprecation

ekoleda-codaio: Wow, much faster. Thanks!

Makefile
ekoleda-codaio: Removed parallel, since it's fast enough now when using the `/dist` version.

ekoleda-codaio:
> `coda validate` basically just calls the zod validation in upload_validation.ts. So I think it would be fine/commit to make this block merges.

TIL! Added to the regular build job instead.

Makefile
jonathan-codaio: Ah I see, we're doing different things here. `checkDeprecationWarnings` is a Coda-specific flag for checking SDK deprecation warnings. I guess it's good to keep checking those in sample code. (In the `packs` repo, we have deprecated syntax everywhere so it's not meaningful to check it today.) You're ignoring deprecated stuff at the `node` level, so yeah keep doing what you're doing.

jonathan-codaio (approved):

PR #2354, opened 2022-11-30
Fix credentailsLocation validation

ekoleda-codaio:
See https://github.com/coda/packs-sdk/pull/2349/files#r1036262881


jonathan-codaio (approved):
Thanks!

huayang-codaio:
good catch.

PR #2384, opened 2022-12-08
Update Todoist samples to v2

ekoleda-codaio:
Todoist has sunset their v1 API, and this PR upgrades to v2. The endpoints and objects are mostly the same, with some small tweaks. Also includes some other small changes I introduced in the cards branch.


documentation/samples/packs/todoist/formula.ts
```diff
@@ -20,17 +20,17 @@ const TaskSchema = coda.makeObjectSchema({
...
   },
   displayProperty: "name",
   idProperty: "taskId",
 });
 
-// Formula that looks up rich metadata about a task given it's URL.
+// Formula that looks up rich metadata about a task given it's ID.
 pack.addFormula({
-  name: "GetTaskById",
+  name: "Task",
```
gary-codaio, line 33: Is this okay that it's a dup formula name

gary-codaio: Ah I think I see that these samples are structured as separate files and then full is everything together, disregard

gary-codaio (approved):

documentation/samples/packs/todoist/formula.ts
ekoleda-codaio: > h I think I see that these samples are structured as separate files and then full is everything together

Yes, that's correct. The shorter ones are used to highlight specific concepts.

PR #2386, opened 2022-12-10
Just a missing word and updated error message

Mandera:
Mostly wanted to say that the guide is very intuitive to follow, nice work!


ekoleda-codaio (approved):
Thanks for the kind words and for the fixes! I didn't realize we have a nicer error message now.

PR #2391, opened 2022-12-14
Add a brand to MetadataContext so it will never fall victim to duck typing

patrick-codaio:
Context: https://codaproject.slack.com/archives/C01835XEBSS/p1670975809120289?thread_ts=1667494841.972689&cid=C01835XEBSS


api.ts
```diff
@@ -1220,12 +1220,13 @@ export function makeMetadataFormula(
...
     // regular pack functions (which this is)
     execute([search, serializedFormulaContext], context) {
-      let formulaContext = {};
+      let formulaContext: any = {};
       try {
         formulaContext = JSON.parse(serializedFormulaContext);
       } catch (err: any) {
         //  Ignore.
       }
+      formulaContext.__brand = 'MetadataContext';
```
jonathan-codaio, line 1229: So the id with brands is that they should only exist within the type system and not affect anything at runtime. Usually you'd do this with a cast.

Here I think you can just declare `let formulaContext = {} as MetadataContext;`

requirements.txt
```diff
@@ -4,7 +4,7 @@
 bracex==2.3.post1 ; python_version >= '3.7'
 cairocffi==1.4.0 ; python_version >= '3.7'
 cairosvg==2.5.2
-certifi==2022.12.7 ; python_version >= '3.6'
+certifi==2022.9.24 ; python_version >= '3.6'
```
jonathan-codaio, line 7: Hmm weird, this looks like it regressed. Would just revert this file.

jonathan-codaio (commented):
Nice. Does this work ok on the `coda` side?

huayang-codaio (commented):
does this PR only change some type definition? How would it fix the issue?

huayang-codaio:
let me read the code further.

patrick-codaio:
@huayang-codaio The actual fix will be in the Coda repo after I can use this typing update.

requirements.txt
patrick-codaio: Done

api.ts
patrick-codaio: Done, and testing in coda repo now...

huayang-codaio (approved):

patrick-codaio:
@jonathan-codaio Verified the fix in coda repo: https://github.com/coda/coda/pull/83618

jonathan-codaio (approved):

PR #2398, opened 2022-12-27
Document PKCE options

ekoleda-codaio:


docs/guides/basics/authentication/oauth2.md
```diff
@@ -171,6 +171,35 @@ Additionally, in order to prevent abuse, Coda enforces the `authorizationUrl` an
...
 
 
+## Advanced settings
+
+Depending on the OAuth provider you are connecting to, you may need to utilize these additional settings when configuring OAuth.
+
+
+### Proof Key for Code Exchange (PKCE)
+
+Some OAuth providers offer or require the use of [Proof Key for Code Exchange (PKCE)][auth0_pkce] during the OAuth flow. It is an additional flow layered on top of the standard Authorization Code flow that increases security by helping prevent against certain forms of attack. To use it in a Pack you simply need to enable the feature using the `useProofKeyForCodeExchange` field of the authentication configuration.
```
huayang-codaio, line 181: I wonder if it's better to use the RFC link directly here https://www.rfc-editor.org/rfc/rfc7636 

auth0 is just another implementation of the oauth framework imo.

huayang-codaio (approved):
Thanks!

docs/guides/basics/authentication/oauth2.md
ekoleda-codaio: Ya, good call. I really liked their description of PKCE, but it's not a very canonical source. I went with oauth.com instead, which links to the RFC but is a bit easier to read.

PR #2420, opened 2023-01-04
Avoid isolated-vm version 4.4.2.

jonathan-codaio:
Working around https://staging.coda.io/d/_d-GJF-DmEUK#Bugs_tuq-W/r26524&view=modal

There appears to be a bug in 4.4.2. For whatever reason, my installs are still using 4.4.1 so I don't run into this, but it seems like other people are getting 4.4.2 by default.

There aren't a lot of commits in https://github.com/laverdet/isolated-vm/commits/main between the versions, so it's almost certainly the case that https://github.com/laverdet/isolated-vm/commit/dba2aceae18d45106bad1ae0546a6274a38bc7d6 broke something. https://github.com/laverdet/isolated-vm/commit/905abbe0c36f6c143c3512173636443452b024be is a follow-on fix that may or may not address this issue, but there isn't a new patch version out to test with.

Regardless, let's just make sure people who depend on us don't wind up getting 4.4.2. Validated using the (TIL) https://semver.npmjs.com/

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "zod": "3.20.2"
   },
   "optionalDependencies": {
-    "isolated-vm": "^4.4.1"
+    "isolated-vm": "=4.4.1 || >4.4.2"
   },
   "devDependencies": {
     "@brightcove/kacl": "^0.1.11",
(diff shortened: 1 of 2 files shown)


dweitzman-codaio (approved):

huayang-codaio (approved):
Thanks. will keep an eye on this.

PR #2441, opened 2023-01-12
wip PR for adding matchers to FormulaMetadata

spencer-codaio:


types.ts
```diff
@@ -791,8 +791,12 @@ export interface Format {
    */
   instructions?: string;
   /**
+   * to be deprecated
+   *
    * A list of regular expressions that match URLs that the formula implementing this format
    * is capable of handling. As described in {@link Format}, this is a discovery mechanism.
+   *
+   * This will be replaced by the `matchers` property in {@link makeFormula} in a future version.
```
patrick-codaio, line 799: When pasting a URL into a table, and it matches a *formula's* regex, is a column format still required? Options:
1) Yes, a column format is required to have the table-pasting entrypoint.
2) No, we automatically create a "column format" for every formula, and the only purpose of addColumnFormat is to customize it.
3) No, but the column only becomes a normal formula column, not a pack-typed column.

I'm inclined towards #3.

testing/upload_validation.ts
```diff
@@ -1410,22 +1415,31 @@ function validateFormulas(schema: z.ZodObject<any>) {
...
+        if (hasError) {
+          context.addIssue({
+            code: z.ZodIssueCode.custom,
+            path: ['formats', i],
+            message: 'Formats can only be implemented using formulas that take exactly one required parameter.',
+          });
+        }
+
+        if ((formula.matchers?.length || 0 > 0) && (format.matchers?.length || 0 > 0)) {
+          context.addIssue({
```
patrick-codaio, line 1436: I wonder if we should also block an isAction formula from having matchers?

spencer-codaio: oh true good call

types.ts
spencer-codaio: hmm honestly i think 3 is a little too aggressive, esp becuase it doesn't apply at the cell level.

i think 1 is better to have it be explicit

api_types.ts
```diff
@@ -449,6 +449,16 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
    */
   readonly extraOAuthScopes?: string[];
+
+  /**
+   * A list of regular expressions that match URLs that the formula is capable of handling.
+   * As described in {@link makeFormula}, this is a discovery mechanism.
+   *
+   * This will replace the `matchers` property in {@link Format} in a future version.
+   */
+  readonly matchers?: string[];
```
jonathan-codaio, line 459: This should be `RegExp[]` here where the maker specifies it, matching the typing for column format matchers, right? It's only in the compiled metadata that we'd type this as `string[]`, and the compiler serializes it into a string.

@patrick-codaio did a bunch of iterations on this way back and I think after several bugs we landed at the declaration as a RegExp being the way to go.

helpers/metadata.ts
```diff
@@ -70,17 +72,19 @@ function compileSyncTable(syncTable: GenericSyncTable): PackSyncTable {
       listDynamicUrls: compileMetadataFormulaMetadata(listDynamicUrls),
       getter: {
         supportsUpdates: Boolean(executeUpdate),
+        matchers: (matchers || []).map(matcher => matcher.toString()),
```
jonathan-codaio, line 75: We probably don't have plans to the matchers here on sync tables right? When supporting that like the hackathon project it's probably more intuitive to specify them at the level of the sync table rather than the getter. Our current typing doesn't prevent somebody from providing them on the getter even though they're unused, so I can see why this line is here, but I wonder if we should just omitting the matchers from sync tables until properly supported.

jonathan-codaio (commented):
Whenever this moves out of WIP just a friendly reminder to update the changelog.

helpers/metadata.ts
patrick-codaio: Spencer and I were brainstorming about the SDK and I think it's worth having a bit more conversation about where to put matchers on sync tables. 

We thought about how if, e.g., `getSchema` were a property of the formula instead of the table, then the SDK (if not CFL, maybe?) could support dynamic schemas for all formulas. If our proposed `translateUrlIntoDefaultParams` were on the formula, that could allow URL pastes for cards & column formats to also escape the "1 required param" restriction.

We're already dealing with `description` being defined on both the sync table and the formula, and that works out ok, I guess, so maybe we can do something similar for more fields?

jonathan-codaio: Yeah for dynamic schemas for regular formulas, it's much more about the CFL and execution challenges than it is about how to model it in the SDK.

One of the reasons it's awkward to put metadata functions onto formulas is that once compiled, they aren't just functions, they are actually formula objects themselves. So if you declare a `getSchema` on your formula, the output is the return value fo `makeMetadataFormula`, which is itself a formula object of the same type as the formula itself. It creates a recursiveness that seems pretty gnarly, though it could possibly still work, and autocomplete metadata formulas are not dissimilar this.

I think the thing to optimize for would be the intuitiveness for the maker. We moved things like the identity name from nested fields to technically-redundant top-level fields because it was important for clarity. Ditto for descriptions--I wouldn't say it's been great to have them at both levels, and I think if our web of types wasn't so nasty, we probably would have typed things so that you could only specify the description at the top level of a sync table.

I think I wouldn't mind moving to a world where `Formula` represents the essential bits of an executable formula, like `parameters`, `resultType`, and `execute` that can be reused as the type for getters and metadata formulas, and `StandaloneFormula` represents an actual building block, and so adds `name`, `description`, `matchers`, etc.

types.ts
patrick-codaio: I'm not necessarily opposed to #1, but just in case you don't know - the pack magic for URL pasting in tables only works on a completely blank column - if the format of the column has been set at all (e.g., by some content being put into it), then the pack magic won't trigger. That's why #3 didn't seem aggressive to me.

api_types.ts
spencer-codaio: cool thanks i was getting confused about the two different types here

types.ts
```diff
@@ -791,8 +791,12 @@ export interface Format {
    */
   instructions?: string;
   /**
+   * to be deprecated
```
ekoleda-codaio, line 794: I think "to be deprecated" isn't very valuable, and we should omit it, or actually deprecate the field now. If using the `matchers` property works today, then we should deprecate this now (aka it's not the preferred way but it still works) and then in a future version remove it.

types.ts
spencer-codaio: actually im not really sure what you mean by #3, how does it become a normal formula column and not a pack-typed column? it needs the pack column in order to have the magic of you being able to type in a new value at each cell

and yeah even knowing the constraints on the behavior i think i prefer being explicit here

patrick-codaio: Oh, yes, I wasn't thinking through that carefully enough. I could imagine a popup promotion offering to make a new additional column that uses the URL as input, but you're totally right that we can never do auto-transformation magic in this case.

PR #2450, opened 2023-01-19
Attempt to limit/block external collaborators from opening issues

chris-codaio:
Instead, forward users to our forum, help center, and bug bounty locations where we can help them.

Not sure if this will completely work, but giving it a try.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,11 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Coda Community Forum
+    url: https://community.coda.io/c/15
+    about: Please ask and answer questions here.
+  - name: Coda Support
+    url: https://help.coda.io/
+    about: Coda's support site.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,11 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Coda Community Forum
+    url: https://community.coda.io/c/15
+    about: Please ask and answer questions here.
+  - name: Coda Support
+    url: https://help.coda.io/
+    about: Coda's support site.
```
ekoleda-codaio, line 8: Perhaps combine these into one link to the Pack SDK support page:

https://coda.io/packs/build/latest/support/

It covers both as well as additional options.

ekoleda-codaio (approved):

.github/ISSUE_TEMPLATE/config.yml
chris-codaio: Done

PR #2454, opened 2023-01-26
Improve authentication reference docs

ekoleda-codaio:
- Add inline examples and links to guides and samples for each authentication type.
- Add a new sample for `Custom` authentication (Vonage SMS).
- Fix the sample template to wrap source code in `{% raw %}` tags, to prevent placeholder strings in the code from being treated as Jinja variables.
- Hide `AWSAssumeRoleAuthentication` and `VariousAuthentication` since they aren't intended to be used by 3rd party developers.
- Improve reference docs for `UserVisibleError` and `StatusCodeError`.
- Improve `StatusCodeError` sample code to use the `StatusCodeError.isStatusCodeError()` function.


ekoleda-codaio:
Ping @alan-codaio

alan-codaio (approved):

PR #2456, opened 2023-02-05
Ensure ESLint config is being applied as expected

osg74:
### Overview
The `lint` and `lint-fix` Makefile targets don't currently seem to detect and enforce the `.eslintrc.js` config as expected.  To fix this with the current `xargs eslint` implementation we need to explicitly pass the eslint config as well.

A further improvement would be to call `eslint` directly instead -- we can then rely on the corrected `.eslintignore` file rather than piped `grep` calls for eliminating irrelevant files, e.g.:

```
${ROOTDIR}/node_modules/.bin/eslint -c .eslintrc.js --ext .ts .
```

### Verification
- The original bug was discovered when running `eslint` with the `--debug` option and the subsequent fixes were also verified this way along with explicit result output using the `--format tap` option.
- Final build results for these changes are also [green](https://app.circleci.com/pipelines/github/osg74/packs-sdk/18/workflows/25b8137e-a77b-4cc7-9f22-9acd94074aef/jobs/21/steps).

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,4 +2,6 @@
 .githooks/*
 .idea/*
 .vscode/*
+dist/*
 node_modules/*
+*.d.ts
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,3 +7,6 @@ dist
 node_modules
 *.d.ts
 *.js
+
+# This has a separate self-contained .eslintrc.js configuration
+documentation/samples
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,6 +2,8 @@
 .githooks/*
 .idea/*
 .vscode/*
+.yarn*
 dist/*
 node_modules/*
 *.d.ts
+*.js
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,8 +5,3 @@
 .yarn*
 dist
 node_modules
-*.d.ts
-*.js
-
-# This has a separate self-contained .eslintrc.js configuration
-documentation/samples
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,9 +1,9 @@
-.circleci/*
-.githooks/*
-.idea/*
-.vscode/*
+.circleci
+.githooks
+.idea
+.vscode
 .yarn*
-dist/*
-node_modules/*
+dist
+node_modules
 *.d.ts
 *.js
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,12 @@
 *yarn*
-.coda*
 .DS_Store
+.cache
+.coda*
+.devcontainer
+.npmrc
 .vscode
-node_modules
 artifacts
 dist/test
 local-docs
+node_modules
 site
-.npmrc
-.devcontainer
-.cache
--- a/.gitignore
+++ b/.gitignore
@@ -1,15 +1,12 @@
-.coda
-.coda.json
+*yarn*
+.coda*
 .DS_Store
 .vscode
 node_modules
 artifacts
 dist/test
-yarn-error.log
 local-docs
 site
-.coda-pack.json
-.coda-credentials.json
 .npmrc
 .devcontainer
 .cache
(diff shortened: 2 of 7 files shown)


osg74:
I know the Contribution guide mentions that SDK PRs are not accepted 🙈 but thought this might be a fix worth sharing anyway.

.eslintignore
```diff
@@ -1,5 +1,12 @@
...
+.idea
+.vscode
+.yarn*
+dist
+node_modules
+*.d.ts
+*.js
+
+# This has a separate self-contained .eslintrc.js configuration
+documentation/samples
```
ekoleda-codaio, line 12: There is a separate config for these files, but if you add this line then those files aren't linted at all. In the main branch, running `make lint` catches errors in those files, using the local config. This should be removed.

.gitignore
```diff
@@ -1,15 +1,12 @@
-.coda
-.coda.json
+*yarn*
```
ekoleda-codaio, line 1: Are these changes relevant to PR? It would seem tangential to update this file. Additionally, some of these wildcards seem too broad.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Changed
 
 - Updated `PrecannedDateRange` to match the date range picker in the Coda UI: added `Last90Days`, `Last180Days`, `Last365Days`, `Next90Days`, `Next180Days`, `Next365Days`, `Last7AndNext7Days`, `Last30AndNext30Days`, deprecated `Last3Months`, `Last6Months`, `Next3Months`, and `Next6Months`, and removed `ThisWeekStart`, `ThisMonthStart`, and `ThisYearStart`, which never actually worked.
+- Improved `eslint` calls in `Makefile` to ensure the associated ESLint config files are applied as expected.
```
ekoleda-codaio, line 10: We tend to only include changes in the changelog if consumers of the SDK will notice them. A change to the eslint config only affects maintainers, so I'd omit it.

Makefile
```diff
@@ -81,7 +81,7 @@ bootstrap:
 
 .PHONY: lint
 lint:
-	find . -name "*.ts" | grep -v /dist/ | grep -v /node_modules/ | grep -v "\.d\.ts" | xargs ${ROOTDIR}/node_modules/.bin/eslint
+	${ROOTDIR}/node_modules/.bin/eslint -c .eslintrc.js -f tap --ext .ts .
```
ekoleda-codaio, line 84: While including `-f tap` may be useful for testing, it changes the output of this make rule and should probably be removed. Here and below.

.eslintignore
```diff
@@ -1,5 +1,12 @@
-.circleci/*
-.githooks/*
-.idea/*
-.vscode/*
-node_modules/*
+.circleci
+.githooks
+.idea
+.vscode
+.yarn*
```
ekoleda-codaio, line 5: What is this meant to match against?

.eslintignore
```diff
@@ -1,5 +1,12 @@
...
+dist
+node_modules
+*.d.ts
```
ekoleda-codaio, line 8: Why wouldn't we want to lint these?

.eslintignore
```diff
@@ -1,5 +1,12 @@
...
+*.js
```
ekoleda-codaio, line 9: While this likely to be true going forward, I think it would be safer to just list the files to ignore, since there are a small number.

ekoleda-codaio (changes requested):
Thanks for wanting to contribute. I have noticed that VSCode gets a little angry about certain config files not matching these rules, so there certainly is a problem to be solved.

CHANGELOG.md
osg74: OK, good to know -- I'll remove this.

Makefile
osg74: This was included more for illustration, but I'll remove the explicit output for the final version.

.eslintignore
osg74: Agreed, though the original lint command in the Makefile excluded *.d.ts files in the various grep filters, hence the addition here.  I'll remove line 8 in this case.

.eslintignore
osg74: Prevents parsing of any yarn files and esp. the .yarn directory though passing an `--ext .ts` option to the `eslint` command is arguably more efficient.

.eslintignore
osg74: The linter gave some warnings about parsing .js files when it wasn't expecting them but, as above, adding `--ext .ts` seems to operate better than using an ignore statement in this case.

.eslintignore
osg74: Indeed -- I'll remove this as it was added in error along with a `root` directive in the lower level eslint config.

.gitignore
osg74: I'll revert this.

osg74:
I'll cancel this PR and rather submit a cleaned up version based on the recommendations discussed above.

PR #2460, opened 2023-02-14
Add option to pass API token as CLI argument

ekoleda-codaio:
I've had at least two developers ask for this ([1](https://community.coda.io/t/apikey-environment-variable-instead-of-coda-json-file/37612?u=eric_koleda), [2](https://community.coda.io/t/feedback-on-passing-coda-api-key-as-a-command-line-argument-when-running-coda-upload-command/32748?u=eric_koleda)), to make it easier to set up auto-deployments of their Pack in a GitHub action. It was possible to work around this be generating a `.coda.json` file on the fly, but this is a lot simpler.

### DIFF
--- a/cli/coda.ts
+++ b/cli/coda.ts
@@ -85,6 +85,11 @@ if (require.main === module) {
       command: 'clone <packIdOrUrl>',
       describe: 'Clone an existing Pack that was created using Pack Studio',
       builder: {
+        apiToken: {
+          string: true,
+          alias: 't',
+          desc: 'API token to use for the operation. Use the `register` command to define a default token.',
+        },
         codaApiEndpoint: {
           string: true,
           hidden: true,
@@ -151,11 +156,6 @@ if (require.main === module) {
           alias: 'n',
           describe: 'Notes about the contents of this Pack version',
         },
-        codaApiEndpoint: {
-          string: true,
-          hidden: true,
-          default: DEFAULT_API_ENDPOINT,
-        },
         intermediateOutputDirectory: {
           string: true,
           alias: 'o',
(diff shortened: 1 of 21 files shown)


cli/coda.ts
```diff
@@ -85,6 +85,11 @@ if (require.main === module) {
       command: 'clone <packIdOrUrl>',
       describe: 'Clone an existing Pack that was created using Pack Studio',
       builder: {
+        apiToken: {
+          string: true,
+          alias: 't',
+          desc: 'API token to use for the operation. Use the `register` command to define a default token.',
+        },
```
jonathan-codaio, line 92: Want to just split this out into a shared variable?

cli/create.ts
```diff
@@ -27,22 +28,26 @@ export async function handleCreate({
...
   // TODO(alan): we probably want to redirect them to the `coda register`
   // flow if they don't have a Coda API token.
-  const apiKey = getApiKey(codaApiEndpoint);
-  if (!apiKey) {
-    printAndExit('Missing API token. Please run `coda register` to register one.');
+  if (!apiToken) {
+    apiToken = getApiKey(codaApiEndpoint);
+    if (!apiToken) {
+      return printAndExit('Missing API token. Please run `coda register` to register one.');
+    }
```
jonathan-codaio, line 50: Seems fine if you want to move this all to a helper and then these are all just one-line calls to that.

ekoleda-codaio: What could that call look like? I could turn the `printAndExit()` call into an exception, but then I'd still need to add exception handling, which would eat away at any simplicity gains.

jonathan-codaio: I think you can just copy-paste that as-is into a helper. `printAndExit()` is just doing `process.exit(1)`, so just terminates the script immediately and doesn't need any handling.

Something like:

```typescript
function assertApiToken(codaApiEndpoint: string, cliApiToken?: string): string {
  if (cliApiToken) {
    return cliApiToken;
  }
  const apiKey = getApiKey(codaApiEndpoint);
  if (!apiKey) {
    return printAndExit('Missing API token. Please run `coda register` to register one.');
  }
  return apiKey;
}
```

ekoleda-codaio: Ah, I see now. I've gone ahead and done that, and for some other duplicated code I noticed as well.

jonathan-codaio (approved):
Awesome

PR #2461, opened 2023-02-16
Add utility to unmap keys

chris-codaio:
Needed for mapping arrays of object keys from Coda->Pack.

PTAL @coda/packs

### DIFF
--- a/handler_templates.ts
+++ b/handler_templates.ts
@@ -332,22 +332,30 @@ export function transformBody(body: any, schema: Schema | undefined): any {
   return body;
 }
 
-function unmapKeys(obj: {[key: string]: any}, schema?: Schema): object {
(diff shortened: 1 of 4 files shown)


handler_templates.ts
```diff
@@ -384,6 +392,12 @@ export function untransformBody(body: any, schema: Schema | undefined): any {
   return body;
 }
 
+export function untransformKeys(keys: string[], schema: Schema | undefined): string[] {
```
huayang-codaio, line 395: Unlike the other methods (mapKeys vs unmapKeys, transformBody vs untransformBody), it's unclear by name what this method is used for and when it should be used. Can you add some documentation to this method?

chris-codaio: Added comment plus some tests which I realized I hadn't committed.

huayang-codaio (approved):

PR #2465, opened 2023-02-18
Revert "don't import our crypto shim if crypto is supported natively"

chris-codaio:
Reverts coda/packs-sdk#2350

Seeing if this is the cause of the linter slowdown


huayang-codaio (approved):

chris-codaio:
No evidence this is the root cause - seeing the problem with and without this change.

PR #2466, opened 2023-02-18
Revert lint slowdown

chris-codaio:
Bisection seems to be pointing to https://github.com/coda/packs-sdk/pull/2414 as the place where linting started slowing down.  Let's see if that proves out.


huayang-codaio (approved):

chris-codaio:
Pulling this PR over to the `coda` repo results in a fast lint run: https://github.com/coda/packs-sdk/pull/2398
Anything after that results in a slow lint run.

chris-codaio:
I tried reverting the next PR: https://github.com/coda/packs-sdk/pull/2414
That didn't make lint fast again.

Perhaps something in 2414 and something in later PRs both tickled whatever is slowing lint down?  Does that actually make sense?

chris-codaio:
Next, I tried rolling package.json and yarn.lock both back to what was present in https://github.com/coda/packs-sdk/pull/2398.  That didn't do the trick either.  So maybe it isn't dependency related but is actually some code change that was also tickled by 2414?

PR #2469, opened 2023-02-22
Remove semver from crypto_shim

alan-codaio:
Since it was unused.


dweitzman-codaio:
I beat you to the bunch and did this in another PR to avoid bringing more code into bundles while upgrading the sdk in `coda`: https://github.com/coda/packs-sdk/pull/2468

alan-codaio:
Ah whoops, thanks for doing that ahead of me!

PR #2470, opened 2023-02-28
Add design guide

ekoleda-codaio:
[Preview](https://coda-packs-sdk--2470.com.readthedocs.build/en/2470/guides/design/)

### DIFF
--- a/docs/guides/design.md
+++ b/docs/guides/design.md
@@ -166,7 +166,7 @@ When designing your schema, select user-friendly names for your properties. The 
 
 ??? example "Example: Todoist task schema"
 
-    The Todoist API returns [up to 20 fields][todoist_task] for a task, but for most use cases only a few are required. Additionally the name "content" is replaced with "name", and the field "id" renamed to "taskId".
+    The Todoist API returns up to 20 fields for a task, but for most use cases only a few are required. Additionally the name "content" is replaced with "name", and the field "id" renamed to "taskId".
 
     === "API response"
         ```json
--- a/docs/guides/design.md
+++ b/docs/guides/design.md
@@ -21,7 +21,7 @@ Unlike other types of integrations, a Coda Pack doesn’t prescribe an exact end
 
 - Prefer parameters over hard-coding specific patterns.
   {: .yes}
-- Return [structured data][data_types], so users can to chain formulas together.
+- Return [structured data][data_types], so users can chain formulas together.
   {: .yes}
 
 <section class="tutorial-row" markdown>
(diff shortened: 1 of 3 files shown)


docs/guides/design.md
```diff
@@ -0,0 +1,281 @@
...
+
+### Design the schema
+
+Examine the data returned for each item in the collection and determine what to expose to users in Coda. Select the fields most important to users and start there. You can always add more fields later without breaking anything.
+
+When designing your schema, select user-friendly names for your properties. The field in the API may use technical terminology or refer to an older name no longer in use by the product.
+
+??? example "Example: Todoist task schema"
+
+    The Todoist API returns [up to 20 fields][todoist_task] for a task, but for most use cases only a few are required. Additionally the name "content" is replaced with "name", and the field "id" renamed to "taskId".
```
dweitzman-codaio, line 169: The `[up to 20 fields][todoist_task]` doesn't seem to render as an actual link

Screen Shot 2023-03-01 at 11 08 34 AM

dweitzman-codaio (approved):
Spotted one rendering issue, but looks good otherwise

docs/guides/design.md
ekoleda-codaio: Good catch! I removed the link, since the example response below is illustrative enough.

docs/guides/design.md
```diff
@@ -0,0 +1,281 @@
...
+
+
+## General guidance
+
+No matter what kind of Pack you are building, there are some basic rules to keep in mind.
+
+
+### Build building blocks
+
+Unlike other types of integrations, a Coda Pack doesn’t prescribe an exact end-to-end experience. Instead it provides a new set of building blocks, like formulas or buttons, that a user can deploy to improve their docs. These building blocks need to provide sufficient flexibility so that they can be combined in novel and bespoke ways.
```
jonathan-codaio, line 20: Might be worth also mentioning that users however do in fact benefit from seeing specific examples of your building blocks in action and key use cases already illustrated, and for that we have Featured Docs which are absolutely a best practice when it comes to shipping your pack.

jonathan-codaio (commented):
Great stuff as usual. I always crave this kind "but here's how we think you should really write stuff in practice" content but find it almost always absent in platform documentation.

docs/guides/design.md
ekoleda-codaio: Great callout. I added a line at the end of this section, with a link out to Joe's guide on featured docs.

jonathan-codaio (approved):
🎉

PR #2476, opened 2023-03-02
Set --frozen-lockfile for yarn install on CI

dweitzman-codaio:
To catch situations where a person (or a tool like renovate) hasn't updated `yarn.lock` before sending a PR

Tested by temporarily having a commit on this branch where package.json and yarn.lock don't match: https://app.circleci.com/pipelines/github/coda/packs-sdk?branch=dweitzman-yarn-frozen

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,11 @@ bs: bootstrap
 _bootstrap-node:
 	mkdir -p ${YARN_CACHE_DIR}
 	yarn config set cache-folder ${YARN_CACHE_DIR}
-	yarn install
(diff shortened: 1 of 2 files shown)


jonathan-codaio (approved):
Yeah, I think Nigel made this change for `coda` a while back, probably makes sense to do it everywhere for the same reasons.

dweitzman-codaio:
Yeah, this came to my attention now because the renovate PRs are failing to update `yarn.lock` but still passed CI testing

PR #2478, opened 2023-03-03
See if we can make renovate rebuild generated files

dweitzman-codaio:
I tried to test this but I don't think the test worked. It's probably simplest to just merge a PR and see if it works

I'll wait until https://github.com/coda/renovate-config/pull/9 is merged before submitting this


chris-codaio (approved):
Ooo, cool.  Looking forward to seeing if we can get this to work!

PR #2480, opened 2023-03-03
Tell renovate to avoid isolated-vm 4.4.2

dweitzman-codaio:


jonathan-codaio (approved):
Great, thanks for this!

PR #2513, opened 2023-03-04
Remove "mime" dependency which appears unused

dweitzman-codaio:
Originally added here: https://github.com/coda/packs-sdk/commit/84b08b590ace235324bb6319c0147d58c673e53c#r102979739


ekoleda-codaio (approved):
That looks like it might have been fat finger in VSCode that added it automatically. I don't see any reason to have included it. Good catch!

PR #2519, opened 2023-03-06
Upgrade python deps with "pipenv update"

dweitzman-codaio:
`requirements.txt` is actually a generated file in packs-sdk, so the renovate PRs to modify it will get undone by anyone running `make bs`

Seems like we'll need to manually update Pipfile dependencies from time to time with `pipenv update`, unless renovate figures out how to do the right thing after we hide the requirements.txt file from it?...


jonathan-codaio (approved):
Not important enough to spend time on now, but I wonder if we should just change requirements.txt to just be manually updated like one would expect, and engineers would keep it updated as needed when readthedocs changes vs generating it. I think it's a bit counterintuitive right now.

PR #2520, opened 2023-03-08
Remove yarn "resolutions"

dweitzman-codaio:
After doing this, "yarn audit" reports no dependencies with known security issues. Seems like the deps are sufficiently up-to-date that we don't need any explicit "resolutions" right now


jonathan-codaio (approved):
Nice

PR #2521, opened 2023-03-08
Migrate from yarn to pnpm

dweitzman-codaio:
Seems like pnpm is generally faster and less likely to get into a state where you have to force-delete a cache, and it has some handy features like `pnpm audit --fix` which can automatically update `package.json` to avoid any dependencies with known security issues

There's no rush to review or merge this. It's not urgent or blocking anything.


.eslintrc.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
     project: ['./tsconfig.eslint.json'],
     sourceType: 'module',
   },
-  plugins: ['ban', 'filenames', 'local', '@typescript-eslint'],
+  plugins: ['ban', 'filenames', '@kr-modules/coda-rules', '@typescript-eslint'],
```
dweitzman-codaio, line 17: pnpm makes it harder than yarn to use undeclared dependencies, so I moved the local eslint rules into a subpackage

Makefile
```diff
@@ -22,19 +18,18 @@ bs: bootstrap
...
-	  yarn install; \
+	  pnpm install; \
 	else \
-	  yarn install --frozen-lockfile; \
+	  pnpm install --frozen-lockfile; \
 	fi
-	# Install a symlink of the working directory as @codahq/packs-sdk, so that the sample code compiles.
-	yarn unlink || true # Remove any existing links providing @codahq/packs-sdk
-	yarn link  # Provide @codahq/packs-sdk from this directory
-	yarn link "@codahq/packs-sdk"  # Consume the link whenever @codahq/packs-sdk is imported
```
dweitzman-codaio, line 37: Replaced this `yarn link` dance with a subpackage as well

bundles/thunk_bundle.js
```diff
@@ -6207,7 +5802,7 @@ module.exports = (() => {
           }
           name = JSON.stringify("" + key);
           if (name.match(/^"([a-zA-Z_][a-zA-Z_0-9]*)"$/)) {
-            name = name.substr(1, name.length - 2);
+            name = name.slice(1, -1);
             name = ctx.stylize(name, "name");
           } else {
             name = name.replace(/'/g, "\\'").replace(/\\"/g, '"').replace(/(^"|"$)/g, "'");
```
dweitzman-codaio, line 5808: I can't explain all the differences here, but it's ~400 lines shorter and some of the differences I looked into seemed to be from using newer versions of dependencies

helpers/object_utils.ts
```diff
@@ -3,12 +3,9 @@ export function deepFreeze<T extends {}>(obj: T): Readonly<T> {
 
   for (const k of Object.keys(obj)) {
     const key = k as keyof T;
-    if (
-      obj[key] !== null &&
-      (typeof obj[key] === 'object' || typeof obj[key] === 'function') &&
-      !Object.isFrozen(obj[key])
-    ) {
-      deepFreeze(obj[key]);
+    const value = obj[key];
```
dweitzman-codaio, line 6: Seems like somehow this PR caused the typechecker to stop believing that `typeof obj[key] === 'object` implies it's still and object on the next line. Solved by extracting a local variable for `obj[key]` to persuade the typechecker that its type is consistent.

scripts/dev/install-pnpm.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
dweitzman-codaio, line 1: Copied from the coda repo with one change: for some reason I had to do `sudo node` instead of just `node`. No idea why (without the sudo there are permissions errors writing `pnpm` to /usr/local/bin).

testing/fetcher.ts
```diff
@@ -1,9 +1,9 @@
 import type {AWSAccessKeyCredentials} from './auth_types';
 import type {AWSAssumeRoleCredentials} from './auth_types';
-import type {Credentials as AWSCredentials} from '@aws-sdk/types';
 import {AssumeRoleCommand} from '@aws-sdk/client-sts';
 import type {Authentication} from '../types';
 import {AuthenticationType} from '../types';
+import type {AwsCredentialIdentity} from '@aws-sdk/types';
```
dweitzman-codaio, line 6: Only types are changed in this file, not logic

dweitzman-codaio:
FYI to @neal-codaio I solved the eslint-plugin-local problem differently for this repo vs what https://github.com/coda/coda/pull/85077 did. The solution here involves pnpm workspaces. I don't know if workspaces are compatible with the bazel rules.

.circleci/config.yml
```diff
@@ -42,18 +42,21 @@ commands:
   bootstrap_node:
     steps:
       - restore_cache:
-          name: Restore Cache - yarn
+          name: Restore Cache - pnpm
           keys:
-            - v2-dependencies-{{ checksum "package.json" }}
+            - pnpm-v1-dependencies-{{ checksum "package.json" }}
```
neal-codaio, line 47: If you add `&pnpm-cache-key` between the- and pnpm-... you can specify the name once and use it on L59, by doing `key: *pnpm-cache-key`.  This is done in the coda config file if you want examples.

Makefile
neal-codaio: I think this can be simplified further.  IIUC, pnpm uses `--frozen-lockfile` by default on CI.  Also I notice you already have that on L27, so I think you can replace the entire if/else here.

neal-codaio:
> FYI to @neal-codaio I solved the eslint-plugin-local problem differently for this repo vs what [coda/coda#85077](https://github.com/coda/coda/pull/85077) did. The solution here involves pnpm workspaces. I don't know if workspaces are compatible with the bazel rules.

I'm happy to do the same in the coda repo if it works and is cleaner.  This seemed fine, but I'm not sure.  I think we just got something that worked, not necessarily what is best.  I suspect we'll have to iterate on these improvements.  Thanks for doing this.  Looks like we're going in the right direction.

.gitignore
```diff
@@ -13,3 +13,4 @@ site
 .npmrc
 .devcontainer
 .cache
+yarn.lock
```
jonathan-codaio, line 16: Probably don't need/want this?

jonathan-codaio (commented):
Nice, overall this LGTM. Would want to add a CHANGELOG entry as well, since external developers occasional contribute to the repo and so the pnpm change will affect their environments.

FYI @ekoleda-codaio as well esp for the new way to link @codahq/packs-sdk back onto itself.

Given that the thunk is changing, would just want this tested with some amount of diligence by building and executing several packs, and probably on an adhoc too.

Makefile
dweitzman-codaio: I think the extra call with `--frozen-lockfile` was making me think CI detection wasn't working. Simplified!

.circleci/config.yml
dweitzman-codaio: That's definitely better. Done

.gitignore
dweitzman-codaio: Removed

scripts/dev/install-pnpm.sh
neal-codaio: `make bs` in coda repo runs with sudo already.  Perhaps it doesn't in packs-sdk repo?

dweitzman-codaio: Ah, that must be it

dweitzman-codaio:
> Given that the thunk is changing, would just want this tested with some amount of diligence by building and executing several packs, and probably on an adhoc too.

I made sync tables for a few packs and verified that they worked with the new thunk, and also when recompiled using the pnpm packs-sdk. I didn't look that deep into all the features of each pack, but the tables looked right and images were correctly synced for the gDrive pack. Seems promising enough to merge?...

dweitzman-codaio:
Added a CHANGELOG entry

jonathan-codaio (approved):
Yeah, I would feel comfortable merging.

PR #2522, opened 2023-03-08
Update final shape of two-way related params

chris-codaio:
PTAL @coda/packs @jonathan-codaio @ekoleda-codaio @huayang-codaio @dweitzman-codaio @alan-codaio


runtime/thunk/thunk.ts
```diff
@@ -86,14 +91,18 @@ function doFindAndExecutePackFunction<T extends FormulaSpecification>({
     }
     case FormulaType.Sync: {
       const formula = findSyncFormula(manifest, formulaSpec.formulaName);
-      return formula.execute(params, executionContext as SyncExecutionContext);
+      // for some reason TS can't tell that the return type is correctly GenericSyncFormulaResult.
+      return formula.execute(params, executionContext as SyncExecutionContext) as any;
```
huayang-codaio, line 95: maybe adding two method signatures instead of doing the type casting here. at least it prevents returning unexpected types. 

```
async function doFindAndExecutePackFunction<T extends FormulaSpecification>({
  params,
  formulaSpec,
  manifest,
  executionContext,
  updates,
}: FindAndExecutionPackFunctionArgs<T>): Promise<PackFunctionResponse<T>>;
async function doFindAndExecutePackFunction<T extends FormulaSpecification>({
  params,
  formulaSpec,
  manifest,
  executionContext,
  updates,
}: FindAndExecutionPackFunctionArgs<T>): Promise<
  GenericSyncFormulaResult | GenericSyncUpdateResultMarshaled | PackFormulaResult
>
```

huayang-codaio (approved):

api.ts
```diff
@@ -812,6 +809,60 @@ export interface SyncUpdateResult<K extends string, L extends string, SchemaT ex
...
+/**
+ * Type definition for a single marshaled update success result returned by a sync table update function.
+ * @hidden
+ */
+export interface SyncUpdateSingleResultMarshaledSuccess<
+  K extends string,
+  L extends string,
+  SchemaT extends ObjectSchemaDefinition<K, L>,
+> {
+  success: true;
```
dweitzman-codaio, line 831: I don't have a strong opinions either way, but something like `updateOutcome: "Success" | "Error"` could offer more wiggle room in the future than `success: boolean`, and maybe a tiny bit of defense against someone doing `if (result.success)` and not noticing that `success` was undefined instead of false.

dweitzman-codaio (approved):
Probably want to have @ekoleda-codaio double-check as well before merging

api.ts
chris-codaio: This is all internal right now (post marshaling once we leave the pack-owned code), so probably doesn't really matter since we can change it in the future as needed.

dweitzman-codaio: Even if we changed it, if it's returned by the lambda I think we'd have to support the old values unless either the SDK version is end of life or we rebuild the lambdas

(I think there's a workflow to rebuild all the lambdas, but personally I'd be afraid to run it for fear of breaking an existing pack)

api.ts
```diff
@@ -783,14 +783,11 @@ export type GenericSyncUpdate = SyncUpdate<any, any, any>;
...
-export interface SyncUpdateSingleResult<
+export type SyncUpdateSingleResult<
   K extends string,
   L extends string,
   SchemaT extends ObjectSchemaDefinition<K, L>,
-> {
-  newValue?: ObjectSchemaDefinitionType<K, L, SchemaT>;
-  error?: Error;
-}
+> = ObjectSchemaDefinitionType<K, L, SchemaT> | Error;
```
jonathan-codaio, line 790: Just for my understanding, does `ObjectSchemaDefinitionType` refer to "an object that fulfills the given schema"?

api.ts
jonathan-codaio: Yeah once this winds up in a built pack it becomes a pain to migrate. I'd prefer erring on the side of extensible shapes as well.

runtime/bootstrap/index.ts
```diff
@@ -151,7 +153,11 @@ export async function executeThunk<T extends FormulaSpecification>(
   },
   packBundlePath: string,
   packBundleSourceMapPath: string,
-): Promise<T extends SyncFormulaSpecification ? GenericSyncFormulaResult : PackFormulaResult> {
+): Promise<
+  T extends SyncUpdateFormulaSpecification ? GenericSyncUpdateResult
```
jonathan-codaio, line 157: Optional, it seems like this return type is actually the same as `PackFunctionResponse<T>` so could reuse your new named type if you want.

runtime/thunk/thunk.ts
jonathan-codaio: Yeah overloads seem like a good way to preserve the strong typing amid all the generics.

api.ts
```diff
@@ -812,6 +809,60 @@ export interface SyncUpdateResult<K extends string, L extends string, SchemaT ex
...
+  finalValue: ObjectSchemaDefinitionType<K, L, SchemaT>;
+}
+
+/**
+ * Type definition for a single marshaled update failure result returned by a sync table update function.
+ * @hidden
+ */
+export interface SyncUpdateSingleResultMarshaledError {
+  success: false;
+  error: Error;
```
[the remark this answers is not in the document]
huayang-codaio: yeah it's supposed to reinstantiate an error class on the receiver side. however I recall there's some known bugs here. not sure if this path is affected.

api.ts
chris-codaio: Yep - it's the same type we return from 1-way sync table pull requests.

runtime/bootstrap/index.ts
chris-codaio: I'll have to add some type imports in a shared location that doesn't yet contain any, but seems doable.  Done.

api.ts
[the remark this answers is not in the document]
dweitzman-codaio: If it's going through error marshaling (and it probably is?...) then errors can survive marshaling but some detail is lost along the way. For an arbitrary error type, the error type and any fields other than the string message would be lost.

We could work around that by adding more fields here in the future, though. If we wanted to associate the error with particular columns, maybe it'd look like this: `{outcome: SyncUpdateResultType.Error, error: Error, affectedColumns: string[]}`

runtime/thunk/thunk.ts
chris-codaio: Cool.

api.ts
chris-codaio: I'm not sure we need another state here in the future and I don't think the added complexity in the handling path is worth it, but I'll do it anyways.

jonathan-codaio (approved):
Nice

ekoleda-codaio (approved):

PR #2524, opened 2023-03-13
Update contributing instructions for pnpm

ekoleda-codaio:
Google Cloud shell doesn't include `pnpm` installed, so updating the instructions to including installing it manually.


dweitzman-codaio (approved):
Thanks!

PR #2525, opened 2023-03-13
Fix XML handling when `isBinaryResponse: true`

ekoleda-codaio:
This was fixed on the server in o/bug/27140, but was still broken in the testing fetcher used locally.

### DIFF
--- a/testing/fetcher.ts
+++ b/testing/fetcher.ts
@@ -138,7 +138,7 @@ export class AuthenticatingFetcher implements Fetcher {
     }
 
     try {
-      if (isXmlContentType(response.headers['content-type'])) {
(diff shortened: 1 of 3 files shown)


testing/fetcher.ts
```diff
@@ -270,7 +270,7 @@ export class AuthenticatingFetcher implements Fetcher {
     if (!this._credentials) {
       throw new Error(
         `${this._authDef.type} authentication is required for this pack, but no local credentials were found. ` +
-          'Run "coda auth path/to/pack/manifest to set up credentials."',
```
ekoleda-codaio, line 273: I think my VSCode is handling these continued lines somewhat differently, but it lints fine. Is there something I can do to bring this in line with the rest of the team?

PR #2526, opened 2023-03-13
Switch to a patch file for eslint-plugin-local so `npm install` still works

dweitzman-codaio:
The reverts the workspace parts of https://github.com/coda/packs-sdk/pull/2521 and adds a patch file instead

`npm install` is sometimes used by non-Codans who want to test or develop using unreleased SDK versions so it's helpful to support npm if we can


patches/eslint-plugin-local@1.0.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/index.js b/index.js
+index 899cbc994c0db5ec6c9e0eb960772ecc8f8f052a..c179aebcafaad4399e70b4966b6c5e10ca420820 100644
+--- a/index.js
++++ b/index.js
+@@ -1 +1,7 @@
+-module.exports = require('../../.eslintplugin')
++// Escape from pnpm subdirectory to load .eslintplugin in project root
++const path = require('path')
```
dweitzman-codaio, line 8: The other repo uses `node:path` instead of just `path` here, but that seems to break on CI for packs-sdk

jonathan-codaio (approved):

PR #2531, opened 2023-03-16
Make sure propertyAutocomplete has the same connectionRequirement as getSchema

dweitzman-codaio:
I'm not sure if anything is actually going to check this formula setting, but it seems like the right thing to do either way

Missed this in https://github.com/coda/packs-sdk/pull/2463


jonathan-codaio (approved):
Yeah, actually I think this turns out to be important. Without this setting, the formula will probably not be invoked with a connection and so won't be able to make authenticated requests.

If this was working for you before this, it's worth checking your calc service code to make sure it's respecting the connection requirement declarations. If it was always assuming there was a connection, things could break in the opposite case, when a connection is not used. (Parameter autocomplete was originally afflicted by this, it was built assuming you'd always have a connection and then had bugs when people built packs that used autocomplete but didn't require authentication.)

dweitzman-codaio:
> If this was working for you before this, it's worth checking your calc service code to make sure it's respecting the connection requirement declarations.

Seems like `_getAuthContextForFormula` should theoretically raise an error if the formula's `connectionRequirement` is inconsistent with the `InvokeMetadata` parameters

I think it wasn't raising any errors for me previously because `makePropertyAutocompleteFormula` is copied from `makeMetadataFormula` and both of them seem to basically default to setting `ConnectionRequirement.Optional` and since `maybeRewriteConnectionForFormula` wasn't being called it would just stay as `Optional`

PR #2533, opened 2023-03-16
Eliminate a dependency on the deprecated "request" package

dweitzman-codaio:
Context: https://codaproject.slack.com/archives/C02GSTENV19/p1678999880117889


jonathan-codaio (approved):
Ah that's silly that this was still there.

PR #2552, opened 2023-04-04
Install pnpm on packs-sdk bootstrap

dweitzman-codaio:
I think https://github.com/coda/coda/pull/89202 inadvertently caused trouble for people trying to use pnpm in packs-sdk

One side effect of this PR is that `make bs` requires `sudo` now

I'm not entirely convinced this is the right or best approach. Seems like there's still some weirdness here which `pnpm` exists in my homebrew path

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ _bootstrap-install-pnpm:
 
 .PHONY: _bootstrap-node
 _bootstrap-node:
-	pnpm install
+	npx pnpm install
 
 .PHONY: _bootstrap-python
 _bootstrap-python:
@@ -61,6 +61,7 @@ _bootstrap-githooks: clean-githooks
 
 .PHONY: bootstrap
 bootstrap:
+	$(MAKE) MAKEFLAGS= _bootstrap-install-pnpm
 	$(MAKE) MAKEFLAGS= _bootstrap-node
 	$(MAKE) MAKEFLAGS= _bootstrap-system-packages
 	$(MAKE) MAKEFLAGS= _bootstrap-python
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: 5.4
 
 patchedDependencies:
   eslint-plugin-local@1.0.0:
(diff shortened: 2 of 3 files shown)


pnpm-lock.yaml
dweitzman-codaio, line 1: Renovate upgraded this to a pnpm v8 lockfile, but here we're installing pnpm v7

Makefile
```diff
@@ -61,6 +61,7 @@ _bootstrap-githooks: clean-githooks
 
 .PHONY: bootstrap
 bootstrap:
+	$(MAKE) MAKEFLAGS= _bootstrap-install-pnpm
```
neal-codaio, line 64: Is npx available if we haven't installed node yet?  Seems like it wouldn't be.

pnpm-lock.yaml
neal-codaio: Should we do this in a separate PR?

Makefile
dweitzman-codaio: Good question. I think `make bs` in the sdk repo doesn't actually install node and assumes you've installed it already

pnpm-lock.yaml
dweitzman-codaio: Yeah, we should. I'm on my way to the airport but will try to get this updated sometime today.

If anyone out there is blocked on this, you can probably use `brew install pnpm` as a temporary workaround but you may need to rerun it if `make bs` in another repo uninstalls pnpm in the meantime

jonathan-codaio (commented):
Seems reasonable to me.

What's the background on us seeming to prefer pnpm 7? Do we care enough to do that here too or just leave that for the `coda` repo?

neal-codaio:
> Seems reasonable to me.
> 
> What's the background on us seeming to prefer pnpm 7? Do we care enough to do that here too or just leave that for the `coda` repo?

We can and should upgrade to pnpm 8. However, they changed the lockfile format.  Originally the bazel rules didn't support the new format.  That has been fixed.  AFAIK, we can upgrade to pnpm 8 whenever we chose.  We just need to ensure that we change all the places that it's referenced.  I think that's these [two files](https://github.com/search?q=repo%3Acoda%2Fcoda+pnpm%407&type=code) + the pnpm-lock.yml file.  If a PR that includes those 3 changes works and integration tests pass, we can merge that PR.  Then folks *should* just have to `make bs` after rebasing with this change.  Though more testing would be nice (not sure how practical though).

dweitzman-codaio:
> Seems reasonable to me.
> 
> What's the background on us seeming to prefer pnpm 7? Do we care enough to do that here too or just leave that for the `coda` repo?

I think the coda repo is still using 7 because of an issue with the aspect bazel rules and since this is global I think we'll need to match the coda repo version

Whenever the coda repo is ready to upgrade, we can upgrade here too

jonathan-codaio:
Ah so if the issue is bazel specifically, seems like packs-sdk can jump right to pnpm 8 since we don't care about bazel in this repo (unless the issue is transitive when pulled into `coda`?)

dweitzman-codaio:
> Ah so if the issue is bazel specifically, seems like packs-sdk can jump right to pnpm 8 since we don't care about bazel in this repo (unless the issue is transitive when pulled into `coda`?)

The issue is not that it's transitive, but both repos do the equivalent of `sudo node - add --global pnpm@7` to install pnpm into the _global_ node package folder

I briefly tried installing into the local node modules instead of global but noticed two potential issues:
- If someone forgets to run `source ./setup-env.sh` they could easily accidentally run the wrong version
- Seems like node adds files under `./node_modules` using a different file structure than pnpm, so then when you try to run `pnpn install` afterwards it can fail saying that `node_modules/` has an unsupported layout

dweitzman-codaio:
I'm going to wait until https://github.com/coda/coda/pull/90015 is merged in the coda repo to upgrade pnpm to version 8 there and then do make this PR also use version 8

The reason to switch to 8 now is that renovate PRs are already using it can it causes lockfile flapping / messy-looking diffs in PRs if people are switching back and forth between pnpm 7 and 8 lockfiles

Makefile
```diff
@@ -24,7 +24,7 @@ _bootstrap-install-pnpm:
 
 .PHONY: _bootstrap-node
 _bootstrap-node:
-	pnpm install
+	npx pnpm install
```
neal-codaio, line 27: Do you want to force this to v8?

neal-codaio (approved):
I'm fine with this.  While this probably isn't the best approach, it's probably better than the current state.  I think the longer term would be if we can use pnpm via bazel.  That's likely more work, so I think just making it more consistent is a better thing until we can get everything else working.  I've approved the PR in the coda repo.

dweitzman-codaio:
Closing since pnpm install on bootstrap was handled in https://github.com/coda/packs-sdk/pull/2629

PR #2554, opened 2023-04-06
Add an optional searchDynamicUrls metadata formula to dynamic sync tables.

jonathan-codaio:
The use case is where you have a non-trivial number of results from listDynamicUrls so it's impractical to navigate to the option you want directly.

We can't just implement naive text search over the listDynamicUrls results, because those are hierarchical, and we only fetch child results when you expand a single parent, so we would not be searching over all the results. Deeply nested search is what users actually want--to go to the final leaf result directly rather than just filtering among parents.

Originally I wanted to see if I could pass the search query as a parameter to listDynamicUrls, so that pack makers would only have one function to implement since these will make similar API calls. But the `search` parameter for listDynamicUrls is already claimed by the `parentUrl`. And trying to stick the query somewhere else, especially when the main single param to a metadata formula is already called `search` seemed just too confusing and hacky.

And there's a significant advantage to having a dedicated metadata formula for this, which is that we can simply check for the existence of this formula to determine whether we should render a search box in the dynamic url list UI. Not all packs will have this, it's up to the pack maker to implement this, largely dependent on the API being used having some kind of search endpoint.

I left the return value interface the same as for listDynamicUrls. You return an array of results, and nested results are allowed, I don't see any reason to exclude them or invent a new interface, even though non-leaf results would be an uncommon use case.

This PR should amount to a copy-paste-rename of everywhere that listDynamicUrls appears.

PTAL @ekoleda-codaio @huayang-codaio @chris-codaio @coda/packs


huayang-codaio (approved):
Indeed. Took me a while to figure out that `search` is already used by `parentUrl`. 

If people come up with the request to search for urls under a specific `parentUrl`, are we on the same page to deny the request?

Code changes look good to me.

jonathan-codaio:
> If people come up with the request to search for urls under a specific parentUrl, are we on the same page to deny the request?

Yeah it's I guess the main question if this simple SDK shape is sufficient. I was imaging that yeah, we would just say that searches are global and not in the context of a parentUrl. I think searching is kind of a shortcut way of not even dealing with the hierarchy at all. And the UI would be unlovely if there were search boxes nested within the hierarchy.

chris-codaio:
I believe this approach will work well for the SFDC pack which desperately needs this.

That said, it may be worth considering a way to merge a hierarchy position with the search by passing in an optional parentId to the search function so search can scope if the pack finds this necessary?  There are plenty of APIs out there that don't support universal object search.  Unclear how we'd thread the needle in the UX to make this work.

jonathan-codaio:
> I believe this approach will work well for the SFDC pack which desperately needs this.
> 
> That said, it may be worth considering a way to merge a hierarchy position with the search by passing in an optional parentId to the search function so search can scope if the pack finds this necessary? There are plenty of APIs out there that don't support universal object search. Unclear how we'd thread the needle in the UX to make this work.

Yeah, for that case I was thinking that we could shove `parentUrl` into the metadata context that would normally have the sync table params. I wouldn't suggest that we do this to start, but if needed we at least have a mechanism for it.

I think it's not too gross to have `search` be the actual query string, and the `context` object be one or more other random inputs that we jam in but do our best to carefully document. I didn't want to put the search query in the metadata context for listDynamicUrls though because it just felt very backwards when the main parameter is already called `search`.

jonathan-codaio:
I'm going to wait for @ekoleda-codaio to be back and opine just so we're all happy with the SDK shape before proceeding.

ekoleda-codaio (approved):
LGTM! I think a lot of Pack will benefit from this feature.

PR #2556, opened 2023-04-10
Try bumping patch version.

jonathan-codaio:
When I try to bump to this version in `coda`, `modules/server/base/test/static_error_pages_test.ts` fails with:

```
Error: global leak(s) detected: '__esDecorate', '__runInitializers', '__propKey', '__setFunctionName'
    at Runner.checkGlobals (node_modules/.aspect_rules_js/mocha@10.2.0/node_modules/mocha/lib/runner.js:398:21)
    at Runner.<anonymous> (node_modules/.aspect_rules_js/mocha@10.2.0/node_modules/mocha/lib/runner.js:165:12)
    at Runner.emit (node:events:525:35)
    at node_modules/.aspect_rules_js/mocha@10.2.0/node_modules/mocha/lib/runner.js:822:14
    at done (node_modules/.aspect_rules_js/mocha@10.2.0/node_modules/mocha/lib/runnable.js:310:5)
    at node_modules/.aspect_rules_js/mocha@10.2.0/node_modules/mocha/lib/runnable.js:371:11
```

I think it's now succeeding once I bump the semantic version. David you had some mysterious problem where you couldn't make `coda` happy without bumping the semantic version, was it the same as this?

Will confirm that this alone fixes things on the CI before merging.

PTAL @dweitzman-codaio @huayang-codaio @coda/packs

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.3.1",
+  "version": "1.3.2",
   "license": "MIT",
   "workspaces": [
     "dev/eslint"


dweitzman-codaio:
The problem I solved by bumping the version was the full lint time doubling, but in that case there were no failing tests

dweitzman-codaio:
> Also as a note: I think @ekoleda-codaio observed previously that the release scripts get upset if we skip patch versions, so it may turn out that we need to release 1.3.1 rather than skipping it before we can release 1.3.2

Or another option when the time comes it just to do a full minor release instead of a patch release

jonathan-codaio:
It fixes the issue 🤷 

Yeah that's super annoying about the missing semantic versions in releases. Going to minor version I guess is ok, but it does obfuscate whether a given release was just minor fixes.

PR #2558, opened 2023-04-10
Fix breaking changes in MkDocs Material 9

ekoleda-codaio:
In #2519 the MkDocs Material library was upgraded to a new major version, which introduced some [breaking changes](https://github.com/squidfunk/mkdocs-material/issues/4714). Namely, the click to copy button on code blocks and the edit source buttons are now opt-in. 

This PR re-enables those features, as well as takes advantage of the new view source button. Previously the edit button was being used to view the source, so the `edit_uri` has been fixed to that it opens the true edit URL and the new view source button has been enabled.

### Before
Screen Shot 2023-04-10 at 6 12 05 PM

### After
Screen Shot 2023-04-10 at 6 12 13 PM


jonathan-codaio (approved):
Oh nice! I was looking to copy an example the other day and was wondering where the button went.

PR #2560, opened 2023-04-12
Add 'placeholder' property to PropertyIdentifierDetails

cqian-codaio:
Adds a placeholder value for properties that we can use for cards when there is no value for that property.

Task: https://staging.coda.io/d/_dL5PQJ5pk51#All-Tasks_tu43U/r1292&view=modal

PTAL @gary-codaio @patrick-codaio @jonathan-codaio cc @erickoledadevrel


schema.ts
```diff
@@ -870,6 +870,10 @@ export interface PropertyIdentifierDetails {
    * If set to '', the label will be omitted.
    */
   label?: string;
+  /**
+   * An optional placeholder value, in case the property value is null or undefined.
```
jonathan-codaio, line 874: Would probably just be really explicit in the comment, e.g. "which will be rendered when the property value is null or undefined."

And then just to be clear, if the property value is empty/falsey, will the placeholder be used or not?

jonathan-codaio (approved):

schema.ts
cqian-codaio: Yeah that's tricky. I feel if we get one of these cases:
- undefined
- null
- []
- ""
Then it should show the placeholder.

jonathan-codaio: Yeah, I leave it up to you all to figure out what makes the most sense. Would just be very explicit in the documentation about which cases apply here.

ekoleda-codaio: I think you should clarify when this placeholder is used. Only when the label is also shown (subtitles) or anytime the value is shown (card title, snippet, etc)? Is the placeholder value returned when the property is accessed via CFL?

patrick-codaio (commented):
LGTM beyond Eric's comments

schema.ts
cqian-codaio: Yeah I think we can do card title, subtitle, and snippet

cqian-codaio: @erickoledadevrel updated this part

ekoleda-codaio: Looks good!

PR #2566, opened 2023-04-18
Compile-time checks for exhaustiveness on PropertyIdentifierDetails.

jonathan-codaio:
Trying to make it so we can catch this kind of thing proactively.

PTAL @cqian-codaio @patrick-codaio @coda/packs 

CC @oleg-codaio in case you know how to make this go all the way to a `ensureEmptyObject` generic helper, I couldn't quite get it that last mile, but this was at least a one-liner. I tried using `IsEmptyObject` borrowed from `coda` but couldn't connect that properly to a generic assertion helper.


globals.d.ts
```diff
@@ -0,0 +1,8 @@
+interface ObjectConstructor {
```
jonathan-codaio, line 1: Copied this from `coda`. Makes it so that `Object.keys()` is strongly typed, otherwise it's just `string[]` which is not helpful.

jonathan-codaio:
Changed this to a simpler approach that doesn't require adding to globals.d.ts. While the core typescript compiled fine with that, I had trouble getting dts-bundle-generator and the documentation compilation step to recognize those globals.

patrick-codaio (approved):

PR #2567, opened 2023-04-19
[CrossDoc] Add support for a fixed id on properties to support column name changes

chris-codaio:


schema.ts
```diff
@@ -242,6 +242,12 @@ interface BaseSchema {
   /** @hidden */
   mutable?: boolean;
 
+  /**
+   * Optional fixed id for this property, used to support renames of properties over time. If specified,
+   * changes to the name of this property will not cause the property to be treated as a new property.
+   */
+  fixedId?: string;
```
jonathan-codaio, line 249: This, `mutable`, and `autocomplete` only apply to top-level properties, not nested properties. I had David ensure that the documentation for `autocomplete` was clear on that, we should probably do the same for `fixedId` and `mutable`.

schema.ts
```diff
@@ -242,6 +242,12 @@ interface BaseSchema {
   /** @hidden */
   mutable?: boolean;
 
+  /**
```
dweitzman-codaio, line 245: Can you also add a `// TODO(dweitzman): Use fixedId for the autocomplete name when available to support property renames.` above the `replaceInlineAutocompleteFunctionsWithNamedAutocompleteFunctions()` function definition in api.ts?

schema.ts
dweitzman-codaio, line 249: If you introduced a `fixedId` that's equal to the original property name, would that allow you to introduce a `fixedId` on a column that never previously had one without breaking anything?

I'm thinking about scenarios where it's impossible to introduce a new `fixedId` in a backwards-compatible way and/or it's impossible to both introduce one and rename a column without doing multiple pack releases spread out over time.

chris-codaio: Specifying a `fixedId` on its own will never cause a rename - it's just an ID that helps map fields across renames.  Adding `fixedId` should always be backwards compatible, but you're correct in that you'd need to roll out `fixedId` first so that the existing schema can be forward mapped with a subsequent change.

chris-codaio: Can you point me to that documentation?

schema.ts
chris-codaio: Done

schema.ts
chris-codaio: I don't know how I would do it in a single pack version.

chris-codaio: Based on conversation with David, it sounds like there isn't any documentation about this, at least not yet.

chris-codaio: I think the most common use case for this stuff is dynamic schema, where the source of truth changes property names in place.  Cross-doc is likely to be the primary user of this feature for a while.

chris-codaio: I can add a quick note to this effect.

dweitzman-codaio: Makes sense to me that dynamic schemas would be the main use case. How much do we care about unintuitive behavior for the static schema case?

If I'm understanding the PRs correctly then a static schema pack author would do column renaming like this:
1) Release an updated version of an existing pack to introduce `fixedId` values in the static schema
2) Wait until all sync tables in docs for that pack have been synced at least once to use the updated static schema
3) Release an updated version that renames the property in the static schema.

That flow feels like it has a messy step 2, if I'm not misunderstanding how the PRs will work.

chris-codaio: You are understanding correctly.  Worst case, if they don't wait for sync tables to re-sync after step 1, is existing behavior.

dweitzman-codaio: Cool.

`fixedId` makes sense to me for dynamic schemas, but I wonder if something more like `originalPropertyId` or `previousPropertyId` would be a simpler API for static schemas if we wanted to support column renames on those and, if so,  would it make sense to restrict `fixedId` to dynamic schemas only for now.

jonathan-codaio (approved):
LGTM for my comments

PR #2570, opened 2023-04-21
Remove EmailSchema.autocomplete

dweitzman-codaio:
This is technically a breaking change, but there was almost no practical use for EmailSchema.autocomplete so it seems unlikely this would impact anyone in practice.

The motivation for doing this now is to free up the name "autocomplete" on BaseSchema for sync table row editing.


ekoleda-codaio (approved):
LGTM, but perhaps wait for Jonathan to LGTM as well.

jonathan-codaio (approved):
The irony is once we have two way, we probably want to bring this back, just with a different name, in order to have full fidelity in the editing experience for Email columns.

dweitzman-codaio:
> The irony is once we have two way, we probably want to bring this back, just with a different name, in order to have full fidelity in the editing experience for Email columns.

Yeah, I was thinking of renaming it in this PR but it seemed fine to save it for a later PR when we're ready to actually implement that editing experience

PR #2573, opened 2023-04-26
Slate marshaling support

chris-codaio:
PTAL @coda/packs @jonathan-codaio @huayang-codaio


huayang-codaio (commented):
Is this only to be used in cross-doc?

chris-codaio:
> Is this only to be used in cross-doc?

Yes, for the moment.

testing/validation.ts
```diff
@@ -184,6 +184,8 @@ function checkPropertyTypeAndCodaType<ResultT extends any>(
         case undefined:
           // no need to coerce current result type
           return [];
+        case ValueHintType.CodaInternalRichText:
+          return [{message: `CodaInternalRichText is not supported in external packs.`}];
```
jonathan-codaio, line 188: Interesting, so are we able to hard-block here in the SDK because the cross-doc getSchema call never calls our validation code? Not sure if that's great if so haha but love that we're able to reliably block other usage.

schema.ts
```diff
@@ -167,6 +167,8 @@ export enum ValueHintType {
    * Indicates to render a boolean value as a toggle.
    */
   Toggle = 'toggle',
+  /** @hidden */
+  CodaInternalRichText = 'codaInternalRichText',
```
huayang-codaio, line 171: I would like to see more code documentation here to remind us what this field is used for. 

If we were ever to expose this to external developers, I think we'd use a new hint type name anyway. 

cc @ekoleda-codaio for sdk changes & docs for sanity check.

huayang-codaio (approved):

testing/validation.ts
chris-codaio: Yeah, that's the idea.  Seems to work properly in my prototype.

schema.ts
chris-codaio: Sure - I can add some more comments, but leave `@hidden` in place.

ekoleda-codaio: LGTM. It's hard to hide things completely, but the name here also reflects the limited availability.

testing/validation.ts
chris-codaio: Though to be clear, this is only when running client-side pack validation of runtime results.  It doesn't block uploading such a pack.  I can bifurcate some of our validation logic to differentiate internal vs. external usage with more work if that's desired.

jonathan-codaio: Oh, can you make sure this gets validated in upload_validation.ts? That gets called both on the client and on the upload.

chris-codaio: Yeah, I'll work on a `superRefine` to catch this case if I can wrestle it to the ground.

testing/upload_validation.ts
```diff
@@ -1112,6 +1127,20 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
       validateImageProperty();
       validateSnippetProperty();
       validateSubtitleProperties();
+    })
+    .superRefine((data, context) => {
+      const schemaHelper = objectSchemaHelper(data as GenericObjectSchema);
+      const internalRichTextPropertyTuple = Object.entries(schemaHelper.properties).find(([_key, prop]) => 
+        prop.type === ValueType.String && prop.codaType === ValueHintType.CodaInternalRichText);
+      if (internalRichTextPropertyTuple && !isValidUseOfCodaInternalRichText(data.identity?.packId)) {
```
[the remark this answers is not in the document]
chris-codaio: It does call this path, so we'll need the exemption.

testing/upload_validation.ts
```diff
@@ -884,6 +893,11 @@ function isValidIdentityName(packId: number | undefined, name: string): boolean
   return isValidObjectId(name);
 }
 
+function isValidUseOfCodaInternalRichText(packId: number | undefined): boolean {
+  // CrossDoc pack is allowed to use this type hint.
+  return packId === 1053;
```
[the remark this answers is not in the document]
chris-codaio: Yeah - fat fingered the input.

PR #2575, opened 2023-04-27
Refactor 2-way autocomplete so functions can be defined inline in the schema

dweitzman-codaio:
Supported in this PR:
- Defining either a list of hard-coded autocomplete options in-place in a schema, a JS function in-place for static schemas, or setting `AutocompleteType.Dynamic` to invoke the sync table's global `autocomplete` function for any dynamic properties
- Type errors if a property should be a number, string, or object but tries to hard-code values or a function returning a different one of those types
- Numbers and strings can also return a display name instead of just the raw string or number
- If a new pack version is released, an autocomplete function on the latest version will still run for a column synced at a previous pack version as long as the property name hasn't changed. This is because in-place autocomplete functions get named after the property they're defined on.
- If a dynamic `getSchema` formula tries to return a result with a function-valued `autocomplete` in it, return `UserVisibleMessage` explaining that dynamic schemas must use `AutocompleteType.Dynamic`

Not yet supported in this PR:
- Typing to let people return a cache TTL on the result. This can be added later

Not going to be supported, either now or in the future:
- For object schemas, in-place hard-coded options and in-place function values don't have typechecking on their return types. Didn't see a way to make that work
- For dynamic schema functions, there's also no typechecking of return types since it could be anything

The tsdocs are all marked as `@hidden` until launch

Pack example with a static schema:
```
export const UserSchema = coda.makeObjectSchema({
  type: coda.ValueType.Object,
  displayProperty: "name",
  idProperty: "internalId",
  properties: {
    name: { type: coda.ValueType.String },
    internalId: { type: coda.ValueType.String, required: true },
  },
  mutable: true,
  autocomplete: () => {
    return [
      { name: "Person 1", internalId: "abc123" },
      { name: "Person 2", internalId: "def456" },
    ];
  },
});

const WidgetsSchema = coda.makeObjectSchema({
  properties: {
    type: {
      type: coda.ValueType.String,
      mutable: true,
      autocomplete: () => {
        return ["animal", "vegetable", "mineral"];
      },
    },
    subtype: {
      type: coda.ValueType.String,
      mutable: true,
      autocomplete: (context) => {
        switch (context.propertyValues["type"]) {
          case "animal":
            return ["cheetah", "falcon", "lizard"];
          case "vegetable":
            return ["celery", "carrot", "tree"];
          case "mineral":
            return ["steel", "concrete", "sand"];
        }
        return [];
      },
    },
  // ...
```

Example with a dynamic schema:
```
pack.addDynamicSyncTable({
  name: "DynamicTasks",
  description: "The tasks in the selected project.",
  identityName: "DynamicTask",
  getName: async function (context) {
    return "Dynamic tasks name";
  },
  getDisplayUrl: async function (context) {
    return "https://example.com/";
  },
  getSchema: async function (context) {
    return coda.makeObjectSchema({
      properties: {
        taskType: {
          type: coda.ValueType.String,
          mutable: true,
          autocomplete: coda.AutocompleteValueType.Dynamic,
        },
        // runtimeErrorType: {
        //   type: coda.ValueType.String,
        //   mutable: true,
        //   autocomplete: () => {
        //     // This raises a user-visible runtime error explaining the need to use AutocompleteType.Dynamic
        //     // instead.
        //     return ["a", "b", "c"];
        //   },
        // },
      },
      displayProperty: "taskType",
      idProperty: "taskType",
      featuredProperties: ["taskType"],
    });
  },
  autocomplete: (context) => {
    if (context.propertyName === "taskType") {
      return ["Task Type 1", "Task Type 2"];
    }
    throw new coda.UserVisibleError(
      `Cannot autocomplete property ${context.propertyName}`
    );
  },
  formula: {
    name: "SyncTasks",
    description: "Sync tasks in a project.",
    parameters: [],
    execute: async function () {
      const result = [{ taskType: "someTaskType" }];
      return {
        result,
      };
    },
    executeUpdate: async function () {
      return { result: [] };
    },
  },
});
```


api.ts
```diff
@@ -1423,23 +1440,46 @@ export function makeMetadataFormula(
...
-    name: 'getPropertyAutocompleteMetadata',
-    description: 'Gets property autocomplete',
-    execute([], context) {
-      return execute(context as PropertyAutocompleteExecutionContext);
-    },
+
+  type ResultT = SchemaType<ArraySchema<SchemaT>>;
+
+  // The type SchemaType<ArraySchema<T>> is equivalent to Array<SchemaType<T>>, but typescript doesn't know
+  // that unless we do a cast.
```
huayang-codaio, line 1464: :(

api.ts
```diff
@@ -2067,8 +2137,23 @@ export function makeDynamicSyncTable<
     formula,
     connectionRequirement,
     dynamicOptions: {getSchema, entityName, defaultAddDynamicColumns},
-    propertyAutocomplete,
   });
+
+  if (autocomplete) {
+    table.namedAutocompletes ??= {};
```
huayang-codaio, line 2143: TIL javascript now supports this syntax.

api.ts
```diff
@@ -2067,8 +2137,23 @@ export function makeDynamicSyncTable<
     formula,
     connectionRequirement,
     dynamicOptions: {getSchema, entityName, defaultAddDynamicColumns},
-    propertyAutocomplete,
```
huayang-codaio, line 2070: this is the unused autocomplete field used by 2-way right?

runtime/thunk/thunk.ts
```diff
@@ -142,53 +143,57 @@ async function doFindAndExecutePackFunction<T extends FormulaSpecification>({
           break;
         case MetadataFormulaType.PropertyAutocomplete:
           const syncTable = syncTables?.find(table => table.name === formulaSpec.syncTableName);
-          const autocompleteFn = ensureExists(syncTable?.propertyAutocomplete);
-          const propertyValues = {};
+          const autocompleteFormula = syncTable?.namedAutocompletes?.[formulaSpec.autocompleteName];
```
huayang-codaio, line 146: where would `autocompleteName` come from? specifically:
- maybe some helper method of autocomplete name is needed somewhere.
- is it possible that an identity level autocomplete function would respond to a static property? (in which case we may have two autocomplete functions responding to a single property)

huayang-codaio (commented):
I think a test or formula of identity level autocomplete (i.e. for dynamic properties) would be helpful. (only saw static autocompletes in fake_v2)

api.ts
dweitzman-codaio: Yeah, that field was added by me for the earlier implementation where you'd always define `propertyAutocomplete` on a sync table itself. That kind of still exists here but it's only intended for dynamic sync tables and got renamed to just `autocomplete`

api.ts
```diff
@@ -2203,3 +2288,49 @@ export function maybeRewriteConnectionForFormula<
...
+  // Converting JS functions to strings happens on inputSchema instead of the deep copied version because the
+  // deep copy will have already thrown away any JS functions.
+  const namedAutocompletes: SyncTableAutocompleters = {};
+
+  for (const propertyName of listPropertiesWithAutocompleteFunctions(inputSchema)) {
+    schema.properties[propertyName].autocomplete = propertyName as AutocompleteReference;
+    namedAutocompletes[propertyName] = makePropertyAutocompleteFormula({
+      execute: inputSchema.properties[propertyName].autocomplete as any,
+      schema: normalizeSchema(schema.properties[propertyName]),
+      name: `${identityName}.${propertyName}.Autocomplete`,
```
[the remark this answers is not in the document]
dweitzman-codaio: Yes. This particular `name` field isn't used for anything except maybe making it clear what you're looking at in a debugging, since the formula isn't exposed for use anywhere outside the PropertyAutocomplete metadata function.

runtime/thunk/thunk.ts
dweitzman-codaio: The name here is the same string value that gets stored into the schema itself, which would be `propertyName as AutocompleteReference` by the convention of what `makeSyncTable()` does.

> is it possible that an identity level autocomplete function would respond to a static property? (in which case we may have two autocomplete functions responding to a single property)

A person could technically set a static property up with `autocomplete: AutocompleteType.Dynamic` to point it at the sync table's general dynamic autocomplete function instead of a property-specific one, even for a non-dynamic sync table.

dweitzman-codaio:
> I think a test or formula of identity level autocomplete (i.e. for dynamic properties) would be helpful. (only saw static autocompletes in fake_v2)

Added a test for makeSyncTable()

api.ts
```diff
@@ -231,6 +234,19 @@ export class MissingScopesError extends Error {
...
+/**
+ * A map of named autocomplete methods for a particular sync table. The names need to match
+ * the values stored in the object schema. For the name, we use the property's name so that
+ * it'll be consistent across pack versions. In the future if we want to support packs
+ * being able to rename an existing property, we could try to set the names to the old
+ * property names. Alternatively, we could just say that autocomplete will briefly stop
+ * working until the sync table is refereshed so its schema matches the current pack release's
+ * schema.
+ */
+interface SyncTableAutocompleters {
```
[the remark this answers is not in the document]
dweitzman-codaio: For the moment it's not needed by anything in the `coda` repo so I figured it was better to err on the side of not exporting things if there's no need for it

api.ts
dweitzman-codaio: Earlier iterations of this SDK code with different API surfaces had much sadder typescript casts, so I was actually kind of happy that this was the only weird one

However... when we add the ability to run a `cacheTtl` in addition to the result array, the types here will be an imperfect reflection of what the formula can actually return.

If we want to allow both a simple scenario where people just return a raw array as well as a more complex scenario where you can return that array wrapped in an object with additional information like a `cacheTtl`, that'll be too sophisticated for the pack schema types to express natively.

runtime/thunk/thunk.ts
dweitzman-codaio: For completeness of discussion, there's another option that's different from what the code has now:

Instead of putting the property name itself into the `autocomplete` value in the schema, we could put some sentinel value like `__coda_static__` and then use the property name at runtime to map back to a particular autocomplete formula from `namedAutocompletes`

My motivation for filling in a string property name value during `makeSyncTable()` instead of just a marker like `__coda_static__` and resolving the property name runtime later was mostly to eliminate uncertainty and/or annoyance related to schema normalization: if I set up `syncTable.namedAutocomplete[foo] = ...` and `property.Foo.autocomplete = "foo"` at the same moment in `makeSyncTable` then there's no possibility for inconsistency about whether one name is normalized and the other isn't

api.ts
```diff
@@ -1423,23 +1440,46 @@ export function makeMetadataFormula(
...
+    throw new Error(`Value for execute must be a function`);
   }
-  return makeObjectFormula({
-    name: 'getPropertyAutocompleteMetadata',
-    description: 'Gets property autocomplete',
-    execute([], context) {
-      return execute(context as PropertyAutocompleteExecutionContext);
-    },
+
+  type ResultT = SchemaType<ArraySchema<SchemaT>>;
```
jonathan-codaio, line 1461: Interesting, didn't know you could declare a type dynamically within a function (thought it always happened in the generics declaration, e.g. `<SchemaT extends Schema, ResultT extends SchemaType<ArraySchema<SchemaT>>>`

api.ts
```diff
@@ -1423,23 +1440,46 @@ export function makeMetadataFormula(
...
+
+  // The type SchemaType<ArraySchema<T>> is equivalent to Array<SchemaType<T>>, but typescript doesn't know
+  // that unless we do a cast.
+  const executeRetyped = execute as PropertyAutocompleteMetadataFunction<SchemaType<ArraySchema<SchemaT>>>;
+
+  const innerExecute = async ([]: ParamValues<[]>, context: ExecutionContext): Promise<ResultT> => {
```
jonathan-codaio, line 1467: This is just because we have to bend this to something matching the shape of a regular `MetadataFormula`?

api.ts
```diff
@@ -1423,23 +1440,46 @@ export function makeMetadataFormula(
...
+    const result = await executeRetyped(context as PropertyAutocompleteExecutionContext);
+    return result;
+  };
+
+  const formulaDefn: ArrayFormulaDef<[], SchemaT> = {
+    connectionRequirement: ConnectionRequirement.Optional,
```
jonathan-codaio, line 1473: Any reason this shouldn't just inherit the connection requirement from e.g. the builder, via a parameter to this helper?

api.ts
```diff
@@ -1699,11 +1739,41 @@ export interface SyncTableOptions<
    * sync tables that have a dynamic schema.
    */
   dynamicOptions?: DynamicOptions;
+
   /**
-   * A function to autocomplete values for properties marked with `mutable` and `autocomplete`.
+   * An autocomplete function to use for any dynamic schema properties.
+   * The name of the property that's being modified by the doc editor
+   * is available in autocomplete function's context parameter.
```
jonathan-codaio, line 1746: Nit, should it be "in _the_ autocomplete function's"?

api.ts
```diff
@@ -1699,11 +1739,41 @@ export interface SyncTableOptions<
...
+   *       `Cannot autocomplete property ${context.propertyName}`
+   *     );
+   *   },
+   *   ...
+   * ```
+   *
    * @hidden
    */
-  propertyAutocomplete?: PropertyAutocompleteMetadataFunction;
+  autocomplete?: PropertyAutocompleteMetadataFunction<any>;
```
jonathan-codaio, line 1776: Should this go in `dynamicOptions`, since that's the set of all things that apply only to dynamic schemas?

api.ts
```diff
@@ -2203,3 +2296,49 @@ export function maybeRewriteConnectionForFormula<
   }
   return formula;
 }
+
+// This helper method finds any inline autocomplete functions in a static sync table schema.
+// These functions will need to be extracted into the "namedAutocompletes" property on the sync
+// table and replaced with strings.
+function listPropertiesWithAutocompleteFunctions(schema: ObjectSchemaDefinition<string, string>): string[] {
+  const result: string[] = [];
+  for (const propertyName of Object.keys(schema.properties)) {
```
jonathan-codaio, line 2305: There can be nested properties, but presumably we don't need to recurse here because only top-level properties are editable and hence only those can actually execute their autocomplete functions? Probably worth an explanation in a comment.

api.ts
```diff
@@ -2203,3 +2296,49 @@ export function maybeRewriteConnectionForFormula<
...
+
+    result.push(propertyName);
+  }
+  return result;
+}
+
+// Finds any inline autocomplete functions within the inputSchema and replaces them
+// with strings references into the returned namedAutocompletes.
+function replaceInlineAutocompleteFunctionsWithNamedAutocompleteFunctions({
+  inputSchema, // DO NOT MUTATE inputSchema!
```
jonathan-codaio, line 2322: Is there anything we do with `Readonly<T>` or something to help enforce this?

api.ts
```diff
@@ -2203,3 +2296,49 @@ export function maybeRewriteConnectionForFormula<
...
+  identityName: string;
+}): SyncTableAutocompleters {
+  // Converting JS functions to strings happens on inputSchema instead of the deep copied version because the
+  // deep copy will have already thrown away any JS functions.
+  const namedAutocompletes: SyncTableAutocompleters = {};
+
+  for (const propertyName of listPropertiesWithAutocompleteFunctions(inputSchema)) {
+    schema.properties[propertyName].autocomplete = propertyName as AutocompleteReference;
+    namedAutocompletes[propertyName] = makePropertyAutocompleteFormula({
+      execute: inputSchema.properties[propertyName].autocomplete as any,
```
jonathan-codaio, line 2337: Seems like in a future PR we can support static declarations of enumerable values by just taking an array that's given as the `autocomplete` value and wrapping it in a function that returns it here. Which is what I think we do for parameter autocomplete already.

api.ts
```diff
@@ -2203,3 +2296,49 @@ export function maybeRewriteConnectionForFormula<
...
+      schema: normalizeSchema(schema.properties[propertyName]),
+      name: `${identityName}.${propertyName}.Autocomplete`,
```
jonathan-codaio, line 2339: @chris-codaio is exploring the ability to declare an id on a schema property in addition to the name, to make properties renamable, so when that happens, we should probably switch this to name the formula based on the id if present.

schema.ts
```diff
@@ -225,16 +244,34 @@ interface BaseSchema {
...
+   *        return results.map(user => {display: user.name, value: user.id})
+   *      },
+   *   },
+   * }
+   * ```
    *
    * @hidden
    */
-  autocomplete?: boolean;
+  autocomplete?: PropertySchemaAutocomplete<any>;
```
jonathan-codaio, line 274: Any reason not to put this on BaseSchema?

testing/upload_validation.ts
```diff
@@ -616,7 +618,7 @@ const imageAttributionNodeSchema = z.object({
 const basePropertyValidators = {
   description: z.string().optional(),
   mutable: z.boolean().optional(),
-  autocomplete: z.boolean().optional(),
+  autocomplete: z.union([z.string(), z.array(z.object({}).passthrough())]).optional(),
```
jonathan-codaio, line 621: Just confirming that if you declare `autocomplete` on a schema property, the compiled metadata json will contain only a formula metadata object with the `execute` stripped out, and not some accidentally serialization of the `execute` function?

jonathan-codaio (commented):
Good stuff

runtime/thunk/thunk.ts
dweitzman-codaio, line 1: A note to reviewers: I recommend clicking "hide whitespace changes" in github when reviewing this file

schema.ts
dweitzman-codaio: I think the place where you commented is in BaseSchema. The reason for making more strictly-typed definitions of the `autocomplete` property on the string and number sub-interfaces of BaseSchema was to support stricter compiler-time checks on an inline autocomplete function or inline array of static values

api.ts
dweitzman-codaio: In terms of why the code looks like this, the answer is copy-paste.

I'll update the PR to keep the connection requirements for autocomplete formulas updated in the same way as other metadata formulas on a sync table

api.ts
dweitzman-codaio: Done

api.ts
dweitzman-codaio: I originally had it in dynamicOptions, but then realized we support dynamic properties in static sync tables.

For example, the Jira issues table actually uses `makeSyncTable` rather than `makeDynamicSyncTable`. Seems like dynamic sync tables always want you to give them a URL link before they finish being created and static sync tables don't?

api.ts
dweitzman-codaio: Correct, I've added a note:

```
// Not that we won't detect autocomplete functions within nested object schemas, but that's not necessary
// here: the only types you can autocomplete in a 2-way schema are the top-level ones.
```

api.ts
dweitzman-codaio: I did a quick test and it seems like `Readonly` only protects the top-level properties and not deeply nested ones?

I've added in a `Readonly<>` wrapper just to emphasize to the reader that they shouldn't change it, but I haven't removed the comment

api.ts
dweitzman-codaio: Technically this PR already supports static declarations of enumerable values, although I haven't added that support into the `coda` code.

I haven't wrapped static value lists in a formula because it seems like we might want to be able to just read them directly in the doc without invoking a pack over the network

The `T[]` below is what allows for static declarations of values:
```
/**
 * A function or set of values to return for property autocomplete.
 * @hidden
 */
export type PropertySchemaAutocomplete<T extends PackFormulaResult> =
  | PropertyAutocompleteMetadataFunction<T[]>
  | T[]
  | AutocompleteType
  | AutocompleteReference;
```

api.ts
dweitzman-codaio: Makes sense to me. I feel like the simplest way to handle that is to just let autocomplete briefly break when switching to an ID until people run the first post-upgrade re-sync.

schema.ts
jonathan-codaio: Oh oops, missed this.

testing/upload_validation.ts
dweitzman-codaio: `JSON.stringify({'stringProp': 'foo', 'functionProp': () => ['bar']})` returns `'{"stringProp":"foo"}'`, so if someone put an inline autocomplete into a sub-object-schema that we don't transform during `makeSyncTable` it should just get silently dropped during JSON serialization.

The downside is that if there's a bug in `makeSyncTable` we could silently drop actual `autocomplete` functions. Hopefully the unit test will prevent someone from accidentally breaking that

api.ts
jonathan-codaio: Right, but when you wan to declare a dynamic schema on a static table, you still use `dynamicOptions`. Here's the Jira def: https://github.com/coda/packs/blob/5d18547bb9c13965cdd1114978150f32f2e3aa7c/packs/atlassian/jira/common/formulas.ts#L522

testing/upload_validation.ts
jonathan-codaio: Checking my understanding, I think what happens so far is that if you pass a function in the autocomplete field, and it's a top-level property, we'll replace the function with a string that's the autocomplete name. If the value is not a function, we'll leave it alone. If the property is not top-level, we'll leave alone whatever is there.

Then I think zod will error out on some values that were left alone, but allow possibly garbage values through if they have the shape of an array of objects. Is that intentional, the static declarations of return values can be any object because it's valid to return an autocomplete result of any shape? And also if you want to statically declare an array of scalar autocomplete values you have to wrap them in an object to make zod happy?

Maybe we can defer this to another PR, but if we're enforcing a {display, value} kind of shape for static values we should enforce that with zod.

I'm not sure what to do about non-top-level fields, whether to silently allow them through if they have the right shape, or strip them out silently since they're unused, or error out on them. Can handle separately though.

bundles/thunk_bundle.js
```diff
@@ -5824,6 +5824,26 @@ module.exports = (() => {
...
+      dynamicSchema.forEach((item) => throwOnDynamicSchemaWithJsAutocompleteFunction(item));
+      return;
+    }
+    if (typeof dynamicSchema === "object") {
+      for (const key of Object.keys(dynamicSchema)) {
+        throwOnDynamicSchemaWithJsAutocompleteFunction(dynamicSchema[key], key);
+      }
+    }
+    if (typeof dynamicSchema === "function" && parentKey === "autocomplete") {
+      throw new UserVisibleError(
```
ekoleda-codaio, line 5841: I'd advise against a user-visible error here, since the Pack user can't do anything to remedy this mistake. I think it's better for the user to get a generic error message, and they will reach out to the Pack maker, who if they get access to the doc can see this error in the logs.

chris-codaio:
Looking good to me.  I'm inclined to checkpoint this work and start writing live code to see how it feels in practice across a few packs.

dweitzman-codaio:
> Looking good to me. I'm inclined to checkpoint this work and start writing live code to see how it feels in practice across a few packs.

Sounds good. I'll update the adhoc with these changes plus the corresponding server-side changes

chris-codaio:
> > Looking good to me. I'm inclined to checkpoint this work and start writing live code to see how it feels in practice across a few packs.
> 
> Sounds good. I'll update the adhoc with these changes plus the corresponding server-side changes

Two-way is now available in head + staging; I think we want to move off of the adhoc and into those environments now.

jonathan-codaio (approved):
Sweet

api.ts
dweitzman-codaio: Ah, makes sense. Moved `autocomplete` into `dynamicOptions` for non-dynamic sync tables. For a dynamic sync table it remains top-level

api.ts
dweitzman-codaio: The code is now updated to set the connection requirement based on the parent sync table, with two unit tests to check behavior depending on whether `setUserAuthentication()` is called before or after `addSyncTable()`

testing/upload_validation.ts
dweitzman-codaio: > Checking my understanding, I think what happens so far is that if you pass a function in the autocomplete field, and it's a top-level property, we'll replace the function with a string that's the autocomplete name. If the value is not a function, we'll leave it alone. If the property is not top-level, we'll leave alone whatever is there.

Yes, that matches my understanding 

> Then I think zod will error out on some values that were left alone, but allow possibly garbage values through if they have the shape of an array of objects

Yes, was accurate. I've updated the code so that Zod validates the shape correctly for strings, numbers, and booleans, and also allows objects through with any shape.

The updated code has a downside, though: if someone ignores typescript errors and lets Zod catch the problem, the list of errors can be pretty long and potentially hard to make sense of. I think all the union types makes Zod unsure when there's a problem about what the type was supposed to be, and it seems to err on the side of giving more possible errors instead of fewer

My inclination is to go-ahead for now with possibly hard-to-read Zod errors and try to improve them in the future if we find that people in practice are ignoring typescript and the documentation and putting strange nonsense values into the autocomplete field.

bundles/thunk_bundle.js
dweitzman-codaio: Changed to `throw Error()` so a pack maker can still see a meaningful error but a pack user will just see that the pack had an internal error

api.ts
huayang-codaio: we can also freeze the object of pass in a copy. not sure if it's worth.

dweitzman-codaio: I think deep freezing an input parameter in-place would technically be a caller-visible mutation to that parameter, since afterwards the caller wouldn't be able to mutate the value either.

PR #2592, opened 2023-05-16
Tweak autocomplete metadata formula types

oleg-codaio:
This PR follows up on #2589, which improved typing for optional parameters and revealed a discrepancy between the optionality of the search parameter in autocomplete and non-autocomplete formulas. Fixing that up is more involved, so this PR rolls back some of that typing, while improving `autocompleteSearchObjects` to make sure only valid keys are passed in for the display and value properties.

PTAL @coda/packs


api.ts
```diff
@@ -1562,7 +1562,8 @@ export declare type GenericMetadataFormulaMetadata = Omit<GenericMetadataFormula
  */
 export type MetadataFunction = (
   context: ExecutionContext,
-  search?: string,
+  // TODO(oleg): this should be optional unless in an autocomplete context.
+  search: string,
```
jonathan-codaio, line 1566: Kinda worried about this, we're lying to external developers, which I think is ok if you lie in the conservative direction and say something might be optional that isn't, but when we claim that this will always be defined, people will start writing code that does stuff like `search.toLowerCase()` that will error out when this is undefined. If this were just for internal use I'd be ok with it but I think it would be problematic to have this in an external SDK release.

oleg-codaio: I agree, but we already had this code in place before https://github.com/coda/packs-sdk/pull/2589. This is just reverting this change here since fixing this requires a broader fix than I'm prepared to make here.

jonathan-codaio (approved):

PR #2593, opened 2023-05-17
Add placeholders to snippets

ekoleda-codaio:
Following up my change to the Monaco implementation to allow for placeholders in snippets, this PR adds placeholders to the snippets. I've spun up an adhoc with the changes if you want to test the snippets, ping me for the URL.

Changes to the snippet system
- Created some custom placeholder syntaxes, to allow placeholders to be valid typescript code in various places.
- Replace the placeholders with the default value when the snippets are shown in the documentation.

Changes to individual snippets
- Improve authentication snippets:
  - Make the generic user and system auth snippets not tied to any specific auth type.
  - Create snippets for each major auth type.
 - Add snippets for object schemas designed for sync tables and cards.
- Make the snippets more generic and add clearer "TODO" comments where follow-up is needed.
- Remove the schema from the sync table snippet, relying on users to compose snippets instead.


documentation/samples/snippets/card.ts
```diff
@@ -1,59 +1,67 @@
...
-  imageProperty: "<Property name>", // Image shown on the card.
-  linkProperty: "<Property name>", // Link opened when the card is clicked.
 });
 
 // A formula that accepts a URL and returns an object matching the schema above.
 pack.addFormula({
-  name: "<User-visible name of formula>",
-  description: "<Help text for the formula>",
+  name: "$1",
+  description: "${6:My description.}",
```
jonathan-codaio, line 38: How come this goes from 1 to 6?

Is it because the user will fill in the value for $1 on line 8 and it'll autopopulate it here?

documentation/scripts/documentation_compiler.ts
```diff
@@ -128,6 +129,29 @@ function compileSnippetEmbed(codeFile: string) {
   fs.writeFileSync(path.join(snippetDirPath, `${snippetFileName}.html`), exampleSnippetEmbed);
 }
 
+function formatCodeSnippet(code: string, removePlaceholders=false) {
```
jonathan-codaio, line 132: I don't quite understand what this is doing--is this allowing the snippets to have custom syntax that is then transformed into Monaco syntax? If so why not just use the Monaco syntax directly? I'm probably misunderstanding.

jonathan-codaio (commented):
Yes!! Yep would love the adhoc link, probably easier to review by trying these than starting at the characters.

documentation/samples/snippets/card.ts
ekoleda-codaio: Yes, in this case `$1` is just referring to the value previously entered above.

documentation/scripts/documentation_compiler.ts
ekoleda-codaio: The Monaco syntax is `${1:Placeholder}`. That works fine in strings, but if you are trying to use it as a variable name or object key it's invalid typescript. I like the fact that all of our snippets are `.ts` files and compiled with the SDK and didn't want to loose that. So I created a few variants on the format that are also valid identifiers, like `$1$Placeholder$`, etc.

jonathan-codaio: Ah I see. Mind adding essentially this explanation to a comment above this function then?

documentation/samples/snippets/sync_table.ts
```diff
@@ -2,44 +2,32 @@ import * as coda from "@codahq/packs-sdk";
...
-      let response = await context.fetcher.fetch({
-        method: "GET",
-        url: url,
-      });
-      let items = response.body.items;
-      // Adjust the items to fit the schema if required.
+    execute: async function (args, context) {
+      let [
+        // TODO: Unpack the parameter values.
+      ] = args;
```
jonathan-codaio, line 23: Nit, very optional, aesthetically this layout isn't my favorite, I guess maybe it feels bulky and doesn't match the one-liner that real code would have when there aren't huge number of params.

I wonder if something like is easier to look at and work with.

```typescript
// TODO: Unpack the parameter values like `like [param1, param2] = args;`
let [] = args;
```

jonathan-codaio (approved):
The demo is so so good.

documentation/scripts/documentation_compiler.ts
ekoleda-codaio: Done.

documentation/samples/snippets/sync_table.ts
ekoleda-codaio: I've found myself gravitating to the multi-line approach because:
- It has a nice vertical symmetry with the parameters array above.
- Makes it easy to scan which parameters have default values.
That said, it does look weird when there is only a single parameter, so I've changed it to use the single line approach you outlined.

PR #2594, opened 2023-05-18
Generate snippets file for Visual Studio Code

ekoleda-codaio:
Updates the documentation compiler to produce a `.code-snippets` file that is compatible with VSCode, and updates `coda init` to copy that file into the project's `.vscode` folder. Some design choices:

- Trigger tokens are prefixed with a forward slash to mirror the Pack Studio. This is not required in VSCode, but felt like it was better for consistency.
- The `pack.code-snippets` file is copied once during init, so updating the SDK version won't automatically bring in the latest snippets. I feel like this OK, but perhaps symlinks are an option?

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,13 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+### Added
+
+- Added the command `coda extensions` to the CLI for installing developer extensions that help with building Packs. Currently it only supports Visual Studio Code (`coda extensions vscode`), creating a code snippets file which provides the same slash commands as the Pack Studio.
+
 ### Changed
 
-- Using the empty string as a `tokenPrefix` with OAuth2 authentication will result in no prefix being used in the `Authentication` header. Previously, the empty string would be treated the same as `undefined` which would lead to the default prefix of `Bearer` being used. Note that this change took effect for live packs on April 28, 2023 indepedently of the SDK version; in this SDK version the behavior changed only here in the CLI execution simulator (the `coda execute` command).
-
-### Changed
-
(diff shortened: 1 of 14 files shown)


jonathan-codaio (commented):
Awesome! I was going to ask if this was a possibility, and here it is...

Kinda seems like adding the VSCode snippets should be its own command, so you can do it whenever you want including after an SDK update, and do it independently of `coda init`. Fine if `coda init` also does it on your behalf (though there's an argument that if you don't use VSCode you'll have this weird artifact that you'll need to .gitignore, though if our template's gitignore already has this then this is no problem).

So maybe we should have a `coda add-vs-code-snippets` command or something?

ekoleda-codaio:
Fair point, `coda init` has been somewhat agnostic to date (although it does choose `npm` over `yarn` or `pnpm`). What about making it depend on a flag, like `coda init --vscode`? Having an entire CLI command dedicated to copying this file feels like a bit overkill to me personally, and I doubt that people will really ever want to refresh the file.

jonathan-codaio:
Well, I'm also thinking about people who already have a pack they've built but now just want to add the snippets. Will running `coda init` in an existing project work ok or will it attempt to overwrite files, or just bail? As long as the use case of "awesome I want these snippets in my existing pack env" is supported, I don't really care what the syntax is.

ekoleda-codaio:
I've reverted the changed to `init` and added a new `extensions` command instead:

```
coda extensions <tools..>

Installs developer extensions for working with Packs.

Positionals:
  tools  Which tools to install extensions for.
                            [array] [required] [choices: "vscode"] [default: []]

Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]

Not enough non-option arguments: got 0, need at least 1
```

jonathan-codaio (approved):
Sweet

PR #2599, opened 2023-05-23
Do not squash fixedIds in schema normalization

patrick-codaio:
Without this, some fields drop fixedIds when Coda is processing the results of getSchema invocations.


dweitzman-codaio (approved):
Schema normalization strikes again

alan-codaio (approved):
🙏🏼

requirements.txt
```diff
@@ -35,7 +35,7 @@ python-dateutil==2.8.2 ; python_version >= '2.7' and python_version not in '3.0,
 pyyaml==6.0 ; python_version >= '3.6'
 pyyaml-env-tag==0.1 ; python_version >= '3.6'
 regex==2022.10.31 ; python_version >= '3.6'
-requests==2.31.0 ; python_version >= '3.7' and python_version < '4'
+requests==2.28.2 ; python_version >= '3.7' and python_version < '4'
```
jonathan-codaio, line 38: Mind reverting? This was part of a renovate security update this morning.

chris-codaio (approved):
Thanks!

schema.ts
```diff
@@ -1565,6 +1565,7 @@ export function normalizeSchema<T extends Schema>(schema: T): T {
     const normalized: ObjectSchemaProperties = {};
     const {
       id,
+      fixedId,
```
jonathan-codaio, line 1568: Can do later, but given how dangerous this method is proving to be, we should use the new exhaustiveness-checking technique and destructure all of the properties of this schema, so TS will catch if a new property is added but not handled here.

It would amount to adding this here:

```typescript
      identity,
      description,
      attribution,
      fixedId,
      type,
      codaType,
      properties,
      includeUnknownProperties,
      autocomplete,
      mutable,
      // eslint-disable-next-line @typescript-eslint/naming-convention
      __packId,
      ...rest
    } = schema as GenericObjectSchema;
    ensureNever<keyof typeof rest>();
```

and then using those destructured fields below instead of referencing them as `schema.foo`.

patrick-codaio: Added! I went ahead and alphabetized the property lists too, for faster scanning.

requirements.txt
patrick-codaio: Thanks, I didn't include it initially, but I'm trying to figure out why circle is saying there are dist changes...

jonathan-codaio (approved):
Thanks!

PR #2600, opened 2023-05-24
Replace animated webp files with mp4 video files

ekoleda-codaio:
The mp4 files are about 30% of the size and should help with our web vitals score. This change also:

- Enables the `mkdocs-video` plugin to allow for the easy embedding of mp4 videos.
- Allows the `screenshot` class to be used on video elements.
- Reruns the existing image optimization script over to catch some newer screenshots.
- Removes the outdated webp handling in the image optimization script.
- Fixes a deprecation warning with the social cards config that was causing the build to fail.


abhishek-codaio (approved):
🎉

PR #2601, opened 2023-05-24
Add "propertySchema" to PropertyAutocompleteFormulaSpecification

dweitzman-codaio:
This seems like a simple and relatively future-friendly way to put the schema of the property currently being autocompleted somewhere a pack maker can easily access it.

There are still some unresolved questions about whether propertyName and syncTable.schema should or shouldn't be normalized, denormalized to the author-written values, or use the fromKey values when they're set, but this PR may buy us more time before we need to sort that out.

If we decide to start denormalizing schemas for the context it would change the `propertySchema` here, but we'd also have a breaking change with `context.syncTable.schema` to sort out.

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -2,6 +2,7 @@ import type {$Values} from './type_utils';
 import type {ArraySchema} from './schema';
 import type {Continuation} from './api';
 import type {MetadataFormula} from './api';
+import type {Schema} from './schema';
 
 /**
  * Markers used internally to represent data types for parameters and return values.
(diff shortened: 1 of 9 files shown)


jonathan-codaio (approved):
So it seems like whether or not the schema is denormalized doesn't affect this change, it would just affect what is provided to the formulaSpec at runtime? (Though either way we would want to update the documentation comment to describe it unambiguously.)

dweitzman-codaio:
> So it seems like whether or not the schema is denormalized doesn't affect this change, it would just affect what is provided to the formulaSpec at runtime? (Though either way we would want to update the documentation comment to describe it unambiguously.)

Yes, `propertySchema` would be populated regardless of whether it or `context.syncTable.schema` are normalized or not

PR #2629, opened 2023-06-02
Upgrade to pnpm 8

dweitzman-codaio:
Also set `engines` in package.json to catch people using the wrong pnpm version and get Renovate to update the lockfile correctly when making PRs for upgraded dependencies


neal-codaio (approved):

huayang-codaio (approved):

scripts/dev/install-pnpm.sh
```diff
@@ -1,25 +1,6 @@
...
-  ERROR=$?
-  RETRY=$((RETRY + 1))
-  if [ "${RETRY}" -gt ${MAX_RETRIES} ]; then
-    log error "Bailing out after ${RETRY} retries... ¯\_(ツ)_/¯"
-    exit ${ERROR}
-  fi
-  log warn "Waiting to retry ${RETRY} of ${MAX_RETRIES}..."
-  sleep ${RETRY_DELAY_SECS}
-done
+sudo npm install -g @pnpm/exe@~8.5.1
```
huayang-codaio, line 6: this is a bit aggressive but I guess we have no option?

huayang-codaio: Is it an option to stay with a legacy version of pnpm and renovate?

dweitzman-codaio: The `sudo` is an attempt to fix a CI error. It wasn't necessary for running locally on my mac. If it works with `sudo` on CI I could condition that to only happen on CI

huayang-codaio: Oh I meant the whole `npm install -g` thing.

dweitzman-codaio: Ah, well I think there's a bit of a chicken-egg problem there where we need `pnpm` to install node deps but we can't install it locally without `pnpm` because then the `node_modules` layout is already wrong an `pnpm` doesn't know what to do with it. Installing pnpm globally instead of locally is one way around that

dweitzman-codaio:
Updated the PR to install pnpm under `./pnpm_install/bin` instead of in the user's normal `PATH`, which seems to make CI happy

PR #2578, opened 2023-05-01
Update dependency yargs to v17.7.2

krypton-build-user:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [yargs](https://yargs.js.org/) ([source](https://togithub.com/yargs/yargs)) | dependencies | patch | [`17.7.1` -> `17.7.2`](https://renovatebot.com/diffs/npm/yargs/17.7.1/17.7.2) |

---

### Release Notes

<details>
<summary>yargs/yargs</summary>

### [`v17.7.2`](https://togithub.com/yargs/yargs/blob/HEAD/CHANGELOG.md#&#8203;1772-httpsgithubcomyargsyargscomparev1771v1772-2023-04-27)

[Compare Source](https://togithub.com/yargs/yargs/compare/v17.7.1...v17.7.2)

##### Bug Fixes

-   do not crash completion when having negated options ([#&#8203;2322](https://togithub.com/yargs/yargs/issues/2322)) ([7f42848](https://togithub.com/yargs/yargs/commit/7f428485e75e9b1b0db1320216d1c31469770563))

</details>

---

### Configuration

📅 **Schedule**: Branch creation - "before 3am on the first day of the month" (UTC), Automerge - At any time (no schedule defined).

🚦 **Automerge**: Enabled.

♻ **Rebasing**: Whenever PR becomes conflicted, or you tick the rebase/retry checkbox.

🔕 **Ignore**: Close this PR and you won't be reminded about this update again.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Self-Hosted Renovate Bot](https://togithub.com/coda/renovate-config). Find all logs [here](https://app.circleci.com/pipelines/github/coda/renovate-config?branch=main).

PR #2630, opened 2023-06-02
[renovate] Attempt to fix postUpgradeTasks so renovate can update build artifacts

dweitzman-codaio:


dweitzman-codaio:
Example error: https://github.com/coda/packs-sdk/pull/2578#issuecomment-1574359638

This also depends on https://github.com/coda/renovate-config/pull/12

PR #2637, opened 2023-06-05
Update dependency qs to v6.11.2

krypton-build-user:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [qs](https://togithub.com/ljharb/qs) | dependencies | patch | [`6.11.1` -> `6.11.2`](https://renovatebot.com/diffs/npm/qs/6.11.1/6.11.2) |

---

### Release Notes

<details>
<summary>ljharb/qs</summary>

### [`v6.11.2`](https://togithub.com/ljharb/qs/blob/HEAD/CHANGELOG.md#&#8203;6112)

[Compare Source](https://togithub.com/ljharb/qs/compare/v6.11.1...v6.11.2)

-   \[Fix] `parse`: Fix parsing when the global Object prototype is frozen ([#&#8203;473](https://togithub.com/ljharb/qs/issues/473))
-   \[Tests] add passing test cases with empty keys ([#&#8203;473](https://togithub.com/ljharb/qs/issues/473))

</details>

---

### Configuration

📅 **Schedule**: Branch creation - "before 3am on the first day of the month" (UTC), Automerge - At any time (no schedule defined).

🚦 **Automerge**: Enabled.

♻ **Rebasing**: Whenever PR becomes conflicted, or you tick the rebase/retry checkbox.

🔕 **Ignore**: Close this PR and you won't be reminded about this update again.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Self-Hosted Renovate Bot](https://togithub.com/coda/renovate-config). Find all logs [here](https://app.circleci.com/pipelines/github/coda/renovate-config?branch=main).

PR #2646, opened 2023-06-06
Skip building docs for renovate PRs

dweitzman-codaio:
This is a follow-up to https://github.com/coda/packs-sdk/pull/2630

I'm not sure how to get `make _bootstrap-doc-tools` to work in the renovate container, but it's possible that building compiled JS files and typescript definitions will still work.

Failure error message example: https://github.com/coda/packs-sdk/pull/2637#issuecomment-1579226886

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -30,7 +30,9 @@
   },
 
   postUpgradeTasks: {
-    commands: ['make bs _bootstrap-doc-tools build'],
-    fileFilters: ['dist/**/*', 'documentation/**/*'],
+    // NOTE(dweitzman)): _bootstrap-doc-tools doesn't work in the renovate container
+    // at the moment so we skip building docs and just compile code.
+    commands: ['make bs clean compile'],
+    fileFilters: ['**/*.js', '**/*.d.ts'],
   },
 }


renovate.json5
```diff
@@ -30,7 +30,9 @@
   },
 
   postUpgradeTasks: {
-    commands: ['make bs _bootstrap-doc-tools build'],
-    fileFilters: ['dist/**/*', 'documentation/**/*'],
+    // NOTE(dweitzman)): _bootstrap-doc-tools doesn't work in the renovate container
+    // at the moment so we skip building docs and just compile code.
```
huayang-codaio, line 34: this doesn't affect building docs in main branch or PRs right?

huayang-codaio (approved):

renovate.json5
dweitzman-codaio: No, the other builds on the repo's circleci won't be affected. This configuration is only used by Renovate while making PRs to upgrade dependencies, which is a separate circleci. A failure here also doesn't stop Renovate from making a PR, it just makes a PR with a comment saying that the artifacts may not be correct

PR #2647, opened 2023-06-07
Add documentation for two-way sync.

ekoleda-codaio:
Preview:
- https://coda-packs-sdk--2647.com.readthedocs.build/en/2647/guides/blocks/sync-tables/two-way/
- https://coda-packs-sdk--2647.com.readthedocs.build/en/2647/samples/topic/two-way/


api.ts
```diff
@@ -966,11 +966,19 @@ export interface SyncFormulaResult<K extends string, L extends string, SchemaT e
...
+  /**
+   * The previous value of the row.
+   */
   previousValue: ObjectSchemaDefinitionType<K, L, SchemaT>;
+  /**
+   * The new value of the row, with the user edits applied.
+   */
   newValue: ObjectSchemaDefinitionType<K, L, SchemaT>;
+  /**
+   * The fields of the object that have been updated.
```
chris-codaio, line 980: Nit: we use `object` here but `row` above.  Should these be consistent?

docs/guides/blocks/sync-tables/two-way.md
```diff
@@ -0,0 +1,293 @@
...
+pack.addSyncTable({
+  name: "Tasks",
+  // ...
+  formula: {
+    // ...
+    execute: async function (args, context) {
+      // Pull down the data ...
+    },
+    executeUpdate: async function (args, updates, context) {
+      // Push up the changes ...
```
chris-codaio, line 90: Should this say `// Push up the changes and return the updated data ...`?

docs/guides/blocks/sync-tables/two-way.md
```diff
@@ -0,0 +1,293 @@
...
+      result: [task],
+    };
+    ```
+
+
+## Batching updates {: #batching}
+
+By default row updates are processed one at a time, which makes the coding simple but can impact performance when many rows are being updated at once. You can greatly improve performance by batching multiple updates into a single `executeUpdate` call and processing them in parallel.
+
+The batch size can be adjusted by setting the `maxUpdateBatchSize` field of the sync formula. You'll need to adjust your `executeUpdate` function to be able to handle multiple updates at once. Use a batch update endpoint if your API provides it, or if not make fetch requests [in parallel][fetcher_parallel] to realize performance gains.
```
chris-codaio, line 153: Should we call out the 60 second timeout the pack has to execute here?

docs/guides/blocks/sync-tables/two-way.md
```diff
@@ -0,0 +1,293 @@
...
+    ```
+
+
+## Handling errors
+
+When a row fails to update Coda will show an error message attached to that row. The pending edits will be retained, so that they can adjust them and try again.
+
+<img src="../../../../images/two_way_error.png" srcset="../../../../images/two_way_error_2x.png 2x" class="screenshot" alt="Error attached to a row.">
+
+To associate an error with a row return it at the corresponding index in the `result` array. For tables that use single row updates (`maxUpdateBatchSize == 1`) if you throw a `UserVisileError` it will be associated with the row automatically. In all other cases a thrown error will be treated as a failure of the entire sync and cancel any remaining row updates.
```
chris-codaio, line 286: s/`UserVisileError`/`UserVisibleError`

chris-codaio (approved):
This is great!!!

api.ts
ekoleda-codaio: Good catch, done.

docs/guides/blocks/sync-tables/two-way.md
ekoleda-codaio: Done.

docs/guides/blocks/sync-tables/two-way.md
ekoleda-codaio: Good catch, fixed.

ekoleda-codaio:
@dweitzman-codaio - Can you give this a quick look to make sure I didn't miss anything in regards to property autocomplete?

docs/guides/blocks/sync-tables/two-way.md
```diff
@@ -0,0 +1,300 @@
...
+        for (let attr of attriutes) {
+          properties[attr] = {
+            // ...
+            autocomplete: coda.AutocompleteType.Dynamic,
+          }
+        }
+        // ...
+      },
+      autocomplete: async function (context) {
+        let attr = context.propertyName;
```
dweitzman-codaio, line 252: Now that we're writing documentation, we'll need to make a decision about whether `propertyName` is the normalized property name, the `fromKey`, or the key originally written by the pack maker, along with the related question about whether to denormalize schemas in the formula context also

We'll probably need to schedule some time to continue discussing and see if we can agree on a path forward.

dweitzman-codaio (commented):
I haven't read through everything, but for autocomplete I have one nit and one unresolveled decision to call out

docs/guides/blocks/sync-tables/two-way.md
```diff
@@ -0,0 +1,300 @@
...
+        let patterns = await searchPatterns(context, search);
+        return patterns;
+      },
+    },
+  },
+  // ...
+});
+```
+
+For sync tables with dynamic schemas, you aren't allowed to define the search function directly on the property itself. Instead use the special value `AutocompleteType.Dynamic`, which tells Coda to call the sync table's `autocomplete` function. This is a single function that handles the auto-complete for all dynamic properties. It's defined directly on a dynamic sync table, or within the `dynamicOptions` of a regular sync table. The function can determine which property to provide autocomplete for by inspecting `context.propertyName`.
```
jonathan-codaio, line 234: Nit, optional, "aren't allowed" makes it sound a bit arbitrary, could frame it as "you aren't able to include code in dynamic responses" or something.

docs/guides/blocks/sync-tables/two-way.md
```diff
@@ -0,0 +1,300 @@
...
+          return options;
+        },
+      },
+    });
+    ```
+
+
+## Handling errors
+
+When a row fails to update Coda will show an error message attached to that row. The pending edits will be retained, so that they can adjust them and try again.
```
jonathan-codaio, line 288: Nit, "so that they" seems to be a referencing a user who is not mentioned, maybe change to "so that the user can adjust them".

jonathan-codaio (approved):
Another super awesome addition to the documentation!

docs/guides/blocks/sync-tables/two-way.md
```diff
@@ -0,0 +1,300 @@
...
+For sync tables with dynamic schemas, you aren't allowed to define the search function directly on the property itself. Instead use the special value `AutocompleteType.Dynamic`, which tells Coda to call the sync table's `autocomplete` function. This is a single function that handles the auto-complete for all dynamic properties. It's defined directly on a dynamic sync table, or within the `dynamicOptions` of a regular sync table. The function can determine which property to provide autocomplete for by inspecting `context.propertyName`.
+
+=== "Dynamic sync table"
+
+    ```ts
+    pack.addDynamicSyncTable({
+      // ...
+      getSchema: async function (context) {
+        let attributes = await getCustomAttributes(context);
+        for (let attr of attriutes) {
```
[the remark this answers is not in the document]
ekoleda-codaio: Good catch, fixed.

docs/guides/blocks/sync-tables/two-way.md
ekoleda-codaio: Ya, I've added a TODO to go back and document this once we have it finalized.

docs/guides/blocks/sync-tables/two-way.md
ekoleda-codaio: Changed to "aren't able", but I don't want to get to much into "why" since it's complicated and really doesn't matter.

docs/guides/blocks/sync-tables/two-way.md
ekoleda-codaio: Good suggestion, done.

PR #2648, opened 2023-06-07
Fix isolated-vm install with standalone pnpm

dweitzman-codaio:
In https://github.com/coda/packs-sdk/pull/2629 I switched packs-sdk to a standalone pnpm install because pnpm requires a newer node version than the repo supports

An unintended side effect of using the standalone pnpm was that the node-gyp part of the isolated-vm install broke (https://github.com/pnpm/pnpm/issues/4347).

A workaround is to explicitly name node-gyp as a dev dependency of isolated-vm in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,6 +152,13 @@
     "overrides": {
       "vm2@<3.9.15": ">=3.9.15",
       "xml2js@<0.5.0": ">=0.5.0"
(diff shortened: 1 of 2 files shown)


huayang-codaio (approved):
👍

PR #2649, opened 2023-06-08
Introduce AuthenticationType.MultiHeaderToken.

jonathan-codaio:
Most of the use cases we've seen of people requesting exemptions to use Custom auth are just cases where the API they're using forces them to use multiple headers. So let's just offer that as a first-class auth type, much safer to do it ourselves and easier for the maker.

PTAL @ekoleda-codaio @huayang-codaio @dweitzman-codaio @coda/packs


test/auth_test.ts
```diff
@@ -170,6 +170,45 @@ describe('Auth', () => {
...
+          mockPromptForInput,
+          'Enter the token to use for the "Header2" header for this Pack:\n',
+          {mask: true},
+        );
+        sinon.assert.calledOnceWithExactly(mockPrint, 'Credentials updated!');
+
+        assertCredentialsFileExactly({headers: {Header1: 'token1', Header2: 'token2'}});
+      };
+
+      it('defaultAuthenticatio', () => execTest(createFakePack({defaultAuthentication: auth})));
```
dweitzman-codaio, line 208: Nit: missing the `n` at the end of `defaultAuthenticatio`

testing/upload_validation.ts
```diff
@@ -390,6 +391,25 @@ const defaultAuthenticationValidators: Record<AuthenticationType, z.ZodTypeAny>
...
+      .array(
+        zodCompleteStrictObject<MultiHeaderTokenAuthentication['headers'][number]>({
+          name: z.string(),
+          description: z.string(),
+          tokenPrefix: z.string().optional(),
+        }),
+      )
+      .refine(
+        headers => {
+          const keys = headers.map(header => header.name);
```
dweitzman-codaio, line 406: Might be worth making `head.name` lowercase here since HTTP headers ought to be case insensitive and it would be weird to have both `Header-Foo` and `header-foo` in the list

dweitzman-codaio (approved):

types.ts
```diff
@@ -55,6 +55,12 @@ export enum AuthenticationType {
    * @see {@link CustomHeaderTokenAuthentication}
    */
   CustomHeaderToken = 'CustomHeaderToken',
```
dweitzman-codaio, line 57: Not super important, but a question: do we plan to deprecate `CustomHeaderToken` and/or make `MultiHeaderToken` require at least one header?

If `MultiHeaderToken` can do everything `CustomHeaderToken` can, seems like there's a possible argument to consolidate

test/auth_test.ts
jonathan-codaio: Thanks, fixed.

testing/upload_validation.ts
jonathan-codaio: Yep great point, done.

types.ts
jonathan-codaio: Yeah, I think it might be reasonable to deprecated CustomHeaderToken. It is functionally redundant now. It is easier to use for the novice though, but maybe not so much to justify 3 flavors of header auth.

dweitzman-codaio: Hypothetically if we were thinking of consolidating, it might make sense to use a name that doesn't have `Multi` in it if it can support a single header just as well

I'm not proposing any name changes here, but figured it was a thought worth mentioning

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.3.6",
+  "version": "1.3.7",
```
ekoleda-codaio, line 3: Is this version bump that workaround for the `coda` repo? I seem to recall that the changelog tool doesn't like it when we skip versions.

types.ts
ekoleda-codaio: We could even go a step further and have a `Tokens` auth, that allows you to prompt for any number of query and/or header tokens. Not something we necessarily should do now, but there's probably an even wider solution space to consider.

ekoleda-codaio (approved):

types.ts
jonathan-codaio: Yeah, all fair, had been thinking about this too. I think if `CustomHeaderToken` wasn't already taken I would have used that or something similar, so this was kind of an effort to distinguish these.

What do people think about `GenericHeaderToken` or `GenericHeaderToken`, or is that too similar to `CustomHeaderToken`?

package.json
jonathan-codaio: Yeah it was for that workaround. Let me try again tomorrow without and see if it works ok.

Even with it I had trouble getting VSCode to find my new stuff, and wound up restarting the TS Server manually. So maybe that workaround would work for the version issue too.

types.ts
jonathan-codaio: @dweitzman-codaio @ekoleda-codaio Thoughts on `GenericHeaderToken` vs `MultiHeaderToken`?

dweitzman-codaio: Would it be crazy to reuse `AuthenticationType.CustomHeaderToken` and just have a different `BaseAuthentication` subclass so that either of these would work:

```
// Old version, implies type `CustomHeaderTokenAuthentication`
pack.setUserAuthentication({
  type: coda.AuthenticationType.CustomHeaderToken,
  headerName: "Authorization",
  tokenPrefix: "SSWS",
})

// New version, implies the new type `MultiHeaderTokenAuthentication` but has the same `AuthenticationType` value
pack.setUserAuthentication({
  type: coda.AuthenticationType.CustomHeaderToken,
  headers: [{
    name: "Authorization",
    description: "Foo",
    tokenPrefix: "SSWS",
  }]
})
```

It'd make our internal code more more complex since we'd have to check for `auth.headers` vs `auth.headerName` to know what we're dealing with, but it might keep the SDK surface simpler

jonathan-codaio: It's not a completely crazy idea. But I think breaking the use of a strict discriminated union would involve some amount of pain for us in code, for makers, and for documentation.

Where we're at right now is definitely goofy, but I'm not sure this case is common enough to bend over backwards for cleanliness.

ekoleda-codaio:
I thought I made this comment, but don't see it on GitHub: Does this PR also need to update the listing page's security tab to handle this new auth type?

types.ts
dweitzman-codaio: That's fair.

Since `GenericHeaderToken` feels inconsistent with existing naming and `CustomHeaderTokens` would be confusing since `CustomHeaderToken` already exists, I'm thinking that maybe having both `MultiHeaderToken` and `CustomHeaderToken` exist separately could be the least awkward option.

PR #2651, opened 2023-06-10
Fallback to reading api key from env var

abhishekamit:
When running in codespaces containers are ephemeral so its much more natural to store this in its secret management solution

### DIFF
--- a/cli/config_storage.ts
+++ b/cli/config_storage.ts
@@ -50,6 +50,7 @@ export function getApiKey(codaApiEndpoint: string): string | undefined {
       }
     }
   }
+  return process.env.CODA_PACKS_API_KEY;
 }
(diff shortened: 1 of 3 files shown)


cli/config_storage.ts
```diff
@@ -50,6 +50,7 @@ export function getApiKey(codaApiEndpoint: string): string | undefined {
       }
     }
   }
+  return process.env.PACKS_API_KEY;
```
jonathan-codaio, line 53: Nit, what about `CODA_PACKS_API_KEY`?

jonathan-codaio (approved):
Makes sense.

Mind adding a mention in the CHANGELOG about this?

abhishekamit:
> Makes sense.
> 
> Mind adding a mention in the CHANGELOG about this?

Done

PR #2655, opened 2023-06-12
Introduce ValueHintType.SelectList

dweitzman-codaio:
Context: http://go/select-list-value-hint

The overall idea is:
- You must set ValueHintType.SelectList before setting "autocomplete"
- In a follow-up PR, "autocomplete" will be renamed to "options". This is because options may be useful for the filter bar on a non-editable table or column.
- For now, this PR is restricting `SelectList` so it only works for strings and objects. We may add back booleans and numbers later so people can use the "display" properties to give friendly names, but this PR seems big enough already.

Some details about this PR:
- For a ValueType.String, the "autocomplete" field will only be defined in typescript when codaType is SelectList.
- For ValueType.Object, "autocomplete" will always be available on typescript and it'll just be a schema validation error to try to use it without the right codaType, to avoid getting into too big a mess with the typescript for ObjectSchemaDefinition.
- For Array values, you'd do something like this:
```
multiSelectProp: {
  type: ValueType.Array,
  mutable: true,
  items: {
    type: ValueType.String,
    codaType. ValueHintType.SelectList,
    // Renaming of "autocomplete" to "options" will be in another PR
    options: ['foo', 'bar', 'baz']
    // Potentially confusing detail: setting "mutable" at this level would compile, but does nothing.
  }
}
```

Only tested through unit tests so far, but I'll start working on the corresponding server-side code and test them both together before merging


api.ts
```diff
@@ -2516,7 +2525,11 @@ export function maybeRewriteConnectionForFormula<
 function listPropertiesWithAutocompleteFunctions(schema: ObjectSchemaDefinition<string, string>): string[] {
   const result: string[] = [];
   for (const propertyName of Object.keys(schema.properties)) {
-    const {autocomplete} = schema.properties[propertyName];
+    const propertySchema = maybeUnwrapArraySchema(schema.properties[propertyName]);
+    if (propertySchema?.codaType !== ValueHintType.SelectList) {
+      continue;
+    }
```
huayang-codaio, line 2531: would it be better to throw if autocomplete is defined but the coda type isn't select list? or maybe you already did this somewhere else. I'll keep reading.

schema.ts
```diff
@@ -1411,19 +1452,28 @@ export function makeObjectSchema<
...
 
       // Autocomplete gets manually copied over because it may be a function, which deepCopy wouldn't
       // support.
       if (autocompleteFunction) {
-        schema.properties[typedKey].autocomplete = autocompleteFunction;
+        const schemaCopyForAutocomplete = maybeUnwrapArraySchema(schema.properties[typedKey]);
+        if (schemaCopyForAutocomplete) {
+          assertCondition(schemaCopyForAutocomplete.codaType === ValueHintType.SelectList);
+          schemaCopyForAutocomplete.autocomplete = autocompleteFunction;
+        }
```
huayang-codaio, line 1476: 👍

testing/upload_validation.ts
```diff
@@ -1194,12 +1202,19 @@ const genericObjectSchema: z.ZodTypeAny = z.lazy(() =>
...
+  }, 'You must set "codaType" to ValueHintType.SelectList when setting an "autocomplete" property.')
+  .refine((schema: Schema) => {
+    const schemaForAutocomplete = maybeUnwrapArraySchema(schema);
+    return (
+      schemaForAutocomplete?.codaType !== ValueHintType.SelectList ||
+      !schemaForAutocomplete?.autocomplete ||
+      // NOTE: This is intentionally schema.mutable rather than schemaForAutocomplete.mtuable.
+      schema.mutable
+    );
+  }, `"mutable" must be true to set "autocomplete"`);
```
huayang-codaio, line 1217: this error message may need another copy.

huayang-codaio (commented):
lgtm overall.

schema.ts
```diff
@@ -1244,6 +1268,23 @@ export function isArray(val?: Schema): val is ArraySchema {
...
+): BooleanSchema | NumberSchema | StringSchema | GenericObjectSchema | undefined {
+  if (!isArray(val)) {
+    return val;
+  }
+
+  if (!isArray(val.items)) {
+    return val.items;
+  }
+
+  return undefined;
```
chris-codaio, line 1285: Nit: Superfluous return statement - just remove it.

chris-codaio (approved):
Cool!

ekoleda-codaio:
SDK surface LGTM.

schema.ts
```diff
@@ -777,14 +789,24 @@ export interface DurationSchema extends BaseStringSchema<ValueHintType.Duration>
   maxUnit?: DurationUnit;
 }
 
+/**
+ * A schema representing a value with selectable options.
+ */
+export interface StringWithOptionsSchema
+  extends BaseStringSchema<ValueHintType.SelectList>,
+    PropertyWithAutocompleteWithOptionalDisplay<string> {
+  /** Instructs Coda to render this value as a hyperlink. */
```
jonathan-codaio, line 798: Nit, copy paste issue in comment.

api.ts
dweitzman-codaio: There's a zod error in that scenario, but it was missing a unit test which I just added. The zod validation error will say, `You must set "codaType" to ValueHintType.SelectList when setting an "autocomplete" property`

schema.ts
dweitzman-codaio: Fixed

schema.ts
dweitzman-codaio: Done

testing/upload_validation.ts
dweitzman-codaio: A follow-up PR will both rename "autocomplete" to "options" and also remove this restriction entirely since we want people to be able to set up options for view-only filters.

I updated the check here instead of removing it just to avoid scope creep in this PR

dweitzman-codaio:
A few more changes:
* Allow `autocomplete` when the `codaType` is `Reference`. In that code, that mainly involved changing `schema.codaType === ValueHintType.SelectList` to use a new `unwrappedSchemaSupportsAutocomplete(schema)` function which checks for either `SelectList` or `Reference`
* I removed the requirement to set `mutable` when using `autocomplete` now instead of waiting for a future PR since it seemed easier than updating the logic in this PR and then immediately deleting it in a follow-up PR

Corresponding PR is https://github.com/coda/coda/pull/93298

huayang-codaio (approved):

PR #2657, opened 2023-06-15
Rename property "autocomplete" to "options"

dweitzman-codaio:
This will support a better user experience with read-only tables or columns and the filter bar.

In sync table parameters, I renamed "autocomplete" to "propertyOptions" so it's a little more clear that the options are related to properties rather than the overall sync table.

Example usage:

```
properties: {
  color: {
    type: coda.ValueType.String,
    codaType: coda.ValueHintType.SelectList,
    mutable: true,
    options: ['red', 'green', 'blue'], // Previously "autocomplete"
  },
  // ...
```

A sync table might look like this now (`autocomplete` -> `propertyOptions`):

```
makeSyncTable({
  name: 'SomeSync',
  // ...
  dynamicOptions: {
    propertyOptions: context => {  // Previously "autocomplete"
      return context.propertyName === 'foo' ? ['foo1', 'foo2'] : [];
    },
  },
```

For packs compiled prior to this change, any `autocomplete` will start to be ignored once the server-side part of this rolls out


chris-codaio (approved):
Assuming this is a glorified search/replace, I'll just rubber stamp.

schema.ts
```diff
@@ -329,9 +324,6 @@ export interface BaseNumberSchema<T extends NumberHintTypes = NumberHintTypes> e
   type: ValueType.Number;
   /** An optional type hint instructing Coda about how to interpret or render this value. */
   codaType?: T;
-
```
dweitzman-codaio, line 332: Note: here and in upload validation I removed some `autocomplete` declarations for numbers and booleans since for the moment we're only supporting strings and objects

schema.ts
```diff
@@ -799,9 +791,6 @@ export interface StringWithOptionsSchema
     PropertyWithAutocompleteWithOptionalDisplay<string> {
   /** Instructs Coda to render this value as a select list. */
   codaType: ValueHintType.SelectList;
-
-  /** @hidden */
-  autocomplete: PropertySchemaAutocompleteWithOptionalDisplay<string>;
```
dweitzman-codaio, line 804: Note: removed the property because it's already defined by the `PropertyWithAutocompleteWithOptionalDisplay<string>` parent interface.

test/schema_test.ts
```diff
@@ -278,7 +309,13 @@ describe('Schema', () => {
           fromKey: 'another',
           properties: {
             Boo: {type: schema.ValueType.String, fromKey: 'boo'},
-            Baz: {type: schema.ValueType.Number, fromKey: 'baz', mutable: true, autocomplete: AutocompleteType.Dynamic},
+            Baz: {
+              type: schema.ValueType.String,
```
dweitzman-codaio, line 313: Changed this test to use to a `ValueType.String` since `Number` doesn't support `options`

test/schema_test.ts
```diff
@@ -233,7 +234,32 @@ describe('Schema', () => {
...
       // Make sure the function is preserved in spite of the missing value hint across
       // the deep copy that makeObjectSchema() does.
-      assert.deepEqual(normalizedBoo.items.autocomplete(), ['1']);
+
+      const options = normalizedBoo.items.options;
+      assert.isFunction(options);
+      assert.deepEqual(options({} as PropertyOptionsExecutionContext), ['1']);
+    });
+
+    it('options functions are disallowed in dynamic sync tables', () => {
```
dweitzman-codaio, line 243: Added a test for `throwOnDynamicSchemaWithJsOptionsFunction` while I was in here

dweitzman-codaio (commented):
It's mostly glorified find and replace. I'm adding comments in a few places with actual code changes

api_types.ts
```diff
@@ -881,25 +881,25 @@ export enum PrecannedDateRange {
 /**
  * @hidden
```
ekoleda-codaio, line 882: Should this be un-hidden now? I may have missed it in my PR that added documentation.

runtime/types.ts
```diff
@@ -20,7 +20,7 @@ export enum MetadataFormulaType {
   SyncGetDisplayUrl = 'SyncGetDisplayUrl',
   SyncGetTableName = 'SyncGetTableName',
   SyncGetSchema = 'SyncGetSchema',
-  PropertyAutocomplete = 'PropertyAutocomplete',
+  PropertyOptions = 'PropertyOptions',
```
ekoleda-codaio, line 23: Will changing the enum value break any existing Packs using this? Fine to break things in an alpha, but would want to give a heads up.

ekoleda-codaio (approved):

runtime/types.ts
dweitzman-codaio: Existing `autocomplete` values will stop working for alpha testers, although not specifically from this enum change.

Packs should still run but will start behaving as if no `autocomplete` value was set. Makers can get autocomplete in their packs working again just be renaming to `options` and setting `codaType: ValueHintType.SelectList` on the property.

api_types.ts
dweitzman-codaio: I'll make a follow-up PR to unhide this and a few other things

PR #2662, opened 2023-06-22
Bump SDK version 1.4.0 to 1.4.1

dweitzman-codaio:
Same reason as https://github.com/coda/packs-sdk/pull/2467 -- somehow this impacts lint performance

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.4.1",
+  "version": "1.4.1-prerelease.1",
   "license": "MIT",
   "workspaces": [
     "dev/eslint"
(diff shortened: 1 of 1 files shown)


neal-codaio (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.4.0",
+  "version": "1.4.1",
```
jonathan-codaio, line 3: Per my Slack comment, how do we feel about something like `1.4.1-prelease1`, so we can increment the suffix each time we make a change rather than having to make our actual public semantic versions non-contiguous?

I think the `semver` checks we do in `coda` are ok with this syntax, but obviously needs a little bit of verification.

dweitzman-codaio: Changed to `1.4.1-prerelease.1`

jonathan-codaio (approved):

PR #2665, opened 2023-06-26
Save metadata during "coda build"

dweitzman-codaio:
I don't think there are any other commands that can print the metadata, but it can be helpful for debugging.

It's marked as "hidden" since it seems useful to the packs team but not necessarily for pack makers.


jonathan-codaio:
Cool, this is definitely super useful.

Thoughts on doing this as part of `coda build`, since that already outputs compiled output to a temp directory? Should we just add the metadata to the temporary output?

dweitzman-codaio:
> Cool, this is definitely super useful.
> 
> Thoughts on doing this as part of `coda build`, since that already outputs compiled output to a temp directory? Should we just add the metadata to the temporary output?

Makes sense to me. Updated the PR

testing/compile.ts
```diff
@@ -271,6 +273,12 @@ export async function compilePackBundle({
     throw await processVmError(err, tempBundlePath);
   }
 
+  // Write the generated metadata. It's not used by the upload command, but
+  // it's helpful for debugging upload validation errors.
+  const metadata = await importManifest<PackVersionDefinition>(tempBundlePath);
+  const tempMetadataPath = path.join(intermediateOutputDirectory, 'metadata.json');
+  fs.writeFileSync(tempMetadataPath, JSON.stringify(metadata));
```
jonathan-codaio, line 280: This should be writing the result of `compilePackMetadata`, right?

dweitzman-codaio: Good catch, fixed

jonathan-codaio (approved):

huayang-codaio (commented):
can be very useful. thank you

PR #2666, opened 2023-06-26
Allow older SDK versions to upload after autocomplete breaking change

dweitzman-codaio:
I didn't fully test the scenario of an old SDK with a newer server, and the previous SDK versions were inadvertently setting "namedAutocompletes" even for people not explicitly using the feature 😢 

People using old SDK versions are currently hitting errors uploading packs to Coda


jonathan-codaio (approved):
Should CP  a bump with this into `coda` after today's snsap.

PR #2669, opened 2023-07-08
Fix video loading on certain browsers

ekoleda-codaio:
As per bug-33085, in Chrome on Android and Arc on Mac the animated-gifs-turned-mp4s weren't playing. The problem appears to be related to the codec and settings used to encode the video. Through searching and testing it appears that the pixel format was incompatible. I ran the script below to change the pixel format for all videos using `ffmpeg`.

```sh
find . -maxdepth 1 -type f -name "*.mp4" | xargs -I {} sh -c 'ffmpeg -y -loglevel error -i {} -c:v libx264 -pix_fmt yuv420p -vf "scale=trunc(iw/2)*2:trunc(ih/2)*2" -c:a copy {}.converted.mp4 && echo "Converted {} to {}.converted.mp4"' && find . -maxdepth 1 -type f -name "*.converted.mp4" | xargs -I {} sh -c 'mv {} $(dirname {})/$(basename {} .converted.mp4)'
```

[Preview](https://coda-packs-sdk--2669.com.readthedocs.build/en/2669/tutorials/get-started/web/#navigate-to-the-pack-studio)


jonathan-codaio (approved):
Nice sleuthing.

Does it make any sense to host these on YouTube and just embed them?

ekoleda-codaio:
> Does it make any sense to host these on YouTube and just embed them?

I had considered it, especially since this world of video compatibility is quite opaque to me. However these were originally animated gifs, and the desired UX is a short, looping screen recording showing a couple of clicks. Requiring the user to click a play button, and maybe get an ad, seems like it would be out of place for a few seconds of content.

My thought was to try this approach for now, but if we continue to run into issues self-hosting these to look for a more robust alternative.

PR #2673, opened 2023-07-13
Partially update typescript and related dependencies

patrick-codaio:
These dep updates were failing on their own. 

I hit a snag with typescript 5.1, but updating it past 5.0 still seemed worthwhile.


requirements.txt
```diff
@@ -1,47 +1,47 @@
 # Autogenerated by 'make bs'
 # If the format has changed, update MIN_PIPENV_VERSION in Makefile
 -i https://pypi.org/simple
-bracex==2.3.post1 ; python_version >= '3.7'
```
patrick-codaio, line 4: I'm working on figuring out what the new MIN_PIPENV_VERSION should be

docs/reference/sdk/interfaces/core.MultiHeaderTokenAuthentication.md
```diff
@@ -39,7 +39,7 @@ When requiresEndpointUrl is set to true this should be the root domain that all
 For example, this value would be "example.com" if specific endpoints looked like {custom-subdomain}.example.com.
 
 For packs that make requests to multiple domains (uncommon), this should be the domain within
-[networkDomains](core.PackVersionDefinition.md#networkdomains) that this configuration applies to.
+[networkDomains](core.PackDefinition.md#networkdomains) that this configuration applies to.
```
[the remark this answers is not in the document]
ekoleda-codaio: Very interesting. I believe the `core.` prefix was needed then because the types crossed modules (`core` vs `development`), and I think it failed to resolve the link without it. With this PR they are resolving, just in strange ways.

Looking at the typedoc change log, I saw this entry:

> Added --useTsLinkResolution option (on by default) which tells TypeDoc to use TypeScript's @link resolution.

Does turning that off also resolve the issue? If not there are a number of other altered links in the `docs/` directory they also need addressing then.

testing/validation.ts
```diff
@@ -251,10 +251,16 @@ function tryParseEmail(result: unknown, schema: BaseStringSchema): ValidationErr
 function tryParseSlider(result: unknown, schema: SliderSchema | ProgressBarSchema) {
   const value = result as number;
   const {minimum, maximum} = schema;
-  if (typeof minimum !== 'string' && value < (minimum ?? 0)) {
+  if (typeof minimum === 'string') {
+    return {message: `Slider minimum value must be a number, but got string: "${minimum}".`};
```
dweitzman-codaio, line 255: I don't think a string value here is technically disallowed?

My understanding is that the slider minimum value can basically be a formula which is why it has a type of `string | number`, even though nothing else in packs SDK allows people to embed formulas.

I think maybe @jonathan-codaio knows more context on sliders with string min/max values?...

docs/reference/sdk/interfaces/core.MultiHeaderTokenAuthentication.md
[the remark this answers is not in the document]
patrick-codaio: That *does* address the issue, though it comes with a couple other small changes, but I think they're fine.

testing/validation.ts
patrick-codaio: Ok I can just revert this file, it's an unrelated change.

docs/reference/sdk/interfaces/core.MultiHeaderTokenAuthentication.md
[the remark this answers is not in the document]
ekoleda-codaio: Ya, that LGTM. I'll take an action item to look into `--useTsLinkResolution ` and perhaps migrate to it in a future PR.

requirements.txt
patrick-codaio: Not having luck with this. If my env is the only one making this format change, I can just revert this file manually.

docs/reference/sdk/classes/core.PackDefinitionBuilder.md
```diff
@@ -385,7 +385,7 @@ pack.setUserAuthentication({
...
+| `authDef` | `Object` |
```
dweitzman-codaio, line 388: I wonder if typedoc would make more of an effort to document the parameters to `setUserAuthentication()` here if the `AuthenticationDef` type was re-exported from `index.ts`. Or even if it doesn't, it might be useful to manually link to `AuthenticationDef` from the parameter description or something.

I don't have strong feelings about this, but if there's a simple way to continue directing people looking at this parameter to something more descriptive than "object" that could be nice.

dweitzman-codaio (approved):

docs/reference/sdk/classes/core.PackDefinitionBuilder.md
patrick-codaio: I introduced a new `UserAuthenticationDef` type so that there's something other than "object" here.

PR #2676, opened 2023-07-13
Add sample Pack demonstrating property options

ekoleda-codaio:


docs/samples/topic/two-way.md
```diff
@@ -396,4 +396,237 @@ pack.setUserAuthentication({
...
+        continuation: continuation,
+      };
+    },
+    maxUpdateBatchSize: 10,
+    executeUpdate: async function (args, updates, context) {
+      // Make all of the updates in parallel to improve performance.
+      let jobs = updates.map(async update => {
+        return updateExpense(context, update);
+      });
+      let completed = await Promise.allSettled(jobs);
```
dweitzman-codaio, line 561: Side note: I don't know if we formed any conclusion on whether whether/how to make a helper function or option to do the `Promise.allSettled` but unless someone has a better way I'd be in favor of making a standard helper function for this at least

dweitzman-codaio (approved):

docs/samples/topic/two-way.md
ekoleda-codaio: No objections here!

PR #2677, opened 2023-07-20
More specific typing for PropertyOptionsExecutionContext.propertySchema

dweitzman-codaio:
Adjusting the typing here allows packs to read "fromKey" without any typescript casts.


huayang-codaio (approved):

patrick-codaio:
LGTM, though I'm no typescript expert.

dweitzman-codaio:
I'm not sure how best to resolve the typedoc validation issue here. CI is correct that `CodaInternalRichTextSchema` is not a documented type.

The hidden rich text type is actually already mentioned in the `StringSchema` type reference doc, which somehow hasn't triggered this test failure.

I think I'll leave this PR alone until Eric is back and I may merge a typecast in another repo in the meantime with a TODO to clean it up after some version of this PR is merged

jonathan-codaio (commented):
Can you try adding that token in `intentionallyNotExported` in `typedoc.js`? I think that powers exemptions from this check.

dweitzman-codaio:
> Can you try adding that token in `intentionallyNotExported` in `typedoc.js`? I think that powers exemptions from this check.

Seems like `intentionallyNotExported` still prints a warning since the types are exported, just not documented, and printing any warning is enough to make the test mad

dweitzman-codaio:
Rebased this and the typedoc problem seems to have mysteriously gone away. Now the documentation just gives the type of `propertyOptions` as `object` rather than a giant list of all the union types it could contain, avoiding references to `@hidden` internal types.

I don't know why the behavior is different now -- maybe a dependency upgrade?

I've also added an explicit `{@link Schema}` to help people who are browsing around and want to find the relevant typedoc for the field.

PR #2678, opened 2023-07-20
[2-way] Select list option to allow user-added strings

dweitzman-codaio:
For string select lists we can support user-added strings if the pack maker allows it.

I don't particular like the name `customOptions`, so any suggestions for better names would be welcome

Objects are not supported because we'd need some way to ask the pack to construct and object from a string. In the future I think it could be nice to support some kind of a generic object builder so people can type in pack objects, but that's not in scope now.


schema.ts
```diff
@@ -790,6 +790,8 @@ export interface StringWithOptionsSchema
     PropertyWithAutocompleteWithOptionalDisplay<string> {
   /** Instructs Coda to render this value as a select list. */
   codaType: ValueHintType.SelectList;
+
+  customOptions?: boolean;
```
patrick-codaio, line 794: how about allowNewValues / allowNewOptions ?

huayang-codaio (approved):

PR #2679, opened 2023-07-24
[2-way autocomplete] Update typing to allow cacheTtlSecs on return values

dweitzman-codaio:
This is purely a typescript typing change to avoid compilation errors in pack code.

normalizePropertyOptionsResults() already does the right thing at runtime by wrapping arrays in this type if it wasn't returned by the function.


huayang-codaio (approved):
makes sense to me.

api_types.ts
```diff
@@ -912,10 +912,13 @@ export type OptionsReference = string & {
   __brand: 'OptionsRef';
 };
 
-// TODO(dweitzman): Update the type to allow options functions to return a cache TTL.
+type PropertyOptionsMetadataResult<ResultT extends PackFormulaResult[]> =
+  | ResultT
+  | {results: ResultT; cacheTtlSecs?: number};
```
ekoleda-codaio, line 917: For sync table `execute` and `executeUpdate` the key holding the results is `result` (singular). Perhaps use the same here?

dweitzman-codaio: Consistency makes sense to me. Renamed here and updated the normalization logic. 

To avoid a breaking change, the hidden internal type `PropertyOptionsAnnotatedResult` still uses `results` (plural) but pack makers won't ever need that type.

PR #2681, opened 2023-07-26
Start tracking originalKey on object properties

patrick-codaio:
To see how this works out in practice, here are the accompanying changes:
- in `packs`: https://github.com/coda/packs/pull/3692
- in `coda`: https://github.com/coda/coda/pull/95483


api.ts
```diff
@@ -2151,9 +2153,10 @@ export function makeSyncTable<
     });
   }
 
-  const formulaSchema = getSchema
+  const normalizedSchema = normalizeSchema(schema);
```
patrick-codaio, line 2156: this is to make sure we don't call normalize twice on the same schema

schema.ts
```diff
@@ -1615,75 +1645,90 @@ export function normalizeSchema<T extends Schema>(schema: T): T {
...
+    };
+    if (fixedId) {
+      normalizedProps.fixedId = fixedId;
+    }
+    if (mutable) {
+      normalizedProps.mutable = mutable;
+    }
+    if (required) {
+      normalizedProps.required = required;
+    }
```
jonathan-codaio, line 1702: Any reason not to just pass these through as-is like before? In particular, for the boolean ones, this will technically be a change in output, because values specified as `false` will now be omitted. I would vote for not churning more than necessary here.

schema.ts
```diff
@@ -1706,16 +1752,21 @@ export function makeReferenceSchemaFromObjectSchema(
...
-    mutable,
-    options,
-  });
+  };
+  if (mutable) {
+    referenceSchema.mutable = mutable;
+  }
+  if (options) {
+    referenceSchema.options = options;
+  }
```
jonathan-codaio, line 1768: Ditto here about churning this.

api.ts
patrick-codaio: I added some tests, including a bug fix I found by using this SDK version in the packs repo.

schema.ts
```diff
@@ -883,11 +867,41 @@ export interface ObjectSchemaProperty {
...
+   * changes to the name of this property will not cause the property to be treated as a new property.
+   * Only supported for top-level properties of a sync table.
+   * Note that fixedIds must already be present on the existing schema prior to rolling out a name change in a
+   * new schema; adding fixedId and a name change in a single schema version change will not work.
+   * @hidden
+   */
+  fixedId?: string;
+
+  /**
+   * For internal use only.
```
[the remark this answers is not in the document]
patrick-codaio: Totally agree about completely splitting the maker-facing types from the post-build types, I put in a few TODOs about that. I played with it and it seems like a pretty big change to try to tackle right now.

Updated comment.

schema.ts
patrick-codaio: Ah, definitely need to at least pass through `false`. I just reverted this churn and the other.

schema.ts
```diff
@@ -276,22 +276,6 @@ interface BaseSchema {
    * explain the purpose or contents of any property that is not self-evident.
    */
   description?: string;
-
-  /**
-   * Whether this object schema property is editable by the user in the UI.
-   */
-  /** @hidden */
-  mutable?: boolean;
```
patrick-codaio, line 284: @chris-codaio @dweitzman-codaio @huayang-codaio Is moving these a bad idea for some reason I'm not seeing? I think the resulting changes in other repos look sensible to me:
- in `packs`: https://github.com/coda/packs/pull/3692
- in `coda`: https://github.com/coda/coda/pull/95483

schema.ts
```diff
@@ -1247,6 +1273,9 @@ export function makeAttributionNode<T extends AttributionNode>(node: T): T {
 
 /**
  * The union of all of the schema types supported for return values and object properties.
+ *
+ * TODO(patrick): GenericObjectSchema is designed to be a runtime type, as it requires identities
```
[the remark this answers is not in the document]
patrick-codaio: Great point, done.

schema.ts
dweitzman-codaio: This doesn't feel unreasonable to me.

Seems like it doesn't break anything interesting in the other repos, and even if we're forgetting about something there's no irreversible harm from shuffling typescript types around.

jonathan-codaio (approved):

schema.ts
huayang-codaio: Makes sense.

schema.ts
```diff
@@ -905,6 +920,12 @@ export type ObjectSchemaProperties<K extends string = never> = {
 /** @hidden */
 export type GenericObjectSchema = ObjectSchema<string, string>;
 
+/**
+ * When an object schema is being used as a property of another object, then it additionally has
+ * ObjectSchemaProperty properties.
+ */
+export type PropertyObjectSchema = GenericObjectSchema & ObjectSchemaProperty;
```
huayang-codaio, line 927: I would probably either skip this, or create `PropertyBooleanSchema`, `PropertyNumberSchema`, etc as well.

schema.ts
```diff
@@ -1122,6 +1143,12 @@ export type ObjectSchemaDefinitionType<
   T extends ObjectSchemaDefinition<K, L>,
 > = ObjectSchemaType<T>;
 
+/**
+ * When an object schema is being used as a property of another object, then it additionally has
+ * ObjectSchemaProperty properties.
+ */
+export type PropertyObjectSchemaDefinition = ObjectSchemaDefinition<string, string> & ObjectSchemaProperty;
```
huayang-codaio, line 1150: ditto.

huayang-codaio (approved):
lgtm. thanks for the cleanups when you worked on this PR.

PR #2700, opened 2023-08-09
Only use v8 serialization internally within the thunk

dweitzman-codaio:
_**This is a draft PR, not ready for review yet**_

### DIFF
--- a/helpers/legacy_marshal/index.ts
+++ b/helpers/legacy_marshal/index.ts
@@ -91,7 +91,7 @@ function processValue(val: any, depth: number = 0): any {
   return serializedValue;
 }
 
-export function legacyMarshalValue(val: any): string | undefined {
+export function marshalValueForAnyNodeVersion(val: any): string | undefined {
   // Instead of passing a replacer to `JSON.stringify`, we chose to preprocess the value before
   // passing it to `JSON.stringify`. The reason is that `JSON.stringify` may call the object toJSON
   // method before calling the replacer. In many cases, that means the replacer can't tell if the
@@ -124,7 +124,7 @@ export function legacyWrapError(err: Error): Error {
   //     'to actually fetch from the remote API.';
   // }
 
-  return new Error(legacyMarshalValue(err));
+  return new Error(marshalValueForAnyNodeVersion(err));
 }
 
 export function legacyUnwrapError(err: Error): Error {
--- a/helpers/legacy_marshal/index.ts
+++ b/helpers/legacy_marshal/index.ts
@@ -91,7 +91,7 @@ function processValue(val: any, depth: number = 0): any {
   return serializedValue;
 }
 
-export function marshalValueForAnyNodeVersion(val: any): string | undefined {
+export function marshalValueForAnyNodeVersion(val: any): string {
   // Instead of passing a replacer to `JSON.stringify`, we chose to preprocess the value before
   // passing it to `JSON.stringify`. The reason is that `JSON.stringify` may call the object toJSON
   // method before calling the replacer. In many cases, that means the replacer can't tell if the
--- a/runtime/common/marshaling/index.ts
+++ b/runtime/common/marshaling/index.ts
@@ -179,7 +179,7 @@ export function marshalValueToStringForSameOrHigherNodeVersion(val: any): string
 }
 
 export function unmarshalValueFromString(marshaledValue: string): any {
-  if (marshaledValue.startsWith('/')) {
+  if (marshaledValue?.startsWith('/')) {
     // Looks like a v8-serialized value
     return unmarshalValue(deserialize(marshaledValue));
   }
--- a/runtime/common/marshaling/index.ts
+++ b/runtime/common/marshaling/index.ts
@@ -4,6 +4,7 @@ import {MissingScopesError} from '../../../api';
 import {StatusCodeError} from '../../../api';
 import {deserialize} from './serializer';
 import {format} from 'util';
+import {legacyUnmarshalValue} from '../../../helpers/legacy_marshal';
 import {serialize} from './serializer';
 
 // We rely on the javascript structuredClone() algorithm to copy arguments and results into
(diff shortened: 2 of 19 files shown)


runtime/common/marshaling/index.ts
```diff
@@ -173,12 +174,17 @@ export function marshalValue(val: any): MarshaledValue {
...
-export function marshalValueToString(val: any): string {
+export function marshalValueToStringForSameOrHigherNodeVersion(val: any): string {
   return serialize(marshalValue(val));
 }
 
 export function unmarshalValueFromString(marshaledValue: string): any {
-  return unmarshalValue(deserialize(marshaledValue));
+  if (marshaledValue?.startsWith('/')) {
+    // Looks like a v8-serialized value
+    return unmarshalValue(deserialize(marshaledValue));
```
chris-codaio, line 184: Maybe a try/catch around `deserialize` in case that fails?

dweitzman-codaio: What error handling were you thinking here?

I don't think we'd want to suppress any errors here for fear of moving incorrect data around. This can't be a legacy marshaled value if it starts with "/" since that can't be the beginning of JSON string.

chris-codaio: Ah, I suppose that's sufficient then.

helpers/legacy_marshal/index.ts
```diff
@@ -91,7 +91,7 @@ function processValue(val: any, depth: number = 0): any {
   return serializedValue;
 }
 
-export function legacyMarshalValue(val: any): string | undefined {
+export function marshalValueForAnyNodeVersion(val: any): string {
```
jonathan-codaio, line 94: Do you want to rename the unmarshaler too?

test/test_helper.ts
```diff
@@ -6,9 +6,13 @@ export const testHelper = {
    * @param promise The original promise.
    * @returns The inverted promise, which fulfills if rejected and rejects if fulfilled.
    */
-  async willBeRejected<ErrorT = any>(promise: Promise<any>): Promise<ErrorT> {
+  async willBeRejected<ErrorT = any>(promise: Promise<any> | (() => Promise<any>)): Promise<ErrorT> {
     try {
-      await promise;
+      if (typeof promise === 'function') {
```
jonathan-codaio, line 11: Nit, how come we need this? Should be able to just call the function at the call site if it truly returns a promise.

jonathan-codaio (commented):
Overall LGTM once heavily tested.

runtime/common/marshaling/index.ts
```diff
@@ -173,12 +185,72 @@ export function marshalValue(val: any): MarshaledValue {
...
+  if (useUnsafeVersionCompatibilityHack) {
+    return maybeChangeWireVersionOnBase64EncodedV8SerializedData(serialized);
+  }
+
+  return serialized;
 }
 
 export function unmarshalValueFromString(marshaledValue: string): any {
-  return unmarshalValue(deserialize(marshaledValue));
+  if (marshaledValue?.startsWith('/')) {
```
huayang-codaio, line 248: if `marshaledValue` could be undefined, maybe update the parameter type as well.

runtime/common/marshaling/index.ts
```diff
@@ -173,12 +185,72 @@ export function marshalValue(val: any): MarshaledValue {
...
+// This is a temporary hack for ease of node upgrades. It turns out node's documentation
+// is incorrect when it says v8.serialize is backwards compatible: moving from node 14/16
+// to node 18 changes the v8 serialize wire version from 13 to 15, which can't be read
+// by node 14 or 16. If we look at what actually changed in the wire format, it shouldn't
+// impact us as long as we don't try to let people serialize buffer views.
+// The changes were:
+// - https://chromium-review.googlesource.com/c/v8/v8/+/3417189 (support for shared strings)
+// - https://chromium-review.googlesource.com/c/v8/v8/+/3386802 (a change to buffer views)
+// The hacky temporary strategy here is to manually change the wire version back from 14 or 15
+// to 13 by decoding and re-encoding the first few bytes in the string.
```
huayang-codaio, line 197: amazing. thank you.

huayang-codaio (approved):
makes sense to me. 

do you plan to rebuild all (or selected) the pack bundles after releasing this change?

huayang-codaio:
re testing: can use an adhoc.

test/marshaling_test.ts
```diff
@@ -72,6 +83,52 @@ describe('Marshaling', () => {
     assert.deepEqual(transform(new SomeClass('hi')), {message: 'hi'});
   });
 
+  it('can write values readable on node 14 from node 18', async () => {
```
dweitzman-codaio, line 86: FYI, I've added a test case for the node18 backwards compatibility by hard-coding serialized output here. I manually ran just this one test case on node18 and it passed, showing that for at least these cases it's harmless to swap the wire version. I copied these test cases from above, but happy to add more if anything else seems worthwhile

test/test_helper.ts
dweitzman-codaio: Oops, I should have used `assert.throws()`. Removed this

runtime/common/marshaling/index.ts
dweitzman-codaio: I can get rid of the question mark here. The `undefined` I had encountered was actually a weird edge case where typed arrays would get dropped / turned into undefined but since those are officially not supported anymore I just dropped that test case

dweitzman-codaio:
> makes sense to me.
> 
> do you plan to rebuild all (or selected) the pack bundles after releasing this change?

No. The wacky serialization backwards-compatibility hack is so we can avoid that:
- If old packs return old serialized data, that's fine: a server running a newer node version can still read old data (it's forwards-compatible)
- The server will write data to packs using the serialization hack in a 14-compatible format

The only problem would be if we have node14 lambdas still alive for a loooong time and we want to upgrade the server yet again and there's a v8.serialization change we can't hack our way around. AWS is deprecating lambdas with node 14 in November, so we'll probably get kicked off node14 by AWS and need to rebuild everything at node18 long before we run into serialization compatibility issues.

helpers/legacy_marshal/index.ts
dweitzman-codaio: Renamed to internalUnmarshalValueForAnyNodeVersion

chris-codaio (approved):

PR #2701, opened 2023-08-09
make endpoint key available in external metadata

gary-codaio:
so we can check it when we consider if we want to show oauth as configurable for pack configurations


jonathan-codaio (approved):
I think this is making it available in the _external_ metadata (it already exists in the internal metadata and is referenceable on the server, just not on the client).

PR #2703, opened 2023-08-14
[Node18] Ensure we build docker images for x86_64 architecture

chris-codaio:
We've been running into problems where we build the SDK from a Mac arm machine, but then try to deploy that build to lambda in AWS which runs x86_64.

For now, this code assumes we always want x86_64.  We may run into Rosetta2 emulator issues when trying to run this on an arm64 dev machine, in which case we'll likely need to have two versions of this compilation step run, then be sure to select the correct one during test time vs. run time.  Will see how that lands and followup as necessary.

PTAL @coda/packs @dweitzman-codaio

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ do-compile-isolated-vm:
 	mkdir build-isolated-vm && \
 		cd build-isolated-vm && \
 		npm init -y && \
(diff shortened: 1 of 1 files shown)


dweitzman-codaio (approved):

huayang-codaio:
Nice

PR #2732, opened 2023-09-01
Add `MetadataContext` as a Sync property

alan-codaio:
SDK change needed to support reading `MetadataContext` in the `PropertyOptionsMetadataFunction` (although this change can allow other formulas to also read `MetadataContext`


api_types.ts
```diff
@@ -701,6 +702,11 @@ export interface Sync {
    * The dynamic URL is likely necessary for determining which API resources to fetch.
    */
   dynamicUrl?: string;
+
+  /**
+   * {@link MetadataContext}
+   */
+  metadataContext?: MetadataContext;
```
dweitzman-codaio, line 709: Maybe a name like `parameters` would make it more clear what'll be here? cc @ekoleda-codaio

api_types.ts
```diff
@@ -701,6 +702,11 @@ export interface Sync {
    * The dynamic URL is likely necessary for determining which API resources to fetch.
    */
   dynamicUrl?: string;
+
+  /**
```
dweitzman-codaio, line 706: Another thing to consider here is whether we want to add this for all ExecutionContext types or whether we might want to initially only do it for `PropertyOptionsExecutionContext` by giving it a `PropertyOptionsSync` sub-interface. That'd avoid the need to worry about populating this for pre-existing formulas but still leave the door open to generalize it later

patrick-codaio: Even if we don't support it on all metadata formulas *yet*, I personally think it's better to go with all execution contexts for sync tables, as I think we should get there eventually. 

Actually, I daresay it should go with all execution contexts for formulas & actions too, so maybe even move it higher?

api_types.ts
patrick-codaio: Yeah... I'm curious if @alexd-codaio had a reason for not using an interface name more like `Parameters` when he first introduced MetadataContext?

api_types.ts
patrick-codaio: Actually, looking at this a little more, why can't `PropertyOptionsMetadataFormula` extend `MetadataFunction` and get this for free?

alan-codaio: I believe that the reasoning is that, the `MetadataFunction` interface is a bit overloaded (it has a number of parameters that only pertain to specific types of metadata functions, like `search` only being relevant to autocomplete functions). There was also some mention of the positionality of the arguments, although I can't remember off the top of my head the details about that discussion since `MetadataContext` doesn't hold any data about the positional arguments. 

So the thought was to declare `PropertyOptions` as a separate interface to keep things cleaner.

alan-codaio: More on the positionality: I remember @dweitzman-codaio was talking about a scenario where we applied varargs with the same name i.e. (`cake`, `cake_flavor`, `cake`, `cake_flavor`...) and attempted to do an autocomplete accounting for the aggregate of this info — `MetadataContext` would need a different interface shape to properly handle this scenario. (But I can't remember why this specific example isn't related to not extending `MetadataFunction`)

patrick-codaio: ah, yeah we should make MetadataFunction able to take in varargs

api_types.ts
```diff
@@ -775,6 +776,11 @@ export interface ExecutionContext {
    * Information about state of the current sync. Only populated if this is a sync table formula.
    */
   readonly sync?: Sync;
+
+  /**
+   * {@link core.MetadataContext}
+   */
+  parameters?: MetadataContext;
```
patrick-codaio, line 783: looks like everything else is readonly here... any reason not to put that on this too?

alan-codaio: oh yeah this should e readonly too

api_types.ts
```diff
@@ -775,6 +776,11 @@ export interface ExecutionContext {
    * Information about state of the current sync. Only populated if this is a sync table formula.
    */
   readonly sync?: Sync;
+
+  /**
+   * {@link core.MetadataContext}
+   */
+  readonly parameters?: MetadataContext;
```
ekoleda-codaio, line 783: I've moved towards referring to these as "arguments" instead of "parameters" in my sample code, which I think is more semantically correct. So I probably prefer `arguments` here instead.

ekoleda-codaio, line 783: As mentioned in another comment, `MetadataContext` doesn't support multiple sets of vararg arguments, which is not ideal. If we are going to introduce a new way to access arguments across all contexts, it would be great if it also resolved that issue.

alan-codaio: that's reasonable, although I'd prefer to do it in a separate followup. (this one's already had its scope increased from being only accessible by pack property options metadata formulas to all formulas)

alan-codaio: I was talking to @patrick-codaio on Friday, and he pointed out that even though the docs refer to these as arguments, in the code they are still called parameters. Maybe for the sake of consistency we'd want to keep this as parameters, and then do a whole pass to rename to arguments later?

api_types.ts
dweitzman-codaio: Re: why `PropertyOptionsMetadataFunction` doesn't extend `MetadataFunction`, there are two reasons:

1) The existing argument types of `MetadataFunction` wouldn't work well for property autocomplete so we wanted `context` versions of some of the same parameters anyway. For example, `MetadataFunction` has a `search` argument used for parameter autocomplete, but for property autocomplete we use `Object.defineProperty()` trickery to spy on whether the search text is used for implicitly tracking what dependencies the formula has for cache keys. We wouldn't be able to spy whether the `MetadataFunction` `search` function parameter is or isn't accessed, so I didn't use it.

2) More generally, trying to force every kind of metadata function to use the same set of positional arguments and parameters doesn't seem as clean, simple, consistent, and flexible as just putting context into the context object in whatever shape seems nice without regard to to typing of formula parameters.

Today, a get schema or a get connection name metadata function always gets a "search" positional argument which serves no purpose. Because of that, if we wanted to introduce new positional arguments people might end up writing functions like `getSchema: (context, _unusedSearch, _unusedDeprecatedFormulaContext, newerBetterFormulaContextInputThatSupportsVarArgs)`. In terms of typescript annoyances, I assume the only reason `makeMetadataFormula` needs to deserialize `formulaContext` from a string is that it has `ParameterType.String` to be a valid formula parameter and whoever wrote `makeMetadataFormula` was trying to appease typescript by cramming a non-string into a string argument. `ExecutionContext` and its subinterfaces aren't limited to just pack formula parameter types, so it's simpler and more flexible to use than formula arguments.

api_types.ts
ekoleda-codaio: Sounds good to me.

ekoleda-codaio: It's very hard to change the SDK surface in a breaking way later, but if we can tie it to the `parameters` => `arguments` change above then that would give us an out.

api_types.ts
```diff
@@ -776,6 +777,11 @@ export interface ExecutionContext {
...
```
dweitzman-codaio, line 784: If I remember correctly, when we originally discussed this you proposed putting it within the `Sync` object on execution context since it was going to hold the sync table parameters.

Is there an intent here to use the same field to support other, non-sync parameters on different types of formulas?

alan-codaio: yeah exactly, as per @patrick-codaio 's suggestion here: https://github.com/coda/packs-sdk/pull/2732#discussion_r1316207564

dweitzman-codaio: What would we put in here for a non-sync formula? A duplicate version of the existing formula parameters indexed by parameter name strings instead of a list with positional indexes?

alan-codaio: I thought the general north star here would be to move this shared object into a standard place, instead of having it in separate different places for different formulas (so, when we eventually have an updated formula parameters shape that can handle the multiple set of varargs arguments, we could simply have all formulas pull from execution context instead of a positional array of arguments). 

Perhaps I am misunderstanding some of my earlier conversations with you and/or @patrick-codaio , or this PR is too ambitious?

dweitzman-codaio: I'm not opposed to moving arguments from positional parameters to context for more types of arguments in more types of formulas, but my feeling is that sort of thing would be orthogonal to this PR which is only trying to put sync table parameters into the context.

`MetadataContext` is the only type we have for sync table parameters at the moment, so it seems like for now we're stuck with it. If we want to replace it with a new, incompatible type in the future we'd need to think about and design what that would look like.

patrick-codaio: I'm fine with being less ambitious in this PR, I just thought we should be at least talking about the long term north star.

alan-codaio: Sounds good — moving it back into the `Sync` context for now. @dweitzman-codaio is that ok with you? I saw your original comment about putting it in a custom sync field for property options only, but I think we already get the parameters for normal sync formulas in `calc_service._getParameters`

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Added
 
 - For CLI development, there is now a `--allowOlderSdkVersion` param for the `coda upload` command that builds a new version. Coda will soon default to preventing a Pack build to have an older SDK version than the prior version, under the assumption that it most often happens when multiple dev environments conflict with each other. This new option is a bypass for that protection, for the rare case when you actually want to downgrade the SDK.
+- Added a new `parameters` property to our `Sync` interface so that all sync formulas, not just metadata formulas, can read their sync table parameters.
```
patrick-codaio, line 10: Perhaps specify that what sync formulas are not metadata formulas, to break it down all the way for makers? Is it just 2way autocomplete?

patrick-codaio (approved):

api_types.ts
dweitzman-codaio: I'm fine with having the property defined in typescript for all sync table formulas. The only downside to putting it directly in Sync is that it's more work to ensure it's populated for all the formula types, but it seems like a manageable amount of work that's worth doing.

CHANGELOG.md
dweitzman-codaio: Technically property options is a metadata formula. I don't think there are any non-metadata formulas that would be gaining new access to sync table parameters from this?...

Could simplify the changelog entry to something like "Added a new `parameters` property in the `Sync` interface with sync table parameters."

api_types.ts
```diff
@@ -790,10 +796,10 @@ export interface SyncExecutionContext extends ExecutionContext {
 }
 
 /**
- * Sub-class of {@link ExecutionContext} that is passed to the `options` function of
+ * Sub-class of {@link SyncExecutionContext} that is passed to the `options` function of
  * mutable sync tables for properties with `options` enabled.
  */
-export interface PropertyOptionsExecutionContext extends ExecutionContext {
+export interface PropertyOptionsExecutionContext extends SyncExecutionContext {
```
dweitzman-codaio, line 802: Smart!

dweitzman-codaio (approved):

PR #2734, opened 2023-09-14
Stop supporting '=' in sync table identity names

huayang-codaio:
According to my map reduce result, there's no sync tables in all envs that is using a `=` character in their identity name. And the `=` is not a valid character in coda object ids. So let's simply get rid of it. 

PTAL @patrick-codaio @dweitzman-codaio @alan-codaio 

cc @taj-codaio @chris-codaio

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Added
 
 - For CLI development, there is now a `--allowOlderSdkVersion` param for the `coda upload` command that builds a new version. Coda will soon default to preventing a Pack build to have an older SDK version than the prior version, under the assumption that it most often happens when multiple dev environments conflict with each other. This new option is a bypass for that protection, for the rare case when you actually want to downgrade the SDK.
+- "=" character is no longer supported in sync table identity names.
 
 ## [1.5.1] - 2023-07-31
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Added
 
 - For CLI development, there is now a `--allowOlderSdkVersion` param for the `coda upload` command that builds a new version. Coda will soon default to preventing a Pack build to have an older SDK version than the prior version, under the assumption that it most often happens when multiple dev environments conflict with each other. This new option is a bypass for that protection, for the rare case when you actually want to downgrade the SDK.
+- "=" character is no longer supported in sync table identity names.
```
patrick-codaio, line 10: Can this be in a different section than "added"?

testing/upload_validation.ts
```diff
@@ -907,7 +907,7 @@ function buildMetadataSchema({sdkVersion}: BuildMetadataSchemaArgs): {
     }),
   );
 
-  const Base64ObjectRegex = /^[A-Za-z0-9=_-]+$/;
+  const Base64ObjectRegex = /^[A-Za-z0-9_-]+$/;
```
patrick-codaio, line 910: Can we rename this ValidCodaObjectIdRegex?

patrick-codaio (approved):

dweitzman-codaio (approved):

PR #2735, opened 2023-09-14
Enable returning Dates for DateTime values

oleg-codaio:
I noticed I wasn't able to return `Date` objects for `DateTime` values - seems if we support `Date` values we should also support `DateTime`? Or we should just rip out support for these hints entirely and stick to `.toISOString()`.

PTAL @coda/packs


ekoleda-codaio:
Should it include `ValueHint.Time` as well?

oleg-codaio:
Hmm, not sure about `Time`. We'd essentially just discard the date component of `Date` and only look at the relative time?

ekoleda-codaio:
Ya, I believe that's how it works already. This part of the code just helps with type checking.

ekoleda-codaio (approved):
LGTM, but you may want to get a review from an engineer that better understands the type system of the SDK.

jonathan-codaio (commented):
Did it work in all of these cases at runtime when testing (using an `as any` cast presumably), it was just TS complaining?

oleg-codaio:
It worked as expected with `DateTime`, but for completeness sake I just put together a basic pack:

```ts

pack.addFormula({
  name: "DateNow",
  description: "The current date.",

  parameters: [],

  resultType: coda.ValueType.Object,
  schema: {
    type: coda.ValueType.Object,
    properties: {
      value: {type: coda.ValueType.String, codaType: coda.ValueHintType.Date, required: true},
    }
  },

  execute: async function ([]) {
    return {value: new Date()};
  },
});

pack.addFormula({
  name: "TimeNow",
  description: "The current time.",

  parameters: [],

  resultType: coda.ValueType.Object,
  schema: {
    type: coda.ValueType.Object,
    properties: {
      value: {type: coda.ValueType.String, codaType: coda.ValueHintType.Time, required: true},
    }
  },

  execute: async function ([]) {
    return {value: new Date() as any};
  },
});

pack.addFormula({
  name: "DateTimeNow",
  description: "The current date/time.",

  parameters: [],

  resultType: coda.ValueType.Object,
  schema: {
    type: coda.ValueType.Object,
    properties: {
      value: {type: coda.ValueType.String, codaType: coda.ValueHintType.DateTime, required: true},
    }
  },

  execute: async function ([]) {
    return {value: new Date() as any};
  },
});
```



image

Looks like only `DateTime` works as expected. If I update the Pack to instead return only date strings / time strings, then things look as expected:

image

So looks like the formula engine / SDK isn't truncating Date/Time values properly. The same applies to date time strings passed in too though (i.e., if I update the code above to do `new Date().toISOString()`).

jonathan-codaio:
Hmm ok. There's some recent code indicating potentially wrong behavior in the post-processing of pack results for dates/times: https://github.com/coda/coda/blob/232cf6aab344902c8cd7c523fb5d9db4a1dbf9d7/modules/common/calc-service/object_wrapper.ts#L185

I would maybe just restrict this change to DateTime for now, and at some later point we can go try to clean things up for dates and times. Technically those would be breaking changes so I wouldn't want to cause collateral damage right now.

I think the JS `Date` vs `string` cases are actually the same because I think this stuff comes over the wire and into object_wrapper.ts as ISO date strings unless I'm mistaken.

jonathan-codaio (approved):
Thanks for verifying!

chris-codaio:
Please make sure we can handle this in the two-way edit direction as well :)

oleg-codaio:
> Please make sure we can handle this in the two-way edit direction as well :)

I'm not as familiar with two-way, mind validating and let me know if you see anything broken?

oleg-codaio:
Chatted with @chris-codaio who has concerns that marshalling/unmarshalling might not work as expected with `Date`s being passed for `DateTime` values. Looking at the code, given `Date`s work correctly, it follows `DateTime` values should as well, too. However, I don't have the time to investigate this further in depth.

Will reassign this to the 2 way/Packs teams, if y'all would like to validate this further before merging. Otherwise, the workaround for Packs is to keep returning strings.

PR #2736, opened 2023-09-15
Avoid validation errors on unused property options in non-sync formulas

dweitzman-codaio:
The solution used here is to do a `deepCopy()` on the schema in `makeFormula` so unsupported types like js functions get silently dropped.

I'm not super thrilled about silently dropping data, but there's at least some precedent for it: `makeSyncTable` already does a deepCopy (see https://github.com/coda/packs-sdk/blob/38f6a7378f7b16ff5383ddd4ede3c86a53ee7653/api.ts#L2115), which made the logic for moving JS options functions into formulas more complicated.


huayang-codaio (approved):
Honest I just realized that deepCopy will do this trick. Can you add a comment at these two lines just to remind us that the deepCopy is for the sole purpose of stripping out functions?

dweitzman-codaio:
> Honest I just realized that deepCopy will do this trick. Can you add a comment at these two lines just to remind us that the deepCopy is for the sole purpose of stripping out functions?

Good call. Done

PR #2768, opened 2023-10-03
Update Todoist sample with 2way

ekoleda-codaio:
Update the "full" Todoist sample to include two-way sync support. Includes:

- Adjusting the schemas to use `fromKey`, since that requires less translation to and from representations.
- Add `#region` comments in the code to make cold folding easier
  - This would be better structured into separate files, but I want to keep it one file for Pack Studio users.
- Change how circular schema references are created, to reduce duplication.
- Change the Task priority property to a sting like "P1", to match the UI.
- Remove actions `UpdateTask` and `MarkComplete`, since those can now be done via two-way sync.
- Add action `SetDueDate`, since that can't be done via two-way sync (the due date is a rich object).


documentation/samples/packs/todoist/full.ts
```diff
@@ -397,92 +394,59 @@ pack.addFormula({
 });
 
 pack.addFormula({
```
dweitzman-codaio, line 396: As an alternative to excluding due date from 2-way, what do you think of either:
- Treating a time of midnight as a date without a time, or
- Splitting it across two columns so there's a datetime column and an `endOfDate` column

Either of those approaches could allow for updating with 2-way without the additional action formula

ekoleda-codaio: > Treating a time of midnight as a date without a time

I thought about this, but it would require editors and viewers to understand a fairly subtle behavioral difference between 12:00 and 12:01.

> Splitting it across two columns so there's a datetime column and an `endOfDate` boolean/checkbox column

I tried something similar (`dueDisplay`, `dueDate`, `dueDateTime`) and it didn't feel like a great UX. Plus Todoist's system for due dates is actually even more complex, allowing for recurrences, etc. I chose to omit that for brevity, but all said I think the correct way to display this complex bundle of information is as a sub-object. Artificially flattening it out into N different columns helps with editing, but hurts consuming, IMO.

I imagine that is circumstance (having a rich object you want to allow users to modify in some ways) will be somewhat common, so I think there is also some utility in demonstrating how to handle that case. Assuming we think this is how we would recommend developers solve that case.

dweitzman-codaio (approved):

PR #2769, opened 2023-10-03
Remove tip about ID properties

ekoleda-codaio:
With the new sync table storage there is no longer an issue with using a property named `id`. Assigning to on-call.


patrick-codaio (approved):
Oh that is so good to know!

PR #2777, opened 2023-10-05
Add missing changelog entries

dweitzman-codaio:
One about a validation fix from #2736
The other about the switch node 18 #2737

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Changed
 
 - "=" character is no longer supported in sync table identity names.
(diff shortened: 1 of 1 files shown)


huayang-codaio (approved):

huayang-codaio:
Do we have any other documentation that needs to be updated?

dweitzman-codaio:
> Do we have any other documentation that needs to be updated?

For node 18 specifically? I think the command line tools should all still work regardless of what node version people have installed. Seems like `coda execute` opportunistically tries to use `isolated-vm` if it's installed and working, but falls back to not using it if necessary.

cc @ekoleda-codaio in case you're aware of any documentation that mentions node versions

PR #2779, opened 2023-10-13
Clarify that card subtitles fallback to featured properties

ekoleda-codaio:
Assigning to on-call.

### DIFF
--- a/docs/guides/blocks/cards.md
+++ b/docs/guides/blocks/cards.md
@@ -92,6 +92,8 @@ const ProductSchema = coda.makeObjectSchema({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


ekoleda-codaio:
CC current on-call.

gary-codaio (approved):

docs/guides/blocks/cards.md
```diff
@@ -92,6 +92,8 @@ const ProductSchema = coda.makeObjectSchema({
 });
 ```
 
+If `subtitleProperties` are not defined the card will fall back to using the `featuredProperties` configured on the schema, if any. Featured properties are also used to determine the initial columns to show in a sync table. If you want a schema to include include `featuredProperties` but not display a subtitle, set `subtitleProperties` to an empty array.
```
gary-codaio, line 95: typo: include include

ekoleda-codaio: Good catch, fixed!

PR #2783, opened 2023-10-17
Attempting to upgrade mkdocs

dweitzman-codaio:
DO NOT REVIEW (not ready yet)

This is partly because if we upgrade enough mkdocs-related things it'll make some dependency-related security warnings go away

The changes here are mostly because mkdocs 1.5 does more aggressive checking for bad links and we previously had links in markdown that didn't work

I'm not sure if/how those links worked in the html docs, but I guess I can get that after marking a PR


documentation/scripts/documentation_compiler.ts
```diff
@@ -76,7 +76,7 @@ function compileExamples() {
         throw new Error(`${exampleFooterLink} is not a valid path`);
       }
       exampleFooterLink = `${SdkReferenceLink}${exampleFooterLink}`;
-      learnMoreLink = example.linkData.url;
+      learnMoreLink = example.linkData.url! + (example.linkData.url?.endsWith('/') ? 'index.md' : '.md');
```
dweitzman-codaio, line 79: This is hacky, but I'm using a trailing slash to indicate links should go to `/index.md` vs no trailing slash to mean `/foo` means `/foo.md`

documentation/generated/examples.json
```diff
@@ -12,7 +12,7 @@
       "url": "/guides/blocks/column-formats"
     },
     "exampleFooterLink": "https://coda.io/packs/build/latest/guides/blocks/column-formats",
-    "learnMoreLink": "/guides/blocks/column-formats",
+    "learnMoreLink": "/guides/blocks/column-formats.md",
```
ekoleda-codaio, line 15: I don't think it will work to have the `.md` links in this file, since it's used to power the Help side panel in the Pack Studio.

Instead of changing `documentation_compiler.ts` to add the extension, instead add it in `documentation/example_page_template.md` which only affects the generated markdown here.

PR #2785, opened 2023-10-19
Pin deps in Pipfile so renovate can upgrade them

dweitzman-codaio:


requirements.txt
```diff
@@ -1,47 +1,46 @@
...
-pygments==2.15.1; python_version >= '3.7'
-pymdown-extensions==10.0.1; python_version >= '3.7'
+pygments==2.16.1; python_version >= '3.7'
+pymdown-extensions==10.3; python_version >= '3.8'
 python-dateutil==2.8.2; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
-pyyaml==6.0; python_version >= '3.6'
+pyyaml==6.0.1; python_version >= '3.6'
 pyyaml-env-tag==0.1; python_version >= '3.6'
-regex==2023.5.5; python_version >= '3.6'
+regex==2023.10.3; python_version >= '3.7'
```
dweitzman-codaio, line 37: Note: I haven't pinned all the indirect dependencies, which doesn't seem worth the trouble for now but we can always do it in the future if/when we want to

jonathan-codaio (approved):

PR #2789, opened 2023-10-19
Add OAuth2 client credentials type

cqian-codaio:
Adds OAuth2ClientCredentials authentication type

Docs: https://staging.coda.io/d/Packs-Platform-IA-go-packs_d36WK4zgrYx/Client-Credentials_suTiX?utm_source=slack#_lu5jL

PTAL @jonathan-codaio @gary-codaio @erickoledadevrel - any feedback on the new authentication type naming


chris-codaio (commented):
Worth ensuring someone on the security team is looking at these PRs as well (particularly implementing the OAuth flow)?

types.ts
```diff
@@ -463,6 +470,46 @@ export interface MultiQueryParamTokenAuthentication extends BaseAuthentication {
   }>;
 }
 
+/**
```
cqian-codaio, line 473: This doc says only tokenUrl may be necessary. I saw we had a bunch of other parameters under the OAuth2Authentication that may be relevant to token exchange so included them here in a shared interface

types.ts
```diff
@@ -81,6 +81,13 @@ export enum AuthenticationType {
    * @see {@link OAuth2Authentication}
    */
   OAuth2 = 'OAuth2',
+  /**
+   * Authenticate using OAuth2 client credentials.
+   * The API must use a (largely) standards-compliant implementation of OAuth2.
+   *
```
cqian-codaio, line 87: For all these comments I mostly just copy/pasted from the OAuth 2 docs so let me know if I should change anything

types.ts
```diff
@@ -81,6 +81,13 @@ export enum AuthenticationType {
    * @see {@link OAuth2Authentication}
    */
   OAuth2 = 'OAuth2',
+  /**
+   * Authenticate using OAuth2 client credentials.
+   * The API must use a (largely) standards-compliant implementation of OAuth2.
```
jonathan-codaio, line 86: I would just change this line to be a link to https://oauth.net/2/grant-types/client-credentials/

This comment existed because for regular OAuth there are all kinds of APIs that _almost_ implement but do horrible quirky stuff. Some of it we support, some of it we don't. I don't think we've seen an implementation of client credentials that isn't compliant yet, which makes sense because it's a much simpler flow.

types.ts
```diff
@@ -463,6 +470,46 @@ export interface MultiQueryParamTokenAuthentication extends BaseAuthentication {
   }>;
 }
 
+/**
+ * Base interface for OAuth definitions.
+ */
+interface BaseOAuthAuthentication extends BaseAuthentication {
```
jonathan-codaio, line 476: I don't think I'd bother generalizing this yet. We haven't seen that APIs do anything non-compliant yet so I don't think we need to offer all the customizations like tokenPrefix and credentialsLocation. Those are just workarounds for bad OAuth2 implementations. Let's start out supporting only standards-compliant implementations and go from there. So I think we only need to offer `tokenUrl` here.

cqian-codaio:
> Worth ensuring someone on the security team is looking at these PRs as well (particularly implementing the OAuth flow)?

Yup we'll start adding them as reviewers once we start implementing the logic

testing/auth.ts
```diff
@@ -78,6 +78,8 @@ export function setupAuth(manifestDir: string, packDef: BasicPackDefinition, opt
     case AuthenticationType.OAuth2:
       ensureExists(packDef.defaultAuthentication, 'OAuth2 only works with defaultAuthentication, not system auth.');
       return handler.handleOAuth2();
+    case AuthenticationType.OAuth2ClientCredentials:
+      return printAndExit('This authentication type is not yet implemented');
```
ekoleda-codaio, line 82: Add a TODO here?

testing/fetcher.ts
```diff
@@ -411,6 +411,8 @@ export class AuthenticatingFetcher implements Fetcher {
           headers: requestHeaders,
         };
       }
+      case AuthenticationType.OAuth2ClientCredentials:
+        throw new Error('Not yet implemented');
```
ekoleda-codaio, line 415: Ditto.

types.ts
ekoleda-codaio: Huh, I was about to leave an opposite comment :-) IMO the developer experience really sucks when you hit an auth wall, and carrying forward a few more of these config options now could save a lot of developer pain down the road. I don't know of any cases yet where these extra options will be needed for the client credentials flow, but I worry that when they do come up they'll not rise to the top of any prioritization queue.

ekoleda-codaio: At the very least we should add in the `scopes` option, since that's part of the spec:

https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2

And one of the APIs I logged in go/oauth-needs requires it:

https://bitwarden.com/help/public-api/#bearer-access-tokens

types.ts
ekoleda-codaio: Above I would say something like "Authenticate using the OAuth2 client credentials flow. This is a less common type of OAuth2, which involves exchanging a client ID and secret for a temporary access token."

Likewise, I'd updated the `OAuth2` entry above with something like "Authenticate using the OAuth2 authorization code flow. This is the most common type of OAuth2, which involves the user approving access to their account before being granted a token."

types.ts
cqian-codaio: Yeah I'll include scope at least
I also found that Fedex doesn't require it, but Microsoft does 
https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow#get-a-token

cqian-codaio: Oh I didn't know you had examples of APIs using this
Downdetector requires the client id/secret in the headers: https://downdetectorapi.com/v2/docs/#tag/Authentication/paths/~1tokens~1verify/get
I think we can just go with having those extra options be common to both then?

testing/auth.ts
cqian-codaio: Done

testing/fetcher.ts
cqian-codaio: Done

ekoleda-codaio (approved):

types.ts
```diff
@@ -492,33 +547,14 @@ export interface OAuth2Authentication extends BaseAuthentication {
...
-   * the documentation for the API you are connecting to.
-   */
-  scopes?: string[];
   /**
    * The delimiter to use when joining {@link scopes} when generating authorization URLs.
    *
    * The OAuth2 standard is to use spaces to delimit scopes, and Coda will do that by default.
    * If the API you are using requires a different delimiter, say a comma, specify it here.
    */
   scopeDelimiter?: ' ' | ',' | ';';
```
jonathan-codaio, line 556: Any reason not to include this one if we're including `scopes`?

jonathan-codaio (commented):
Ah shoot, I was really hoping this simpler flow would not be subject to all the standards-incompliant quirks and complexity. Alas.

types.ts
ekoleda-codaio: Oh ya, good catch!

cqian-codaio: Adding this to the next PR

PR #2797, opened 2023-10-23
Support `executeUpdate`-specific OAuth scopes

dweitzman-codaio:
The motivation here is so packs can have read-only oauth scopes by default but prompt for write access in response to a 2-way update error message


patrick-codaio (approved):
Cool!

PR #2798, opened 2023-10-23
Add client credentials auth to coda CLI

cqian-codaio:
Adds functionality to request an access token and refresh it from the cli

Video showing execution of fedex pack, using the updated endpoint:
https://coda.d.pr/i/o7ulz7

PTAL @gary-codaio @jonathan-codaio

### DIFF
--- a/testing/auth.ts
+++ b/testing/auth.ts
@@ -20,7 +20,7 @@ import {getPackAuth} from '../cli/helpers';
 import {launchOAuthServerFlow} from './oauth_server';
 import {makeRedirectUrl} from './oauth_server';
 import * as path from 'path';
-import {performOAuthClientCredentialsServerFlow} from './oauth_server';
+import {performOAuthClientCredentialsServerFlow} from './oauth_helpers';
 import {print} from './helpers';
 import {printAndExit} from './helpers';
 import {promptForInput} from './helpers';
@@ -44,7 +44,8 @@ export async function setupAuthFromModule(
   return setupAuth(manifestDir, manifest, opts);
 }
 
-export function setupAuth(manifestDir: string, packDef: BasicPackDefinition, opts: SetupAuthOptions = {}): void {
+export async function setupAuth(manifestDir: string, packDef: BasicPackDefinition, opts: SetupAuthOptions = {})
+    : Promise<void> {
   const auth = getPackAuth(packDef);
   if (!auth) {
     return printAndExit(
@@ -251,7 +252,7 @@ class CredentialHandler {
     assertCondition(this._authDef.type === AuthenticationType.OAuth2);
     const existingCredentials = this.checkForExistingCredential() as OAuth2Credentials | undefined;
     print(
-      `*** Your application must have ${makeRedirectUrl(this._oauthServerPort)} whitelisted as an OAuth redirect url ` +
+      `*** Your application must have ${makeRedirectUrl(this._oauthServerPort)} allowlisted as an OAuth redirect url ` +
         'in order for this tool to work. ***',
     );
     const {clientId, clientSecret} = this._promptOAuth2ClientIdAndSecret(existingCredentials);
@@ -292,7 +293,7 @@ class CredentialHandler {
     });
   }
 
-  handleOAuth2ClientCredentials() {
+  async handleOAuth2ClientCredentials() {
     assertCondition(this._authDef.type === AuthenticationType.OAuth2ClientCredentials);
     const existingCredentials = this.checkForExistingCredential() as OAuth2ClientCredentials | undefined;
     const {clientId, clientSecret} = this._promptOAuth2ClientIdAndSecret(existingCredentials);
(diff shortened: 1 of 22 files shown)


testing/auth.ts
```diff
@@ -229,25 +230,32 @@ class CredentialHandler {
...
     const clientSecret = ensureNonEmptyString(newClientSecret || existingCredentials?.clientSecret);
 
+    return {clientId, clientSecret};
+  }
+
+  handleOAuth2() {
+    assertCondition(this._authDef.type === AuthenticationType.OAuth2);
+    const existingCredentials = this.checkForExistingCredential() as OAuth2Credentials | undefined;
+    print(
+      `*** Your application must have ${makeRedirectUrl(this._oauthServerPort)} whitelisted as an OAuth redirect url ` +
```
jonathan-codaio, line 254: Nit, this was already here, but mind fixing up to use `allowlisted` per inclusive language guidelines?

testing/auth.ts
```diff
@@ -284,6 +292,44 @@ class CredentialHandler {
...
+    const manifestScopes = this._authDef.scopes || [];
+    const requestedScopes =
+        this._extraOAuthScopes.length > 0 ? [...manifestScopes, ...this._extraOAuthScopes] : manifestScopes;
+
+    performOAuthClientCredentialsServerFlow({
+      clientId,
+      clientSecret,
+      scopes: requestedScopes,
+      authDef: this._authDef,
+    }).then(({accessToken, expires}) => {
```
jonathan-codaio, line 317: Hmm let's not use then/catch. If you want to kick this off in this non-async method, would just make a new private method that does it, can just `void` that method when you call it to orphan the promise. Would leave a TODO that we should go back make input handling async if we're going to do stuff like this inside of it.

Either way this feels a little brittle though because I think the script can exit while this orphaned promise is still running, and so your CLI output might look weird. Not sure if there's a way to wait for all promise to settle before exciting the CLI main.

testing/fetcher.ts
```diff
@@ -186,17 +189,28 @@ export class AuthenticatingFetcher implements Fetcher {
...
+
+    if (requestFailure.statusCode !== HttpStatusCode.Unauthorized ||
+        !([AuthenticationType.OAuth2, AuthenticationType.OAuth2ClientCredentials].includes(this._authDef.type))) {
       return false;
     }
-    const {accessToken, refreshToken} = this._credentials as OAuth2Credentials;
-    if (!accessToken || !refreshToken) {
-      return false;
+
+    if (this._authDef.type === AuthenticationType.OAuth2) {
```
jonathan-codaio, line 198: Would do this as an exhaustive `switch`

testing/fetcher.ts
```diff
@@ -246,14 +260,46 @@ export class AuthenticatingFetcher implements Fetcher {
...
+      expires,
+      scopes
+    }
+  }
+
+  private async _refreshOAuthCredentials() {
+    assertCondition(this._authDef && [AuthenticationType.OAuth2, AuthenticationType.OAuth2ClientCredentials]
+        .includes(this._authDef.type));
+    let credentials: OAuth2Credentials | OAuth2ClientCredentials;
+    if (this._authDef.type === AuthenticationType.OAuth2) {
```
jonathan-codaio, line 296: Ditto exhaustive switch

testing/oauth_server.ts
```diff
@@ -22,6 +25,67 @@ interface TokenCallbackResponse {
   data: {[key: string]: string};
 }
 
+async function requestOAuthAccessToken(
```
jonathan-codaio, line 28: MIght move this to an oauth_helpers.ts file or something, because the client credentials flow won't be using the oauth server.

testing/oauth_server.ts
```diff
@@ -164,3 +193,34 @@ class OAuthServerContainer {
     this._server?.close();
   }
 }
+
+export async function performOAuthClientCredentialsServerFlow({
```
jonathan-codaio, line 197: Ditto here for moving this out of this file.

testing/oauth_server.ts
```diff
@@ -164,3 +199,38 @@ class OAuthServerContainer {
...
+  clientSecret,
+  authDef,
+  scopes,
+  afterTokenExchange,
+}: {
+  clientId: string;
+  clientSecret: string;
+  authDef: OAuth2ClientCredentialsAuthentication;
+  scopes?: string[];
+  afterTokenExchange?: (params: AfterTokenOAuthClientCredentialsExchangeParams) => void;
```
cqian-codaio, line 214: This is very awkward in that we have a callback and it returns the credentials. The callback is necessary so we can properly stub this function in the tests. The regular returns value is for when this gets called in an async function during the refresh. We could just use the callback in both places, though it may cause the code in the refresh case to look more awkward

jonathan-codaio: Hmm I don't think the real reason is because of tests, I think it's because in the regular OAuth2 case you're waiting for user input. You launch a server that idles while the user goes to their browser to go through the approval flow on the web, and only when they complete that is there sufficient info to run the token exchange, so by necessity it's an interactive flow.

Client credentials isn't interactive so I don't think we should be modeling it the same way. The code to make the exchange should just be a normal async function that gets awaited.

cqian-codaio: I tried to model it like that initially. However `setupAuth`, which will eventually call this function, wasn't async, so I thought we needed either a `.then` or a callback to set the credentials afterwards. `.then` didn't work since the promise would let go of execution and `executeFetch` here would get the incomplete credentials:
https://github.com/coda/packs-sdk/pull/2798/files#diff-56a0e2bb7f87e922174aa0681e794985f029bc4718805e6434f006248eac5509R1343

 `setupAuth` is only called from the `setupAuthFromModule` which is async, so I'm wonder if we can just make it async as well

cqian-codaio: I've switched `setupAuth` to be async and removed the callback, and it looks to be working fine

jonathan-codaio (approved):
Nice

PR #2805, opened 2023-10-25
Add client credentials as system auth

cqian-codaio:
Missed adding this, since we want client credentials to be available as both user auth and system auth

PTAL @gary-codaio @jonathan-codaio


jonathan-codaio (approved):
Ah oops, I should have caught this too.

PR #2806, opened 2023-10-26
Update two-way.md

josh-supsync:
2-way sync is now publicly available


ekoleda-codaio:
Good catch! This page is actually auto-generated from other files, and the source text is here:

https://github.com/coda/packs-sdk/blob/main/documentation/samples/packs/two-way/README.md

Can you remove it from there as well? I'm also happy to make this change in my own workspace, if that's simpler.

ekoleda-codaio:
Closing this PR, and I'm carrying it forward in #2808.

PR #2809, opened 2023-10-31
Export helpers for creating isolated-vm Isolate

dweitzman-codaio:
The goal here is to avoid the need to directly import isolated-vm in other repos, which avoids the need to manually upgrade isolated-vm in those repos


shaonan-codaio (approved):
Thank you so much!
LGTM as I don't see it'll break anything

PR #2829, opened 2023-11-02
Fix todoist example pack on items with no due date

dweitzman-codaio:
Was testing something using the todoist pack and came across this issue

### DIFF
--- a/documentation/samples/packs/todoist/full.ts
+++ b/documentation/samples/packs/todoist/full.ts
@@ -257,6 +257,9 @@ function formatTaskForAPI(task: any) {
     // Convert the priority back to a number.
     result.priority = 5 - Number(result.priority.substring(1));
   }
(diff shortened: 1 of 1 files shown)


ekoleda-codaio:
Do you know under what circumstances this would happen? The `due` property isn't even mutable, so if I send `null` or `undefined` in shouldn't get it back again?

dweitzman-codaio:
> Do you know under what circumstances this would happen? The `due` property isn't even mutable, so if I send `null` or `undefined` in shouldn't get it back again?

Might be that you have to create items with no due date and then when you mutate them you need to pipe the null through

ekoleda-codaio:
I was able to repro for the general case, and filed bug 34286 to track it. This seems like a platform issue and not something the Pack maker should need to handle.

PR #2831, opened 2023-11-03
Get the pnpm version from package.json

dweitzman-codaio:
This PR also re-enables pnpm upgrades, since pnpm is now a standalone install and won't conflict with other global node installations


neal-codaio (approved):

huayang-codaio:
Thanks!

PR #2835, opened 2023-11-30
Create a .nvmrc file so "nvm use" will pick an appropriate node version

dweitzman-codaio:
Personally this is what I do. Might be useful to other people as well?...


setup-env.sh
```diff
@@ -9,3 +9,9 @@ if [[ "$(uname -s)" == "Darwin" ]] && [[ "$(arch)" == "arm64" ]]; then
 fi
 
 export PATH=${BASEDIR}/build/node/bin:${BASEDIR}/node_modules/.bin:${BASEDIR}/.pnpm_install/bin:${PATH}
+
+if command -v nvm &> /dev/null
+then
+  # Gets version from .nvmrc
+  nvm use
```
dweitzman-codaio, line 16: This will instruct people to run `nvm install v18` if they haven't already, so I figured it's probably fine not to mention that directly in the CONTRIBUTING instructions

jonathan-codaio (approved):
Great

PR #2883, opened 2023-12-18
Add option to cache non-GET requests in the `FetchRequest` interface

alan-codaio:
We don't allow caching non-`GET` requests by default, but some makers may want to cache the requests for non-side-effecting requests (for example, querying a GraphQL API). Adds the option to cache these results here when making those fetch calls.

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -580,8 +580,6 @@ export interface FetchRequest {
    * If true, Coda will cache the POST request and return the same response for subsequent requests.
    * This is mainly used for POST requests that do not have side effects, such as querying a
    * GraphQL API.
-   *
-   * You will also need to set `cacheTtlSecs` to a non-zero value for this to work.
    */
   cachePOSTRequest?: boolean;
 }
--- a/api_types.ts
+++ b/api_types.ts
@@ -576,6 +576,14 @@ export interface FetchRequest {
    * You may inspect the `Location` header of the response to observe the indicated redirect URL.
    */
   ignoreRedirects?: boolean;
+  /**
+   * If true, Coda will cache the POST request and return the same response for subsequent requests.
+   * This is mainly used for POST requests that do not have side effects, such as querying a
+   * GraphQL API.
+   *
+   * You will also need to set `cacheTtlSecs` to a non-zero value for this to work.
+   */
+  cachePOSTRequest?: boolean;
 }
 
 /**
--- a/api_types.ts
+++ b/api_types.ts
@@ -556,8 +556,20 @@ export interface FetchRequest {
    * instead of making the HTTP request. If left unspecified, Coda will automatically
    * cache all GET requests for approximately 5 minutes. To disable the default caching,
    * set this value to `0`.
+   *
+   * If you are trying to cache a POST, PUT, PATCH, or DELETE request, you must also
+   * set {@link FetchRequest.forceCache} to true.
    */
   cacheTtlSecs?: number;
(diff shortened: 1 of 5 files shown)


api_types.ts
```diff
@@ -576,6 +576,12 @@ export interface FetchRequest {
    * You may inspect the `Location` header of the response to observe the indicated redirect URL.
    */
   ignoreRedirects?: boolean;
+  /**
+   * If true, Coda will cache the POST request and return the same response for subsequent requests.
+   * This is mainly used for POST requests that do not have side effects, such as querying a
```
jonathan-codaio, line 581: Nit, would strengthen this to say something like "This _must_ be used only for POST requests that do not have side effects..."

api_types.ts
```diff
@@ -576,6 +576,12 @@ export interface FetchRequest {
...
+   * GraphQL API.
+   */
+  cachePOSTRequest?: boolean;
```
jonathan-codaio, line 584: Curious for @ekoleda-codaio 's thoughts on this. Context is that I asked for this because the GitHub pack is now expensively making multiple graphql queries in order to implement some parameter autocompletes. Normally this wouldn't be quite so bad because we would cache these http responses for 5 minutes and subsequent autocomplete requests would operate on cached data.

But because graphql is basically a GET request with a payload and so is implemented as a POST, and we currently refuse to cache POST requests, any scenario like this will not get the benefit of caching.

Today we only allow caching of GET requests. We could have a property like `forceCache` that generally forces us to consider caching, but I can't see a world in which we would allow caching for PATCH, PUT or DELETE, those would seem to only be used for true mutations. For HEAD, maybe we could default allow caching if it seemed important, dunno. So it really seems like POST is a special snowflake that is usually a mutation but sometimes just a GET with a payload, and so allowing cacheability only for POSTs and naming the parameter to indicate that might be the right answer.

(I think maybe the naming I was leaning towards was `isCacheablePost`.)

huayang-codaio: I don't think GraphQL requires to use POST. The spec says that a graphql server should be accepting both GET and POST for queries (https://graphql.org/learn/serving-over-http/#get-request). On the other hand, mutations always have to come in as POSTs. 

Anyways, there are always some of those non-GET APIs that have no side effect. Like most search APIs are POSTs to explicitly avoid caching. I personally would be in favor of using `forceCache` as Jonathan mentioned, leaving it to the Pack maker to decide.

jonathan-codaio: I think even if we were building from scratch I'd consider still having a second property for this; it would be very easy to shoot yourself in the foot and throw cache ttls all over the place or reuse a fetcher helper that applies one and all the sudden you're randomly dropping mutations on the floor based on when our cache decides to kick in. So being explicit here seems healthy.

I'd be fine with something equivalent to forceCache if that's the preference.

alan-codaio: I like having forceCache as an explicit option. If we don't want to use an explicit option, we could also log when POST requests are being cached in the Pack logs.

alan-codaio: We may want to just log these calls anyway, since I'd imagine a sloppy Pack inadvertently caching POST calls will lead to customers writing into us

alan-codaio: Cool, just updated, let me know if the description wording makes sense!

ekoleda-codaio (approved):
LGTM!

huayang-codaio (approved):

PR #2884, opened 2023-12-20
Introduce `displayName` field for sync table schema properties

patrick-codaio:


jonathan-codaio:
Looks reasonable to me. Should be nice and simple on the SDK side, I think the more interesting questions will come out of the prototyping on the `coda` side.

As for validation, you can already name a column like `$$[...]` and it gets escaped when used in formulas, so shouldn't the same apply to names that arrive via a pack schema?

alan-codaio (approved):

jonathan-codaio:
You might need to bump the semantic version to 1.7.5-prerelease.2 to avoid the mysterious bug that makes the lint job super slow when you bump to this commit hash on the `coda` side.

PR #2885, opened 2023-12-27
Add warning when running a Pack with multiple domains

ekoleda-codaio:
This change is based on feedback from Pack makers who became frustrated when they sunk a lot of time into developing a Pack only to realize at the end that it required Coda approval since it uses multiple domains.


huayang-codaio (approved):
Makes sense. Thanks!

PR #2886, opened 2023-12-27
Advanced property options feature to ignore properties when caching

dweitzman-codaio:
Normally we do implicit caching where the properties read during the propertyOptions formula are incorporated into the cache key. Some APIs can force packs to make an overly-specific request while getting more general information.

As a hypothetical example, imagine that some API requires you to pass in an email ID when fetching the all labels that your account can use, but the list of labels is actually the same for all emails. You might read `email_id` from the set of properties but then return it in the new `propertiesNotUsed` setting from this PR to prevent it from becoming part of the cache key
   
This feature is undocumented and not an officially-supported API, even at typescript level where it's not exposed to things like vscode autocomplete: for now it's only needed by advanced makers trying to improve caching performance with unusual APIs. If/when there's enough demand we could make it an official feature.

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -1525,7 +1525,7 @@ interface PropertyOptionsNormalizedResults {
   // Undocumented feature for situations when you might be using an awkward API that
   // only lets you fetch data using an overly-specific property when a more general
   // one would have sufficed.
-  propertiesNotUsed?: string[];
+  unusedProperties?: string[];
 }
 
 function normalizePropertyOptionsResultsArray(
--- a/api.ts
+++ b/api.ts
@@ -1521,6 +1521,11 @@ export type PropertyOptionsResults = PropertyOptionsMetadataResult<any | Propert
 interface PropertyOptionsNormalizedResults {
   cacheTtlSecs?: number;
   results: Array<{display: string | undefined; value: any}>;
+
+  // Undocumented feature for situations when you might be using an awkward API that
+  // only lets you fetch data using an overly-specific property when a more general
+  // one would have sufficed.
+  propertiesNotUsed?: string[];
 }
 
 function normalizePropertyOptionsResultsArray(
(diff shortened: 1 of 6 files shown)


api.ts
```diff
@@ -1521,6 +1521,11 @@ export type PropertyOptionsResults = PropertyOptionsMetadataResult<any | Propert
 interface PropertyOptionsNormalizedResults {
   cacheTtlSecs?: number;
   results: Array<{display: string | undefined; value: any}>;
+
+  // Undocumented feature for situations when you might be using an awkward API that
+  // only lets you fetch data using an overly-specific property when a more general
+  // one would have sufficed.
+  propertiesNotUsed?: string[];
```
patrick-codaio, line 1528: A quick brainstorm of other name options:
- unusedProperties
- cacheIgnoreProperties
- cacheKeyExcludeProperties

And one other possibility - instead of a negative framing, we could in theory make this `cacheKeyProperties` and have it override our property spying completely. Perhaps easier to explain/reason about? Perhaps harder for the pack maker to keep accurate as the pack evolves?

Very curious for @ekoleda-codaio 's thoughts.

dweitzman-codaio: I like the conceptual simplicity of a positive framing like `cacheKeyProperties` as a way to opt out of implicit caching entirely, although I'm also nervous that if the only way to ignore a key is opting out of implicit caching entirely it significantly raises the technical skills needed to just ignore one or two properties

My inclination for now is to go with something like `unusedProperties` for simplicity while relying on the fact that this is will be an undocumented feature that we can always change later. Open to a different approach, though.

patrick-codaio: Sounds good to me.

patrick-codaio (approved):

api.ts
dweitzman-codaio: Cool, renamed the undocumented setting to `unusedProperties`

PR #2890, opened 2024-01-02
Fix buffer request bodies with `coda execute --vm=false`

dweitzman-codaio:
There was a bug due to `Buffer.isBuffer()` not working if you created a non-native buffer in pack code and try to check if it's a buffer using native buffers in node_fetcher.ts

This forces non-native Buffers even for node_fetcher, ensuring consistent behavior

### DIFF
--- a/testing/execution.ts
+++ b/testing/execution.ts
@@ -6,6 +6,7 @@ import {FormulaType} from '../runtime/types';
 import type {GenericSyncFormulaResult} from '../api';
 import type {MetadataContext} from '../api';
 import type {MetadataFormula} from '../api';
+import {Buffer as NonNativeBuffer} from 'buffer/';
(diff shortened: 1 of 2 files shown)


testing/execution.ts
```diff
@@ -6,6 +6,7 @@ import {FormulaType} from '../runtime/types';
 import type {GenericSyncFormulaResult} from '../api';
 import type {MetadataContext} from '../api';
 import type {MetadataFormula} from '../api';
+import {Buffer as NonNativeBuffer} from 'buffer/';
```
dweitzman-codaio, line 9: The trailing slash here is intentional: https://www.npmjs.com/package/buffer#usage

huayang-codaio: TIL

huayang-codaio (approved):

PR #2891, opened 2024-01-02
Require unique sync formula names

dweitzman-codaio:
There's a bug with autocomplete when multiple sync tables use the same formula name. It's probably possible to fix the autocomplete bug in isolation, but it seems reasonable to just disallow duplicate sync table formula names in the first place.

### DIFF
--- a/dist/testing/upload_validation.js
+++ b/dist/testing/upload_validation.js
@@ -1207,6 +1207,7 @@ function buildMetadataSchema({ sdkVersion }) {
             .superRefine((data, context) => {
             var _a, _b;
             const identityNames = [];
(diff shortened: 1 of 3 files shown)


huayang-codaio (approved):
surprise. :D

dweitzman-codaio:
@ekoleda-codaio I think you were in favor of this, but wanted to confirm since I can't find any slack conversations about it

ekoleda-codaio:
Ya, I mentioned this in the bug report:

> The solution may not be to allow this case, but rather to through a validation error when multiple sync tables use the same formula name.

So LGTM!

PR #2897, opened 2024-01-10
Fix default value for `search` argument when executing metadata formulas

ekoleda-codaio:
The `search` argument of metadata functions should always be defined, but the new support in the `execute` function wasn't setting a default value when it wasn't specified on the command line.

### Before

image

### After

image

### DIFF
--- a/testing/execution.ts
+++ b/testing/execution.ts
@@ -460,6 +460,8 @@ async function executeFormulaOrSyncWithRawParamsInVM<T extends FormulaSpecificat
       // to an autocomplete metadata formula), because at execution time it gets passed as a serialized
       // JSON string anyway which is already parsed by the compiled pack definition.
       params = rawParams as ParamValues<ParamDefs>;
+      // Default the search string (first arg) to an empty string.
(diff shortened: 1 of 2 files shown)


testing/execution.ts
```diff
@@ -460,6 +460,8 @@ async function executeFormulaOrSyncWithRawParamsInVM<T extends FormulaSpecificat
       // to an autocomplete metadata formula), because at execution time it gets passed as a serialized
       // JSON string anyway which is already parsed by the compiled pack definition.
       params = rawParams as ParamValues<ParamDefs>;
+      // Default the search string (first arg) to an empty string.
+      params[0] ??= '';
```
jonathan-codaio, line 464: Can you make this same change at line 513 too? We unfortunately have slightly different execution paths for with-vm and without, the reason why we can't reuse I can't recall atm.

jonathan-codaio (approved):

testing/execution.ts
ekoleda-codaio: Done!

PR #2899, opened 2024-01-12
Add docs for CLI metadata formula execution

ekoleda-codaio:


docs/guides/development/cli.md
```diff
@@ -157,6 +157,38 @@ npx coda execute path/to/pack.ts Items --dynamicUrl=https://example.com/api/tabl
...
+npx coda execute <sync name>:searchDynamicUrls [query]
+npx coda execute <sync name>:getName
+npx coda execute <sync name>:getDisplayUrl
+npx coda execute <sync name>:getSchema [unused] [params JSON]
+
+# Two-way sync
+npx coda execute MySync:update [params..] [sync update array json]
+
+# Authentication
+npx coda execute Auth:getConnectionName
```
jonathan-codaio, line 180: Nit, we also have `Auth:getConnectionName`

docs/guides/development/cli.md
```diff
@@ -157,6 +157,38 @@ npx coda execute path/to/pack.ts Items --dynamicUrl=https://example.com/api/tabl
 ```
 
 
+### Running metadata functions
+
+More complicated Packs may have a lot of code outside of the core `execute` function, used to generate metadata like parameter autocomplete values, dynamic schemas, etc. You can run those functions using the CLI as well, using the same `execute` command but passing in additional information in name of the formula or sync table.
+
+```sh
+# Parameter autocomplete
+npx coda execute <formula name>:autocomplete:<paramName> [query] [params JSON]
```
jonathan-codaio, line 166: What do you think about including a specific example here, that shows what "params JSON" actually looks like on the CLI? I think it can get confusing to know what's being asked for in terms of quoting and the like.

And then if we're doing concrete examples, maybe one that explains, hey in most cases don't even worry about the JSON, it's only relevant if your implementation looks at other parameters, and otherwise you don't even have to pass a value for that.

jonathan-codaio (commented):
Great!

docs/guides/development/cli.md
ekoleda-codaio: Did you mean to type `getConnectionId`? If so, I omitted that since we currently hide that function from developers, since it isn't used anywhere:

https://github.com/coda/packs-sdk/blob/8d4254faafb034c62fba03837de82d85b5a2f754/types.ts#L237-L246

jonathan-codaio: Oh yep, and thanks for the reminder, makes sense.

docs/guides/development/cli.md
ekoleda-codaio: Done. Also rearranged the content a bit, PTAL.

jonathan-codaio (approved):
Looks great!

PR #2901, opened 2024-01-22
Expose system connection authentication in external packs info

kelsey-codaio:
Expose system connection authentication in external packs info

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Added
 
 - `coda execute` now allows for executing metadata formulas and sync updates.
+- Add `systemConnectionAuthentication` to `ExternalPackVersionMetadata`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Added
 
 - `coda execute` now allows for executing metadata formulas and sync updates.
-- Add `systemConnectionAuthentication` to `ExternalPackVersionMetadata`.
 
 ### Changed
 
--- a/compiled_types.ts
+++ b/compiled_types.ts
@@ -144,7 +144,7 @@ export type ExternalSyncTable = PackSyncTable;
 
 type BasePackVersionMetadata = Omit<
   PackVersionMetadata,
-  'defaultAuthentication' | 'systemConnectionAuthentication' | 'formulas' | 'formats' | 'syncTables'
+  'defaultAuthentication' | 'formulas' | 'formats' | 'syncTables'
 >;
 
 // Further stripped-down version of `PackVersionMetadata` that contains only what the browser needs.
--- a/compiled_types.ts
+++ b/compiled_types.ts
@@ -144,7 +144,7 @@ export type ExternalSyncTable = PackSyncTable;
 
 type BasePackVersionMetadata = Omit<
   PackVersionMetadata,
-  'defaultAuthentication' | 'formulas' | 'formats' | 'syncTables'
+  'defaultAuthentication' | 'systemConnectionAuthentication' | 'formulas' | 'formats' | 'syncTables'
 >;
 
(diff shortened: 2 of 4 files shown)


patrick-codaio (approved):
Wow, can't believe that wasn't there before, thanks!

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Added
 
 - `coda execute` now allows for executing metadata formulas and sync updates.
+- Add `systemConnectionAuthentication` to `ExternalPackVersionMetadata`.
```
jonathan-codaio, line 10: For internal-only stuff we needn't update the changelog, that's meant for public consumption.

jonathan-codaio (changes requested):
Hmm what are we trying to accomplish here?

First off we can't just pass through systemConnectionAuthentication, this metadata object is meant to be a plain JSON object but the authentication definition has code in it (like getConnectionName), so anything like that needs to be transformed first.

I think the reason this wasn't included before was that end users aren't supposed to know or care about system authentication, so it doesn't need to be in external metadata.  Do end users need to know all the sudden?

kelsey-codaio:
> Hmm what are we trying to accomplish here?
> 
> First off we can't just pass through systemConnectionAuthentication, this metadata object is meant to be a plain JSON object but the authentication definition has code in it (like getConnectionName), so anything like that needs to be transformed first.
> 
> I think the reason this wasn't included before was that end users aren't supposed to know or care about system authentication, so it doesn't need to be in external metadata. Do end users need to know all the sudden?

Yeah, make sense. I was trying to address [this issue](https://staging.coda.io/d/_dL5PQJ5pk51?encodedLocationContext=eyJjb21tZW50VGhyZWFkSWQiOiJyLXE2ZzItSkh6aTYiLCJwYWdlSWQiOiJzZWN0aW9uLW5STVVxVnlRZjgiLCJ0eXBlIjoidGFibGUiLCJvYmplY3RJZCI6InRhYmxlLXV1dTFnTDJqcmwiLCJncmlkSWQiOiJncmlkLUUweXpvR1E0M1UiLCJyb3dJZCI6ImktN1dCQjZJVFYwViIsInZpZXdNb2RlIjoiZnVsbCJ9), which required knowing whether or not a pack was using system authentication from `syncTableGrid`. Thoughts on how else we could accomplish this? @jonathan-codaio

jonathan-codaio:
Hmm can we condition editability on

```typescript
pack.authentication &&
pack.authentication.type != ExternalPackAuthenticationType &&
syncFormula.connectionRequirement &&
syncFormula.connectionRequirement != ConnectionRequirement.None
```

I'm not sure that we need to know about system auth here. If `pack.authentication` is blank or None, it means there is no authentication that the end user needs to care about, so any connectionRequirement is either erroneous or an implementation detail and we can just ignore.

kelsey-codaio:
> Hmm can we condition editability on
> 
> ```ts
> pack.authentication &&
> pack.authentication.type != ExternalPackAuthenticationType.None &&
> syncFormula.connectionRequirement &&
> syncFormula.connectionRequirement != ConnectionRequirement.None
> ```
> 
> I'm not sure that we need to know about system auth here. If `pack.authentication` is blank or None, it means there is no authentication that the end user needs to care about, so any connectionRequirement is either erroneous or an implementation detail and we can just ignore.

The problem we were running into was that [Eric Koleda said](https://staging.coda.io/d/_dL5PQJ5pk51?encodedLocationContext=eyJjb21tZW50VGhyZWFkSWQiOiJyLXE2ZzItSkh6aTYiLCJwYWdlSWQiOiJzZWN0aW9uLW5STVVxVnlRZjgiLCJ0eXBlIjoidGFibGUiLCJvYmplY3RJZCI6InRhYmxlLXV1dTFnTDJqcmwiLCJncmlkSWQiOiJncmlkLUUweXpvR1E0M1UiLCJyb3dJZCI6ImktN1dCQjZJVFYwViIsInZpZXdNb2RlIjoiZnVsbCJ9) it should work even if the user doesn't explicitly set `connectionRequirement`. (Right now, `syncTableGrid.connectionRequirement` is undefined when using the example pack code snippet in the bug.) @jonathan-codaio

PR #2902, opened 2024-01-22
Remove guidance against `id` properties

ekoleda-codaio:
Previously the documentation advised against naming schema properties `id` since they would conflict with a hidden sync table column with the same. With the recent updates to the sync table infrastructure this is no longer the case, so the recommendation no longer makes sense.

This PR removed the best practice and updates sample code and snippets to use `id` properties. It also fixes a small bug with the sync table snippet.


ekoleda-codaio:
Assigning to on-call: @gary-codaio

gary-codaio (approved):

PR #2914, opened 2024-02-29
Incremental SDK type changes

patrick-codaio:


test/typedoc_coverage_test.ts
```diff
@@ -88,6 +88,9 @@ function hasComment(data: ReflectionData): boolean {
 }
 
 function messageForEntity(data: ReflectionData): string {
+  if (!data.kindString) {
+    throw new Error(`No kindString for ${data.name}, did you forget to add /** ... */ comments?`);
```
patrick-codaio, line 92: Drive by improvement:

Before:
Screenshot 2024-02-29 at 9 55 55 AM
After:
Screenshot 2024-02-29 at 9 55 47 AM

api.ts
```diff
@@ -957,15 +959,87 @@ export function isSyncPackFormula(fn: BaseFormula<ParamDefs, any>): fn is Generi
...
+  SchemaT extends ObjectSchemaDefinition<K, L>,
+> extends BaseSyncFormulaResult<K, L, SchemaT> {
+  /**
+   * Once there is no additional continuation returned from a pack sync formula, this may be returned instead, to give
+   * metadata about the entirety of the sync execution.
+   *
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  syncExecutionCompletionMetadata?: Omit<
```
jonathan-codaio, line 1005: Nit, I would probably give this type a name. It'll eventually show up in documentation too and it'll be clearer to be a named thing.

api.ts
```diff
@@ -2184,13 +2258,30 @@ export function makeSyncTable<
...
+    }
     return {
-      result: responseHandler({body: result || [], status: 200, headers: {}}, appliedSchema),
-      continuation,
-    } as SyncFormulaResult<K, L, SchemaT>;
+      result,
+      syncExecutionCompletionMetadata: {
+        // It's ok if this is empty
+        ...syncResult.syncExecutionCompletionMetadata,
+        syncExecutionCompletionMetadataVersion: CurrentSyncExecutionCompletionMetadataVersion,
```
jonathan-codaio, line 2282: I guess maybe this is ok, I guess I had only been thinking about persisting the completion data on the `coda` side and wanting those payloads to be versioned, as an implementation detail. Versioning the SDK output might be useful but feels maybe orthogonal?

dweitzman-codaio: Seems like CurrentSyncExecutionCompletionMetadataVersion in the metadata is kind of implied by the SDK version

Do we expect them to diverge and be able to change independently?

api_types.ts
```diff
@@ -704,12 +704,18 @@ export interface Sync {
    * value returned in the `continuation` property of result of the prior sync.
    */
   continuation?: Continuation;
+
+  // TODO(patrick): Unhide this
+  /** @hidden */
+  prevSyncContinuation?: Continuation;
```
dweitzman-codaio, line 710: The name `prevSyncContinuation` doesn't feel that distinct from `continuation`

I wonder if it's worth doing a quick naming brainstorm & voting exercise in a doc? Personally I might lean toward a name more like `incrementalSyncBase` to clearly distinguish it from non-incremental sync.

dweitzman-codaio (commented):
The type stuff makes sense to me. Had some questions about versioning and naming

api.ts
patrick-codaio: I'll just delete all this stuff. Y'all are right, the versioning can all be done on the Coda side.

api_types.ts
patrick-codaio: Ah you're right, I'll make a doc.

api.ts
```diff
@@ -958,14 +959,26 @@ export function isSyncPackFormula(fn: BaseFormula<ParamDefs, any>): fn is Generi
...
+  /**
+   * Once there is no additional continuation returned from a pack sync formula, this may be returned instead, to give
+   * metadata about the entirety of the sync execution.
+   *
+   * This is ignored if there is also a {@link continuation} on this object.
+   *
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  syncExecutionCompletionMetadata?: SyncExecutionCompletionMetadata;
```
ekoleda-codaio, line 981: Nitpick: The key name seems overly verbose. Would `metadata` suffice? Or perhaps `completion`?

patrick-codaio: I like how `completion` matches `continuation`, I'll use that.

ebo-codaio (approved):
Overall looks good

api_types.ts
patrick-codaio: NARRATOR: He did not make a doc

ebo-codaio (approved):

dweitzman-codaio (approved):

api.ts
```diff
@@ -2333,13 +2346,32 @@ export function makeSyncTable<
...
     const appliedSchema = context.sync.schema;
+    const result = responseHandler({body: syncResult.result || [], status: 200, headers: {}}, appliedSchema) as Array<
+      ObjectSchemaDefinitionType<K, L, SchemaT>
+    >;
+    const {continuation, completion: syncExecutionCompletionMetadata} = syncResult;
+
+    if (continuation) {
+      if (syncExecutionCompletionMetadata) {
+        // eslint-disable-next-line no-console
+        console.warn('Ignoring syncExecutionCompletionMetadata because there is also a continuation.');
```
jonathan-codaio, line 2363: I get a bit nervous about doing this kind of thing in the SDK because it gets bundled with the pack and we don't have any control over stopping it if we don't like it. Probably better to do this on the `coda` side if we even care to log it.

patrick-codaio: Ok I'll do a followup PR

patrick-codaio: https://github.com/coda/packs-sdk/pull/2970

PR #2922, opened 2024-03-01
add optional labels for properties to be hoisted out for better interpretation in coda

gary-codaio:
Adds a way for a pack maker to give some additional context for some types of fields


schema.ts
```diff
@@ -1165,6 +1165,51 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
    * {@link ValueHintType.ImageAttachment} or {@link ValueHintType.ImageReference} hints
    */
   imageProperty?: PropertyIdentifier<K>;
+  /**
+   * The name of a property within {@link ObjectSchemaDefinition.properties} that can be interpreted as the creation
+   * datetime of the object.
+   *
+   * Must be a {@link ValueType.String} property with the
```
jonathan-codaio, line 1172: We accept numbers as dates/datetimes too.

schema.ts
```diff
@@ -1165,6 +1165,51 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+  modifiedByProperty?: PropertyIdentifier<K>;
+  /**
+   * A list of property names from within {@link ObjectSchemaDefinition.properties} for the properties of the object
+   * that can be interpreted as people who are relevant to the object, not including the creator or last modifier.
+   *
+   * Each property must be a {@link ValueType.String} property with the {@link ValueHintType.Email} hint or
+   * a {@link ValueType.Object} with the {@link ValueHintType.Person} hint
+   * @hidden
+   */
+  relevantPeopleProperty?: Array<PropertyIdentifier<K>>;
```
jonathan-codaio, line 1212: Wondering if we just collect all the fields that are already tagged as ValueHintType.Email vs introducing a new thing? Granted you might not be able to disambiguate creator/modifier but that might not wind up that important in the end.

(For ValueHintType.Person, I think we'd just suggest that when the id property on the person corresponds to an email, make sure to use ValueHintType.Email on that property, and only focus on emails as identifiers for now. Later on we can deal with third-party user identifiers if it makes sense to.)

gary-codaio: Is the main reason not to use the Person type because it's not guaranteed to have an email?  I was thinking we could collect these objects anyways and filter out those without an email in the object.

On the point of collecting all fields vs asking to label, happy to do that and eliminate relevantPeopleProperty, though I might say keeping a created by and last modified by to distinguish those actions would be my leaning

jonathan-codaio: Yeah I'm not sure what we'd do with it if it doesn't have an email. Though maybe the name is useful enough?

Yeah, my suggestion was indeed to keep createdBy and modifiedBy but just infer the rest from the existing hint type(s).

dweitzman-codaio: Makes sense to me for general references to include any .Email or .Person field. Would probably need to add a helper function to pull them out, unless it fits naturally into some other function that's already parsing the schema

jonathan-codaio (approved):

dweitzman-codaio (approved):

patrick-codaio (approved):

PR #2923, opened 2024-03-02
Sync table hierarchy

patrick-codaio:


api_types.ts
```diff
@@ -391,6 +395,22 @@ export type SuggestedValueType<T extends UnionType> = T extends ArrayType<Type.d
...
+
+  // TODO(patrick): add more options here for date ranges, etc
+}
+
+/**
+ * A pointer to a particular property in another sync table.
+ */
+interface SyncTableRelation {
+  tableName: string;
+  propertyKey: string;
```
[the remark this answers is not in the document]
patrick-codaio: E.g., when pull requests have Repos for a parent, which value from a Repo object should be passed into the current param?

johnli-codaio (commented):
High level I think it looks great! Waiting to see what the bullpen brings but I think this seems fine to me.

Just for my own understanding; for the `crawlStrategy` I understand the parentTable as a necessary part of it, but it looks like we are fetching the property value of the parent sync table to do what exactly? Is this the whole `repoUrl` thing that we were thinking about?

Another question I had was if crawl strategies are going to be exposed in the process_csb_ingestion workflow? Would that be easy to do?

patrick-codaio:
@johnli-codaio propertyKey is indeed the repoUrl thing. And I'm expecting that the process_csb_ingestion workflow uses the `getSyncTableHierarchy` function.

johnli-codaio (approved):

testing/upload_validation.ts
```diff
@@ -227,6 +228,44 @@ export function validateSyncTableSchema(
   );
 }
 
+/**
+ * Returns a map of sync table names to their parent sync table names, or undefined if the hierarchy is invalid.
+ * Example valid return: { Child: 'Parent' }
```
ebo-codaio, line 233: are we going to use this function in the workflow? if we would like to create child ingestion from parent one, is it better to return `{Parent: 'Child'}` instead?

ebo-codaio: also thinking through what's needed during the child ingestion creation, we might also need to know how to build up the execution parameters for child ingestion. Would you have any thoughts on where to store the data? should it be in this mapping or we store it in the schema somewhere? 

The parameters could be `{parentPropertyParam: rowData[parentPropertyKey], grandParentPropertyParam: parentParameters[grandParentPropertyParam]}`

patrick-codaio: We should be able to parse the necessary data out of the childKey, right @johnli-codaio ?

johnli-codaio: Yeah I can include these params inside the childKey, but wouldn't these execution params also just be passed into the `ingestionParam` column in our `csb_ingestion_configs` table?

We would need the SDK to expose the execution params somehow for it's children (ex: repoUrl for pull requests, calendarId for events, etc.) I think?

johnli-codaio: To me I thought the childKey was mostly for identification and deduping purposes, not necessarily for holding ingestion params, but I could use it for ingestion params as well (at the very least, enough to identify the entity we're crawling, like pullRequestId or calendarId, etc.)

ebo-codaio (approved):
Overall looks good, this will provide details to the crawling in workflow

testing/upload_validation.ts
ebo-codaio: exactly, we need to store `parentPropertyParam`, `parentPropertyKey`, and maybe grandparent ones later to build the `ingestionParam` column. Where do we generate `childKey`?

johnli-codaio: I'm thinking we can generate childKey in the `process_csb_ingestion` workflow, before we insert a new ingestion and upsert a new config. You can add a TODO for me to handle that.

ebo-codaio: But `ingestionParam` also needs be built in the same place in order to upsert child ingestion? 
image

johnli-codaio: I'm thinking what we do is:

1. get Sync Hierarchy
2. Calculate childKey based on syncHierarchy and some identification of the entity we are crawling (repoUrl, calendarId, etc.)
3. Build ingestionParameters and all config stuff 
4. We might need a new SQL query (insert_child_ingestion) or something like that that would be a TXN that would:
  a. Insert the child ingestion first. If it conflicts (a childKey already exists) then we just return. Keeps the SQL idempotent
  b. Insert the child ingestion config with the ingestionParams.

ebo-codaio: Yeah, this looks good to me, just would like to find an easy way to get `parentPropertyParam` and `parentPropertyKey` in the SDK, current function only returns function names mapping

patrick-codaio: I think it might make sense to double-store the data in the child key and the ingestion params, though I'm not sure.

But to answer Evan, sorry I misunderstood you before, I was imagining something like:
```
const currentPageOfSyncResults = [...]
const childTables = getSyncTableHierarchy(pack)[currentTable];
for (const childTable of childTables) {
  const syncTableDef = pack.syncTables.find(...);
  const crawlDefinedParams = syncTableDef.getter.parameters.filter(p=>p.crawlStrategy...);
  for (const parentRow of currentPageOfSyncResults) {
    const crawlParams = {};
    for (const paramDef of crawlDefinedParams) {
      crawlParams[paramDef.name] = parentRow[paramDef.crawlStrategy.parentTable.propertyKey];
    }
    // create key
    // insert ingestion/config
  }
}
```

ebo-codaio: Oh, got it, that makes sense. We would have to double store if it's required by dedupe. Thanks!

api.ts
```diff
@@ -498,7 +500,22 @@ export function makeParameter<T extends ParameterType, O extends ParameterOption
...
+    if (crawlStrategyDef.parentTable) {
+      const {tableName, propertyKey} = crawlStrategyDef.parentTable;
+      crawlStrategy = {
+        parentTable: {
+          tableName,
+          propertyKey: normalizeSchemaKey(propertyKey),
+        },
+      };
+    } else {
+      crawlStrategy = crawlStrategyDef;
```
johnli-codaio, line 514: Curious if this `else` will ever be hit? when would `crawlStrategyDef` be defined but not `parentTable`?

test/upload_validation_test.ts
```diff
@@ -2172,6 +2175,321 @@ describe('Pack metadata Validation', async () => {
...
+          const metadata = createFakePack({
+            id: 1013,
+            syncTables: [table2, table1],
+          });
+          await testHelper.willBeRejectedWith(validateJson(metadata), /Sync table parent hierarchy is invalid/);
+          const hierarchy = getSyncTableHierarchy(metadata);
+          assert.isUndefined(hierarchy);
+        });
+
+        it('does not allow multiple parents', async () => {
```
johnli-codaio, line 2434: I could imagine a world where a child table relies on two parents though? Perhaps none of our data sources have that right now?

Could iterate on this in case that does happen! not blocking

johnli-codaio (approved):

api.ts
patrick-codaio: Not in this PR, but my plan is to add date range crawling, etc into `crawlStrategy` (that's why I defined it at the param level)

test/upload_validation_test.ts
patrick-codaio: Yeah, I reviewed all our 1st party sync tables, and the only ones with multiple required fields were either:
- parent & grandparent
- date ranges
- bools
- enum, defined statically by an autocomplete function

I'm hopeful that we can go a long time without needing to support that.

testing/upload_validation.ts
```diff
@@ -1704,6 +1797,16 @@ function buildMetadataSchema({sdkVersion}: BuildMetadataSchemaArgs): {
         }
       });
     })
+    .superRefine((data, context) => {
+      const crawlHierarchy = getSyncTableHierarchy(data as PackVersionDefinition);
+      if (!crawlHierarchy) {
+        context.addIssue({
+          code: z.ZodIssueCode.custom,
+          path: ['syncTables'],
+          message: 'Sync table parent hierarchy is invalid',
```
jonathan-codaio, line 1806: Can do later but ideally these messages are easily actionable, so would be good to pinpoint specific invalid cases, e.g. `crawlStrategy.parentTable "Issus" not found in this pack` and have the error path point at the table and parameter with the error.

So may want to reframe `getSyncTabeHierarchy` as `validateCrawlHierarchy` or something and pass the zod context so you can traverse using it.

patrick-codaio: Great idea, done!

testing/upload_validation.ts
```diff
@@ -227,6 +228,120 @@ export function validateSyncTableSchema(
...
+        continue;
+      }
+      if (param.crawlStrategy.parentTable) {
+        const {tableName, propertyKey} = param.crawlStrategy.parentTable;
+
+        const tableSchema = syncTableSchemasByName[tableName];
+        if (!tableSchema) {
+          context?.addIssue({
+            code: z.ZodIssueCode.custom,
+            path: ['syncTables', index],
```
jonathan-codaio, line 262: The path to the violation would actually be `['syncTables', index, 'parameters', paramIndex, 'crawlStrategy', 'parentTable']` I think?

testing/upload_validation.ts
```diff
@@ -227,6 +228,120 @@ export function validateSyncTableSchema(
...
+            message: `Sync table ${syncTable.name} expects parent table ${tableName} to exist.`,
+          });
+          return undefined;
+        }
+
+        const property = tableSchema.properties[propertyKey];
+        if (!property) {
+          context?.addIssue({
+            code: z.ZodIssueCode.custom,
+            path: ['syncTables', index],
```
jonathan-codaio, line 272: Ditto

testing/upload_validation.ts
```diff
@@ -227,6 +228,120 @@ export function validateSyncTableSchema(
...
+          return undefined;
+        }
+
+        // TODO(patrick): Validate the types match
+
+        // We only allow one parent per table.
+        if (parent && parent !== tableName) {
+          context?.addIssue({
+            code: z.ZodIssueCode.custom,
+            path: ['syncTables', index],
```
jonathan-codaio, line 284: This one is probably ok since both crawlStrategies are valid on their own but it's the presence of both that creates the violation.

testing/upload_validation.ts
```diff
@@ -227,6 +228,120 @@ export function validateSyncTableSchema(
...
+  // Verify that there's no cycle
+  if (_hasCycle(result)) {
+    context?.addIssue({
+      code: z.ZodIssueCode.custom,
+      path: ['syncTables'],
+      message: `Sync table parent hierarchy is cyclic`,
+    });
+    return undefined;
+  }
+  return result;
```
jonathan-codaio, line 309: Do we use the result anymore?

jonathan-codaio (approved):

testing/upload_validation.ts
patrick-codaio: Not here, but the coda code will need it.

testing/upload_validation.ts
```diff
@@ -227,6 +228,120 @@ export function validateSyncTableSchema(
...
+    syncTableSchemasByName[syncTable.name] = syncTable.schema;
+  }
+  for (const [index, syncTable] of pack.syncTables.entries()) {
+    let parent: string | undefined;
+    for (const param of syncTable.getter.parameters) {
+      if (!param.crawlStrategy) {
+        continue;
+      }
+      if (param.crawlStrategy.parentTable) {
+        const {tableName, propertyKey} = param.crawlStrategy.parentTable;
```
dweitzman-codaio, line 256: Might be worth calling this local variable `parentTableName` for clarity?

testing/upload_validation.ts
```diff
@@ -227,6 +228,120 @@ export function validateSyncTableSchema(
...
+            path: ['syncTables', index],
+            message: `Sync table ${syncTable.name} expects parent table ${tableName}'s schema to have the property ${propertyKey}.`,
+          });
+          return undefined;
+        }
+
+        // TODO(patrick): Validate the types match
+
+        // We only allow one parent per table.
+        if (parent && parent !== tableName) {
```
dweitzman-codaio, line 281: Maybe a name in the spirit of `expectedParentTableName` or `alreadyDiscoveredParentTableName` would be more clear than `parent`?

testing/upload_validation.ts
```diff
@@ -227,6 +228,120 @@ export function validateSyncTableSchema(
...
+ * Returns a map of sync table names to their child sync table names, or undefined if the hierarchy is invalid.
+ * Example valid return: { Parent: 'Child' }
+ * {} is also a valid result, when there are no sync tables, or no parent relationships.
+ * @hidden
+ */
+export function validateCrawlHierarchy(
+  pack: PackVersionDefinition,
+  context?: z.RefinementCtx,
+): Record<string, string[]> | undefined {
+  const result: Record<string, string[]> = {};
```
dweitzman-codaio, line 241: Maybe a name like `parentToChildrenTableNames` could be more clear than `result`? It's a large enough function that the docstring might not always be visible as people are reading through it.

testing/upload_validation.ts
```diff
@@ -227,6 +228,120 @@ export function validateSyncTableSchema(
...
+      }
+      if (subtreeHasCycle(child, subtree, visited)) {
+        return true;
+      }
+    }
+    visited.delete(currentKey);
+    return false;
+  }
+
+  for (const key of Object.keys(tree)) {
```
dweitzman-codaio, line 333: If you're feeling cheeky this could maybe be shortened to something like `return subtreeHasCycle('__CODA_INTERNAL_ROOT__', Object.keys(tree), new Set<string>())`

testing/upload_validation.ts
patrick-codaio: I'll just add 'parameters' without an index.

testing/upload_validation.ts
patrick-codaio: Good point

testing/upload_validation.ts
patrick-codaio: Very smart.

testing/upload_validation.ts
patrick-codaio: Changing to `firstDiscoveredParentTable`

testing/upload_validation.ts
patrick-codaio: Great point.

testing/upload_validation.ts
patrick-codaio: So cheeky! I love it.

PR #2924, opened 2024-03-06
Permissions

patrick-codaio:


test/schema_test.ts
```diff
@@ -371,6 +371,38 @@ describe('Schema', () => {
...
+        properties: {
+          name: {type: ValueType.String, required: true},
+          id: {type: ValueType.Number, required: true},
+          other: {type: ValueType.Boolean},
+        },
+        displayProperty: 'name',
+        idProperty: 'id',
+      });
+
+      const thingReferenceSchema = schema.makeReferenceSchemaFromObjectSchema(thingSchema);
```
patrick-codaio, line 388: (I noticed we didn't have any test for a schema that still sets its own identity)

api.ts
```diff
@@ -1127,6 +1128,23 @@ export interface SyncFormulaDef<
...
    */
   updateOptions?: Pick<CommonPackFormulaDef<ParamDefsT>, 'extraOAuthScopes'>;
+
+  /**
+   * Given a set of rows, return permission lists, keyed by the row's idProperty value.
+   *
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  getPermissions?(rows: Array<ObjectSchemaDefinitionType<K, L, SchemaT>>): Promise<Record<string, Permission[]>>;
```
sam-codaio, line 1138: Nice love this. Only thing to think through is if this could take in just the id field or if it needs the entire row? Just want to think about _when_ we will be syncing permissions. Like if we want to sync them on a different cadence than the rows will we still have access to the entire row content?

dist/bundle.d.ts
```diff
@@ -1854,6 +1854,24 @@ export interface ObjectSchemaDefinition<K extends string, L extends string> exte
...
 	 */
 	modifiedByProperty?: PropertyIdentifier<K>;
+	/**
+	 * This property should be typed as either an Email or a Person.
+	 * @hidden
+	 */
+	permissionUserProperty?: K;
+	/**
+	 * This property should be typed one of: an Email, a Person, or a reference to a sync table which has
+	 * isUserPrincipalTable=true.
```
sam-codaio, line 1864: I think we mean isGroupPrincipalTable=true?

sam-codaio (commented):
Looking good a minor comment or two

schema.ts
```diff
@@ -1792,14 +1843,18 @@ export function normalizeObjectSchema(schema: GenericObjectSchema): GenericObjec
       ? subtitleProperties.map(subProp => normalizeSchemaPropertyIdentifier(subProp, normalizedProperties))
       : undefined,
     titleProperty: titleProperty ? normalizeSchemaPropertyIdentifier(titleProperty, normalizedProperties) : undefined,
-    createdAtProperty: createdAtProperty ?
```
patrick-codaio, line 1795: sorry for the distracting auto-formats

api.ts
```diff
@@ -1144,6 +1145,23 @@ export interface SyncFormulaDef<
...
+
+  /**
+   * If the table implements {@link getPermission}, the maximum number of rows that will be sent to that
+   * function in a single batch. Defaults to 10 if not specified.
+   *
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  maxPermissionBatchSize?: number;
```
neal-codaio, line 1164: Would this be better as a parameter to getPermissions?

schema.ts
```diff
@@ -1203,6 +1211,44 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
   modifiedByProperty?: PropertyIdentifier<K>;
 
   // TODO(dweitzman): Only support options in the typing when the codaType is ValueHintType.SelectList.
+
+  // TODO(patrick): Unhide this.
+  /**
+   * This property should be typed as either an Email or a Person.
+   * @hidden
+   */
+  permissionUserProperty?: K;
```
neal-codaio, line 1220: I'm not sure if this is possible.  Would it be desirable to have a single `permissionPropery` which is an object?  That way we can change it easier and also not have to prefix every property w/permission.

schema.ts
```diff
@@ -1203,6 +1211,44 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+export interface Permission {
+  // The simplest ACL is just a user email.
+  userEmail?: string;
+
+  // If there is an isUserPrincipal table, you can use the idProperty of its rows to reference them.
+  userReference?: string | number;
+
+  // If more than 1 sync table isGroupPrincipal, this specifies the identityName of the group
+  // being referenced.
+  groupIdentityName?: string;
```
neal-codaio, line 1242: If a user can be in multiple groups, how are they specified?  Each group must have a `Permission`?  Should any of the groups properties be arrays?

api.ts
patrick-codaio: Not sure how that would work... the pack maker defines the number, but coda code invokes getPermissions...

schema.ts
patrick-codaio: Oh, interesting... so the validation of it would just be dynamic based on what kind of principal the table is marked as? That could work...

schema.ts
patrick-codaio: A user's membership in groups is defined by the isGroupPrincipalTable's permission property, not here.

Here's an example of what I'm imagining one item could return as its permission list:
```
[
  {userEmail: 'creator@foo.com'},
  {groupIdentityName: 'Team', groupReference: 'project-x-team'},
  {groupIdentityName: 'Team', groupReference: 'project-y-team'},
  {groupIdentityName: 'Department', groupReference: 'engineering'}
]
```

Not necessarily the most efficient, I grant you.

PR #2926, opened 2024-03-07
Add documentation for displayName

ekoleda-codaio:


alan-codaio (approved):

docs/guides/advanced/schemas.md
```diff
@@ -547,6 +547,30 @@ for (let customField of customFeilds) {
 ```
 
 
+### Column names {: #column-names}
+
+By default sync table column names are derived from the [normalized name](#normalization) of the corresponding schema property. For example, a property with the key `contact_email` would be normalized to the property name `contactEmail` which would result in a column named "Contact email". This is just the default name for the column however, and users can later rename it in their doc if they so desire.
```
patrick-codaio, line 552: technically i think the normalized form is `ContactEmail`

docs/guides/advanced/schemas.md
```diff
@@ -547,6 +547,30 @@ for (let customField of customFeilds) {
...
+      type: coda.ValueType.String,
+      displayName: "iPhone version",
+    },
+    // ...
+  },
+  // ...
+});
+```
+
+This display name will only be used to set the default column name for that property, and won't have any affect on how the property appears in the formula editor, etc. Changes to the `displayName` will rename existing columns upon their next sync, as long as the user hasn't manually changed the column name.
```
patrick-codaio, line 571: Hrm, I think this could be improved. If you're referencing the column, the formula editor does use the display name. It's only if you're digging into the Value column that you'll see the raw keys, and I wouldn't expect most end users to be doing that. CC @alan-codaio

docs/guides/advanced/schemas.md
ekoleda-codaio: Whoops, good catch. Will send a follow-up PR.

docs/guides/advanced/schemas.md
ekoleda-codaio: This section talks about schemas in general, not just in sync tables. If you use a formula that returns a schema you'll only have access to the raw keys, and I don't think that's an edge case. "In the formula editor" wasn't very clear though, since you can access columns backed back schema properties that way. I'll send a follow-up PR.

PR #2928, opened 2024-03-10
Allow prereleases

patrick-codaio:


jonathan-codaio (approved):
Amusingly you will probably need to bump the prerelease version here in package.json in order to roll this out in `coda` 😛

PR #2929, opened 2024-03-11
Update SDK to handle user entity sync tables

sam-codaio:
* Create user entity table to be used with user mapping
* Require 2 fields (userId and userEmail) when on any sync table marked as the user table. Require userEmail to be email or person. Require userId to be string or number.


schema.ts
```diff
@@ -178,6 +178,12 @@ export enum ValueHintType {
    * Indicates to render a value as a select list.
    */
   SelectList = 'selectList',
+
+  /**
+   * Indicates an external user id
```
neal-codaio, line 183: Is external clear in this context?  External to what?  Or not internal to what?

schema.ts
```diff
@@ -178,6 +178,12 @@ export enum ValueHintType {
...
+   * @hidden
+   */
+  UserId = 'userId',
```
neal-codaio, line 186: Should this be called ExternalUserId or something more specific depending on the answer to the comment above?

neal-codaio: And if so, I'm not sure how much that applies to.

sam-codaio:
Should we limit packs to 1 and only 1 user entity table?

gary-codaio (approved):
Great!

api.ts
```diff
@@ -293,6 +294,12 @@ export interface SyncTableDef<
    * @hidden
    */
   namedPropertyOptions?: SyncTablePropertyOptions;
+
+  /**
+   * See {@link SyncTableOptions.roles}
+   * @hidden
+   */
+  roles?: TableRole[];
```
jonathan-codaio, line 302: Per the other discussion it seems like it was unnecessarily complex to make this an array without a clear use case.

api.ts
```diff
@@ -1972,6 +1979,14 @@ export interface SyncTableOptions<
    * sync tables that have a dynamic schema.
    */
   dynamicOptions?: DynamicOptions;
+
+  /**
+   * Used to mark a sync table as having a specific purpose
```
jonathan-codaio, line 1984: We'll probably do a sweep through the tsdoc comments before we would publish them but there's always a chance we miss some of these, so I'd commit everything up to the style of our published documentation. So always complete sentences and properly punctuated.

I might do something like "Used to indicate that the entities in this table have a specific semantic meaning known to Coda, for example, that the rows being synced each represent a user."

schema.ts
```diff
@@ -1201,7 +1211,21 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
    * @hidden
    */
   modifiedByProperty?: PropertyIdentifier<K>;
-
+  /**
+   * The name of a property within {@link ObjectSchemaDefinition.properties} that can be interpreted as the email
```
jonathan-codaio, line 1215: Probably `For cases when the object being synced represents a user, the name of the property within {@link ObjectSchemaDefinition.properties} that identifies the email address of the user.`

schema.ts
```diff
@@ -1201,7 +1211,21 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   *
+   * Must be a {@link ValueType.String} property with the {@link ValueHintType.Email} hint or
+   * a {@link ValueType.Object} with the {@link ValueHintType.Person} hint
+   * @hidden
+   */
+  userEmailProperty?: PropertyIdentifier<K>;
+  /**
+   * The name of a property within {@link ObjectSchemaDefinition.properties} that can be interpreted as the user id
```
jonathan-codaio, line 1223: Probably `For cases when the object being synced represents a user, the name of the property within {@link ObjectSchemaDefinition.properties} that identifies the id of the user in the service being synced from.`

api.ts
sam-codaio: Moved back to being singular!

schema.ts
```diff
@@ -1201,7 +1211,24 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+  userEmailProperty?: PropertyIdentifier<K>;
+  /**
+   * For cases where the object being synced represents a user, the name of the property within
+   * {@link ObjectSchemaDefinition.properties} that identifies the id of the user in the service
+   * being synced from.
+   *
+   * Must be a {@link ValueType.String} or {@link ValueType.Number} property
+   * @hidden
+   */
+  userIdProperty?: PropertyIdentifier<K>;
```
ekoleda-codaio, line 1231: Would it be safe to assume / require that the `idProperty` of the schema of a table with the `Users` role be the user ID? Hard to imagine a case where a Pack maker would want / need a different ID for the row.

schema.ts
```diff
@@ -1201,7 +1211,24 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * a {@link ValueType.Object} with the {@link ValueHintType.Person} hint
+   * @hidden
+   */
+  userEmailProperty?: PropertyIdentifier<K>;
+  /**
+   * For cases where the object being synced represents a user, the name of the property within
+   * {@link ObjectSchemaDefinition.properties} that identifies the id of the user in the service
+   * being synced from.
+   *
+   * Must be a {@link ValueType.String} or {@link ValueType.Number} property
```
ekoleda-codaio, line 1228: Would this allow for properties with the hint `Email` to be used as well? I imagine in some systems the only user ID is the user email.

schema.ts
```diff
@@ -1201,7 +1211,24 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
-
+  /**
+   * For cases where the object being synced represents a user, the name of the property within
+   * {@link ObjectSchemaDefinition.properties} that identifies the email address of the user.
+   *
+   * Must be a {@link ValueType.String} property with the {@link ValueHintType.Email} hint or
+   * a {@link ValueType.Object} with the {@link ValueHintType.Person} hint
+   * @hidden
+   */
+  userEmailProperty?: PropertyIdentifier<K>;
```
ekoleda-codaio, line 1222: I wonder if we could make this more generic, like: 
- `emailProperty`
- `primaryEmailProperty`
- `contactEmailProperty`
Then we could use the same field for tables with other roles (`Group`, etc).

schema.ts
sam-codaio: Yeah this would allow that

schema.ts
sam-codaio: I was not positive here. There can be multiple uniqueish identifiers in a system. For example in github right now we return the login id (not the numeric id) as the id of the users table https://github.com/coda/packs/blob/main/packs/github/main/schemas.ts. I would think generally for our purposes we would prefer to use the numeric id although it could be argued either way.

schema.ts
sam-codaio: Discussed f2f. Will likely leave as is here for now

schema.ts
jonathan-codaio: Yeah that's interesting, it does seem like the idProperty should always be the user id in a users table. I'd be open to omitting this and adding it back if it comes up for real.

jonathan-codaio (approved):

testing/upload_validation.ts
```diff
@@ -1407,6 +1410,25 @@ function buildMetadataSchema({sdkVersion}: BuildMetadataSchemaArgs): {
           );
         };
 
+        const validateUserEmailProperty = () => {
+          return validateProperty(
+            'userEmailProperty',
+            userEmail =>
+              (userEmail.type === ValueType.Object || userEmail.type === ValueType.String) &&
```
patrick-codaio, line 1417: I'd rather this be
```
(userEmail.type === ValueType.String && userEmail.codaType === ValueType.Email) ||
(userEmail.type === ValueType.Object && userEmail.codaType === ValueType.Person)
```

test/upload_validation_test.ts
```diff
@@ -4659,4 +4660,116 @@ describe('Pack metadata Validation', async () => {
...
+          syncTables: [syncTable],
+          defaultAuthentication: {
+            type: AuthenticationType.None,
+          },
+        }),
+      );
+      await doValidateJson(metadata);
+    });
+
+    it('Fail when required fields are missing for users', async () => {
```
patrick-codaio, line 4702: I think you didn't test the situation where the Users role is specified but the schema doesn't specify both `userEmailProperty` and `userIdProperty`... (and I think that test will fail if I'm reading your validation code correctly)

schema.ts
patrick-codaio: Would it be reasonable to keep this in but leave it optional, and if not present, assume the `idProperty` is the `userIdProperty`?

schema.ts
```diff
@@ -1048,7 +1048,17 @@ export type PropertyIdentifier<K extends string = string> = K | string | Propert
...
+  | 'linkProperty'
+  | 'imageProperty'
+  | 'snippetProperty'
+  | 'subtitleProperties'
+  | 'createdAtProperty'
+  | 'createdByProperty'
+  | 'modifiedAtProperty'
+  | 'modifiedByProperty'
+  | 'userEmailProperty'
+  | 'userIdProperty'
```
patrick-codaio, line 1061: FWIW it's nice to take advantage of diffs like this as opportunities to alphabetize without unnecessary thrash

test/upload_validation_test.ts
sam-codaio: Writing the test but we are validating`Error: Sync table schemas with role users must set a userEmailProperty`

sam-codaio: It fails in makeSyncTable (since that is where I put the validation) not in upload_validation. Can add an error there as well

PR #2930, opened 2024-03-12
Add Prerelease tag for Users Role

sam-codaio:
* Create Prerelease tag for users Role
* Add additional testing in upload_validation_test
* Fix logic in upload validation on property type

### DIFF
--- a/test/upload_validation_test.ts
+++ b/test/upload_validation_test.ts
@@ -4735,6 +4735,32 @@ describe('Pack metadata Validation', async () => {
       await validateJsonAndAssertFails(metadata);
     });
 
+    it('Fail when both required fields are missing for users', async () => {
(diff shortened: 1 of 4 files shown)


test/upload_validation_test.ts
```diff
@@ -4735,6 +4735,32 @@ describe('Pack metadata Validation', async () => {
       await validateJsonAndAssertFails(metadata);
     });
 
+    it('Fail when both required fields are missing for users', async () => {
+      assert.throws(() =>
```
jonathan-codaio, line 4739: Probably good to add a failure message or regex as the second arg here to make sure it's throwing for the expected reason.

jonathan-codaio (approved):

patrick-codaio (approved):

PR #2931, opened 2024-03-15
Create executeGetPermissions function in sync tables

sam-codaio:
* Create a getPermissions function that can be used to fetch a set of permissions for a set of entities
* Create a Permissions type that uses userId, groupId, domainName and public as the possible principals
* Does NOT include defining permission inheritance hierarchy


api.ts
```diff
@@ -1151,6 +1153,33 @@ export interface SyncFormulaDef<
...
+
+  /**
+   * The javascript function that implements fetching permissions for a set of objects
+   * if the objects in this sync table have permissions in the external system.
+   *
+   * TODO(sam) pre merge:
+   * Does this need continuation? What if we fetch like 4 items and we need to continue for the last one?
+   * What are the size limits?
+   * Should we return as a flattened list or a map of lists?
+   * What do we do if there are no permissions on the objects but we just care about the permissions of the parent?
```
sam-codaio, line 1165: This is part of a broader permissions inheritance issue. I think we likely can just add something new to Permissions that is like useParentPermissions. This works as long as the row contains an id that refers to the parent. Otherwise would need to flow in the parameters from the parent sync or something?

schema.ts
```diff
@@ -1232,6 +1232,43 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+/**
+ * Represents a permission in the external system
+ *
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+export interface Permission {
+  /**
+   * Indicates whether the permission is public to all users who have access to the ingestion
```
neal-codaio, line 1244: Do we want to talk about ingestion here?

schema.ts
```diff
@@ -1232,6 +1232,43 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
   // TODO(dweitzman): Only support options in the typing when the codaType is ValueHintType.SelectList.
 }
 
+/**
+ * Represents a permission in the external system
```
neal-codaio, line 1236: Is this only external systems?  Do we need to define what we mean by external system here?  Not sure what we want to say publicly.

schema.ts
```diff
@@ -1232,6 +1232,43 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * The ID of the group associated with the permission.
+   *
+   * TODO(sam): Update with description of where this group id is set
+   */
+  groupId: string;
+
+  /**
+   * The domain name associated with the permission.
+   */
+  domainName: string;
```
neal-codaio, line 1267: Do we need this?  Can we look this up based on the group?

sam-codaio: We need this. Domain Name is a separate auth mechanism then groups. We can potentially remove it for v0 though as we may bring in support later down the line

neal-codaio: So then should groupId and domainName be `| undefined` (or `?`)?  And then we document that at least one must be specified?  Can both be specified?  Should we use an additional type to make these things more clear?

neal-codaio: I guess I'm thinking about this from a non-Coda perspective.  What if someone wants some data to be open to the world, some to be limited by domain, and some to be limited by group, how would they do that?  And do the comments provide enough context to implement what they want.

sam-codaio: Good point these should all be optional and at most 1 should be specified per row

schema.ts
sam-codaio: That is a good question I guess it just means a permission on the entity instead of external system which is technically optional in a pack

api.ts
```diff
@@ -2249,6 +2292,36 @@ export function makeSyncTable<
...
+          .map(([rowId, _count]) => rowId);
+
+        if (oversizedRowIds.length > 0) {
+          throw new Error(
+            `Permissions limit exceeded limit of ${MaxPermissionsPerRow} permissions for row with ids: ${oversizedRowIds.join(
+              ', ',
+            )}`,
+          );
+        }
+        return result;
```
sam-codaio, line 2321: We should likely log the full permissions when this error occurs. Would be good to do this the canonical way

api.ts
```diff
@@ -1108,6 +1110,25 @@ export interface SyncUpdateResultMarshaled<
  */
 export type GenericSyncUpdateResultMarshaled = SyncUpdateResultMarshaled<any, any, any>;
 
+/**
+ * Sets a limit on the number of permissions that can be returned for a single row
+ * in a call to {@link executeGetPermissions}.
+ *
+ * @hidden
```
neal-codaio, line 1117: Did you want TODOs to unhide this and the next occurrence?

api.ts
```diff
@@ -2249,6 +2292,36 @@ export function makeSyncTable<
...
+  const executeGetPermissions = wrappedExecuteGetPermissions
+    ? async function execGetPermissions(
+        rows: Array<ObjectSchemaDefinitionType<K, L, SchemaDefT>>,
+        context: GetPermissionExecutionContext,
+      ) {
+        const result = await wrappedExecuteGetPermissions(rows, context);
+        const {permissions} = result;
+        const permissionCountByRow: {[rowId: string]: number} = permissions.reduce(
+          (acc: {[rowId: string]: number}, permission: Permission) => {
+            acc[permission.rowId] = acc[permission.rowId]++ || 1;
```
neal-codaio, line 2304: Do you want `??` instead of `||`?

api.ts
neal-codaio: What if there are 1000 rows that are oversized?  Do we want to have all this in the error?  Maybe we should add a property onto the error, but only show the first 10 or some number?

api_types.ts
```diff
@@ -748,6 +748,12 @@ export interface Sync {
  */
 export type UpdateSync = Omit<Sync, 'continuation'>;
 
+/**
+ * Information about the current sync, part of the {@link GetPermissionExecutionContext} passed to the
+ * `executeGetPermissions` function of the sync formula.
+ */
```
neal-codaio, line 754: `@hidden`?

api.ts
sam-codaio: Yeah good question. We could set some reasonable limit on what we log here. Will wait on some input from others but fine doing a limit

api.ts
```diff
@@ -1151,6 +1172,27 @@ export interface SyncFormulaDef<
...
   updateOptions?: Pick<CommonPackFormulaDef<ParamDefsT>, 'extraOAuthScopes'>;
+
+  /**
+   * The javascript function that implements fetching permissions for a set of objects
+   * if the objects in this sync table have permissions in the external system.
+   *
+   * TODO(sam): Unhide this
+   * @hidden
+   */
+  executeGetPermissions?(
```
sam-codaio, line 1183: Do we need a new formula type here?

api.ts
```diff
@@ -1151,6 +1172,27 @@ export interface SyncFormulaDef<
...
+    rows: Array<ObjectSchemaDefinitionType<K, L, SchemaT>>,
```
sam-codaio, line 1184: Should we go with a wrapping object here. Might better fit the pattern and leave this more open for modifications in the future

jonathan-codaio (commented):
Haven't had a chance to review the schema changes yet, first set of comments here.

api.ts
```diff
@@ -2264,6 +2337,7 @@ export function makeSyncTable<
       supportsUpdates: Boolean(executeUpdate),
```
jonathan-codaio, line 2337: You probably need an equivalent of this, otherwise we don't know when looking at a pack's metadata whether there is an executeGetPermissions function to call (javascript gets stripped out of the metadata).

schema.ts
```diff
@@ -1232,6 +1232,80 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
   // TODO(dweitzman): Only support options in the typing when the codaType is ValueHintType.SelectList.
 }
 
+/**
+ * Discriminent used to determine the permission type
+ * @hidden
+ */
+export enum PermissionType {
```
jonathan-codaio, line 1239: Should we be a bit more technical/accurate here and call these principal types?

schema.ts
```diff
@@ -1232,6 +1232,80 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+  User = 'user',
+  Group = 'group',
+  Domain = 'domain',
+  Public = 'public',
```
jonathan-codaio, line 1243: What does this actually mean? For our current use case, this will be "organization-wide", should that be the principal that makers actually specify, given that?

api.ts
```diff
@@ -1151,6 +1172,27 @@ export interface SyncFormulaDef<
...
+  /**
+   * The javascript function that implements fetching permissions for a set of objects
+   * if the objects in this sync table have permissions in the external system.
+   *
+   * TODO(sam): Unhide this
+   * @hidden
+   */
+  executeGetPermissions?(
+    rows: Array<ObjectSchemaDefinitionType<K, L, SchemaT>>,
+    context: GetPermissionExecutionContext,
```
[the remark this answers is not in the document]
sam-codaio: We almost definitely should provide that if that is the case that we need it for updates. I would assume the same logic applies here

api.ts
sam-codaio: I do not think there is a real alternative after spleunking through the code in more depth

api.ts
sam-codaio: Yep had made a change here to a request object

api.ts
```diff
@@ -1108,6 +1110,25 @@ export interface SyncUpdateResultMarshaled<
  */
 export type GenericSyncUpdateResultMarshaled = SyncUpdateResultMarshaled<any, any, any>;
 
+/**
+ * Sets a limit on the number of permissions that can be returned for a single row
+ * in a call to {@link executeGetPermissions}.
+ *
+ * @hidden
+ */
+const MaxPermissionsPerRow = 1000;
```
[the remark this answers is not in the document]
sam-codaio: Good question. We thought this was a reasonable limit we can lift/change in the future.

I think at there are a two options I was thinking about for the future: 

1) Just limit the number of entities per batch instead of for a single item in a batch. That way someone can just change the batchSize if there are tons of permissions.
2) Add a continuation concept to the execution context/return if we hit this limit. If we see continuation we pass in the same rows until we do not see one.

Wanted to be fairly strict to start out and we can unwind this if we find the need

api.ts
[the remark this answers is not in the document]
sam-codaio: Added

api.ts
sam-codaio: Added this as well

schema.ts
sam-codaio: Yes

schema.ts
sam-codaio: This is a big question I had. I think the problem is ingestions can be configured in some context (that we have not decided upon fully. Could be workspaces or organizations). Does it make sense for the pack maker to specify which context this thing is available in. I think what they are trying to communicate is `This is available to anyone who has access to the sync`. Happy to discuss

api.ts
patrick-codaio: I'm curious why this was marked resolved? Seems like a valid point?

api.ts
```diff
@@ -1108,6 +1110,49 @@ export interface SyncUpdateResultMarshaled<
...
+
+/**
+ * Type definition for the result of calls to {@link executeGetPermissions}.
+ * @hidden
+ */
+export interface ExecuteUpdatePermissionResult {
```
patrick-codaio, line 1125: Shouldn't this be `ExecuteGetPermissionsResult`?

patrick-codaio: I'd also be fine with `GetPermissionsResult`

api.ts
```diff
@@ -1151,6 +1196,28 @@ export interface SyncFormulaDef<
    * @hidden
    */
   updateOptions?: Pick<CommonPackFormulaDef<ParamDefsT>, 'extraOAuthScopes'>;
+
+  /**
+   * The JavaScript function that implements fetching permissions for a set of objects
```
patrick-codaio, line 1201: seems redundant to say it's javascript?

api.ts
```diff
@@ -1108,6 +1110,49 @@ export interface SyncUpdateResultMarshaled<
...
+ */
+export interface ExecuteGetPermissionsRequest<
+  K extends string,
+  L extends string,
+  SchemaT extends ObjectSchemaDefinition<K, L>,
+> {
+  /**
+   * The list of rows for which to fetch permissions. These rows have been retrieved from the {@link execute} function
+   */
+  rows: Array<ObjectSchemaDefinitionType<K, L, SchemaT>>;
```
jonathan-codaio, line 1146: I think for executeUpdate the input is actually an array of wrapper objects, rather than a wrapper object containing an array, if we're trying to model the exact same thing.

runtime/thunk/thunk.ts
```diff
@@ -112,6 +116,18 @@ async function doFindAndExecutePackFunction<T extends FormulaSpecification>({
...
     }
+    case FormulaType.GetPermissions: {
+      const formula = findSyncFormula(manifest, formulaSpec.formulaName);
+      if (!formula.executeGetPermissions) {
+        throw new Error(`No executeGetPermissions function defined on sync table formula ${formulaSpec.formulaName}`);
+      }
+      const response = await formula.executeGetPermissions(
+        params,
+        ensureExists(getPermissionsRequest),
+        executionContext as SyncExecutionContext,
```
jonathan-codaio, line 127: I think the cast should be for GetPermissionExecutionContext?

schema.ts
```diff
@@ -1232,6 +1232,82 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
   // TODO(dweitzman): Only support options in the typing when the codaType is ValueHintType.SelectList.
 }
 
+/**
+ * Discriminent used to determine the permission type
```
jonathan-codaio, line 1236: Would probably just simplify to something like "The type of principal that can be applied to a permission."

schema.ts
```diff
@@ -1232,6 +1232,82 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+  Public = 'public',
+}
+
+/**
+ * This type definition represents properties required for all permissions.
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+interface BasePermission {
```
jonathan-codaio, line 1252: Was looking at how we model this in the public API, this object is more accurately a BasePrincipal, and the others are UserPrincipal, GroupPrincipal, etc. A _permission_ is a principal joined with an access level (can view, can edit, etc). We may not bother with access levels at this point while things are readonly, but we should use that more accurately vocabulary.

schema.ts
```diff
@@ -1232,6 +1232,82 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+
+/**
+ * This grants access to any user in the world.
+ *
+ * Generally this would apply to an entity where anyone with access to the url can view the item
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+interface PublicPermission extends BasePermission {
```
jonathan-codaio, line 1299: I see in our public API we call this an "Anyone" principal, not sure if we want to be consistent with that, @ekoleda-codaio any opinions?

api.ts
```diff
@@ -2249,6 +2322,37 @@ export function makeSyncTable<
...
+
+        const oversizedRowIds = Object.entries(permissionCountByRow)
+          .filter(([_rowId, count]) => count > MaxPermissionsPerRow)
+          .map(([rowId, _count]) => rowId);
+
+        if (oversizedRowIds.length > 0) {
+          throw new Error(
+            `Objects with ids: ${oversizedRowIds.join(
+              ', ',
+            )} returned more permissions than the maximum allowed of ${MaxPermissionsPerRow} per object`,
```
patrick-codaio, line 2349: FWIW I believe a pack maker using the CLI can route around the makeSyncTable function, so this limit will need to be enforced on the server too.

schema.ts
```diff
@@ -1232,6 +1232,82 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+
+/**
+ * This grants access to a specific group by its group ID.
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+interface GroupPermission extends BasePermission {
+  type: PrincipalType.Group;
+  groupId: string | number;
```
patrick-codaio, line 1277: depending on the result of the slack thread, we might want to bring back groupType

schema.ts
```diff
@@ -1232,6 +1232,82 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+
+/**
+ * This type definition represents properties required for all permissions.
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+interface BasePermission {
+  type: PrincipalType;
+  rowId: string | number;
```
patrick-codaio, line 1254: cc @ekoleda-codaio since this is also deviating from the pattern of executeUpdate's return object structure.

schema.ts
jonathan-codaio: Ah, I skimmed past the row id. I think we should model this in a way where permissions are self-contained objects. The executeGetPermissions return value can pair a permission with an object id. But I don't think that implies they should be part of the same object.

```ts
interface ReturnValueNeedingAGoodName {
  permissions: Permission[];
  rowId: string | number;
}

type Principal = UserPrincipal | GroupPrincipal | AnyonePrincipal;

interface Permission {
  principal: Principal;
  // Later perhaps
  // access: Access;  // CanView, CanEdit, CanComment, whatever
}
```

api.ts
sam-codaio: It is redundant but also consistent

api.ts
sam-codaio: Sure and each wrapper object contains a row in this scenario + whatever else we need to add here? 

Should we have another level of wrapping in case we want to pass in global stuff to this function or is the assumption that this will always come from the context?

schema.ts
sam-codaio: Updated this!

api.ts
sam-codaio: Was thinking maybe we have
```
export interface ExecuteGetPermissionRequestRow<
  K extends string,
  L extends string,
>{
 row: ObjectSchemaDefinitionType<K, L, SchemaT>
 otherRowSpecificStuff: {};
}


/**
 * Type definition for the data passed to the {@link executeGetPermissions} function of a sync table.
 *
 * TODO(sam): Unhide this
 * @hidden
 */
export interface ExecuteGetPermissionsRequest<
  K extends string,
  L extends string,
> {
  /**
   * The list of rows for which to fetch permissions. These rows have been retrieved from the {@link execute} function
   */
  rows: ExecuteGetPermissionRequestRow<K,L,SchemaT>[];

 otherGlobalStuff: {};
}


```

schema.ts
patrick-codaio: Yeah

patrick-codaio: Well, I mean there's a possibility

sam-codaio: yeah I am debating if we can make the assumption that the group id is unique enough. Obviously since the pack maker is in control of everything you _could_ always do something like `groupId: groupTypeOne-id` to represent this if we just had id

patrick-codaio: oh, that's interesting. yeah then leave groupType out for now.

schema.ts
sam-codaio: Happy to be consistent here with the naming

schema.ts
sam-codaio: This has changed a bit based on @jonathan-codaio 's comments but I thought this was a bit more explicit then just row ordering

api.ts
jonathan-codaio: (Mentioned on Slack as well) Yeah this double-wrapping is the most future-proof. The outer wrapping object is not consistent with executeUpdate, not sure if we care and why that wasn't done there. It's certainly verbose while we don't have any other fields, but it's certainly the most future-proof, so I'm supportive.

sam-codaio: Cool just pushed this

jonathan-codaio (approved):
Works for me. Can iterate as necesary.

PR #2933, opened 2024-03-29
Create GroupMembers table role

sam-codaio:
* Create GroupMembers concept in Packs SDK
* Requires 2 fields groupId and userId
* Sync Tables with this role will return a flattened list of all the members of a group
* If desired the developer can add a crawl strategy for this table that works on a 'groups' table and crawls members of each group individually. Alternatively this can just return every single group member from every single group in the system


schema.ts
```diff
@@ -1225,10 +1228,22 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * For cases where the object being synced represents a grou, the name of the property within
+   * {@link ObjectSchemaDefinition.properties} that identifies the id of the group in the service
+   * being synced from.
+   *
+   * This is required for sync tables with role {@link TableRole.GroupMembers}
+   *
+   * Must be a {@link ValueType.String} or {@link ValueType.Number} property
+   * @hidden
+   */
+  groupIdProperty?: PropertyIdentifier<K>;
```
sam-codaio, line 1246: Do we need a groupType if this system uses scim or google groups?

sam-codaio: Actually we would not need that here, that would need to go on the permission.

schema.ts
```diff
@@ -1225,10 +1228,22 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
    * {@link ObjectSchemaDefinition.properties} that identifies the id of the user in the service
    * being synced from.
    *
+   * This is required for sync tables with role {@link TableRole.User} or {@link TableRole.GroupMembers}
    * Must be a {@link ValueType.String} or {@link ValueType.Number} property
    * @hidden
    */
   userIdProperty?: PropertyIdentifier<K>;
+  /**
+   * For cases where the object being synced represents a grou, the name of the property within
```
patrick-codaio, line 1237: you are likely to be eaten by a grou

patrick-codaio (approved):

schema.ts
```diff
@@ -1225,10 +1228,22 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * {@link ObjectSchemaDefinition.properties} that identifies the id of the group in the service
+   * being synced from.
+   *
+   * This is required for sync tables with role {@link TableRole.GroupMembers}
+   *
+   * Must be a {@link ValueType.String} or {@link ValueType.Number} property
+   * @hidden
```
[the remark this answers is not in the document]
sam-codaio: Ahh forgot a TODO here. Will add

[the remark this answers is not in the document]
sam-codaio:
```suggestion
   * TODO(sam): Unhide this
   * @hidden
```

PR #2945, opened 2024-04-02
use the table name instead of getter name since we need to use the table name for policy enforcement

gary-codaio:


gary-codaio:
Abandoning to avoid having to rebundle packs

PR #2949, opened 2024-04-11
Create delegated permission capabilities

sam-codaio:
* Creates a new concept of a "Delegated Permission" in which the permission is actually a pointer to the permissions on a separate object.
* Adds permission type and creates a discriminated union to describe the permission type.
* This leaves the door open for more complex permission types in the future

### DIFF
--- a/schema.ts
+++ b/schema.ts
@@ -1303,15 +1303,51 @@ export interface AnyonePrincipal {
 type Principal = UserPrincipal | GroupPrincipal | AnyonePrincipal;
 
 /**
+ * The type of permission that is being granted
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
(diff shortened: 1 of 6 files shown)


schema.ts
```diff
@@ -1303,15 +1303,51 @@ export interface AnyonePrincipal {
...
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+export interface PrincipalPermission {
+  type: PermissionType.Principal;
   principal: Principal;
 }
 
+/**
+ * This represents a permission that delegates access to another entity
```
ekoleda-codaio, line 1334: Another entity of the same type, I assume? I don't see a way here to specify a foreign type.

sam-codaio: Yeah this would be of the same time as the parent.

PR #2950, opened 2024-04-11
Add inherit parent permissions to crawl

sam-codaio:
* Statically define dependency on parent permissions in the crawl hierarchy
* Validate that the property is the parent's id. If we want to later support multi level inheritance that skips a generation can remove this validation


testing/upload_validation.ts
```diff
@@ -274,6 +274,15 @@ export function validateCrawlHierarchy(
           return undefined;
         }
 
+        if (inheritPermissions && tableSchema.idProperty !== propertyKey) {
```
sam-codaio, line 277: If we want multi level id inheritance we would need to remove this

sam-codaio: Validation would be harder in that case as we would need to understand what it comes from

patrick-codaio (approved):

PR #2952, opened 2024-04-16
Add bodyTextProperty to packs sdk

sam-codaio:
* Add field to indicate that column contains unstructured content


schema.ts
```diff
@@ -1245,6 +1246,18 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * The name of the property within {@link ObjectSchemaDefinition.properties} that can be be interpreted as
+   * a large blob of unstructured content.
+   *
+   * This should consist of text or markdown content.
+   *
+   * Must be a {@link ValueType.String} property.
+   * TODO(sam): Unhide this
+   * @hidden
+   */
+  unstructuredContentProperty?: PropertyIdentifier<K>;
```
sam-codaio, line 1260: @patrick-codaio a big question here is generally this content will NOT work in a sync table (will always blow up the cell size limit). Is that an issue?

ekoleda-codaio, line 1260: I'd consider a different name. `unstructured` doesn't really say much, and feels like a very inside-baseball term. Some alternatives:
- `textContextProperty`
- `bodyTextProperty`
- `contentProperty`
- `bodyProperty`

schema.ts
```diff
@@ -1245,6 +1246,18 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
    * @hidden
    */
   groupIdProperty?: PropertyIdentifier<K>;
+
+  /**
+   * The name of the property within {@link ObjectSchemaDefinition.properties} that can be be interpreted as
+   * a large blob of unstructured content.
```
ekoleda-codaio, line 1252: I'm not sure it will be feasible to return a page of N large blobs of text within the time and memory limits of the Pack for all but very small values of N. For example. I wonder if like fetching permissions, this too should be a per-row function.

schema.ts
patrick-codaio: I suspect it will work ~2/3 of the time, so I vote for not doing anything about this now (other than maybe a build error that a pack maker should not make a field like this a featuredProperty?). We can write code for docs to ignore these columns if we decide to later.

schema.ts
```diff
@@ -1245,6 +1246,18 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   *
+   * This should consist of text or markdown content.
```
patrick-codaio, line 1254: not HTML?

schema.ts
sam-codaio: bodyProperty to me is very clear

schema.ts
sam-codaio: In theory could be HTML. Depends how well our models will support this format

testing/upload_validation.ts
```diff
@@ -1449,6 +1450,14 @@ function buildMetadataSchema({sdkVersion}: BuildMetadataSchemaArgs): {
           );
         };
 
+        const validateUnstructuredContentProperty = () => {
```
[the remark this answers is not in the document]
sam-codaio: We should convert this to markdown before returning

schema.ts
patrick-codaio: I vote `bodyTextProperty` (and `bodyProperty` is a close 2nd).

schema.ts
sam-codaio: @patrick-codaio Any thoughts. I know we discussed this as well. I do like that this is much much less work on our part but I see @ekoleda-codaio's points

schema.ts
```diff
@@ -1245,6 +1246,18 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
    * @hidden
    */
   groupIdProperty?: PropertyIdentifier<K>;
+
+  /**
+   * The name of the property within {@link ObjectSchemaDefinition.properties} that can be be interpreted as
+   * a large blob representing the body of this entity.
```
jonathan-codaio, line 1252: Nit, blob technically stands for "binary large object" but here we mean text and not binary. Would update this to be clear that this is a large text property, of rich or plain text.

schema.ts
jonathan-codaio: String properties already support optional codaType hints of Markdown or Html. So I think we should just look to those hint types to determine how to interpret the text (and if no hint type is provided it is plain text).

jonathan-codaio (commented):
To Eric's point this is not highly scalable and requires the pack to use very small page sizes, possible just pages of 1 item, to stay within limits, but this seemed like the most expedient starting point.

That said, I hadn't considered the blended approach where you still use a sync table which just returns rows of _metadata_, and a new execute function that takes a response row and gets the content. I wonder if that winds up being a better option--the metadata rows can still wind up in the items table but the thing that calls the new execute function that returns body content then has a chance to chunk and write elsewhere.

schema.ts
patrick-codaio: Definitely worth a discussion. The way I see things, there's not a clear boundary between something like a Jira issue description and a google doc. They're both big rich text blobs that we want Brain to understand similarly, and if they're not too big, it would be nice to have them in a doc. So I feel it's still appropriate to let them be "normal" fields.

sam-codaio:
> To Eric's point this is not highly scalable and requires the pack to use very small page sizes, possible just pages of 1 item, to stay within limits, but this seemed like the most expedient starting point.
> 
> That said, I hadn't considered the blended approach where you still use a sync table which just returns rows of _metadata_, and a new execute function that takes a response row and gets the content. I wonder if that winds up being a better option--the metadata rows can still wind up in the items table but the thing that calls the new execute function that returns body content then has a chance to chunk and write elsewhere.

Yeah I had initially considered the blended approach (ALA getPermissions) but decided on this for expediency. I am open to switching to the blended approach as long as we can make a decision soon. Happy to set something up today to make this choice

schema.ts
patrick-codaio: oh, and also think of a future of snowflake-backed sync tables in docs, this data would be more useful then.

schema.ts
sam-codaio: Discussed f2f. We will begin with this approach and add on the function approach as needed. They are not incompatible

patrick-codaio (approved):

PR #2956, opened 2024-04-18
Fix incorrect relative links in the documentation

ekoleda-codaio:
MkDocs 1.5 (recently upgraded) was logging a warning if it found a relative link that should have been to a file.

```
WARNING -  Doc file 'samples/topic/fetcher.md' contains a relative link '../../../guides/basics/fetcher.md', but the target '../guides/basics/fetcher.md' is not found among
           documentation files. Did you mean '../../guides/basics/fetcher.md'?
```

This PR resolves all of those errors by:
- Manually fixing some links.
- Changing how URLs are generated on the Samples and Tutorials listing page, using a helper function in Python.
- Changing how the URLs are generated on the individual samples pages, using a helper function in TypeScript.
- Move the `.cli` snippet files so the relative links line up correctly.

It also changes the log level of the warning so that it will fail a build in the future.


jonathan-codaio (approved):
Thank you!

PR #2962, opened 2024-05-20
Add Domain Principal

sam-codaio:
* Add the domain principal type which will be required in certain packs-sync
* Proposal on adding a type to the user/groups principal types. This will allow us to do special handling for specific group and user types such as avoiding a user mapping table when the user type is email.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.8",
+  "version": "1.7.7-prerelease.1",
   "license": "MIT",
   "workspaces": [
     "dev/eslint"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.7",
+  "version": "1.7.8",
   "license": "MIT",
   "workspaces": [
     "dev/eslint"
--- a/schema.ts
+++ b/schema.ts
@@ -1294,6 +1294,17 @@ export interface GroupPrincipal {
 }
 
 /**
+ * This represents a principal that is anyone in that domain
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+export interface DomainPrincipal {
+  type: PrincipalType.Group;
+  domain: string;
+}
+
+/**
  * This represents a principal corresponding to anyone
  *
  * Generally this would apply to an entity where anyone with access to the url can view the item
@@ -1311,7 +1322,7 @@ export interface AnyonePrincipal {
  * TODO(sam): Unhide this
  * @hidden
  */
-type Principal = UserPrincipal | GroupPrincipal | AnyonePrincipal;
+type Principal = UserPrincipal | GroupPrincipal | AnyonePrincipal | DomainPrincipal;
 
 /**
  * This represents the definition of a permission in the external system.
--- a/schema.ts
+++ b/schema.ts
@@ -1269,6 +1269,7 @@ export enum PrincipalType {
   User = 'user',
   Group = 'group',
   Anyone = 'anyone',
+  Domain = 'domain',
 }
 
 /**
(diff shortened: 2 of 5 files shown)


schema.ts
```diff
@@ -1280,6 +1292,18 @@ export enum PrincipalType {
...
+}
+
+/**
+ * The type of group ID that is being used in a {@link GroupPrincipal}.
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+export enum GroupIdType {
+  GoogleGroup = 'googleGroup',
```
ekoleda-codaio, line 1305: A Google Group doesn't feel like a type of ID, but rather a type of group itself. A Google Group in fact has multiple IDs, both a numeric ID and an associated email address.

schema.ts
```diff
@@ -1280,6 +1292,18 @@ export enum PrincipalType {
 export interface UserPrincipal {
   type: PrincipalType.User;
   userId: string | number;
+  userIdType?: UserIdType;
```
neal-codaio, line 1295: If the userId type is anyone or domain, is userId filled in?

sam-codaio: No that is a different type that does not require userId

neal-codaio: Oh whoops.  I was very confused.  I didn't realize I conflated the UserIdType w/PrincipalType.  You should document the default (presumably id?) if this is undefined.  Or perhaps better, ensure it is defined?

sam-codaio: With the type I was not sure the best answer (for backwards compatibility w/ packs that already provide permissions. We can require it since those packs are all internal)

schema.ts
```diff
@@ -1293,6 +1293,17 @@ export interface GroupPrincipal {
...
 }
 
+/**
+ * This represents a principal that is anyone in that domain
+ *
+ * TODO(sam): Unhide this
+ * @hidden
+ */
+export interface DomainPrincipal {
+  type: PrincipalType.Group;
```
neal-codaio, line 1303: What are the implications of reusing Group?  Is/should domain be a new PrincipalType?

sam-codaio: A group has a list of members which is not how Domain works. Domain is saying if your email address matches this string. To me it is fundamentally different bc of who you are granting access to

neal-codaio: I think you are answering about moving the domain into GroupPrincipal?  I'm thinking the other way, why isn't there a PrincipalType.Domain?

sam-codaio: Oh that is an error

neal-codaio: :-) ok, LGTM now.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.7",
+  "version": "1.7.8",
```
ekoleda-codaio, line 3: I don't think you should be bumping the SDK version, or if you do it should be to a pre-release version.

sam-codaio: @patrick-codaio Are we able to use packs on head w/ prerelease versions?

neal-codaio: I'm fairly certain the answer is yes, we've done that before.  The issue is with using non-released versions.  And the only issue that I know of with that is that lint starts taking ~3x as long (~an hour).

sam-codaio: I thought we made a change where we now use the Live version of packs vs Latest which I thought had to be on a real release?

patrick-codaio: There's a flag that controls whether Coda Brain uses Live or Latest pack versions, and it now defaults to Live everywhere (also you can't Release a pack version built with a pre-release SDK version). 

Are you sure you need this running on head now, and not just your dev? 

Releasing a new version isn't terrible, though it is a bit of a chore - https://staging.coda.io/d/Packs-Platform-IA-go-packs_d36WK4zgrYx/Releasing-Packs-SDK-Versions-o-release-packs-sdk_suerz#_luCH6

sam-codaio: I can start w/ dev for now and then move to head next week

sam-codaio, line 3:
```suggestion
  "version": "1.7.7-prerelease.1",
```

ekoleda-codaio (approved):

PR #2974, opened 2024-06-25
Release new version to pick up popularity property change

chris-codaio:
Trying to address https://github.com/coda/packs/pull/4172#discussion_r1653466443

PTAL @neal-codaio @sam-codaio

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.7-prerelease.3",
+  "version": "1.7.8",
   "license": "MIT",
   "workspaces": [
     "dev/eslint"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.7-prerelease.2",
+  "version": "1.7.7-prerelease.3",
```
sam-codaio, line 3: I think we should just release. Otherwise you can not release new pack versions which I am sure we will want to do

neal-codaio (approved):

package.json
chris-codaio: Yeah, just saw Jonathan's comment come through.  Can go to 1.7.8.

chris-codaio: @sam-codaio - updated to new version.

sam-codaio (approved):

PR #3016, opened 2024-07-02
add entity id property

gary-codaio:
Goal of this is that for some sync tables, it may make sense for some rows to be grouped into entity ids, this property should capture that.


schema.ts
```diff
@@ -1247,6 +1247,12 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
    * @hidden
    */
   groupIdProperty?: PropertyIdentifier<K>;
+  /**
+   * The name of a property within {@link ObjectSchemaDefinition.properties} that represents a unique id for a
+   * parent entity for the object. It is recommended for sync table schemas with a bodyTextProperty to specify an
+   * an entityProperty, which uniquely identifies the entity that groups 1 to multiple rows.
```
patrick-codaio, line 1253: Could you include an example, like a document being broken up into chunks?

patrick-codaio (approved):
LGTM, but can @ekoleda-codaio also approve the name? Slack context: https://codaproject.slack.com/archives/C06HG23CV08/p1718915995227269?thread_ts=1718906405.616409&cid=C06HG23CV08

patrick-codaio:
oh, also can you bump the prerelease version in package.json

ekoleda-codaio:
Sorry for the delay, I was OOO most of last week. I think `parentIdProperty` would make more sense. In the description you already write "represents a unique id for a parent entity for the object". The parent of a DocChunk would be a Doc, for example. I think `idProperty` and `entityIdProperty` would be very hard to distinguish in auto-complete.

gary-codaio:
@ekoleda-codaio thanks, good point. Renamed

schema.ts
```diff
@@ -1247,6 +1247,14 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
    * @hidden
    */
   groupIdProperty?: PropertyIdentifier<K>;
+  /**
+   * The name of a property within {@link ObjectSchemaDefinition.properties} that represents a unique id for a
+   * parent entity for the object. It is recommended for sync table schemas with a bodyTextProperty to specify an
```
ekoleda-codaio, line 1252: In theory, could you specify a `bodyTextProperty` without chunking it into multiple rows? If, perhaps clarify that here?

schema.ts
```diff
@@ -1247,6 +1247,14 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * an entityProperty, which uniquely identifies the entity that groups 1 to multiple rows. For example, a sync table
+   * where each row is a partial chunk of a larger document may want to specify a parent id that represents the
+   * document, while each row's id can be unique to the chunk.
```
ekoleda-codaio, line 1255: Should probably mark with `@hidden`, to match the other new fields.

ekoleda-codaio (changes requested):

ekoleda-codaio (approved):

PR #3020, opened 2024-07-16
Add ResponseTooLarge Error

sam-codaio:
* add response too large error
* bump version to pre-release

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -248,6 +248,32 @@ export class MissingScopesError extends Error {
 }
 
 /**
+ * An error that will be thrown by {@link Fetcher.fetch} when the response body from the external system
+ * exceeds packs platform limits
+ *
+ * This error can be caught and retried by requesting less data from the external system through
+ * a smaller page size or omitting large fields.
+ *
+ * @hidden
+ */
+export class ResponseSizeTooLargeError extends Error {
+  /**
+   * The name of the error, for identification purposes.
+   */
+  override name: string = 'ResponseSizeTooLargeError';
+
+  /** @hidden */
+  constructor(message?: string) {
+    super(message || 'Response size too large');
+  }
+
+  /** Returns if the error is an instance of ResponseSizeTooLargeError. Note that `instanceof` may not work. */
+  static isResponseSizeTooLargeError(err: any): err is ResponseSizeTooLargeError {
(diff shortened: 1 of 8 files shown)


api.ts
```diff
@@ -247,6 +247,32 @@ export class MissingScopesError extends Error {
...
+/**
+ * An error that will be thrown by {@link Fetcher.fetch} when the response body from the external system
+ * exceeds packs platform limits
+ *
+ * This error can be caught and retried by requesting less data from the external system through
+ * a smaller page size or omitting large fields.
+ *
+ * @hidden
+ */
+export class ResponseSizeTooLargeError extends Error {
```
dweitzman-codaio, line 259: Do we want to try to include any information here that would help the pack code understand the size of the response either in absolute terms or relative to the maximum allowed?

I don't think we can do that 100% of the time (like maybe if a server returns data with chunked response encoding we could see that the data is too large, but we don't necessarily know how large it would have been) so I'm fine with making this error vague for now and leaving the option to add more detail in the future if we think it's useful

sam-codaio: I could certainly put an estimate of the response size + the limit that was violated. Not sure if they bytes limit is that helpful but could be worth a shot

dweitzman-codaio: This isn't a blocking comment, so if you have a plan to use this that doesn't depend on getting numbers back in the response I'm fine with that as long we there's no API structure objections from Patrick or Eric

(One thing I thought about re: API was whether this could be expressed as a StatusCodeError, but HTTP only seems to have an official status code for request to big, but not standard error code for response would be too big)

ekoleda-codaio (commented):
API and docs LGTM!

api.ts
sam-codaio: I also do not want to confuse StatusCodeErrors and errors that are imposed by Coda. I think that could be potentially confusing to the consumer

dweitzman-codaio (approved):

api.ts
patrick-codaio, line 259: RequestTooLarge would also be good to add at some point, but doesn't need to block this.

patrick-codaio (approved):

api.ts
sam-codaio: Will add as needed. Starting with this. Once this is going in a real pack can add the other error as well will be very similar

PR #3021, opened 2024-07-17
Explicitly open/close files to resolve EMFILE

sam-codaio:
* Attempt to resolve EMFILE errors by explicitly closing files instead of relying on readFileSync

### DIFF
--- a/runtime/bootstrap/index.ts
+++ b/runtime/bootstrap/index.ts
@@ -296,7 +296,11 @@ export async function registerBundle(
   await context.global.set('module', {}, {copy: true});
 
   // compiling the bundle allows IVM to map the stack trace.
-  const bundle = fs.readFileSync(path).toString();
+
+  // readFileSync does not properly close files in the lambda
+  const fileDescriptor = fs.openSync(path, 'r');
+  const bundle = fs.readFileSync(fileDescriptor).toString();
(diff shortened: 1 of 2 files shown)


gary-codaio (approved):

runtime/bootstrap/index.ts
```diff
@@ -296,7 +296,11 @@ export async function registerBundle(
   await context.global.set('module', {}, {copy: true});
 
   // compiling the bundle allows IVM to map the stack trace.
-  const bundle = fs.readFileSync(path).toString();
+
+  // readFileSync does not properly close files in the lambda
+  const fileDescriptor = fs.openSync(path, 'r');
+  const bundle = fs.readFileSync(fileDescriptor).toString();
```
dweitzman-codaio, line 302: Seems like the readFileSync implementation already calls closeSync: https://github.com/nodejs/node/blob/589b67aa1748974b24df4b4b7281e0cce4e915c7/lib/fs.js#L481

johnli-codaio: @sam-codaio mentioned that there are reports of it not closing properly in lambda. This explicit call will close it is that right?

dweitzman-codaio: If there a stack overflow or something about a lambda-specific bug the code can link to?

PR #3023, opened 2024-07-25
Remove default value for supportsIncrementalSync

ebo-codaio:
Fixing for triggering version error due to default values

thread: https://codaproject.slack.com/archives/C02ERJRBA21/p1721857075068479

PTAL @chris-codaio @coda/packs

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.9-prerelease.1",
+  "version": "1.7.9-prerelease.2",
   "license": "MIT",
   "workspaces": [
     "dev/eslint"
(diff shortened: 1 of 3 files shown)


chris-codaio (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.9-prerelease.1",
+  "version": "1.7.9-prerelease.2",
```
ebo-codaio, line 3: actually not sure whether we need to increase the version here if only validation is changed

chris-codaio: I think we do since you'll need to re-pull this in the Coda repo.

PR #3031, opened 2024-08-02
Move sinon dep to devDependencies

dweitzman-codaio:
Noticed this while looking at https://github.com/coda/packs-sdk/pull/3009


jonathan-codaio (commented):
Actually I think that's intentional, the `mocks` module ships with the SDK to provide batteries-included testing utilities, and the fetcher is a sinon stub.

PR #3032, opened 2024-08-13
Initial SWAG of indexing hints

alexd-codaio:
@jonathan-codaio @neal-codaio @chris-codaio PTAL

Colocate all indexing related properties into a single object.

While I was here, fixed some busted lock file that was busted in https://github.com/coda/packs-sdk/pull/3009.


schema.ts
```diff
@@ -1064,6 +1064,68 @@ export type ObjectSchemaPathProperties = Pick<
...
+  */
+export interface IndexDefinition {
+  /**
+   * A list of properties from within {@link ObjectSchemaDefinition.properties} that should be indexed.
+   */
+  properties: IndexedProperty[];
+  
+  /*
+   * The context properties to be used for indexing.
+   * If unspecified, defaults to the default context properties for the schema.
```
jonathan-codaio, line 1115: What are the default context properties for the schema?

testing/upload_validation.ts
```diff
@@ -1152,6 +1153,90 @@ function buildMetadataSchema({sdkVersion}: BuildMetadataSchemaArgs): {
     }),
   ]);
 
+  const contextPropertiesSchema = z.array(propertySchema).min(1);
+
+  const indexedPropertySchema = z.union([
+    propertySchema,
+    z.object({
```
jonathan-codaio, line 1160: Nit, for every object, you probably want to use our wrapper `zodCompleteStrictObject<...>` which will ensure if you add new properties to the type def that TS will catch that you need to update the validator.

schema.ts
```diff
@@ -1064,6 +1064,68 @@ export type ObjectSchemaPathProperties = Pick<
   | 'popularityRankProperty'
 >;
 
+/**
+ * Specifies how this property should be indexed.
+ * @hidden
+ */
+export enum IndexingStrategy {
```
jonathan-codaio, line 1071: Probably want to export this in index.ts so you can use it in your pack defs.

jonathan-codaio (approved):
Will want to bump to a prerelease version in package.json if you immediately want to start using this in `coda`.

schema.ts
alexd-codaio: Changed this comment.

testing/upload_validation.ts
alexd-codaio: Done.

schema.ts
alexd-codaio: Done.

neal-codaio (commented):
@alexd-codaio  Did you mean to add runtime/native/node18/x86_64/isolated-vm/out/isolated_vm.node?

PR #3034, opened 2024-08-14
add plus and minus 90 days date range

gary-codaio:
Per request from Adil, would like this as a requested date range to use for google calendar


jonathan-codaio (commented):
Would suggesting teeing up a change on the `coda` side first to add this to `date_ranges.ts` and associated files first. Once it appears in the SDK it needs a place to be mapped in the UX in order to compile. So will want to have that ready to go otherwise any other SDK changes coming in behind this will break in `coda` in the meantime.

gary-codaio:
@jonathan-codaio I have https://github.com/coda/coda/pull/115382 ready but needs the packs-sdk version to be bumped for ts to pass

jonathan-codaio (approved):
Cool, can you bump the prerelease version in `package.json` so we don't make `coda` `lint` slow when we bump things over there?

PR #3035, opened 2024-08-21
Fix role not passed through in PackDefinitionBuilder#addSyncTable

oleg-codaio:
`role` was missing here - fixing up to make sure we pass through all parameters going forward.

PTAL @coda/packs


jonathan-codaio (approved):
Good catch! This about to bite me...

PR #3038, opened 2024-08-23
Fix SchemaType when using fromKey

oleg-codaio:
Currently when `SchemaType` detects a schema using at least one `fromKey` attribute on a property, it erases those
properties and joins the whole schema with a record type. Now that later versions of TS make this easier to handle,
improving the type helper to add rich types instead.

PTAL @coda/packs


jonathan-codaio (commented):
Awesome! Huayang tried so hard to get this to work back in the day.

The TS magic is basically impossible to grok, so I'm happy to stamp this, but as a quick check, would you mind putting this commit hash into a branch in the `packs` repo and seeing if TS still compiles?

oleg-codaio:
Yup! This caught a number of issues, see https://github.com/coda/packs/pull/4277.

jonathan-codaio (approved):

PR #3039, opened 2024-08-26
Sketch A: Allow sync formula to return ACLs

patrick-codaio:


chris-codaio:
I suppose this boils down to a special short-cut case of something like the ServiceNow pack where the schema for `KnowledgeArticle` contains the ACLs inline.  Today, we just take in the row and spit out the permissions through a synchronous transform - no APIs involved.  Not the end of the world if we have to implement the getPermissions handler for this case, though, like we do already.  I think I agree with you and others that your Sketch B proposal has more utility.

PR #3040, opened 2024-08-26
Sketch B: Incremental ACL crawl as a separate function

patrick-codaio:


api.ts
```diff
@@ -1176,9 +1193,41 @@ const MaxPermissionsPerRow = 1000;
...
+   */
+  continuation?: Continuation;
+
+  /**
+   * This value will be provided to {@link executeGetIncrementalPermissions} the next time Coda crawls permissions.
+   *
+   * Coda assumes that if you return this, your {@link executeGetIncrementalPermissions} can be called on a
+   * different (potentially more frequent) cadence than incremental data syncs.
+   */
+  incrementalPermissionsContinuation?: Continuation;
```
chris-codaio, line 1230: We call this `Completion` in the normal execute path I think?  Would be good to be consistent on naming.

api_types.ts
```diff
@@ -1013,6 +1004,12 @@ export enum TableRole {
   GroupMembers = 'groupMembers',
 }
 
+export enum IncrementalPermissionCrawlType {
+  RecrawlAllUpdatedItems = 'recrawl_all_updated_items',
+  RecrawlOnlyNewItems = 'recrawl_only_new_items',
```
chris-codaio, line 1009: If we're only visiting new items, are we "recrawling" them or just "crawling" them?  This reads a bit confusing to me.

api_types.ts
```diff
@@ -1013,6 +1004,12 @@ export enum TableRole {
...
+  OnlyUseIncrementalPermissions = 'only_use_incremental_permissions',
```
chris-codaio, line 1010: Aren't all incremental permission crawls "only using incremental permissions"?  I'm confused  If I understand the above comment correctly, should this be named something closer to "OnlyUseFullCrawling"?

api.ts
patrick-codaio: `completion` is an instance of `SyncCompletionMetadata` which has both `incrementalContinuation` and `incrementalPermissionsContinuation` inside of it. I worried that would be confusing to figure out which one should be populated in this instance. cc @ekoleda-codaio

api_types.ts
patrick-codaio: This enum is to control whether the non-incremental `executeGetPermissions` is invoked on sync results as they trickle into the ingestion processing (i.e., before we see the `completion` of the sync). I intend this particular option to indicate "never call executeGetPermissions, even if I'm returning a new row that doesn't yet exist in snowflake". So that's the opposite of "OnlyUseFullCrawling", so clearly I need to improve the naming here :)

PR #3041, opened 2024-08-26
Update SDK to support sibling acls

drew-codaio:
Part of the work for sibling ACLs. Updating permissions to have a new delegated type.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.12-prelease.1",
+  "version": "1.7.13-prerelease.1",
   "license": "MIT",
   "workspaces": [
     "dev/eslint"
--- a/schema.ts
+++ b/schema.ts
@@ -1424,7 +1424,9 @@ type Principal = UserPrincipal | GroupPrincipal | AnyonePrincipal | DomainPrinci
 
 /**
  * The type of permission.
- *
+ * Delegated permissions are permissions that are inherited from another object that aren't
+ * within the same sync table.
+ * Direct permissions are permissions that are directly applied to the object. E.g. a user or group, etc
  * TODO(drew): Unhide this
  * @hidden
  */
@@ -1440,7 +1442,7 @@ export enum PermissionType {
  * @hidden
  */
 export interface DirectPermission {
-  permissionType: PermissionType.Direct;
+  permissionType?: PermissionType.Direct;
   principal: Principal;
 }
 
--- a/schema.ts
+++ b/schema.ts
@@ -1461,7 +1461,7 @@ export interface DelegatedPermission {
  * TODO(drew): Unhide this
  * @hidden
  */
-type Permission = DirectPermission | DelegatedPermission
+export type Permission = DirectPermission | DelegatedPermission
 
 /**
  * This represents the list of permissions on a sync table row.
--- a/schema.ts
+++ b/schema.ts
@@ -1423,14 +1423,45 @@ export interface AnyonePrincipal {
 type Principal = UserPrincipal | GroupPrincipal | AnyonePrincipal | DomainPrincipal;
 
 /**
- * This represents the definition of a permission in the external system.
+ * The type of permission.
+ *
+ * TODO(drew): Unhide this
+ * @hidden
+ */
+export enum PermissionType {
+  Delegated = 'delegated',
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.7.11",
+  "version": "1.7.12",
```
patrick-codaio, line 3: since you'll be releasing this immediately, this is fine, but for future reference, the ideal flow is to make this `1.7.12-prerelease.1` and then as a part of doing the release process, it gets to drop the prerelease.

patrick-codaio:
cc @ekoleda-codaio

package.json
drew-codaio: I'll update this

patrick-codaio:
Another way I thought about doing this is that we could let a sync table schema have a `permissionDelegationProperty` which would need to be a reference to another table.

But I think that could live side-by-side with this, so no objection from me.

jonathan-codaio (commented):
Yeah we can't release this right away I think? Need to update `coda` to add support to recognize support for this and support backwards compatibility, then need to have a PR read to fast merge to update every pack that uses permissions? Unless we change the SDK to be backwards compatible.

schema.ts
```diff
@@ -1424,15 +1424,46 @@ export interface AnyonePrincipal {
...
+
+/**
+ * This represents the definition of a direct permission in the external system.
  *
  * TODO(sam): Unhide this
  * @hidden
  */
-export interface Permission {
+export interface DirectPermission {
+  permissionType: PermissionType.Direct;
```
sam-codaio, line 1444: Can we make this for now permissionType?: PermissionType.Direct?

sam-codaio:
> Yeah we can't release this right away I think? Need to update `coda` to add support to recognize support for this and support backwards compatibility, then need to have a PR read to fast merge to update every pack that uses permissions? Unless we change the SDK to be backwards compatible.

Why not just make the SDK backwards compatible for now (Direct permissions can have no type there and we assume direct?)

drew-codaio:
> > Yeah we can't release this right away I think? Need to update `coda` to add support to recognize support for this and support backwards compatibility, then need to have a PR read to fast merge to update every pack that uses permissions? Unless we change the SDK to be backwards compatible.
> 
> Why not just make the SDK backwards compatible for now (Direct permissions can have no type there and we assume direct?)

good point.

schema.ts
```diff
@@ -1423,15 +1423,46 @@ export interface AnyonePrincipal {
...
 
 /**
- * This represents the definition of a permission in the external system.
+ * The type of permission.
+ *
+ * TODO(drew): Unhide this
+ * @hidden
+ */
+export enum PermissionType {
+  Delegated = 'delegated',
```
ekoleda-codaio, line 1432: Can you include some basic JSDoc describing what the delegated and direct mean, when to use them, etc?

schema.ts
```diff
@@ -1423,15 +1423,48 @@ export interface AnyonePrincipal {
...
 
+/**
+ * This represents the definition of a delegated permission in the external system.
+ * 
+ * TODO(drew): Unhide this
+ * @hidden
+ */
+export interface DelegatedPermission {
+  permissionType: PermissionType.Delegated;
+  delegatedItemId: string | number;
```
sam-codaio, line 1457: Can we add the sync table name here we are referencing as well?

sam-codaio: Would be good to have and would allow more efficient joins in the future potentially.

sam-codaio (approved):

jonathan-codaio:
Make sure to fast follow with `coda` or `packs` changes to incorporate these, the SDK is evolving fast and if changes break TS somebody else's SDK change could be stuck behind getting those fixed. I'll share this with the team as well.

drew-codaio:
> Make sure to fast follow with `coda` or `packs` changes to incorporate these, the SDK is evolving fast and if changes break TS somebody else's SDK change could be stuck behind getting those fixed. I'll share this with the team as well.

My bad still trying to figure out how this whole ecosystem works with the sdk. Will be mindful going forward. Sam and I are working on the fast follow rn.

jonathan-codaio:
No need to apologize, there's lots of unwritten best practices to navigate, just spreading awareness.

PR #3043, opened 2024-08-27
Fix schema validation to allow for arrays of strings when indexing properties

alexd-codaio:
@chris-codaio @sam-codaio @jonathan-codaio PTAL

Missed this in the initial PR.


chris-codaio (approved):
Thanks!

PR #3045, opened 2024-08-27
Define a new ItemReference interface to use with DelegatedPermission.

jonathan-codaio:
PTAL @patrick-codaio @sam-codaio @drew-codaio @coda/packs 
CC @ekoleda-codaio 

Eric we had a little meeting to discuss how to shape this, and how this overlaps with the concept of Identities, which I tried to capture in the comments.

### DIFF
--- a/dist/bundle.d.ts
+++ b/dist/bundle.d.ts
@@ -2118,8 +2118,28 @@ export interface DirectPermission {
  */
 export interface DelegatedPermission {
 	permissionType: PermissionType.Delegated;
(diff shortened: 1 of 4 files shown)


sam-codaio (approved):

drew-codaio (approved):

ekoleda-codaio:
> Eric we had a little meeting to discuss how to shape this, and how this overlaps with the concept of Identities, which I tried to capture in the comments.

Ya, that surface makes sense given where we are.

PR #3049, opened 2024-08-30
Allow relative auth/token URLs for OAuth 2.0 auth

oleg-codaio:
See https://github.com/coda/coda/pull/116158. Enabling relative URLs here so that we can support custom endpoints and automatically prepend the effective endpoint.


testing/upload_validation.ts
```diff
@@ -580,8 +580,10 @@ function buildMetadataSchema({sdkVersion}: BuildMetadataSchemaArgs): {
     }),
     [AuthenticationType.OAuth2]: zodCompleteStrictObject<OAuth2Authentication>({
       type: zodDiscriminant(AuthenticationType.OAuth2),
-      authorizationUrl: z.string().url(),
-      tokenUrl: z.string().url(),
+      /** Accepts relative URLs when requiresEndpointUrl is true. */
+      authorizationUrl: z.string(),
+      /** Accepts relative URLs when requiresEndpointUrl is true. */
+      tokenUrl: z.string(),
```
jonathan-codaio, line 586: Can you do a refine/superRefine so that when requiresEndpoint is false these values are validated as full urls, and when true, validated as relative urls?

ekoleda-codaio:
While I think the ideal SDK surface is just to honor `requiresEndpointUrl`, we may first need to verify that there aren't any existing Packs that already use that with OAuth2 accidentally and don't want to enable this new behavior.

jonathan-codaio:
> While I think the ideal SDK surface is just to honor `requiresEndpointUrl`, we may first need to verify that there aren't any existing Packs that already use that with OAuth2 accidentally and don't want to enable this new behavior.

Did a quick look through our 1PPs. Gong, Servicenow, Snowflake have requiresEndpointUrl with OAuth2, all needing enterprise configs to work, I believe. Maybe those are fine as is because they would never have global oauth creds and so would fail fast, worth verifying.

Salesforce is an interesting one, it uses `endpointKey`. So for that case, we'd want to make sure to disable the endpoint url input since the endpoint will be provided in the token response. And then we would probably want to disallow relative urls in that case, you need an absolute url to get through the flow that ultimately gives you the endpoint.

Would take a one-off job to analyze 3PPs. But with the above fix I'd be comfortable proceeding with the allowlist, if we ever eliminated the allowlist we could do the analysis at that time.

testing/upload_validation.ts
oleg-codaio: Done, added some basic validation.

tsconfig.json
```diff
@@ -68,5 +68,8 @@
     "./scripts/*",
     "./documentation/samples",
     "./documentation/scripts"
-  ]
+  ],
+  "ts-node": {
```
oleg-codaio, line 72: I wasn't able to run `mocha test/upload_validation_test.ts` without this locally.

testing/upload_validation.ts
```diff
@@ -594,6 +596,28 @@ function buildMetadataSchema({sdkVersion}: BuildMetadataSchemaArgs): {
...
+        context.addIssue({
+          code: z.ZodIssueCode.custom,
+          path: [property],
+          message: `${property} must be ${expectedType} URL when requiresEndpointUrl is \
+${requiresEndpointUrl ?? 'not true'}`,
+        });
+      };
+
+      if (
+        (requiresEndpointUrl && !isRelativeUrl(authorizationUrl)) ||
```
jonathan-codaio, line 613: Per the recent comment about Salesforce, if `endpointKey` is provided I think we still need to enforce an absolute url. The combination of `requiresEndpointUrl` and `endpointKey` means "use this single absolute url for the oauth flow; the token exchange response will contain the endpoint keyed by this given property".

oleg-codaio: Updated, let me know if that looks good.

tsconfig.json
jonathan-codaio: Ah ok, yeah we just use `make test-file` normally.

jonathan-codaio (approved):

PR #3051, opened 2024-08-30
Add ability to trigger a full sync

sam-codaio:
* Add item in execute that allows for immediate re-triggering of a full sync

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+### Added
+
+- Internal change to execute response
+
 ## [1.7.13] - 2024-08-29
 
 ### Changed
@@ -748,5 +752,4 @@ await myHelper(context);
 [1.7.12]: https://github.com/coda/packs-sdk/compare/v1.7.8...v1.7.12
 [1.7.11]: https://github.com/coda/packs-sdk/compare/v1.7.8...v1.7.11
 [1.7.10]: https://github.com/coda/packs-sdk/compare/v1.7.8...v1.7.10
-
 [1.7.13]: https://github.com/coda/packs-sdk/compare/v1.7.8...v1.7.13
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ### Added
 
-- Internal change to execute response
+- Internal change to completion
 
 ## [1.7.13] - 2024-08-29
 
--- a/api.ts
+++ b/api.ts
@@ -1035,11 +1035,21 @@ export interface SyncFormulaResult<K extends string, L extends string, SchemaT e
 
   /**
    * Return the list of deleted item ids for incremental sync deletion.
-   * 
+   *
    * TODO(ebo): Unhide this
    * @hidden
    */
   deletedItemIds?: string[];
+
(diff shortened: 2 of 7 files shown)


api.ts
```diff
@@ -1035,11 +1035,21 @@ export interface SyncFormulaResult<K extends string, L extends string, SchemaT e
...
+
+  /**
+   * Returned by an incremental sync if a full sync is needed to recalculate the items.
+   *
+   * This will trigger a full sync so it should be avoided when possible
+   *
+   * TODO(sam): Unhide this
+   * @hidden
+   */
+  triggerFullSync?: boolean;
```
sam-codaio, line 1052: Could also call this something like "has incomplete results" if we want to stay away from this language

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+### Added
+
+- Internal change to execute response
```
patrick-codaio, line 9: let's not put things here yet

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+### Added
```
jonathan-codaio, line 7: Thanks :)

api.ts
```diff
@@ -1035,11 +1035,21 @@ export interface SyncFormulaResult<K extends string, L extends string, SchemaT e
...
    * Return the list of deleted item ids for incremental sync deletion.
-   * 
+   *
    * TODO(ebo): Unhide this
    * @hidden
    */
   deletedItemIds?: string[];
+
+  /**
+   * Returned by an incremental sync if a full sync is needed to recalculate the items.
```
jonathan-codaio, line 1045: So also this will be ignored if returned by a full sync, right?

patrick-codaio, line 1045: If a full sync returns it, what will happen?

api.ts
jonathan-codaio: Hmm, yeah generally we use declarative framings rather than imperative framings. It's up to the system to decide what to do with the information you give it, you don't really give commands. I think it's fine if the documentation says that in practice returning this will trigger a full sync but ultimately you're just describing the result set and it's up to Coda to figure out how to respond to that.

patrick-codaio: I feel like this might most "appropriately" live under `completion`? cc @ekoleda-codaio

patrick-codaio: I'm fine with the name as-is.

ekoleda-codaio: I like completion better as a home, but would that mean storing the incomplete status in the continuation until you finish the incremental sync and issue the completion?

CHANGELOG.md
jonathan-codaio: We at least have to have non-empty content under Unreleased otherwise the release tooling barfs. Keeps coming up, we should follow our normal flow that every change that will eventually need releasing has some content in here, it's fine if it's just boilerplate though.

api.ts
sam-codaio: Yes

CHANGELOG.md
sam-codaio: Yes my understanding is we should add stuff here generally if we plan on releasing? This is pretty vague

jonathan-codaio (approved):

patrick-codaio (approved):

PR #3061, opened 2024-09-04
Add a sync mode param to distinguish permission syncs vs others

patrick-codaio:
Context: https://codaproject.slack.com/archives/C06QGU1AAN9/p1725399881961899


pardeep-codaio (approved):

api_types.ts
```diff
@@ -761,6 +761,12 @@ export interface Sync {
    * {@link core.MetadataContext} The parameters of the sync formula for the sync table.
    */
   readonly parameters?: MetadataContext;
+
+  /**
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  isPermissionAwareSync?: boolean;
```
jonathan-codaio, line 769: We should make this an enum. Not quite sure what the values would be though.

Thinking something like `syncContext` where the values mean Meta, Personal/Shared, and In-doc. Then we can also use this in place of the `!context.docId` heuristic. So really just a naming exercise. Could also pull this out of  the `Sync` object and put NativeIntegration in there so we cover all the cases if we're going to do this. Plus once we support actions in Brain, is it going to be useful to know the invocation context of non-syncs?

Ah, maybe we put some of this in `InvocationLocation`?

```ts
enum InvocationLocationType {
  Brain = 'brain',
  Doc = 'doc',
  NativeIntegration = 'nativeIntegration',
}
```

And then in `Sync`, can do

```ts
enum BrainSyncType {
  Individual = 'individual',
  PermissionAware = 'permissionAware',
}
```

and document that those values are only present for `invocationLocation.invocationLocationType === 'brain'`

patrick-codaio: Brilliant as always. Updated! I named the enum InvocationSource.

api_types.ts
```diff
@@ -761,6 +770,14 @@ export interface Sync {
...
    */
   readonly parameters?: MetadataContext;
+
+  /**
+   * This is only populated when `context.invocationLocation.source === InvocationSource.Brain`.
+   *
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  syncType?: BrainSyncType;
```
jonathan-codaio, line 780: Optional, wondering if we should call this `brainSyncType` if we think it will only ever describe things specific to Brain.

jonathan-codaio (approved):

api_types.ts
ekoleda-codaio: I think I'd prefer the original boolean actually. Decomposing a "type" into it's component features is more flexible, which I think is important when the product is evolving so quickly.

Imagine we add a new sync type which is permission-aware and real-time. With an enum we'd need to add `Realtime` and document that's it's also permission-aware or name is `RealtimePermissionAware`, etc. With decomposed fields you could set `isPermissionAware: true` and `isRealtime: true`.

ekoleda-codaio: That said, I think enums are more future-proof, but perhaps we should scale is to the feature in question vs a broad "type". So perhaps:

```
dataScope: SyncDataScope;

enum SyncDataScope {
  // Sync the data that belongs to the connected account.
  Personal: 'personal',
  // Sync the data for the entire organization (assumes the connected account has elevated privileges).
  Organization: 'organization',
}
```

jonathan-codaio: Ah ok, I think the issue is just that my suggestion didn't model this dimension well. "Type" is too general, I think we just want `permissionMode` or something where the options are still `individual` and `permissionAware` or something. That doesn't preclude adding some orthogonal dimension in the future as illustrated.

It's extremely rare where I find a boolean to wind up being preferable to an enum, you're just penned in by a lack of extensibility. Definitely seemed the case here that BrainSyncType was sloppy modeling though.

jonathan-codaio: Ah I see your SyncDataScope suggestion which I missed when I was responding. That SGTM as well.

PR #3063, opened 2024-09-09
RFC: executionId

patrick-codaio:
Context: https://codaproject.slack.com/archives/C06QGU1AAN9/p1725903732419229

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -883,6 +883,16 @@ export interface ExecutionContext {
    * Information about state of the current sync. Only populated if this is a sync table formula.
    */
   readonly sync?: Sync;
+
+  /**
+   * If this invocation is a part of a crawling execution, like in Coda Brain, then this ID will be provided
+   * to all invocations. That includes invocations of sync `execute` and `executeGetPermissions`, as well as
+   * dynamic table features like `listDynamicUrls`, `getSchema`, and `getName`.
+   *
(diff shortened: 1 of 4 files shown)


pardeep-codaio (approved):

api_types.ts
```diff
@@ -883,6 +883,16 @@ export interface ExecutionContext {
    * Information about state of the current sync. Only populated if this is a sync table formula.
    */
   readonly sync?: Sync;
+
+  /**
+   * If this invocation is a part of a crawling execution, like in Coda Brain, then this ID will be provided
+   * to all invocations. That includes invocations of sync `execute` and `executeGetPermissions`, as well as
+   * dynamic table features like `listDynamicUrls`, `getSchema`, and `getName`.
```
jonathan-codaio, line 890: Ah is this the reason for not putting it inside of `sync`?

jonathan-codaio (commented):
Overall seems reasonable.

The general worry still applies about forcing ourselves to continue to support a particular way of doing things rather than requiring packs to be more stateless. So if we do this this might be one of the last things I'd consider unhiding, until it really proves its necessity.

api_types.ts
patrick-codaio: Yup!

jonathan-codaio (approved):

PR #3064, opened 2024-09-09
Plumbing for executionId

patrick-codaio:


runtime/bootstrap/index.ts
```diff
@@ -242,6 +243,7 @@ export async function injectExecutionContext({
   timezone: string;
   invocationToken?: string;
   sync?: Sync;
+  executionId?: string;
```
jonathan-codaio, line 246: Can you do this to catch this for next time?

```ts
export async function injectExecutionContext({
  context,
  fetcher,
  temporaryBlobStorage,
  logger,
  endpoint,
  invocationLocation,
  timezone,
  invocationToken,
  sync,
  executionId,
  ...rest
}: ExecutionContext & {context: Context; logger: Logger}): Promise<void> {
  ensureNever<keyof typeof rest>();
  ...
```

jonathan-codaio (approved):

runtime/bootstrap/index.ts
patrick-codaio: I don't think that works? 
Screenshot 2024-09-09 at 3 21 43 PM

Maybe you mean something like this?

patrick-codaio: oops didn't wait for upload
Screenshot 2024-09-09 at 3 25 30 PM

PR #3068, opened 2024-09-19
Add primitive for member groups

sam-codaio:
* Support additive nested group membership in the packs-sdk


schema.ts
```diff
@@ -1316,6 +1321,23 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * are required for sync tables with role {@link TableRole.GroupMembers}.
+   *
+   * Must be a {@link ValueType.String} or {@link ValueType.Number} property
+   *
+   * Nested groups are additive. All children members of a nested group will be considered members of the parent group.
+   * TODO(sam): Unhide this
+   * @hidden
+   */
+
+  nestedGroupIdProperty?: PropertyIdentifier<K>;
```
jonathan-codaio, line 1340: Hmm sorry I'm having trouble picturing this. Is the id here that a row in the group members table can be _either_ a user or a child group?

If so, I think that's confusing. I don't think we want tables where the rows are polymorphic. This feels better to express on a Groups table.

sam-codaio: The question to me is do we view parent groups as some separate thing from just group membership in general?

jonathan-codaio: Hmm I guess I was just imagining that there would be a parentIdProperty on a Groups table.

I guess I don't feel strongly. Would be good to have another pack with nested groups to pressure test this with.

api.ts
```diff
@@ -2334,8 +2334,10 @@ export function makeSyncTable<
       throw new Error(`Sync table schemas with role ${TableRole.GroupMembers} must set a groupIdProperty`);
     }
 
-    if (!schemaDef.userIdProperty) {
-      throw new Error(`Sync table schemas with role ${TableRole.GroupMembers} must set a userIdProperty`);
+    if (!schemaDef.userIdProperty && !schemaDef.nestedGroupIdProperty) {
```
patrick-codaio, line 2337: is it valid to have a `nestedGroupIdProperty` without a `userIdProperty`? I'd think no... unless we let them have 2 GroupMembership role sync tables?

schema.ts
```diff
@@ -1316,6 +1321,23 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * One or both of {@link ObjectSchemaDefinition.nestedGroupIdProperty}/{@link ObjectSchemaDefinition.userIdProperty}
+   * are required for sync tables with role {@link TableRole.GroupMembers}.
+   *
+   * Must be a {@link ValueType.String} or {@link ValueType.Number} property
+   *
+   * Nested groups are additive. All children members of a nested group will be considered members of the parent group.
+   * TODO(sam): Unhide this
+   * @hidden
+   */
+
```
patrick-codaio, line 1339: extra line

api.ts
sam-codaio: You can have unlimited group membership role sync tables as of now.

jonathan-codaio (approved):

PR #3069, opened 2024-09-19
Tell a pack invocation retry information about the prior failure

patrick-codaio:


api_types.ts
```diff
@@ -805,6 +807,63 @@ export interface Logger {
...
+   */
+  RateLimitExceeded = 'RateLimitExceeded',
+
+  Unknown = 'Unknown',
+}
+
+interface BaseInvocationError {
+  type: InvocationErrorType;
+  message: string;
+  errorData?: object;
```
patrick-codaio, line 828: maybe this is better as a string, with the understanding that it may or may not be json?

testing/fetcher.ts
```diff
@@ -190,9 +190,11 @@ export class AuthenticatingFetcher implements Fetcher {
       return false;
     }
 
-    if (requestFailure.statusCode !== HttpStatusCode.Unauthorized ||
```
patrick-codaio, line 193: sorry for the auto-formatting diffs

api_types.ts
```diff
@@ -805,6 +807,65 @@ export interface Logger {
...
+
+  Unknown = 'Unknown',
+}
+
+interface BaseInvocationError {
+  type: InvocationErrorType;
+
+  message?: string;
+
+  errorData?: string | object;
```
patrick-codaio, line 830: I'd like to narrow this typing soon, but the current state of our errors isn't very clean and I doubt it's worth the time right now to figure out coercing for all the permutations.

api_types.ts
```diff
@@ -805,6 +807,65 @@ export interface Logger {
...
+  /**
+   * Could mean 3rd party API rate limit or a rate limit imposed by Coda.
+   */
+  RateLimitExceeded = 'RateLimitExceeded',
+
+  Unknown = 'Unknown',
+}
+
+interface BaseInvocationError {
+  type: InvocationErrorType;
```
sam-codaio, line 826: Do we want to start with only the error type? Worried pack authors will rely on message/errorData which feel unstable. 

Fine keeping all the information as well

himanshu-codaio (approved):

himanshu-codaio:
Do we want to start by exposing this data only to the set of whitelisted packs so that random third parties don't start using it while we are working on it?

api_types.ts
```diff
@@ -805,6 +807,65 @@ export interface Logger {
...
+  | ResponseTooLargeInvocationError
+  | UnknownInvocationError;
+
+type MissingInvocationErrorTypes = Exclude<InvocationErrorType, InvocationError['type']>;
+/**
+ * We export this because unfortunately, TS doesn't let you ignore only an noUnusedLocals error without also
+ * suppressing the TS error we *want* to see.
+ * @hidden
+ */
+export type _ensureInvocationErrorEnumCompletion = Assert<MissingInvocationErrorTypes extends never ? true : false>;
```
jonathan-codaio, line 867: Say a bit more about what this is detecting? This is making sure the union type contains everything in our enum?

api_types.ts
jonathan-codaio: Yeah +1, I don't want an SDK contract where people rely on error text being stable, would prefer to only commit to the types.

api_types.ts
```diff
@@ -893,6 +954,15 @@ export interface ExecutionContext {
...
   readonly executionId?: string;
+
+  /**
+   * If this invocation is a retry, this will be populated with information about what went wrong
+   * during the previous attempt.
+   *
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  readonly previousAttemptError?: InvocationError;
```
jonathan-codaio, line 965: I guess ideally we'd put this in `Sync` since that's the only place where "previous" makes any sense, but maybe that's not possible due to desired usage for executeGetPermissions or something?

patrick-codaio: We actually put our auto-retry logic into other places too, like getSchema

patrick-codaio: so this is a different "previous" than a sync's "previous"

api_types.ts
patrick-codaio: Makes sense! Removed both message & data

jonathan-codaio (approved):

himanshu-codaio:
I was thinking a bit more about this overnight, and one additional detail I would like to include is the number of retries, with the array of error codes from each failure.

The main benefit would be a pack can use that info to figure out how aggressively it should try and make sure the attempts succeed. 

For example Gmail fetches 200 messages in a single pass by default. If a failure happens and the pack is retrying, I might want to back off to 20 messages. Only if that also fails I want to back off to a single message. However, that is only possible if the pack knows about the retry count and the failures from each of them. Right now I will have to go all way to the extreme position of fetch a single message because the pack doesn't know anything better.

However, I wouldn't block this PR on it. I would love to get it merged so I can start leveraging it

api_types.ts
himanshu-codaio: Fwiw, I think including the actual failure when the failure was a RateLimit failure would be very useful for packs. 

Google for example includes metadata on when the rate limit would be over and it is safe to start making api calls. Unfortunately from the looks of it any api calls before that time is up continue to fail, and not only do they fail they cause the rate limit expiration to be pushed out.

So if the pack doesn't know what was the response from Google, the pack will have no choice but to re-issue the api calls and that will back fire and will cause the rate limit period to be extended. The only way this will get fixed is if the exponential backoff in our infra backs off far away enough to jump past the deadline given to us by google. This is very inefficient.

Here is an ingestion execution where this happened - https://admin.coda.io/ingestionExecutions/3db6cb4a-1662-4f96-8825-884bc1f697ff and if you look at the consecutive failures we failed 16 times in a row and every failure is causing the retry after time to get pushed out.

image

himanshu-codaio: So I think we should pass the raw failures back from the api calls back to the pack. Yes there is a risk that a pack may rely on some text that they shouldn't but that is up to the pack. We shouldn't penalize packs who know what they are doing because some packs may not know what they should do.

patrick-codaio: For this case, we can probably read from the `retry-after` header and pass that in... although really in that case, we should just not even try to invoke the pack until that time is up. So I don't see much point in adding that additional data?

PR #3071, opened 2024-09-20
Deletion predicates

patrick-codaio:


jonathan-codaio (approved):
LGTM

api_types.ts
```diff
@@ -1084,3 +1084,19 @@ export interface SyncCompletionMetadata {
...
+export interface ItemMatchingPredicate {
+  /**
+   * The name of the property to search for matches.
+   */
+  property: string;
+
+  /**
+   * What value(s) to search for.
+   */
+  matches: string | string[] | number | number[] | boolean | boolean[];
```
sam-codaio, line 1101: By default do we assume these are exact matches? Assuming yes and that if we want regex support we can add something else

sam-codaio (approved):

api_types.ts
patrick-codaio: Exact matches for now, yeah

PR #3073, opened 2024-09-25
Admin authentications

patrick-codaio:


builder.ts
```diff
@@ -289,24 +306,23 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
...
+    this.systemConnectionAuthentication = this._wrapAuthenticationFunctions(
+      systemAuthentication,
+    ) as SystemAuthentication;
+    return this;
+  }
 
+  /**
+   * @hidden
+   */
+  addUserAuthentication(auxAuth: AuxiliaryAuthentication): this {
```
jonathan-codaio, line 318: I think it's confusing to have addUserAuthentication and setUserAuthentication. Would lean into auxiliary more and do addAuxiliaryAuthentication.

types.ts
```diff
@@ -970,6 +977,45 @@ export type VariousSupportedAuthentication =
...
 export type VariousSupportedAuthenticationTypes = $Values<Pick<VariousSupportedAuthentication, 'type'>>;
 
+/**
+ * @hidden
+ */
+export enum ReservedAuthenticationKeys {
+  /**
+   * References the default user authentication of the pack.
+   */
+  Default = 'DEFAULT',
```
jonathan-codaio, line 987: Nit, I think we have a standard enum style in the SDK, can't remember if it's TitleCase or camelCase, but should use it here. If you're trying to use keys that are unlikely to collide with user keys, would go even more distinctive.

types.ts
```diff
@@ -970,6 +977,45 @@ export type VariousSupportedAuthentication =
  */
 export type VariousSupportedAuthenticationTypes = $Values<Pick<VariousSupportedAuthentication, 'type'>>;
 
+/**
+ * @hidden
+ */
+export enum ReservedAuthenticationKeys {
```
jonathan-codaio, line 983: Do we really need these? Built-in vs auxiliary auth types will be in separate parts of the manifest so I don't think there's any risk of collision.

types.ts
```diff
@@ -970,6 +977,45 @@ export type VariousSupportedAuthentication =
...
+
+  /**
+   * Passed into formula context.
+   * Cannot be one of the ReservedAuthenticationKeys.
+   */
+  key: string;
+
+  displayName: string;
+
+  description?: string;
```
jonathan-codaio, line 1010: Would require this, will be important to show the in UX.

api_types.ts
```diff
@@ -941,13 +950,21 @@ export interface ExecutionContext {
    */
   readonly sync?: Sync;
 
+  /**
+   * If this function is being invoked with authentication, this indicates which authentication was used.
+   *
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  readonly authenticationKey?: string;
```
jonathan-codaio, line 959: Probably need a new name here, "authentication key" is very suggestive of a secret credential.

types.ts
jonathan-codaio: Ah I see, you need a way to reference these on the formulas.

api_types.ts
```diff
@@ -494,6 +494,15 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
    * so an end user must have both sets of permissions.
    */
   readonly extraOAuthScopes?: string[];
+
+  /**
+   * TODO(patrick): Without this set, should we allow all or only DEFAULT?
```
jonathan-codaio, line 499: For backwards compatibility, I think we have to take it to mean at a minimum default if present + system if present.

Could choose a different default if auxiliary auth types are defined but that may be confusing.

May be simplest to take "everything" as the default.

api_types.ts
```diff
@@ -494,6 +494,15 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+   *
+   * Ignored if connectionRequirement is None.
```
jonathan-codaio, line 501: When we add validation should probably validate that you can't set these together.

types.ts
```diff
@@ -970,6 +977,45 @@ export type VariousSupportedAuthentication =
...
+   * References the system authentication of the pack.
+   */
+  System = 'SYSTEM',
+}
+
+/**
+ * TODO(patrick): Unhide this.
+ * @hidden
+ */
+export interface AuxiliaryAuthentication {
```
jonathan-codaio, line 999: I can't quite articulate it but I feel like we should either name/frame this as being specific to admins, or add a field to indicate that this definition applies only to admins and force you to set it for now.

We will only offer these to org admins in Brain today. If for some reason we support auxiliary auth for non-admin users, things would break badly if the auth types offered for admins started to get offered to non-admins. So if we're not naming the whole concept in a way to force it to always relate to admins I think we at least need to add a restriction that forces this current crop of defs to be tied to admins if we think we may extend what this is for in the future.

api_types.ts
patrick-codaio: Good point... brainstorming:
- AuthenticationMode
- AuthenticationFlavor
- AuthenticationStrategy
- AuthenticationMethod

And if we don't have the "key" suffix here, maybe we switch the "key" property of AuxiliaryAuth to being "id"?

cc @ekoleda-codaio

patrick-codaio: New idea from @alan-codaio - change "key" to "name" on AuxAuth interface, and then this becomes authenticationName

ekoleda-codaio: I think `authenticationName` is pretty good. There are other instances where we use `name` to refer to the internal label for a thing:

- `SetEndpoint.name`
- `CustomAuthParameter.name`

builder.ts
ekoleda-codaio: What about overloading `setUserAuthentication` to take in an array of authentication types? Similar to how we extended `BaseAuthentication.networkDomain` to take an array of domains.

types.ts
```diff
@@ -970,6 +977,45 @@ export type VariousSupportedAuthentication =
...
+
+  displayName: string;
+
+  description?: string;
+
+  /**
+   * If true, this authentication can be used to sync permissions associated with data
+   * in addition to the data itself.
+   */
+  canSyncPermissions?: boolean;
```
ekoleda-codaio, line 1016: Is this meant to capture the idea of "supports meta syncs"? If so, I don't know if it's quite sufficient. A personal access token _could_ sync permissions, but it couldn't fetch data from the entire org.

builder.ts
patrick-codaio: I think I prefer a separate function because these authentications aren't meant for just _any_ user.

types.ts
patrick-codaio: Calling them "Admin authentications" is growing on me. But I'll throw out one other brainstorm idea:

"organization authentications" - all our organization-level settings are only available admins anyways. I think this makes it pretty clear these will not be used in classic doc usage context. Potential downsides:
- Have we discussed whether we'll allow non-enterprise customers to use meta syncs without making organizations? This assumes the implementation decision of always using organizations, which might not be safe.
- If we are aiming towards supporting a shared slack bot writing connection, we might want that to work at the workspace level, not org level?

types.ts
patrick-codaio: If I rename AuxiliaryAuth=>AdminAuth, does that fix this?

types.ts
patrick-codaio: now: 
```
export enum ReservedAuthenticationNames {
  Default = 'defaultAuthentication',
  System = 'systemAuthentication',
}
```

api_types.ts
patrick-codaio: I'm leaning towards "everything"

types.ts
jonathan-codaio: Yeah it's the Slack but case that makes me rule out "organization authentication". I think "admin" captures that a bit better, because workspaces have admins too and so we maintain some useful ambiguity there. Still an open question if "admin" will always be an appropriate descriptor for these.

We may buy ourselves a little more extensibility if we leave these as "auxiliary auth" and add an `audience` property or something that for now is only allowed to be `audience: AuxiliaryAuthenticationAudience.OrgAdmin` or `CodaBrainAdmin` until we add further support.

types.ts
jonathan-codaio: What did we decide to call meta ingestions in product? We probably want to align with that. I think some bit here is necessary so that we know we can/should offer this as an option for the meta fork in the setup UX.

patrick-codaio: Seems like https://coda.io/d/Brain-Library-Hub-go-brain-library_dotKT55OqGX/Proposal-Naming-of-connections_suR3ShOn is saying we will have "organization" connections that "mirror permissions". Maybe I'm over-indexing on the `permissionSyncMode` that I added before, but I feel like `canSyncPermissions` still works with this framing just fine.

types.ts
patrick-codaio: I hear you... but I'm worried that direction will cut off a different form of flexibility. Imagine we scrap Coda Brain and instead develop a different kind of shared workspace tables product feature. I feel like we'd want to keep an "audience" fully as vague as "admin" for that reason.

types.ts
```diff
@@ -970,6 +985,50 @@ export type VariousSupportedAuthentication =
...
+  /**
+   * User-visible.
+   */
+  description: string;
+
+  /**
+   * If true, Coda will not assume that the user who sets up an account with this authentication
+   * should themselves have access to the data owned by that account.
+   */
+  isServiceAccount?: boolean;
```
patrick-codaio, line 1029: more context: https://coda.io/d/_dtiu4qRK1oG?encodedLocationContext=eyJjb21tZW50VGhyZWFkSWQiOiJyLUI5UnFoZGdVMi0iLCJwYWdlSWQiOiJzZWN0aW9uLTA0cm9oSnZPeUciLCJ0eXBlIjoiY2FudmFzIiwib2JqZWN0SWQiOiJncmlkLXExTExFOWtueFE6OmktUGNNajBhNW9DZDo6Yy1hWjRVX0MzckpLIiwicGFyZW50TG9jYXRpb25Db250ZXh0Ijp7InR5cGUiOiJ0YWJsZSIsIm9iamVjdElkIjoidGFibGUta0dvWE5rRktQeiIsImdyaWRJZCI6ImdyaWQtcTFMTEU5a254USIsInJvd0lkIjoiaS1QY01qMGE1b0NkIiwiY29sdW1uSWQiOiJjLWFaNFVfQzNySksifX0

patrick-codaio:
@jonathan-codaio @ekoleda-codaio @alan-codaio This is ready for proper review now.

api.ts
```diff
@@ -990,7 +990,7 @@ export type TypedPackFormula = Formula | GenericSyncFormula;
 
 export type TypedObjectPackFormula = ObjectPackFormula<ParamDefs, Schema>;
 /** @hidden */
-export type PackFormulaMetadata = Omit<TypedPackFormula, 'execute' | 'executeUpdate'>;
+export type PackFormulaMetadata = Omit<TypedPackFormula, 'execute' | 'executeUpdate' | 'executeGetPermissions'>;
```
patrick-codaio, line 993: unrelated, but seemed more correct

builder.ts
```diff
@@ -289,24 +304,27 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
...
+      systemAuthentication,
+    ) as SystemAuthentication;
+    return this;
+  }
 
+  /**
+   * TODO(patrick): Unhide this
+   * @hidden
+   */
+  setAdminAuthentications(adminAuthentications: AdminAuthentication[]): this {
```
jonathan-codaio, line 318: In the builder pattern, we've used `set()` to mean "set or overwrite a singular value" and `add()` to mean "append to an array value", so I think this should be `addAdminAuthentication()` and take a single value and not an array.

types.ts
```diff
@@ -839,11 +846,9 @@ export interface VariousAuthentication {
 }
 
 /**
- * The union of supported authentication methods.
+ * The union of authentication types that pack makers are allowed to use.
```
jonathan-codaio, line 849: Nit, this comment is confusing, intuitively I think people would expect everything in a public SDK is "usable" by makers. I think the implication here is that VariousAuthentication is only allowed for 1P packs at present?

types.ts
```diff
@@ -859,6 +864,20 @@ export type Authentication =
   | GoogleServiceAccountAuthentication
   | CustomAuthentication;
 
+/**
+ * Exported to avoid Typescript not-used error
+ * @hidden
+ */
+export type _AllowedAuthenticationExtendsBaseAuthenticationCheck = EnsureExtends<
```
jonathan-codaio, line 871: This is fine but I think you could also put this in a test file and get the same effect (and api_test.ts already has some example of expressions meant TS verification that don't do anything at runtime).

testing/upload_validation.ts
```diff
@@ -715,6 +745,22 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
     .filter(([authType]) => authType in variousSupportedAuthenticationTypes)
     .map(([_authType, schema]) => schema);
 
+  const reservedAuthenticationNames = Object.values(ReservedAuthenticationNames).map(value => value.toString());
```
jonathan-codaio, line 748: Interesting, how come `.toString()` was needed?

testing/upload_validation.ts
```diff
@@ -2140,46 +2262,79 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+        if (!isDefined(authNetworkDomains)) {
+          // This is a Various or None auth pack.
           return;
         }
+
+        const readableAuthTitle =
+          name === ReservedAuthenticationNames.Default ? 'setUserAuthentication()' : `authentication ${name}`;
+
+        // Auth network domains must match pack network domains.
+        for (const authNetworkDomain of authNetworkDomains) {
```
jonathan-codaio, line 2285: What is our policy going to be for network domains when there are multiple auth types declared? Without a network domain exemption, I guess it's clear that all authentications must have exactly the same auth domain and it must be the single network domain. But if multiple network domains are allowed, is the auth network domain allowed to vary across auth types? It seems like we either need to require this or our runtime fetcher network domain validation checks need to resolve to the correct auth type to make sure the request is legal (maybe we need to do this anyway).

alan-codaio (approved):

testing/upload_validation.ts
patrick-codaio: Because this :/
Screenshot 2024-10-08 at 12 54 24 PM

testing/upload_validation.ts
patrick-codaio: I'm allowing the auth network domain to vary across auth types, since we have the manual approval gate in place. I agree we need to have tests that our fetcher enforces those various domain restrictions appropriately. cc @alan-codaio

builder.ts
patrick-codaio: Done

types.ts
patrick-codaio: Done

testing/upload_validation.ts
jonathan-codaio: Ah yes, the classic reason we have `array.includes()` on the `coda` side.

jonathan-codaio (approved):

types.ts
patrick-codaio: Done!

PR #3074, opened 2024-09-26
update docs for post setup

gary-codaio:
ran into these gotchas when trying to add a new pack with post setup, let me know if there's other places to make note

### DIFF
--- a/docs/guides/basics/authentication/index.md
+++ b/docs/guides/basics/authentication/index.md
@@ -542,7 +542,7 @@ pack.setUserAuthentication({
 });
 ```
 
-The step's `getOptions` works like [dynamic autocomplete][autocomplete_dynamic], fetching the list of available endpoints and returning a display name and value for each. Note that the returned endpoints must be an object with a name and value field. Further, the value field must be a string that can be parsed by javascript's `new URL(value)` function. After the user enters their credentials they will be prompted to select one of the endpoints in a dialog.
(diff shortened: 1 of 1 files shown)


docs/guides/basics/authentication/index.md
```diff
@@ -542,7 +542,7 @@ pack.setUserAuthentication({
 });
 ```
 
-The step's `getOptions` works like [dynamic autocomplete][autocomplete_dynamic], fetching the list of available endpoints and returning a display name and value for each. After the user enters their credentials they will be prompted to select one of the endpoints in a dialog.
+The step's `getOptions` works like [dynamic autocomplete][autocomplete_dynamic], fetching the list of available endpoints and returning a display name and value for each. Note that the returned endpoints must be an object with a name and value field. Further, the value field must be a string that can be parsed by javascript's `new URL(value)` function. After the user enters their credentials they will be prompted to select one of the endpoints in a dialog.
```
ekoleda-codaio, line 545: "with a name and value field" - I think specifically it's `display` and `value`, and it may be better to clarify that.

ekoleda-codaio, line 545: Nit: "javascript's `new URL(value)` function" - JavaScript the language doesn't technically have a `URL` class, although both the web and Node libraries include it (it's not in the Pack Studio for example). Maybe just say "the value must be a valid URL"?

ekoleda-codaio (commented):
Thanks for helping improve the docs!

docs/guides/basics/authentication/index.md
gary-codaio: Done

gary-codaio: Done

jonathan-codaio (approved):

PR #3075, opened 2024-09-26
Test functions for executeUpdate and executeGetPermissions

sam-codaio:
* Create equivalent test functions for new sync table methods


jonathan-codaio (approved):
Need to bump the semantic version and changelog still though.

PR #3082, opened 2024-10-03
Improve types for mocking

oleg-codaio:
<table><tr><td><details><summary>This PR is part of a stack created with <a href="https://github.com/aviator-co/av">Aviator</a>.</summary>

* **#3084**
* ➡️ **#3082**
* `main`
</details></td></tr></table>

Improving stubs in `MockExecutionContext` to be strongly typed.

Also adding a `transformSyncFormulaResultToGetPermissionsRequest` helper for writing tests working with permissions.

PTAL @coda/packs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ### Changed
 
-- Improve types for testing
+- Improve types used for testing, specifically on `MockExecutionContext` returned by `newMockExecutionContext()`.
+  `fetcher.fetch`, `temporaryBlobStorage.storeUrl`, and `temporaryBlobStorage.storeBlob` should be stubbed using
+  `resolves()` and `rejects()` since the underlying methods return promises
 - Fix `untransformBody` helper for array inputs
 
 ## [1.7.19] - 2024-09-27
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Changed
 
 - Improve types for testing
+- Fix `untransformBody` helper for array inputs
 
 ## [1.7.19] - 2024-09-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 - Add continuation for get permissions request and response
 
+### Changed
+
+- Improve types for testing
+
 ## [1.7.19] - 2024-09-27
 
 ### Changed
--- a/handler_templates.ts
+++ b/handler_templates.ts
@@ -306,7 +306,7 @@ function mapKeys(obj: {[key: string]: any}, schema?: Schema): object {
         continue;
       }
       remappedObject[newKey] = mappedKeys.length > 1 ? deepCopy(obj[key]) : obj[key];
-      const keySchema: Schema & ObjectSchemaProperty | undefined = schema.properties[newKey];
+      const keySchema: (Schema & ObjectSchemaProperty) | undefined = schema.properties[newKey];
       const currentValue = remappedObject[newKey];
       if (Array.isArray(currentValue) && isArray(keySchema) && isObject(keySchema.items)) {
         remappedObject[newKey] = currentValue.map(val => mapKeys(val, keySchema.items));
@@ -367,7 +367,7 @@ function unmapKeys(obj: {[key: string]: any}, schema?: Schema): object {
       continue;
     }
     remappedObject[newKey] = deepCopy(obj[key]);
-    const keySchema: Schema & ObjectSchemaProperty | undefined = schema.properties[key];
+    const keySchema: (Schema & ObjectSchemaProperty) | undefined = schema.properties[key];
     const currentValue = remappedObject[newKey];
     if (Array.isArray(currentValue) && isArray(keySchema) && isObject(keySchema.items)) {
       remappedObject[newKey] = currentValue.map(val => unmapKeys(val, keySchema.items));
@@ -380,8 +380,8 @@ function unmapKeys(obj: {[key: string]: any}, schema?: Schema): object {
 
 export function untransformBody(body: any, schema: Schema | undefined): any {
   if (isArray(schema) && isObject(schema.items)) {
-    const objectBody = body as Record<string, any>;
-    const mappedObjs = unmapKeys(objectBody, schema.items);
+    const objects = body as Array<Record<string, any>>;
+    const mappedObjs = objects.map(obj => unmapKeys(obj, schema.items));
     return mappedObjs;
   }
 
--- a/testing/execution.ts
+++ b/testing/execution.ts
@@ -685,24 +685,6 @@ export async function executeSyncFormulaFromPackDefSingleIteration(
 }
 
 /**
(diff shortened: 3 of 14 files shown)


testing/mocks.ts
```diff
@@ -30,7 +40,9 @@ export function newMockExecutionContext(overrides?: Partial<MockExecutionContext
     timezone: 'America/Los_Angeles',
     invocationToken: v4(),
     fetcher: {
-      fetch: sinon.stub(),
+      fetch: sinon.stub<[FetchRequest], Promise<FetchResponse>>().callsFake(async r => {
+        throw new Error(`Unhandled fetch: ${r.method} ${r.url}`);
```
oleg-codaio, line 44: Updating `newMockExecutionContext` to throw by default (rather than just return an empty value).

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+### Changed
+
+- Improve types for testing
```
jonathan-codaio, line 9: The `returns` --> `resolves` thing seems like the most significant change, would call that out. Will external developers who have written unittests also need to update all usage like you did in the `packs` repo? If so I think we need to pre-announce this and explain how to update things. Clearly what was there before was broken so it's a good change but if it's not backwards-compatible we need to give a heads-up.

testing/execution.ts
```diff
@@ -677,6 +679,24 @@ export async function executeSyncFormulaFromPackDefSingleIteration(
   ) as Promise<GenericSyncFormulaResult>;
 }
 
+/**
+ * Transforms a result from a sync formula execution to one that can be passed to executeGetPermissions.
+ *
+ * @hidden
+ */
+export function transformSyncFormulaResultToGetPermissionsRequest(
```
jonathan-codaio, line 687: Ah this is unfortunate. This is because when you call a sync formula in a test you get back normalized values?

Originally when we created the unittest framework, you got back pre-normalized values, which to me made the most sense because the pack maker is not supposed to care about normalization, it's a thing Coda does on their behalf but they don't really need to know or care about. At some point we made the unittest framework more robust or moved where normalization happens and as a side effect the return values started coming back normalized. So now developers do need to confront this detail.

It's probably worth seeing briefly whether there's a way to get back pre-normalized values during testing so that developers can just ignore this entirely. It might not be feasible though if things are tightly wound up in there.

jonathan-codaio (commented):
CC @ekoleda-codaio

CHANGELOG.md
oleg-codaio: Yeah, if external developers wrote unit tests that didn't correctly return a promise (but instead returned a value), they'd need to update their tests to also use `resolves` to pass type checks. Functionally these tests still work because our code must in general use `await` rather than attempting to do promise chains with `.then()` - the latter would have caused RTEs otherwise.

@ekoleda-codaio can you help with the necessary announcement for this change?

testing/execution.ts
oleg-codaio: Yeah, correct - when you call `executeSyncFormulaFromPackDef` or `executeSyncFormulaFromPackDefSingleIteration` that returns the normalized values. (There are some examples in the `packs` repo.) This happens in `testing/execution.ts` in `findAndExecutePackFunction()`.

If we want to test permissions with the same values as passed in during ingestions then we need to untransform the result. However it looks like @alexd-codaio had a todo to revert these transforms:

https://github.com/coda/packs-sdk/blob/0d5ab265782ed6f9bad9b3fc56d838844be11db1/testing/execution.ts#L95-L96

However if we bypass transforms then extraneous properties aren't filtered out, and we also have to set `validateResult: false` since that seems broken too. So there's a bit of work, we might want to transform and untransform back implicitly (perhaps add a new flag).

jonathan-codaio: Oh good find. Ignoring extraneous properties for a second, could you get by without  transformSyncFormulaResultToGetPermissionsRequest() if you just used `useDeprecatedResultNormalization = false`?

That said, transforming and untransforming I think does simulate what happens in `coda`, so it seems reasonable to do that and not expose an untransformer. I'd rather we just put that behavior behind useDeprecatedResultNormalization, use it explicitly for now, and then announce that we're changing the default.

jonathan-codaio: Err sorry, I guess transform + untransform would become the behavior when useDeprecatedResultNormalization is _false_. You would explicitly set it to false in your new tests for now, and then we can make it part of this announcement and  then change the default.

oleg-codaio: Sounds good, let me update. And yeah easier to work with than dealing with a new helper function, thanks for pointing me here to have another look at the code.

handler_templates.ts
```diff
@@ -380,8 +380,8 @@ function unmapKeys(obj: {[key: string]: any}, schema?: Schema): object {
 
 export function untransformBody(body: any, schema: Schema | undefined): any {
```
oleg-codaio, line 381: Looks like we had a bug here compared to what `transformBody` does (cc @chris-codaio).

testing/execution.ts
oleg-codaio: Updated! Using this on https://github.com/coda/packs/pull/4494.

ekoleda-codaio:
LGTM. I'm not terribly clear on what the delta is for the Pack developer, but I don't see anything to object over. I'm not familiar enough with the implementation to approve the PR, so leaving that to others.

handler_templates.ts
jonathan-codaio: How does this manifest itself? Will this break any existing behavior related to 2-way sync?

CHANGELOG.md
jonathan-codaio: @ekoleda-codaio I think the only developer-facing change to announce would be that if somebody previously wrote tests with `newMockExecutionContext()` and they used `context.fetcher.fetch.returns(...)` they need to update it to use `context.fetcher.fetch.resolves(...)`. This should always have been the case but the old behavior was inaccurate considering these are async methods.

ekoleda-codaio: Thanks for clarifying, I missed this callout. Since tests could only be written with the CLI, and SDK version upgrades are manual and optional for the CLI (compared to the Pack Studio), I think it's fine to release this without an announcement or warning. I also suspect that the number of developers writing tests period is very small, so I expect the blast radius here to be incredibly small. A clear note in the Changelog is likely sufficient, but I can also write up a short community post once the version goes live.

jonathan-codaio: Ok that's fine, let's just go with a more thorough changelog entry.

handler_templates.ts
oleg-codaio: It looks safe to me given how we're calling it in the main repo. The only case I have concerns about is here:

https://github.com/coda/coda/blob/8b056761fdb68f55c89e54c5688a2e066f62f302/modules/server/calc-server/private/calc_service.ts#L2408-L2415

I wasn't able to get this codepath to trigger locally though - tried refreshing both pack sync tables and cross-doc.

oleg-codaio: Wasn't able to repro that code path with 2-way sync but I did update one place where we have test coverage here: https://github.com/coda/coda/pull/118340

jonathan-codaio (approved):
Ok would just flesh out the changelog though re `context.fetcher.fetch.resolves()` so that it's actionable for external developers.

CHANGELOG.md
oleg-codaio: Ok, updated the changelog.

PR #3084, opened 2024-10-07
Add support for validating permissions

oleg-codaio:
<table><tr><td><details><summary><b>Depends on #3082.</b> This PR is part of a stack created with <a href="https://github.com/aviator-co/av">Aviator</a>.</summary>

* ➡️ **#3084**
* **#3082**
* `main`
</details></td></tr></table>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+### Changed
+
+- Improve types for testing
+
 ## [1.7.19] - 2024-09-27
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Changed
 
 - Improve types for testing
+- Fix `untransformBody` helper for array inputs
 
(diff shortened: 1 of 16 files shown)


jonathan-codaio (commented):
Per https://coda.io/d/_dotKT55OqGX?encodedLocationContext=eyJjb21tZW50VGhyZWFkSWQiOiJyLV8yalhqZjdtTngiLCJwYWdlSWQiOiJzZWN0aW9uLTFSeTZLNzV0b3oiLCJ0eXBlIjoiY2FudmFzIiwib2JqZWN0SWQiOiJjYW52YXMtMVJ5Nks3NXRveiJ9, I wouldn't start by changing the SDK, would just make use of getters that return schemas with the same identity and at least prove things out that way.

I'm skeptical that there are that many APIs where you can actually batch lookups-by-id at the API level so ultimately it's going to be very similar to just calling a getter repeatedly and so a new SDK entrypoint that amounts to almost the same thing won't be that interesting. As for scale, you don't have to call the getter for _every_ item id in the sync if the corpus is large, you can randomize the set of ids chosen to spot-check them using a target corpus size for spot checks, that is either absolute (spot check at least 100 items) or percentage based (spot check at least 20% of the corpus).

oleg-codaio:
Alright, per discussion let's defer making these explicit changes in favor of implicitly using getters. We can come back to optimize later as needed.

PR #3085, opened 2024-10-08
Revert to allowing authorizationUrl & tokenUrl to not match networkDomain

patrick-codaio:
Without this, our current Google packs would break, and probably quite a few 3rd party ones.


dweitzman-codaio (approved):

alan-codaio:
No issue with the revert, but this would just block new uploads, correct?

Why not just allow the existing builds but require the correct authentication structure for any new builds?

patrick-codaio:
@alan-codaio I'm tempted... that's why I kept the code. But since it will be an annoying surprise to lots of pack makers, I just didn't want to take on that hassle right now. Just trying to move faster.

PR #3086, opened 2024-10-10
Remove per-row permissions limit

sam-codaio:
* The per row limit is something we are going to hit with larger instances. We should remove this
* Bumping the maximum permission batch size so that we can have more efficient permissions in cases where we do no require an extra fetch call


testing/upload_validation.ts
```diff
@@ -142,7 +142,7 @@ export const Limits = {
   ColumnMatcherRegex: 300,
   NumColumnMatchersPerFormat: 10,
   NetworkDomainUrl: 253,
-  PermissionsBatchSize: 1000,
```
alexd-codaio, line 145: Intentional?

api.ts
```diff
@@ -2439,26 +2430,7 @@ export function makeSyncTable<
...
-          throw new Error(
-            `Objects with ids: ${oversizedRowAccessDefinitions
-              .map(p => p.rowId)
-              .join(', ')} returned more permissions than the maximum allowed of ${MaxPermissionsPerRow} per object`,
-          );
-        }
-        return result;
-      }
-    : undefined;
+  const executeGetPermissions = wrappedExecuteGetPermissions ? wrappedExecuteGetPermissions : undefined;
```
alexd-codaio, line 2433: Do you even need this? Just rename `wrappedExecuteGetPermissions` to `executeGetPermissions` and return it.

alexd-codaio (approved):
LGTM mod nits

api.ts
sam-codaio: We do not

testing/upload_validation.ts
sam-codaio: Yes I think we should change this as well. Double checking with some experts

PR #3094, opened 2024-10-30
Document AWSAssumeRole

ekoleda-codaio:
- Unhide types associated with `AWSAssumeRole`.
- Move auth guide content related to AWS to it's own page.
- Document `AWSAssumeRole` in the guide, including the trust policy and running locally.


docs/guides/basics/authentication/aws.md
```diff
@@ -0,0 +1,124 @@
...
+To use this form of authentication, simply set the authentication type to [`AWSAssumeRole`][sdk_awsassumerole] and specify the AWS service you are connecting to.
+
+```ts
+pack.setUserAuthentication({
+  type: coda.AuthenticationType.AWSAccessKey,
+  service: "s3",
+});
+```
+
+When authenticating, the user will be prompted to enter the Amazon Resource Name (ARN) of the role to assume, as well as an associated external ID.
```
chris-codaio, line 46: The external ID is optional, but recommended.

docs/guides/basics/authentication/aws.md
```diff
@@ -0,0 +1,124 @@
...
+
+<img src="../../../../images/auth_aws_assume_role.png" srcset="../../../../images/auth_aws_assume_role_2x.png 2x" class="screenshot" alt="Prompting the user for their role and external ID">
+
+You can learn more about how to create roles in the [AWS documentation][awsdocs_createrole].
+
+
+### Trust policy
```
chris-codaio, line 53: Note: this needs to be setup by the pack user, not the pack author.

docs/guides/basics/authentication/aws.md
```diff
@@ -0,0 +1,124 @@
...
+        "StringEquals": {
+          "sts:ExternalId": "<INSERT_SECRET_VALUE_HERE>"
+        }
+      }
+    }
+  ]
+}
+```
+
+Coda's AWS user has the ID `arn:aws:iam::029208794193:root`, and this is same for all Packs. To ensure that other Packs can't access your role, generate an external ID (usually a UUID) to act as a shared secret.
```
chris-codaio, line 80: s/user/account

docs/guides/basics/authentication/aws.md
ekoleda-codaio: Ya, I saw that in the AWS docs, but it looks like the external ID is required by our current UX.

image

(Notice how the **Continue** button is grayed out without a value set for **ExternalId**.)

Should I report that as a bug? The AWS docs mention that the external ID is a best practice when the user being granted AssumeRole is not under your control, so it might be best to keep it required?

docs/guides/basics/authentication/aws.md
ekoleda-codaio: Good catch, done.

docs/guides/basics/authentication/aws.md
ekoleda-codaio: Done, here and above.

docs/guides/basics/authentication/aws.md
chris-codaio: We could enforce the best practice - I'm ok with that.  If so, then perhaps above call out that while AWS makes this optional, Coda is requiring it?

chris-codaio (approved):

docs/guides/basics/authentication/aws.md
ekoleda-codaio: Done.

PR #3098, opened 2024-11-01
Support onError for sync formulas

oleg-codaio:
`onError` is supported by standard formulas but not sync table ones, so adding support.

PTAL @coda/packs


api.ts
```diff
@@ -2305,6 +2309,7 @@ export function makeSyncTable<
     execute: wrappedExecute,
     executeUpdate: wrappedExecuteUpdate,
     executeGetPermissions,
```
oleg-codaio, line 2311: Would we also want to use the same one for `executeUpdate` and `executeGetPermissions`, or leave those be to do their own error handling? If the former, we could either define new properties (`onUpdateError` or `onGetPermissionsError`), or add some parameter to `onError` with the `errorSource`.

patrick-codaio: I'm ok with either of those paths forward, and neither are blocking this PR IIUC

api.ts
```diff
@@ -2393,8 +2398,16 @@ export function makeSyncTable<
     params: ParamValues<ParamDefsT>,
     context: SyncExecutionContext,
   ): Promise<SyncFormulaResult<K, L, SchemaT>> {
-    const syncResult = (await wrappedExecute(params, context)) || {};
-    const appliedSchema = context.sync.schema;
+    let syncResult: SyncFormulaResult<K, L, SchemaDefT>;
+    try {
+      syncResult = (await wrappedExecute(params, context)) || {};
+    } catch (err: any) {
```
patrick-codaio, line 2404: since we just talked about `any` in the eng all hands, i'd love to learn - any reason not to use `unknown` here?

patrick-codaio (approved):

api.ts
oleg-codaio: The issue then is we'd need to either type `onError` as `onError(err: unknown)`, or we need some way of asserting that the object is an error. Kind of a pervasive issue throughout our codebases. Did we have a best practice as far as a path forward for errors specifically? I was hoping a future TS release would do something to help with richer typings of errors.

patrick-codaio: Not to my knowledge :(

PR #3099, opened 2024-11-01
Allow paramaterizing sync formula continuation types

oleg-codaio:
Working with sync formulas currently requires a lot of `as` type assertions and `satisfies`. Updating the SDK to instead enable users to define a custom sync context:

```ts
export type MySyncExecutionContext = SyncExecutionContext<ItemsContinuation, ItemsIncrementalContinuation>;

  builder.addSyncTable({
    name: 'Items',
    formula: {
      execute: async ([], context: MySyncExecutionContext) => {
```


docs/reference/sdk/interfaces/core.ExecutionContext.md
```diff
@@ -67,7 +67,7 @@ ___
 
 ### sync
 
-• `Optional` `Readonly` **sync**: [`Sync`](core.Sync.md)
+• `Optional` `Readonly` **sync**: [`Sync`](core.Sync.md)<`unknown`, `unknown`\>
```
patrick-codaio, line 70: My only concern is whether impacts like this on the generated documentation is too distracting. If @ekoleda-codaio signs off, I'm happy.

ekoleda-codaio: Not too distracting, and I don't think many developers consult the reference docs. LGTM.

patrick-codaio (approved):

PR #3100, opened 2024-11-06
Add itemVersionProperty for tracking item state in ingestions

ebo-codaio:
Writeup: https://coda.io/d/go-brainwriteups_dQYLHXsHZWS/Querying-Item-State-From-Packs_suPSd1Ee?utm_source=slack#_luZQmP0N

Add itemVersionProperty for tracking item version in ingestions

PTAL @coda/snowcoda-platform


testing/upload_validation.ts
```diff
@@ -1610,6 +1611,14 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
           );
         };
 
+        const validateItemStateProperty = () => {
+          return validateProperty(
+            'itemStateProperty',
+            itemStatePropertySchema => itemStatePropertySchema.type === ValueType.String,
```
ebo-codaio, line 1617: should we support various types of data?

schema.ts
```diff
@@ -1374,6 +1375,16 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
 
+  /**
+   * The name of the property within {@link ObjectSchemaDefinition.properties} that can be be interpreted as
+   * text representing the state of an item.
+   *
+   * Must be a {@link ValueType.String} property.
+   * TODO(ebo): Unhide this
+   * @hidden
+   */
+  itemStateProperty? : PropertyIdentifier<K>;
```
sam-codaio, line 1386: I do not like this name. What is the "state of an item"?  That could mean anything

ebo-codaio: changed to version

schema.ts
```diff
@@ -1374,6 +1375,16 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
    */
   index?: IndexDefinition;
 
+  /**
+   * The name of the property within {@link ObjectSchemaDefinition.properties} that can be be interpreted as
```
sam-codaio, line 1379: What about 
```
The name of the property within {@link ObjectSchemaDefinition.properties} that can be be interpreted as
a unique representation of a version of the item. 
```

sam-codaio (approved):

schema.ts
patrick-codaio: doesn't need to be unique

schema.ts
```diff
@@ -1374,6 +1375,16 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+   * text/number representing the version of an item.
+   *
+   * Must be a {@link ValueType.String} property.
```
patrick-codaio, line 1382: this doesn't match your validation code later. I think we decided to force this as a string for now, so probably update the validation code to match this?

jonathan-codaio: My thought was that if an API returns a timestamp as a number we should probably avoid the pack maker needlessly creating a string copy of it. So best to proactively support numeric properties even though the contract is that we'll _return_ all versions as strings.

patrick-codaio: Hmmm I defer to @ekoleda-codaio about consistency vs ease-of-use tradeoff.

Either way, let's just make sure comments match reality :)

ebo-codaio: I putted string only initially and forgot to update the comment after. Wdyt the type? @erickoledadevrel

schema.ts
```diff
@@ -1374,6 +1375,16 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
 
+  /**
+   * The name of the property within {@link ObjectSchemaDefinition.properties} that can be be interpreted as
+   * text/number representing the version of an item.
+   *
+   * Must be a {@link ValueType.String} or {@link ValueType.Number} property.
+   * TODO(ebo): Unhide this
+   * @hidden
+   */
+  itemVersionProperty? : PropertyIdentifier<K>;
```
ekoleda-codaio, line 1386: Naming suggestion: `versionProperty`.

The term `item` is redundant, as all of these refer to the item (it's `titleProperty`, not `itemTitleProperty`).

testing/upload_validation.ts
```diff
@@ -1611,6 +1612,15 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
           );
         };
 
+        const validateItemVersionProperty = () => {
+          return validateProperty(
+            'itemVersionProperty',
+            itemStatePropertySchema => itemStatePropertySchema.type === ValueType.String
```
ekoleda-codaio, line 1618: While not public-facing, perhaps remove references to `itemState` if we are no longer using that term.

schema.ts
ekoleda-codaio: One of the things we got right for two-way sync is that the row returned by the code when syncing in is the same as the row value received by the code when executing an update. Returning one type and receiving another is confusing. If we only want to provide strings to the code for now then I would enforce that developers can only return strings in their schema. We can expand that later if we want, but we'd need to support it on both sides.

ebo-codaio: It makes sense to make the type consistent. But if we only support string type, would it make it more complicated to pack makers? For example if the version refers to a numeric timestamp, instead of simply return the property name, the pack maker will have to make change to schema to convert that field into string or add new field with string type.

ebo-codaio: Discussed with Eric and Patrick, the main concern here is like Eric said, it might be confusing to developers on the mismatch of the data type that sent and returned in DB. We can go with string only for the initial version and still have forward compatibility to expand types if needed.

PR #3103, opened 2024-11-09
Limited date range options

patrick-codaio:
Writeup: https://coda.io/d/Brain-Library-Hub-go-brain-library_dotKT55OqGX/Limiting-date-range-param-options_suiIA2S5


api_types.ts
```diff
@@ -338,6 +338,40 @@ export interface ParamDef<T extends UnionType> {
...
+   *
+   * Only supported for Date & DateArray parameters.
+   *
+   * @hidden
+   */
+  // TODO(patrick): Unhide this
+  allowedPrecannedValues?: Array<SuggestedValueType<T>>;
+
+  /**
+   * For a parameter that has an autocomplete providing options, or a one that uses an allowedPrecannedValues
```
jonathan-codaio, line 363: Nit, s/a one/one

testing/upload_validation.ts
```diff
@@ -795,14 +799,50 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+      param => {
+        if (!param.allowedPrecannedValues || !(isArrayType(param.type) && param.type.items === Type.date)) {
+          return true;
+        }
+        const relativeDateRanges = Object.values(PrecannedDateRange);
+        return param.allowedPrecannedValues?.every(
+          (value: unknown) => typeof value === 'string' && relativeDateRanges.includes(value as PrecannedDateRange),
+        );
+      },
+      {message: 'allowedPrecannedValues for a date array parameter can only be a list of relative date ranges.'},
```
jonathan-codaio, line 844: It seems like in the enum definitions, you differentiate "relative" data ranges from other kinds of precanned date ranges, but here it seems like you're describing all date ranges (all of PrecannedDateRange) as "relative".

api_types.ts
```diff
@@ -1107,6 +1139,99 @@ export enum PrecannedDateRange {
...
+ * PrecannedDateRange values were valid.
+ *
+ * We want such packs to be able to use relative date ranges without needing to change their
+ * parameter type, to maintain backwards compatibility.
+ */
+export enum PrecannedDate {
+  Today = 'today',
+  Yesterday = 'yesterday',
+  Tomorrow = 'tomorrow',
+  DaysAgo7 = '7 days ago',
```
jonathan-codaio, line 1195: Any reason to put spaces in these? It doesn't really matter but we have underscores in the existing enums, which may or may not be to match similar `coda` enums. Just trying to have some consistency in the shape of these values.

api_types.ts
```diff
@@ -1107,6 +1139,99 @@ export enum PrecannedDateRange {
   Everything = 'everything',
 }
 
+/**
+ * The set of date ranges whose end dates are today.
+ */
+export const PastRelativeDateRanges = [
```
jonathan-codaio, line 1145: Trying to make sure I understand the vocabulary. The existing vocabulary and comments next to precanned date ranges describe them as "live", meaning "require evaluation at runtime with regards to the current date", which I think is a useful term. When you say "relative" in these places do you mean "live" or do you mean something new like "live and has today as the start or end of the range"?

jonathan-codaio (commented):
Overall LGTM.

api_types.ts
patrick-codaio: Ah, thanks, I was using it for both in different places, inconsistently. For most of these, I will just switch to using "Live" instead of "Relative", but I'm not sure what to do about "PastRelativeDateRanges" and "FutureRelativeDateRanges" (which are the "live and has today as start/end" meaning), because "Relative" doesn't seem obvious enough after sleeping on it. Brainstorm: PastDateRangesUntilNow, PastToNowDateRanges, ...?

cc @ekoleda-codaio

testing/upload_validation.ts
patrick-codaio: fixed!

api_types.ts
patrick-codaio: changed to underscores

api_types.ts
patrick-codaio: I changed these to "UntilNowDateRanges" and "FromNowDateRanges"

jonathan-codaio (approved):

ekoleda-codaio (commented):
Realized I never sent these comments last night.

api_types.ts
```diff
@@ -338,6 +338,40 @@ export interface ParamDef<T extends UnionType> {
...
 
+  /**
+   * In certain contexts, Coda's default behavior is to crawl *all* data available, in which case
+   * the {@link ParamDef.suggestedValue} will not be ideal, as most use cases will prefer efficiency
+   * over completeness. Use this field to specify a default value when Coda is crawling all data.
+   *
+   * @hidden
+   */
+  // TODO(patrick): Unhide this
+  fullCrawlSuggestedValue?: SuggestedValueType<T>;
```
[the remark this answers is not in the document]
patrick-codaio: Reasonable! It would be a little messy with types, needing to pass down the `<T>` but if @jonathan-codaio agrees with you I'm happy to move it.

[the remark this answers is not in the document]
jonathan-codaio: Oh interesting. I guess I had read this as meaning "suggested value when used in a Brain ingestion" but with a less Brain-y name. Is the usage of this truly only specific to when there's a crawl involved? If so, yeah it would seem to make sense to group it with the crawl stuff.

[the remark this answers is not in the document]
patrick-codaio: No, your reading is correct, but I think the question are "ingestions" and "crawls" the same thing or not?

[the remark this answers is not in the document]
ekoleda-codaio: I'm fine either way on that question, I just think if we stick with `crawl` in the name of this new field it should probably live under `crawlStrategy`.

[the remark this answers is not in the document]
jonathan-codaio: I don't see ingestions and crawls as the same. Crawls are a particular implementation strategy that a pack may choose to use.

Agree that having `crawl` in both places with different meanings is probably undesirable.

PR #3106, opened 2024-11-13
Bump dependency versions

dweitzman-codaio:
This PR was written by `pnpm audit --fix`

After this `pnpm audit --prod` is happy, but there are still some dependencies of release-it in dev that have advisories


dweitzman-codaio:
> Yeah it's funny, I tried this too and it finds a bunch of things but _not_ an update to vm2 🤷 . Seems worth doing, maybe we'll still need to manually bump vm2 to resolve that one.

There's no fixed version of vm2 to upgrade to, so it's not clear what we can do other unless we migrate off release-it

PR #3111, opened 2024-11-19
Fix indexing strategy validation for object schemas

randy-codaio:
Noticed this while running some experimentation on dev that indexing strategies in pack schemas were being dropped. The strategy property is being ignored since `{property: 'blah', strategy: 'blah'}` still satisfies the property identifier schema without using `strict()`. Add strict so that schemas with `strategy` defined fall through to the `DetailedIndexedProperty` schema validation. Would've also worked by reordering the zod union.


jonathan-codaio (approved):
Good catch. I think we probably actually want to use strict objects everywhere.

Can you bump the prerelease version to `.3` in package.json? That's needed to incorporate this into `coda`.

PR #3112, opened 2024-11-19
[CSB]: adding full execution id for pack calls to query item version

ebo-codaio:
Pack function requires full execution id info to query item version. 

PTAL @coda/snowcoda-platform

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -996,13 +996,22 @@ export interface ExecutionContext {
 
   /**
    * If this invocation is a part of a crawling execution, like in Coda Brain, then this ID will be provided
-   * to all invocations. That includes invocations of sync `execute` and `executeGetPermissions`, as well as
-   * dynamic table features like `listDynamicUrls`, `getSchema`, and `getName`.
+   * to all invocations as incremental execution id. That includes invocations of sync `execute` 
+   * and `executeGetPermissions`, as well as dynamic table features like `listDynamicUrls`,
+   * `getSchema`, and `getName`.
    *
    * TODO(patrick): Unhide this
    * @hidden
    */
   readonly executionId?: string;
+
+  /**
+   * Provided by a crawling execution, like in Coda Brain, this ID will be the full execution id.
+   *
+   * TODO(ebo): Unhide this
+   * @hidden
+   */
+  readonly fullExecutionId?: string;
 
   /**
    * If this invocation is a retry, this will be populated with information about what went wrong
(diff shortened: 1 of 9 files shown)


api_types.ts
```diff
@@ -996,14 +996,23 @@ export interface ExecutionContext {
...
+   * and `executeGetPermissions`, as well as dynamic table features like `listDynamicUrls`,
+   * `getSchema`, and `getName`.
    *
    * TODO(patrick): Unhide this
    * @hidden
    */
   readonly executionId?: string;
 
+  /**
+   * Provided by a crawling execution, like in Coda Brain, this ID will be the full execution id.
```
patrick-codaio, line 1009: per @jonathan-codaio 's request, "crawls" are only a certain type of ingestion. Perhaps "full sync" would work here?

patrick-codaio: (and can you correct my bad comment above?)

api_types.ts
```diff
@@ -996,14 +996,23 @@ export interface ExecutionContext {
...
+   *
+   * TODO(ebo): Unhide this
+   * @hidden
+   */
+  readonly fullExecutionId?: string;
```
patrick-codaio, line 1014: If we're using "full sync" to mean an ingestion, then it's odd to use "full/incremental" here the way we do internally. I could see this being `baseExecutionId` perhaps...

But then again, maybe this is great as it is, and we should instead figure out a better word for "full sync". How about we expose the word "ingestion" in the SDK comments @jonathan-codaio @ekoleda-codaio ?

api_types.ts
```diff
@@ -996,14 +996,23 @@ export interface ExecutionContext {
 
   /**
    * If this invocation is a part of a crawling execution, like in Coda Brain, then this ID will be provided
-   * to all invocations. That includes invocations of sync `execute` and `executeGetPermissions`, as well as
-   * dynamic table features like `listDynamicUrls`, `getSchema`, and `getName`.
+   * to all invocations as incremental execution id. That includes invocations of sync `execute`
```
patrick-codaio, line 999: I find "as incremental execution id" a little confusing. How about:
```
If this invocation is a part of an ingestion execution, like in Coda Brain, then we provide the ID of the current execution. That execution might be a full execution or an incremental one (in which case {@link fullExecutionId} below would be a different value). ...
```

api_types.ts
patrick-codaio: Relatedly, this would mean "fullSyncSuggestedValue" should maybe become "ingestionSuggestedValue" cc @pardeep-codaio

ebo-codaio:
Close this PR for now, we plan to store this value in fetcher, pack makers do not need to explicitly use this value, but naming discussion is still here https://coda.io/d/go-brainwriteups_dQYLHXsHZWS/11-19-SDK-vocab-for-ingestions_suWTuz2V?utm_source=slack#SDK-ingestion-vocab_tu72mC65/r36

PR #3114, opened 2024-11-21
CB-2904 Remove canSyncPermissions

alan-codaio:
As discussed in https://github.com/coda/coda/pull/120132#discussion_r1842933123, if a sync table uses meta-sync we want to _only_ use admin authentications, not the default authentication. So this PR moves `canSyncPermissions` into the authentications definition that lives in adminAuthentications.

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -1004,20 +1004,12 @@ export enum ReservedAuthenticationNames {
   System = 'systemAuthentication',
 }
 
-export type AdminAuthenticationDefinition = AllowedAuthentication & {
-  /**
-   * If true, this authentication can be used to sync permissions associated with data
-   * in addition to the data itself.
-   */
-  canSyncPermissions?: boolean;
-};
-
 /**
  * TODO(patrick): Unhide this.
  * @hidden
  */
 export interface AdminAuthentication {
-  authentication: AdminAuthenticationDefinition;
+  authentication: AllowedAuthentication;
 
   /**
    * A unique identifier for this authentication configuration. Coda will pass it into formulas via
--- a/types.ts
+++ b/types.ts
@@ -295,12 +295,6 @@ export interface BaseAuthentication {
    * Using multiple authenticated network domains is uncommon and requires Coda approval.
    */
   networkDomain?: string | string[];
-
-  /**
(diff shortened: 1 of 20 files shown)


types.ts
```diff
@@ -1010,12 +1004,20 @@ export enum ReservedAuthenticationNames {
...
+   * in addition to the data itself.
+   */
+  canSyncPermissions?: boolean;
+};
+
 /**
  * TODO(patrick): Unhide this.
  * @hidden
  */
 export interface AdminAuthentication {
```
patrick-codaio, line 1019: why not just move it as a separate property here?

alan-codaio: That works too, but I took my approach because this property used to live in the nested authentication object and this change preserves that structure. It means that I don't need to change anything in the unforked Packs:

```
pack.addAdminAuthentication({
  name: 'AdminOAuth2Authentication',
  description: 'Requests permissions only available to Miro Company Admins',
  displayName: 'Admin Authentication',
  authentication: {...commonAuthenticationDef, scopes: adminAuthScopes, canSyncPermissions: true},
});
```

patrick-codaio: interesting... i'm curious what @ekoleda-codaio thinks

ekoleda-codaio: I was actually just working on documenting admin auth today, and I did find it a bit surprising that `canSyncPermissions` was in the nested auth definition and not in the outer `AdminAuthentication` definition. It would be a simpler interface IMO if it was at the outer level.

patrick-codaio: I agree, I think it's worth the pack code churn to move up a level

ekoleda-codaio: I also found that that only valid value was `true`, which sort of makes it feel like busy work. If there is any value in having this field if there is no variation in the value?

patrick-codaio: the default value is `true`, so it shouldn't be busy work... but sure we could just remove it for now, until we introduce some concept of admin auths that do something different.

alan-codaio: Removed it in afb7cf4e4a7767aeb3388184182304ca5e800718

patrick-codaio (approved):
update package.json version to a prerelease version?

PR #3115, opened 2024-11-22
Add ability to disable decompression of response body

ezhan-codaio:
We are moving response decompression from fetcher to lambda, if decompressing responses cause lambdas to OOM, this would give developers an option to disable decompression in lambda and instead do it in the pack. Additionally, this would allow developers to handle cases where multiple compressions or compression other than gzip and deflate is used.


api_types.ts
```diff
@@ -667,6 +667,10 @@ export interface FetchRequest {
    * You may inspect the `Location` header of the response to observe the indicated redirect URL.
    */
   ignoreRedirects?: boolean;
+  /**
+   * If true, the response {@link FetchResponse.body} will not be automatically decompressed.
+   */
+  disableDecompression?: boolean;
```
jonathan-codaio, line 673: This feels kind of weird as an interface. It seems like this only exists for the case where decompressing causes an OOM, in which case, what is the pack going to do with a compressed response that they can't decompress? The maker is just stuck.

I think we should just have a flag that controls whether or not this functionality is enabled (we have lots of precedent for passing flags into the lambda) and turn it on and monitor for new issues, and deal with whatever we encounter. I think the burden will be on us inside the platform to temporarily add per-pack overrides to stamp our OOMs and then fixing or suggesting fixes to use smaller page sizes or whatever.

jonathan-codaio (commented):
@ekoleda-codaio for thoughts

api_types.ts
ezhan-codaio: SGTM, removed.

ezhan-codaio: SGTM, removed.

api_types.ts
```diff
@@ -689,6 +689,13 @@ export interface FetchResponse<T extends any = any> {
    * to disable any parsing. Note however that this will result in the body being a NodeJS Buffer.
    */
   body?: T;
+  /**
+   * The compression used for body in the raw HTTP response (e.g. gzip, deflate).
+   * 
+   * Note that the response {@link FetchResponse.body} is automatically decompressed if this is gzip or deflate,
+   * otherwise, you will need to handle decompression, including cases where multiple compressions are used.
+   */
+  bodyCompression?: string;
```
jonathan-codaio, line 698: This seems reasonable as an interface (would probably call it simply `compression` to be concise) but are we actually going to make use of it in solving the OOMs? Seems like we should start by just moving the existing automatic decompression into the lambda, unless you had a specific plan to also use this within specific packs?

ezhan-codaio: Need this field so we know which responses we need to decompress in lambda, unless there is a better way?

ezhan-codaio: @patrick-codaio mentioned that it would also be helpful for developers to figure out why lambda is OOMing.

jonathan-codaio: Ah, my bad. I thought you were adding this to the _request_ interface, not the _response_ (and so expecting a bunch of packs to update to start setting this). Nevermind, sorry for the confusion.

Yeah, you do need a bit somewhere to tell the lambda if and how a response was compressed. But I thought at the http level this is communicated in the headers, which are already part of the fetch response. So would just try to use the `Content-Encoding` header or whatever, which is presumably what you'd look at to set this bit in the first place.

ezhan-codaio: Yeah this is basically `headers['Content-Encoding'] === 'gzip' && !config.decompress`, I don't see a good way to pass in runtime config into packs runtime so added this field.

jonathan-codaio: Ah, these are all runtime config settings that are passed into the lambda that you can key new behaviors off of:

https://github.com/coda/coda/blob/4f93caa98bf21eed78ddbcb5cdfceb2fdd775441/modules/common-internal/runtime-config/settings.ts#L1347-L1356

So would add a new property for this here, and then pass that value into the node fetcher to turn off automatic decompression and then look at the same value in `this._packRuntimeOptions` in the fetcher proxy to determine whether should decompress based on the response header.

Let me know if I'm still misunderstanding though.

There's a separate thing we have to address before rolling this out, which is that in order to know whether it's appropriate to do the new behavior in the node fetcher, we need to know if the pack version is up to date enough to have the new logic in the lambda. So need to also look at something like https://github.com/coda/coda/blob/001fdea61d6284d24eb9dd4f17d85cc40a5597b0/modules/common-internal/types/packs/index.ts#L65 in addition to the runtime config to enable the new behavior. (`sdkVersionMatchesFeature()` is used for a similar need when the actual SDK is changing, but in this case we only need to change the lambda so we need only look at the lambda version or whatever). But you can build the new behavior based only on the runtime config and get PRs merged and we can sort out this additional check before turning on the runtime config.

ezhan-codaio: I think that would work, thanks for the pointers!

api_types.ts
ekoleda-codaio: FWIW I think there is value in allowing developer to disable our automatic compression / decompression feature. Specifically, when a Pack is just trying to shuttle data from A to B and doesn't care about the contents, disabling compression could allow it to stay under the memory limits. A Pack may still want to use automatic compression/decompression on other requests where it does have to parse the response.

Axios, Got, and node-fetch all have this option, usually called `decompress: true`. If enabled (the default) the behavior seems to be:
- Automatically add an `Accept-Encoding` header with values we can decompress.
- Automatically decompress payloads we can decompress
- Remove the `Content-Encoding` header if we decompressed the body

jonathan-codaio: I'm fine with essentially just passing through the behavior of `compress` from node-fetch. The issue with this `disableCompression` interface as proposed was that I understood it to still set `Accept-Encoding` headers on the request so _Coda_ gets the benefit of compression during transport between the fetcher and the API, but that it hands the still-compressed value back to the pack to try to make sense of.

Either way, the immediate issue at hand is fixing OOMs with the compression that we apply by default, and adding net-new optional SDK functionality won't help there, so I think we might as well keep those as separate efforts.

PR #3116, opened 2024-11-22
Disallow committing and pushing to `main` directly

ezhan-codaio:


jonathan-codaio:
Hmm I thought we usually do this with branch protection rules. We do have these activated in this repo. Is it just because you're a superuser that you were able to bypass? Do we just need to add another setting to prevent that?

Screenshot 2024-11-22 at 3 50 44 PM

ezhan-codaio:
Yeah admins bypass all branch protection rules, git hooks is the only way to prevent that.

jonathan-codaio (approved):

jonathan-codaio:
> Yeah admins bypass all branch protection rules, we use git hooks in other repos too to prevent that.

I'm a little baffled that GitHub allows this without at least a CLI flag to signal intent.

PR #3117, opened 2024-11-23
Handle getConnectionName, getConnectionUserId for admin authentications

oleg-codaio:
Updating the thunk to properly invoke the correct authentication metadata formulas instead of always using the default.

PTAL @coda/packs / @alan-codaio

### DIFF
--- a/builder.ts
+++ b/builder.ts
@@ -1,4 +1,5 @@
 import type {AdminAuthentication} from './types';
+import type {AdminAuthenticationDef} from './types';
 import type {AllowedAuthentication} from './types';
 import type {AllowedAuthenticationDef} from './types';
 import type {Authentication} from './types';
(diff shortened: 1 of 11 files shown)


builder.ts
```diff
@@ -321,7 +322,7 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
    * TODO(patrick): Unhide this
    * @hidden
    */
-  addAdminAuthentication(adminAuth: AdminAuthentication): this {
+  addAdminAuthentication(adminAuth: AdminAuthenticationDef): this {
```
oleg-codaio, line 325: Adding some automatic metadata formula wrapping.

bundles/thunk_bundle.js
```diff
@@ -6847,6 +6847,17 @@ module.exports = (() => {
     }
   }
   __name(findAndExecutePackFunction, "findAndExecutePackFunction");
+  function getSelectedAuthentication(manifest, authenticationName) {
+    const { defaultAuthentication, adminAuthentications } = manifest;
+    if (!authenticationName || authenticationName === "defaultUserAuthentication" /* Default */) {
```
alan-codaio, line 6852: nit: I think we have `ReservedAuthenticationNames.Default` that you can use here

alan-codaio (approved):

bundles/thunk_bundle.js
oleg-codaio: Oh yeah this is the autogenerated .js file, ` thunk.ts` is using that constant.

PR #3119, opened 2024-11-27
Fix thunk not being able to find system auth

alan-codaio:
I missed this spot although it's interesting we never checked this in the old code?

### DIFF
--- a/bundles/thunk_bundle.js
+++ b/bundles/thunk_bundle.js
@@ -6848,9 +6848,12 @@ module.exports = (() => {
   }
   __name(findAndExecutePackFunction, "findAndExecutePackFunction");
   function getSelectedAuthentication(manifest, authenticationName) {
-    const { defaultAuthentication, adminAuthentications } = manifest;
+    const { defaultAuthentication, systemConnectionAuthentication, adminAuthentications } = manifest;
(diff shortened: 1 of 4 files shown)


jonathan-codaio (approved):

jonathan-codaio:
Copying over my comments from Slack:

https://codaproject.slack.com/archives/C01835XEBSS/p1732737046745139?thread_ts=1732730458.712259&cid=C01835XEBSS

"The issue here is that we actually don't care about looking at the authentication object when executing the code in Eric's repro. but Oleg's change always tries to find the authentication object and asserts that it exists, even though it only uses it in a few cases of metadata formulas that don't apply to system auth. So that change introduced an assertion that's irrelevant to these code paths.

The fix works, totally fine for now, we should consider thinking through whether the thunk should ever even be interested in system auth though. A better fix might push the getSelectedAuthentication() calls down into the cases where it's actually needed."

oleg-codaio (approved):
+1 to @jonathan-codaio's comments, but not blocking. Thanks!

jonathan-codaio (commented):
Forgot to mention that you'll need to bump the prerelease version in package.json in order to bring this into `coda` to start applying the fix to new builds.

PR #3120, opened 2024-11-27
CB-2317 Verify authenticationName is supported when invoking Pack formulas

alan-codaio:
Adds verification to our thunk so that we don't invoke formulas with connections created with unsupported authentication mechanisms. Will add unit tests

### DIFF
--- a/runtime/common/helpers.ts
+++ b/runtime/common/helpers.ts
@@ -4,10 +4,10 @@ import type {GenericSyncFormula} from '../../api';
 
 
 function verifyFormulaSupportsAuthenticationName(
-  formulaName: string,
-  allowedAuthenticationNames: string[] | undefined,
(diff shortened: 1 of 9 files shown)


runtime/common/helpers.ts
```diff
@@ -25,30 +51,53 @@ export function findFormula(packDef: BasicPackDefinition, formulaNameWithNamespa
   }
   for (const formula of formulas) {
     if (formula.name === name) {
+      if (verifyFormulaForAuthenticationName) {
+        verifyFormulaSupportsAuthenticationName(name, formula.allowedAuthenticationNames, authenticationName);
```
patrick-codaio, line 55: why not pass in the formula object instead of the first 2 params?

patrick-codaio (approved):

PR #3125, opened 2024-12-04
Update types for incremental sync

oleg-codaio:
This PR extends `Sync` and `SyncExecutionContext` to add a third type parameter to conditionally define the `sync.continuation` type for incremental syncs, which is often different than the type used for default/full syncs.

In addition, some metadata formulas now pass through their specific `SyncExecutionContext`, allowing type-safe access to `dynamicUrl`, etc.

It also adds a (still hidden) type helper called `TypedSyncFormulaResult` to make it easier to define return types of helper functions used for sync tables.

PTAL @coda/packs


jonathan-codaio (approved):
Not sure why my comment didn't send earlier.

api.ts
```diff
@@ -1268,7 +1281,7 @@ export interface SyncFormulaDef<
    * from a previous invocation, and fetches and returns one page of results, as well
    * as another continuation if there are more result to fetch.
    */
-  execute<ContextReturnT extends ContextT>(
+  execute<const ContextReturnT extends ContextT>(
```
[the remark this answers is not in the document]
oleg-codaio: `const` on parameter types forces TypeScript to stop type widening. I needed it to correctly infer return types as I was working on this, but not it seems like it doesn't actually do anything.

Let me revert.

api_types.ts
```diff
@@ -984,8 +1005,7 @@ export interface ExecutionContext {
   /**
    * Information about state of the current sync. Only populated if this is a sync table formula.
    */
-  readonly sync?: Sync<unknown, unknown>;
-
+  readonly sync?: Sync<unknown>;
```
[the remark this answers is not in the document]
patrick-codaio: your plan is for this to be present for all metadata functions invoked in the context of a sync, right? I was planning to move `executionId` into Sync, assuming this would be available in more than just the sync formula itself.

api_types.ts
```diff
@@ -1295,7 +1328,9 @@ export interface SyncCompletionMetadata<IncrementalContinuationT = Continuation>
    * TODO(sam): Unhide this
    * @hidden
    */
-  hasIncompleteResults?: boolean;
+  hasIncompleteResults: true;
+
+  incrementalContinuation?: never;
```
patrick-codaio, line 1333: This doesn't seem right to me - our original intention with `hasIncompleteResults` was to give packs the option to continue incrementally syncing in a degraded state while a new full sync starts. cc @jonathan-codaio @sam-codaio

sam-codaio: Yeah this is wrong. We should be able to keep incrementally running while we have incomplete results of some kind. Otherwise we have to wait for a full sync for any updates which could take a days. We should keep syncing what we can since often this is just for edge cases where the API does not give us enough information for full fidelity

oleg-codaio: I didn't see any packs returning both incremental results and `hasIncompleteResults: true`, which is why I made this change and why it didn't break anything. Most packs seem to abort at this point since they'd assume a full sync is triggered.

jonathan-codaio: Right, yeah we should allow for the case Sam/Patrick described, thanks for catching. Curious that it's not used in practice yet though.

patrick-codaio: For the (apparently very common) case of a fatal error completely breaking an incremental sync, my thought was to model that as a new IncrementalSyncFatalError for packs to throw.

api_types.ts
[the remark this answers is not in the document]
oleg-codaio: Yeah, exactly. I was thinking we only keep it on `SyncExecutionContext` but it's hard to model unless we have special `makeParameter` functions that work only within sync tables and pass the context through. For now, just keeping as is.

PR #3127, opened 2024-12-09
[Crossdoc] Add reverse lookup information to ObjectSchema

tuan-codaio:
For Crossdoc we receive a lot of feedbacks to improve how we handle lookup & reverse lookup columns. To achieve that we want to add more information to the reverse lookup column in ObjectSchema. See a recent thread https://codaproject.slack.com/archives/C066TBS0PEE/p1733339036860099

In the immediate future it is used to disable adding new items to the reverse lookup column (see https://github.com/coda/coda/pull/121332), but in the future we intend in the future to properly support reverse lookup for Crossdoc table, e.g: if two tables A and B have linked relation columns to each other and both are Crossdoc’d over, then a change in the lookup column in A should be reflected in B. To achieve this we need the full information of where the source column comes from.


schema.ts
```diff
@@ -1383,7 +1383,18 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
+
+  /**
+   * In case this Object schema is that of a reverse lookup column, store the info to link it back to the source column.
+   * Currently only used for Crossdoc
+   * @hidden
+   */
+  reverseLookupInfo?: {
+    sourceGridId: string;
+    sourceColumnId: string;
+    isStandard: boolean;
```
jonathan-codaio, line 1396: It doesn't look like you used these in the `coda` PR? Having grid and column ids in the SDK isn't a great interface because those aren't concepts that exist in the SDK and it's hard to see how they could ever be used by a pack (as opposed to, say, slider mins and maxes which are useful to cross-doc as well as packs).

Can you just have `isReverseLookup?: boolean;` which seems like it would be sufficient for the `coda` PR?

schema.ts
```diff
@@ -1383,7 +1383,18 @@ export interface ObjectSchemaDefinition<K extends string, L extends string>
...
    */
-  versionProperty? : PropertyIdentifier<K>;
+  versionProperty?: PropertyIdentifier<K>;
+
+  /**
+   * In case this Object schema is that of a reverse lookup column, store the info to link it back to the source column.
+   * Currently only used for Crossdoc
+   * @hidden
+   */
+  reverseLookupInfo?: {
```
jonathan-codaio, line 1393: The rest of the top-level object schema properties are ones that could apply to any object schema, but this is specific to references. I think an earlier version of this change defined a new interface for the case `type: ValueType.Object` and `codaType: ValueHintType.Reference` which seems correct, this new property should only live on that interface.

schema.ts
jason-codaio: In order to simulate reverse lookups in a child doc we need the full set of information. The first PR is just disabling the feature while we build proper support.

tuan-codaio:
I ended up injecting fields to the schema directly  for Crossdoc without needing to change the Packs SDK https://github.com/coda/coda/pull/121332, so will close this PR

PR #3128, opened 2024-12-10
Adding sync state service to get latest row version based on given row ids

ebo-codaio:
Discussion: https://coda.io/d/go-brainwriteups_dQYLHXsHZWS/Pack-SDK-Get-item-versions-interface_suFrdPVO?utm_source=slack#_luPjCMl1

Follow up: update `item ids` to `row ids` for ingestion related variables 

PTAL: @patrick-codaio @erickoledadevrel  @jonathan-codaio

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -781,6 +781,29 @@ export interface TemporaryBlobStorage {
   ): Promise<string>;
 }
 
+
+/**
+ * A service for retrieving the sync state in Coda Brain.
+ * @hidden
+ *
+ * TODO(ebo): unhide this
+ */
+export interface SyncStateService {
+  /**
+   * Retrieve the latest version of rows with the given row ids and returns a mapping of row ids to
+   * their latest version, e.g. {"Id-123": "1.0.0"}.
+   *
+   * If the row id is not found, it will not be included in the response.
+   *
+   * If the row version is not defined, it will be set to an empty string.
+   *
+   * @hidden
+   *
+   * TODO(ebo): unhide this
+   */
+  getLatestRowVersions(rowIds: string[]): Promise<{[rowId: string]: string}>;
+}
+
 /**
  * TODO(patrick): Unhide this
  * @hidden
@@ -1062,6 +1085,14 @@ export interface SyncExecutionContext<
    * Information about state of the current sync.
    */
   readonly sync: Sync<ContinuationT, IncrementalContinuationT, IncrementalSyncContinuationT>;
+
+  /**
+   * A service for retrieving the sync state in Coda Brain.
+   * @hidden
+   *
+   * TODO(ebo): unhide this
+   */
+  readonly syncState: SyncStateService;
 }
 
 /**
@@ -1100,6 +1131,14 @@ export interface UpdateSyncExecutionContext extends ExecutionContext {
    * Information about state of the current sync.
    */
   readonly sync: UpdateSync;
+
+  /**
+   * A service for retrieving the sync state in Coda Brain.
+   * @hidden
+   *
+   * TODO(ebo): unhide this
+   */
+  readonly syncState: SyncStateService;
 }
 
 /**
--- a/api_types.ts
+++ b/api_types.ts
@@ -1092,14 +1092,14 @@ export interface SyncExecutionContext<
    *
    * TODO(ebo): unhide this
    */
-  readonly syncState: SyncStateService;
+  readonly syncStateService: SyncStateService;
 }
 
 /**
  * A function to check if a given {@link ExecutionContext} is a {@link SyncExecutionContext}.
  */
 export function isSyncExecutionContext(context: ExecutionContext): context is SyncExecutionContext {
-  return context.hasOwnProperty('sync') && context.hasOwnProperty('syncState');
+  return context.hasOwnProperty('sync') && context.hasOwnProperty('syncStateService');
 }
 
 /**
--- a/api_types.ts
+++ b/api_types.ts
@@ -1096,6 +1096,13 @@ export interface SyncExecutionContext<
 }
 
 /**
+ * A function to check if a given {@link ExecutionContext} is a {@link SyncExecutionContext}.
+ */
+export function isSyncExecutionContext(context: ExecutionContext): context is SyncExecutionContext {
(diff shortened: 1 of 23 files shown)


api_types.ts
```diff
@@ -1047,6 +1070,14 @@ export interface SyncExecutionContext<
...
    */
   readonly sync: Sync<ContinuationT, IncrementalContinuationT, IncrementalSyncContinuationT>;
+
+  /**
+   * A service for retrieving the sync state in Coda Brain.
+   * @hidden
+   *
+   * TODO(ebo): unhide this
+   */
+  readonly syncState: SyncStateService;
```
jonathan-codaio, line 1080: Can we maybe call this `syncStateService`? Isn't `sync` an object that already represents "sync state"?

api_types.ts
```diff
@@ -1085,6 +1116,14 @@ export interface UpdateSyncExecutionContext extends ExecutionContext {
...
    */
   readonly sync: UpdateSync;
+
+  /**
+   * A service for retrieving the sync state in Coda Brain.
+   * @hidden
+   *
+   * TODO(ebo): unhide this
+   */
+  readonly syncState: SyncStateService;
```
jonathan-codaio, line 1126: Do we actually need this here on updates?

api_types.ts
```diff
@@ -776,6 +776,29 @@ export interface TemporaryBlobStorage {
...
+   *
+   * If the row id is not found, it will not be included in the response.
+   *
+   * If the row version is not defined, it will be set to an empty string.
+   *
+   * @hidden
+   *
+   * TODO(ebo): unhide this
+   */
+  getLatestRowVersions(rowIds: string[]): Promise<{[rowId: string]: string}>;
```
jonathan-codaio, line 799: We call things items and item ids in most places, but we use "rows" and row ids" here, how come the difference?

runtime/bootstrap/index.ts
```diff
@@ -286,6 +290,12 @@ export async function injectExecutionContext({
     'executionContext.temporaryBlobStorage.storeUrl',
     temporaryBlobStorage.storeUrl.bind(temporaryBlobStorage),
   );
+
+  await injectAsyncFunction(
+    context,
+    'executionContext.syncState.getLatestRowVersions',
```
jonathan-codaio, line 296: Interesting, I guess `context.syncState.getLatestRowVersions` will technically be available at runtime even for non-syncs, since this code doesn't distinguish between whether this is a sync or not. I guess that's not terrible though as long as the TypeScript is correct.

testing/execution.ts
```diff
@@ -432,7 +432,7 @@ export async function executeFormulaOrSyncWithVM<T extends PackFormulaResult | G
   formulaName: string;
   params: ParamValues<ParamDefs>;
   bundlePath: string;
-  executionContext?: ExecutionContext;
+  executionContext?: SyncExecutionContext;
```
jonathan-codaio, line 435: Hrm looks like we already had this in executeFormulaOrSyncWithRawParamsInVM but this utility isn't specific to syncs but uses `SyncExecutionContext` which is wrong. I guess isn't not causing problems at the moment but we later may need to go clean things up to parameterize this function with the type of execution context for greater accuracy.

testing/fetcher.ts
```diff
@@ -703,6 +704,15 @@ class AuthenticatingBlobStorage implements TemporaryBlobStorage {
   }
 }
 
+class AuthenticatingSyncStateService implements SyncStateService {
```
jonathan-codaio, line 707: Hmm I don't know why I choose to call the existing thing AuthenticatingBlobStorage when it's fake. This should probably be `FakeSyncStateService`.

runtime/bootstrap/index.ts
```diff
@@ -242,13 +244,15 @@ export async function injectExecutionContext({
   context: Context;
   fetcher: Fetcher;
   temporaryBlobStorage: TemporaryBlobStorage;
+  syncState: SyncStateService;
```
jonathan-codaio, line 247: Should probably type this as `SyncStateService | undefined` since ideally we will know enough when setting things up that we can choose not to provide this if it's not a sync.

runtime/bootstrap/index.ts
jonathan-codaio: Edit: per the above, if `syncState` is passed as undefined we should only conditionally inject this.

testing/ivm_helper.ts
```diff
@@ -58,6 +58,7 @@ export async function setupIvmContext(bundlePath: string, executionContext: Exec
     authenticationName: executionContext.authenticationName,
     executionId: executionContext.executionId,
     previousAttemptError: executionContext.previousAttemptError,
+    syncState: executionContext.syncState,
```
jonathan-codaio, line 61: Can we maybe make a type guard `isSyncExecutionContext()` and do:

```ts
syncState: isSyncExecutionContext(executionContext) ? executionContext.syncState : undefined,
```

I know it seems like we always call this with a SyncExecutionContext but I think we should be more accurate so there's less to unwind later if we want to clean up the callers too.

jonathan-codaio (commented):
Sorry for missing the meeting on this, probably some of these were discussed there.

api_types.ts
patrick-codaio: We felt like it would be better to standardize on "rows". Here's the accompanying change, it's pretty small: https://github.com/coda/packs-sdk/pull/3129

api_types.ts
patrick-codaio: I think no

api_types.ts
patrick-codaio: We also already used rows:
Screenshot 2024-12-09 at 5 12 50 PM

testing/fetcher.ts
ebo-codaio: Sounds good

api_types.ts
ebo-codaio: I guess it might be clearer if we include `service` in the name? wdyt @erickoledadevrel

patrick-codaio: I like adding "Service"

patrick-codaio (approved):

PR #3130, opened 2024-12-11
Create page for common errors

ekoleda-codaio:
I had previously listed common errors in the Troubleshooting guide, but when I went to look for it I could never remember where it was. This PR moves that information to it's own page, in the support section.

image

Assigning to on-call.


dweitzman-codaio:
Did we get rid of the CI job that would show a preview of the built docs?

dweitzman-codaio (approved):
I guess you can double-check the docs before deploying them, so I'll stamp this

ekoleda-codaio:
> Did we get rid of the CI job that would show a preview of the built docs?

Yes, it was using a 3P service that we cancelled the contract with.

PR #3131, opened 2024-12-11
Disable the Browserify polyfill for vm

ekoleda-codaio:
Context: https://codaproject.slack.com/archives/C01835XEBSS/p1733862253044739

### DIFF
--- a/test/compile_test.ts
+++ b/test/compile_test.ts
@@ -2,6 +2,7 @@ import {compilePackBundle} from '../testing/compile';
 import {executeFormulaOrSyncWithVM} from '../testing/execution';
 import {newMockSyncExecutionContext} from '../testing/mocks';
 import path from 'path';
+import { readFileSync } from 'fs';
 import sinon from 'sinon';
 import {translateErrorStackFromVM} from '../runtime/common/source_map';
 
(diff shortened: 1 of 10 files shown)


test/compile_test.ts
```diff
@@ -2,6 +2,7 @@ import {compilePackBundle} from '../testing/compile';
 import {executeFormulaOrSyncWithVM} from '../testing/execution';
 import {newMockSyncExecutionContext} from '../testing/mocks';
 import path from 'path';
+import { readFileSync } from 'fs';
```
chris-codaio, line 5: Nit: fix formatting to match other imports (i.e., no spaces around the imported symbol name).

testing/compile.ts
```diff
@@ -76,6 +82,9 @@ async function browserifyBundle({
     debug: true,
     standalone: 'exports',
   });
+  for (const module of ModulesToDisableBrowserifyPolyfill) {
```
chris-codaio, line 85: Any thoughts on how we can actually recognize one of these at compile time and warn the user?

chris-codaio (approved):

test/compile_test.ts
ekoleda-codaio: Done. Also added a eslint rule and use `lint-fix` to fix other instances.

testing/compile.ts
ekoleda-codaio: I'm not sure it's needed, or even desirable. For other Node built-ins without a polyfill (ex: `dns`), Browserify just silently ignores it. This change would treat `vm` as if there was no polyfill, so silently ignoring it seems fine.

Additionally, I doubt that Pack developers will ever intentionally import `vm`, but that it will come along with some transitive dependency they know nothing about and can't easily action, so the warning here could cause more confusion then provide help.

PR #3134, opened 2024-12-13
Default titleProperty to displayProperty

patrick-codaio:


jonathan-codaio (commented):
Any reason to do this in the SDK vs where it gets used? When we do it in the SDK we los the ability to differentiate cases where this was set for real vs where it was defaulted.

noah-codaio:
> Any reason to do this in the SDK vs where it gets used? When we do it in the SDK we los the ability to differentiate cases where this was set for real vs where it was defaulted.

Context in [this](https://codaproject.slack.com/archives/C06QGU1AAN9/p1734022531203339?thread_ts=1730913895.904239&cid=C06QGU1AAN9) Slack thread.

jonathan-codaio:
> > Any reason to do this in the SDK vs where it gets used? When we do it in the SDK we los the ability to differentiate cases where this was set for real vs where it was defaulted.
> 
> Context in [this](https://codaproject.slack.com/archives/C06QGU1AAN9/p1734022531203339?thread_ts=1730913895.904239&cid=C06QGU1AAN9) Slack thread.

Replied there, but:

I agree with the behavior, why do we want to do it in the SDK though? We lose the ability to know if the maker set the property vs us defaulting it.

At the very least, you can do the indirection in ObjectSchemaHelper to that access transparently falls back to displayProperty without actually mutating the pack at compile time.

jonathan-codaio (approved):

PR #3136, opened 2024-12-13
Upgrade sharp-cli to resolve lodash-pick vulnerability.

jonathan-codaio:
Addresses https://github.com/coda/packs-sdk/security/dependabot/74

There is no new version of `lodash.pick` to override to. But we're getting this transitively via `sharp-cli` so just update that. Sharp seemed to resolve this for themselves by downgrading lodash.pick to 3.x.

PTAL @neal-codaio @dweitzman-codaio @coda/packs

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -283,8 +283,8 @@ importers:
         specifier: 4.2.1
         version: 4.2.1
       sharp-cli:
-        specifier: 4.1.1
-        version: 4.1.1
+        specifier: 5.1.0
+        version: 5.1.0
       ts-loader:
         specifier: 9.5.1
         version: 9.5.1(typescript@5.1.6)(webpack@5.95.0)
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -5912,8 +5961,18 @@ packages:
     resolution: {integrity: sha512-0KpjqXRVvrYyCsX1swR/XTK0va6VQkQM6MNo7PqW77ByjAhoARA8EfrP1N4+KlKj8YS0ZUCtRT/YUuhyYDujIQ==}
     dev: true
 
-  /lodash.pick@4.4.0:
```
neal-codaio, line 5915: This removes a newer version of pick and adds a bunch of older versions.  Is that what you want?

jonathan-codaio: > There is no new version of lodash.pick to override to...Sharp seemed to resolve this for themselves by downgrading lodash.pick to 3.x.

neal-codaio (approved):

PR #3144, opened 2025-01-06
Require node20+ for SDK development

dweitzman-codaio:
Since 18 is broken at the moment and we're planning to migrate to 20 soon, seems like we might as well commit and force SDK-based development to use node 20+


setup-env.sh
```diff
@@ -8,10 +8,12 @@ if [[ "$(uname -s)" == "Darwin" ]] && [[ "$(arch)" == "arm64" ]]; then
...
-export PATH=${BASEDIR}/build/node/bin:${BASEDIR}/node_modules/.bin:${BASEDIR}/.pnpm_install/bin:${PATH}
-
 if command -v nvm &> /dev/null
 then
   # Gets version from .nvmrc
   nvm use
 fi
+
+
+export PATH=${BASEDIR}/.pnpm_install/bin:${BASEDIR}/build/node/bin:${BASEDIR}/node_modules/.bin:${PATH}
```
dweitzman-codaio, line 18: Moved this below the `nvm use` because it seems like nvm's node could/does have its own pnpm installed at a potentially different version

jonathan-codaio (approved):

PR #3146, opened 2025-01-24
For content review: DefaultRowLimit as a sync table option

alan-codaio:
Adds `DefaultRowLimit` as a sync table option:

- If not specified we default to 100 rows (as we do today).
- If the customer doc is on the free plan, the table will only have 100 rows, not the `defaultRowLimit`.


gary-codaio:
Not sure about this needing sdk changes. My understanding of was the intention was for it to be a setting set by org admins rather than pack makers.

alan-codaio:
> Not sure about this needing sdk changes. My understanding of was the intention was for it to be a setting set by org admins rather than pack makers.

I guess it's worth clarifying. This was posted in #pack-questions, which is usually for building Packs. (OTOH, people misuse that channel all the time). Pack config is also an overloaded term (I can see it being used as Pack definition). 

Ramesh's ` Matt wants to control the default for new pack installs` sounds like it's more Pack definition to me, but I'll double check with him.

alan-codaio:
I'd also see this as a bit of an odd usage of enterprise controls.. you want to use it set a _minimum_ row limit? I guess if anything we'd want to use this to restrict the amount of data you can pull in, but even that is a bit of a silly example.

api_types.ts
```diff
@@ -1348,11 +1348,21 @@ export type PropertyOptionsMetadataFunction<ResultT extends PackFormulaResult[]>
...
   GroupMembers = 'groupMembers',
 }
 
+/**
+ * Default row limits for a sync table.
+ */
+export enum DefaultRowLimit {
+  OneHundred = 100,
+  OneThousand = 1000,
+  TenThousands = 10000,
```
ekoleda-codaio, line 1363: Extra "s" on here? I think it should it be `TenThousand`.

alan-codaio: Ah yes

PR #3153, opened 2025-02-10
Enable chunking large fetcher responses

oleg-codaio:
Adding an experimental flag to enable fetching larger API responses as chunks.


patrick-codaio (approved):
don't we also want a strong type for the chunked result?

patrick-codaio:
oh, and can you bump the prerelease version in package.json?

PR #3157, opened 2025-02-12
Add parent property; remove unused/deprecated fields

alexd-codaio:
@patrick-codaio @jonathan-codaio @sam-codaio PTAL

Removes two indexing-related properties (`bodyTextProperty` , `parentIdProperty`) and adds a new `parent` property for describing parent/child relationships between items.


schema.ts
```diff
@@ -1135,6 +1134,26 @@ export interface IndexDefinition {
   popularityRankProperty?: PropertyIdentifier<string>;
 }
 
+/**
+ * Defines the parent of an object, which controls various behaviors for this object.
+ * TODO(alexd): Unhide this
+ * @hidden
+ */
+export interface ParentDefinition {
```
ebo-codaio, line 1142: In the parameter `crawlStrategy,` we also have the `parentTable` and `inheritPermissions` properties; I am wondering if this information is redundant?

patrick-codaio: We would only sometimes want crawl parents to also be _this_ kind of parent, but that could be controlled by adding a bool to `SyncTableRelation`... perhaps `isContainer` or some such?

schema.ts
```diff
@@ -1135,6 +1134,26 @@ export interface IndexDefinition {
...
+   * identifies the parent of this object.
+   *
+   * ValueType.Array is not supported is not supported for parentIdProperty.
+   *
+   * If this property references a property with {@link ValueType.Object} and
+   * {@link ValueHintType.Reference}, then it refers to an object in another logical table.
+   *
+   * Otherwise, it refers to an object in the same table.
+   */
+  parentIdProperty: PropertyIdentifier<string>;
```
jonathan-codaio, line 1154: Probably worth writing the actual schema def fragment for a couple real examples.

I know it's temping to use Reference here since the concept already exists but I'm not sure it quite aligns. At least there is the hurdle that a reference is a {name, id} pair and I think much of the time you won't have a name to use for the parent reference, just the id, and we don't care at all about the name for handling subitems, it's purely for display purposes.

In which case, we can either decide to use a Reference and just give a blank name or repeat the id as the name, or change the schema def here so that the parentIdProperty can point to an id scalar even for foreign keys, and then add an extra field here for the identity name of the parent table rather than getting that from the reference def.

schema.ts
```diff
@@ -1135,6 +1134,26 @@ export interface IndexDefinition {
...
+}
```
jonathan-codaio, line 1155: Do you not want to play with the bits for `inheritsPermissions` and `inheritsLifecycle` or whatever? Or first draft is to see what happens without those?

schema.ts
alexd-codaio: FWIW - my canvass of usages of `makeReferenceSchemaFromObjectSchema` within the packs repo shows me lots more places where this could be used (both for the strong and weak entity identity cases). So I think the feature generally might have some legs.

We could choose to relax the constraint that `ObjectSchema` has a required `displayProperty` if we wanted? At least in this context?

Anyway, I don't feel too strongly about it if you want to go that way.

Some usages (including ones we discussed) that seem like they'd work with `parent`:

image
image
image

schema.ts
alexd-codaio: My initial SWAG was to encode whatever behavior I wanted when I used these and systematize them as we used it in more places. But happy to plug in something if you prefer. :)

schema.ts
alexd-codaio: Hmm... not sure I see how to merge them.

`crawlStrategy` is on a parameter def and references strings whereas `parent` is on an `GenericObjectSchema` and references another schema property on the item?

My intent with this is that you'll have a schema like:

```
const docSchema = makeObjectSchema({type: ValueType.Object, idProperty: 'docId', displayProperty: 'title', properties: {docId: {type: ValueType.String}, title: {ValueType.String}}});
const pageSchema = makeObjectSchema({type: ValueType.Object, idProperty: 'pageId', displayProperty: 'name', index: {properties: ['content']}, properties: {pageId: {type: ValueType.String}, name: {ValueType.String}, content: {type: ValueType.String}, doc: makeReferenceSchemaFromObjectSchema(docSchema}}});
```

And then you'll return values like:
```
const rows: Array<SchemaType<typeof pageSchema>> = [
  {id: 'asdf', name: 'Subitems', content: 'Here is some text', doc: {id: 'deadbeef', name: 'Brain Backend vNext'}},
  {id: 'asdf2', name: 'Project Status', content: 'Project Status...', doc: {id: 'deadbeef', name: 'Brain Backend vNext'}},
]
```

Maybe you could have something like... 
```
interface ParentDefinitionBase {
  inheritsPermission: boolean;
  inheritsDelete: boolean;
  // ...
}

export interface CrawlStrategyDefinition extends ParentDefinitionBase {
  parentTable: GenericObjectSchema;
}

export interface ParentDefinition extends ParentDefinitionBase {
  parentIdProperty: PropertyIdentifier<string>;
}
```

patrick-codaio: The only problem I see with re-using the parameter crawl strategy is that some APIs might allow you to crawl all sub-items without needing a parent entity specified. Is that a use case you expect to be common?

Otherwise, we can just enrich the return values on the `coda` side by reading the sync table crawl hierarchy.

sam-codaio: Can we at least think through the cases we have now? 

Confluence - Seems like  it is modeled so that Pages is a crawl child.
Github - Everything is a crawl child of Repos it seems

etc....

sam-codaio (commented):
I am not sure what makes the most sense based on existing comments.

There is a simplicity to putting this on crawl strategy since that would automatically apply this to everything in a child ingestion. But the problem with doing that is it the prescribes a way for modeling ingestions that might not always make sense. You can end up with a ton of ingestions if there are 1000s+ of parents. But if this hits our existing use cases it is really simple to model/understand.

The separate field case is much more flexible but if we have no exact use case might be more complicated for little gain.

schema.ts
alexd-codaio: > The only problem I see with re-using the parameter crawl strategy is that some APIs might allow you to crawl all sub-items without needing a parent entity specified. Is that a use case you expect to be common?

I don't think that parent/child implies `crawlStrategy` or vice versa. Coda will not use `crawlStrategy`, and Airtable/Notion/Confluence don't seem likely either.

IMHO, `crawlStrategy` on a large number of "parents" is not very scalable even if there isn't a "subitem" iteration API (something I think we experienced with the old Drive pack); creating 100k ingestions just doesn't seem like a good pattern.

I imagine that we change the implementation of `crawlStrategy` to write the `parent` onto the child schema if we want.

alexd-codaio: OK - I digested this a bit further and I think I misunderstood the feedback re unifying these. Let me noodle a bit.

jonathan-codaio: Ok curious what the noodling turns up. My initial and maybe naive sense is that this is different than crawlStrategy. A crawl strategy is an instruction about implementation--you're declaring that an effective to ingest what amounts to a tree of entities given the APIs available is to perform a crawl. The subitems thing is simply declaring a conceptual relationship that exists within the data.

So I'm not opposed to keeping them separate. Maybe it turns out that all we're describing in all cases are different shades of a parent-child relationship and so the crawlStrategy amounts to a labeling of the parent reference akin to inheritsPermissions.

schema.ts
jonathan-codaio: That's fine, can add it in later.

alexd-codaio:
Updated with (I think) the feedback addressed re: unifying these.

There are two classes of parent definitions, one that is static (e.g. immutable, bound to sync table definition) and controls crawling / permissions / and lifecycle.

Another that is dynamic (can be specified on each returned object) and controls permissions for now (but perhaps lifecycle).

schema.ts
```diff
@@ -1135,6 +1134,78 @@ export interface IndexDefinition {
...
+
+  /**
+   * The name of the property within {@link ObjectSchemaDefinition.properties} that
+   * identifies the parent of this object.
+   */
+  parentProperty: PropertyIdentifier<string>;
+}
+
+/**
+ * Defines that this object has a static parent (e.g. a parent that is known at schema definition time).
```
jonathan-codaio, line 1178: Sorry I'm having trouble understanding the static vs dynamic concept. What is it about a crawl parent that makes it "known at schema definition time"?

No items are known until runtime so no item relationships are known until runtime. Relationships between Identities (classes of data) can be known at schema definition time. But that seems the same whether there's a crawl or not.

PullRequests has a relationship to Repos that is known at schema definition time, and we happen to want to crawl it. Pages has a relationship to Docs that is known at schema definition time, but we don't want to crawl it.

When we talk about "static" vs "dynamic" it's usually about the dimension of "does a dynamicUrl affect the schema" or "does the end user's account affect the schema" and I suppose we do have to worry about that here although I wasn't thinking about it until now. But the fact that a reference property schema that parentIdProperty points to can have a dynamicUrl feels orthogonal to the rest of this. Most of the time that reference property schema will be defined statically and not contain a dynamicUrl so I don't see what's dynamic about that case.

schema.ts
patrick-codaio: Jonathan makes a good point... i'm ok with these being separate.

Though to clarify my original argument - I wasn't necessarily proposing that we incur the 100k ingestion fanout problem. Instead of that, I was imagining an additional `getSubItems` in addition to `getPermissions`, `execute`, etc on a sync table.

alexd-codaio:
Chatted offline with @jonathan-codaio and updated to something more palatable/understandable.

* Removed the confusing static/dynamic definitions (which were wrong anyway) and updated them with a clearer differentiation.
* Made `parentIdentity` a common field across both flavors
* Added enums for all the places we had bools.

Updated - PTAL @jonathan-codaio @patrick-codaio

schema.ts
```diff
@@ -1135,6 +1134,103 @@ export interface IndexDefinition {
...
+   *
+   * If this property references a property with {@link ValueType.Object} and
+   * {@link ValueHintType.Reference}, then it refers to an object in another logical table.
+   *
+   * Otherwise, it refers to an object in the same table.
+   */
+  parentIdProperty: PropertyIdentifier<string>;
+}
+
+export interface ParentChildParameterMapping {
```
patrick-codaio, line 1178: you're saying that the _output_ schema of a table will define how the _inputs_ of itself should be populated? that seems kinda loopy - outputs should depend on inputs generally. cc @jonathan-codaio

schema.ts
```diff
@@ -1135,6 +1134,103 @@ export interface IndexDefinition {
...
+export interface CrawlableParentDefinition extends BaseParentDefinition {
+  /**
+   * Determines how the parent object should be crawled for children.
+   */
+  crawling: CrawlingBehavior;
+
+  /**
+   * Determines how the lifecycle of the child objects should be handled.
+   */
+  lifecycle: LifecycleBehavior;
```
jonathan-codaio, line 1220: I thought one of the motivations of subitems is that `lifecycle` would apply there too, so that parent deletes could be cascaded?

schema.ts
jonathan-codaio: Sorry can you say a bit more? Is the issue that in a dynamic table case, the output of getSchema will influence how the parameters are interpreted, when normally the parameter values themselves can be used as inputs to `getSchema`? That does seem confusing. But in another perspective, should dynamic tables be allowed to participate in crawling? If so, is there any other way to specify that other than waiting until getSchema in order to express the parent identity using a dynamicUrl?

schema.ts
alexd-codaio: I wasn't planning on building that support out of the gate, so didn't want to put it in the SDK.

schema.ts
patrick-codaio: It's true that we haven't yet had a need to support dynamic tables in our crawls, but I'm nervous to make that harder to support. The current pattern of having the crawl parent defined on the parameters can work just fine for dynamic tables, once we bother to support it on the `coda` side.

Maybe we should think of this like identities - internally we modeled identities as part of the output schema, but later needed to allow them to be set at the table level via `identityName`. We even made `identityName` required. Maybe this whole `parent table` concept should not be under _either_ parameters or output schema, but rather be its own top level concept. cc @ekoleda-codaio

alexd-codaio: I understand the concern re inversion here, but I think it only really matters in the context of dynamic tables (otherwise schema is staticish). And I don't really understand how crawling would work for dynamic tables anyway. (If there was a way of doing dynamic table crawling, I think it'd have to be a new thing we build).

patrick-codaio: I'm pretty concerned even outside of dynamic sync tables. I made a catalyst for tomorrow to give us time to talk things through - https://coda.io/d/_dCum5Z63Pkj#Topics_tucMmc8a/r1935&view=full

schema.ts
```diff
@@ -1135,6 +1134,103 @@ export interface IndexDefinition {
...
+   * The object will inherit permissions from its parent.
+   */
+  Inherit = 'Inherit',
+}
+
+export interface BaseParentDefinition {
+  /**
+   * The identity of the parent object.
+   */
+  parentIdentity: IdentityDefinition;
```
ekoleda-codaio, line 1151: Is this duplicative with `parentIdProperty`? That property would already include the reference to the other table.

api_types.ts
```diff
@@ -382,6 +382,7 @@ export interface ParamDef<T extends UnionType> {
   allowManualInput?: boolean;
 
   // TODO(patrick): Unhide this
+  /** @deprecated use `ParentDefinition` instead */
```
jonathan-codaio, line 385: This won't be considered deprecated anymore.

api_types.ts
```diff
@@ -449,6 +450,7 @@ export type SuggestedValueType<T extends UnionType> = T extends ArrayType<Type.d
   : TypeOfMap<T>;
 
 // TODO(patrick): Unhide this
+/** @deprecated use `ParentDefinition` instead */
```
jonathan-codaio, line 453: Ditto. But SyncTableRelation.inheritPermissions is now deprecated.

jonathan-codaio (commented):
Overall LGTM

schema.ts
alexd-codaio: Yes, deleted

api_types.ts
alexd-codaio: Done.

jonathan-codaio (approved):

api_types.ts
jonathan-codaio: Feel free to mark ` SyncTableRelation.inheritPermissions` as deprecated though unless you're waiting.

alexd-codaio: Done.

PR #3160, opened 2025-02-19
Add new metadata field for SyncFormulaResult to pass to executeGetPer…

drew-codaio:
…missions

Sdk piece to solving this https://codaproject.atlassian.net/browse/CB-2573. 
We want to add a new property that can be passed in memory from the execute call to executeGetPermissions.


api.ts
```diff
@@ -1035,12 +1037,21 @@ export function isSyncPackFormula(fn: BaseFormula<ParamDefs, any>): fn is Generi
...
   /** The list of rows from this page. */
   result: Array<ObjectSchemaDefinitionType<K, L, SchemaT>>;
 
+  /**
+   * The metadata for each row in the result.
+   * This field is used to pass in metadata during the executeGetPermissions call
+   * TODO(drew): Unhide this
+   * @hidden
+   */
+  metadata?: SyncTableMetadata;
```
jonathan-codaio, line 1053: Nit, since this sits side by side with `result`, which is an `Array` type, I think it'll be more readable to type this as an `Array` here, so maybe we only give a name to the item type.

And naming is always hard, I wonder if should call this "pasthrough data" or something to hint about how it actually get used.

And then as discussed if we give it a generic then pack authors can give it a specific type.

```ts
export type SyncPassthroughData = Record<string, any>;

...

export interface SyncFormulaResult<
  K extends string,
  L extends string,
  ContextT extends SyncExecutionContext<any> = SyncExecutionContext,
  PassthroughT extends SyncPassthroughData = SyncPassthroughData,
>
result: Array<ObjectSchemaDefinitionType<K, L, SchemaT>>;
passthroughData? = Array<PassthroughT>;
...
```

jonathan-codaio (commented):
CC @ekoleda-codaio

api.ts
```diff
@@ -1037,10 +1039,19 @@ export interface SyncFormulaResult<
...
   SchemaT extends ObjectSchemaDefinition<K, L>,
   ContextT extends SyncExecutionContext<any> = SyncExecutionContext,
+  PassthroughT extends SyncPassthroughData = SyncPassthroughData,
 > {
   /** The list of rows from this page. */
   result: Array<ObjectSchemaDefinitionType<K, L, SchemaT>>;
 
+  /**
+   * Some additional data that we don't  for each row in the result.
+   * This field is used to pass through data during the executeGetPermissions call
```
jonathan-codaio, line 1049: Would probably do something like:

```
Additional side data to pass through to `executeGetPermissions`. Because each result item
is also passed through, this is only needed for cases where you have data that doesn't need
to be part of the result item but is nonetheless useful for fetching permissions, for example,
the API requests to get item data may include permissions inline, so there is no need to
re-fetch permission data if you can just pass it through.

This array must be the same length as the `result` array.
```

api.ts
```diff
@@ -2470,6 +2507,9 @@ export function makeSyncTable<
       returnValue.deletedRowIds = deletedRowIds ?? deletedItemIds;
       returnValue.deletedItemIds = returnValue.deletedRowIds;
     }
+    if (passthroughData) {
```
jonathan-codaio, line 2510: I don't know why the existing code wasn't just passing those values through, these conditionals seem like no-ops...

testing/execution.ts
```diff
@@ -267,13 +268,16 @@ export async function executeFormulaOrSyncFromCLI({
...
+        passthroughData.push(...response?.passthroughData ?? []);
         executionContext.sync.continuation = response.continuation;
         iterations++;
       } while (executionContext.sync.continuation && result.length < maxRows);
       if (result.length > maxRows) {
         result = result.slice(0, maxRows);
+        passthroughData = passthroughData.slice(0, maxRows);
       }
       printFull(result);
+      printFull(passthroughData);
```
jonathan-codaio, line 280: Should only print this if there is actually any passthrough data, otherwise all sync tables run this way will all of the sudden print an empty array at the end of execution.

FYI @zack-codaio this will be a merge conflict with your change, but should be easy to resolve.

In both cases, since we have more output to print, we should think about whether we should be printing a composite object instead of separate objects on separate log lines. The reason is that if you print a valid composite object, you can pipe the output to `jq` to inspect the output. CC @ekoleda-codaio 

This is fine for this PR though.

testing/execution.ts
```diff
@@ -267,13 +268,16 @@ export async function executeFormulaOrSyncFromCLI({
           : await executeFormulaOrSyncWithRawParams({formulaSpecification, params, manifest, executionContext});
 
         result.push(...response.result);
+        passthroughData.push(...response?.passthroughData ?? []);
```
jonathan-codaio, line 271: Probably good to assert that the passthrough data array has the same length as the result array since we require that.

api.ts
drew-codaio: 🤷 yeah I had no idea why it wasn't working but adding this let me plumb the values through.

testing/execution.ts
drew-codaio: How do I go about asserting this at runtime?

api.ts
```diff
@@ -389,12 +391,20 @@ export interface Continuation {
   [key: string]: string | number | {[key: string]: string | number};
 }
 
+/**
+ * Type definition for some additional data that is returned by a sync table
+ * in addition to the data itself. This data is not stored in Coda, but
+ * is passed to the executeGetPermissions function of the sync table
+ * See {@link SyncFormulaResult.passthroughData}.
+ */
+export type SyncPassthroughData = Record<string, any>;
```
drew-codaio, line 400: @jonathan-codaio I had to unhide this object to get past the documentation errors. Let me know if there's a better way to handle this or if we still want to keep it hidden.

testing/execution.ts
jonathan-codaio: Just right above line 270 something like:

```ts
if (response.passthroughData && response.passthroughData.length !== response.result.length) {
  throw new Error(`Got ${response.result.length} results but only ${response.passthroughData.length} passthrough items (on page ${iteration})`)
}
```

api.ts
jonathan-codaio: Ah hmm yeah I'm not able to get this completely working either.

Can you just add `'SyncPassthroughData'` as an entry in `typedoc.js` in the `intentionallyNotExported` list? Would create a new little section with a comment like `// In-development types that we're not ready to expose yet`

jonathan-codaio: (After hiding all the things again)

testing/execution.ts
drew-codaio: Ah is this the same function that runs the packs code? I didn't realize since it was under /testing.

jonathan-codaio: This is just what runs at runtime in the CLI simulator (`coda execute`). There's conceptually similar code in process_csb_ingestion.ts where we'd want to check this as well, but for now this is just about making sure the simulator makes the assertion.

jonathan-codaio (approved):

patrick-codaio (approved):
LGTM other than Jonathan's comments and any feedback Eric has about the name "passthrough data"

testing/execution.ts
ekoleda-codaio: If we don't print the continuation or completion data, perhaps we shouldn't be printing this either? I could see a case for printing the entire response object, but just selecting this one additional part seems inconguous.

ekoleda-codaio:
After some consideration, I'm not in love with this design. The feature is so narrow (just passing context between `execute` and `executeGetPermissions`) but the naming and placement is fairly broad. Some alternative approaches that come to mind (in order of difficulty):

- Utilize the fetcher cache to avoid making duplicate requests. We normally advise Pack makers to do this when needing to cache data between continuations, etc. No SDK changes needed.
- Allow returning permissions in the `execute` result. If this is the only use case we have in mind, then it would be clearer if we narrowly scoped it to permissions.
- Allow the schema to declare hidden properties. This allows us to reuse all the existing property infrastructure for passing data around without cluttering the user space. May also be useful for `executeUpdate`.
- Introduce a more explicit caching service. Developers have asked for it, and Apps Script has it. It could be scoped to a specific sync.

If we are going to stick with the current design, I'd probably recommend renaming it from `passThroughData` to something like `permissionsContext` or `permissionsContinuation`. This makes it more clear what the scope and purpose of the containing data is.

testing/execution.ts
drew-codaio: Good point. I didn't realize what this function was doing. Let me move my test and test it differently.

jonathan-codaio:
> After some consideration, I'm not in love with this design. The feature is so narrow (just passing context between `execute` and `executeGetPermissions`) but the naming and placement is fairly broad. Some alternative approaches that come to mind (in order of difficulty):
> 
> * Utilize the fetcher cache to avoid making duplicate requests. We normally advise Pack makers to do this when needing to cache data between continuations, etc. No SDK changes needed.
> * Allow returning permissions in the `execute` result. If this is the only use case we have in mind, then it would be clearer if we narrowly scoped it to permissions.
> * Allow the schema to declare hidden properties. This allows us to reuse all the existing property infrastructure for passing data around without cluttering the user space. May also be useful for `executeUpdate`.
> * Introduce a more explicit caching service. Developers have asked for it, and Apps Script has it. It could be scoped to a specific sync.
> 
> If we are going to stick with the current design, I'd probably recommend renaming it from `passThroughData` to something like `permissionsContext` or `permissionsContinuation`. This makes it more clear what the scope and purpose of the containing data is.

Great feedback as always. My unwritten thought process was that it's not inconceivable that we have more execute* methods in the future which also may need passthrough data, so while we are building this for permissions it might be useful to consider it an extensible feature. That said, that's all hypothetical, and so I'm fine that if we keep this architecture we rename it to be clearer.

As for the solid list of alternatives here:

* For fetcher caching, for cases like Drive DWD, I'm not sure it's trivial to arrive at the same set of fetcher requests that got the row data in first place. Certainly it requires a lot of effort to reuse the code in a way where you can make use of the fetcher cache,
* Returning permissions inline seems very attractive to me, but studying the use case here, for Drive and Dropbox, the case is that _some_ permissions can be returned inline, but not all. Even though you can't eliminate all remote requests for permissions, it's common enough that there's significant savings to be realized here. But you can't eliminate `executeGetPermissions` entirely, and I think you would still require passthrough-info to determine there which permissions were returned inline and which still need to be retrieved.
* Hidden schema properties: this works but I find the concept strange. Conceptually what does it mean to return hidden data?
* Generalized caching: this seems worth building, but (1) I think would require some threat modeling to do responsibly, and (2) I don't think escapes the need for passthrough data. If I cache useful info to reuse in `executeGetPermissions`, I need to communicate the cache key or something from which I can derive the cache key. You could say that the cache key is a hash of the item ids in order, perhaps. But I think the intuitive interface would be that pack can make up its own cache keys and pass them around as needed, and outside of continuations that would imply a need for something related to passthrough data.

So I don't see an obvious way to escape something shaped like this. I'd be fine with renaming it `permissionsContext` for clarity.

ekoleda-codaio:
> Returning permissions inline seems very attractive to me, but studying the use case here, for Drive and Dropbox, the case is that some permissions can be returned inline, but not all. Even though you can't eliminate all remote requests for permissions, it's common enough that there's significant savings to be realized here. But you can't eliminate executeGetPermissions entirely, and I think you would still require passthrough-info to determine there which permissions were returned inline and which still need to be retrieved.

True, but we could keep the same basic flow but just enforce that things being passed through are permissions:

```ts
return {
  result: rows,
  rowAccessDefinitions: acls,
}
```

```ts
executeUpdate(args, request, context) {
  const rowRequests = request.rows;
  const rowAccessDefinitions = request.rowAccessDefinitions;
  // Fetch missing row access definitions...
}
```

It wouldn't help in the case where there is some metadata we've already synced that, while not a permission itself, would help us to more easily fetch or compute the permissions later on.

If we do think other `execute*` methods are likely to be added in the future, further enriching the synced rows, then maybe a generic name is OK. 

> Hidden schema properties: this works but I find the concept strange. Conceptually what does it mean to return hidden data?

What I'm drawn to is co-locating all the data about a row, rather than having an array of N rows and then an array of N pass throughs. We already support passing additional data in the result objects outside of the schema definition, it just gets thrown away. What if we could somehow tell Coda "Don't throw away these keys, at least not until we're done processing the row."

jonathan-codaio:
> True, but we could keep the same basic flow but just enforce that things being passed through are permissions:
> 
> ```ts
> return {
>   result: rows,
>   rowAccessDefinitions: acls,
> }
> ```
> 
> ```ts
> executeUpdate(args, request, context) {
>   const rowRequests = request.rows;
>   const rowAccessDefinitions = request.rowAccessDefinitions;
>   // Fetch missing row access definitions...
> }
> ```

This seems to me like it would work, but I'm asking here in the original request thread to confirm:

https://codaproject.slack.com/archives/C06QGU1AAN9/p1740178716145029?thread_ts=1724872359.991309&cid=C06QGU1AAN9

drew-codaio:
@ekoleda-codaio just a heads up. We decided on having this PR continue in the direction of passing through some data. It seems like we will always need to call executeGetPermissions so returning permissions inline won't save us much work. Also there are potential issues with the size of the data returned from permissions (1 row has many permissions). We can also extend the PassthroughData type for future executions if needed.

jonathan-codaio:
> @ekoleda-codaio just a heads up. We decided on having this PR continue in the direction of passing through some data. It seems like we will always need to call executeGetPermissions so returning permissions inline won't save us much work. Also there are potential issues with the size of the data returned from permissions (1 row has many permissions). We can also extend the PassthroughData type for future executions if needed.

Another tipping point for me was the in order to support the one use case that we have, with Google Drive and other file-based packs, is that you have to pass the inline permission as inputs into executeGetPermissions in order to be able to know which items you need to fetch permissions for and which are already handled, and since permissions are typically meant as an _output_ they are largely unbounded in quantity, so it could be problematic to have to provide them as unpaginated inputs. The context/passthrough data is enforced to have the same cardinality as the result page length.

PR #3161, opened 2025-02-19
Update elliptic due to critical vulnerability

neal-codaio:
This should probably be released ASAP if this affects us.  Let me know if you want me to update the release version in here or do anything else.  

Note: there are other moderate octokit issues as well that should be addressed.  There are problems upgrading the `coda` repo which might also be problems here.

PTAL @jonathan-codaio @patrick-codaio

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,7 +150,8 @@
       "eslint-plugin-local@1.0.0": "patches/eslint-plugin-local@1.0.0.patch"
     },
     "overrides": {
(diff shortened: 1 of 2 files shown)


jonathan-codaio:
Thanks. I don't think this affects us at all, we don't use elliptic and I don't even think it gets used transitively. It seems to be getting pulled in via a bunch of browserify stuff, and we don't even use the signing that browserify-sign would presumably be shimming.

PR #3162, opened 2025-02-19
Call executeGetPermissions with actual sync from cli

zack-codaio:
This PR adds the ability to get rowAccessDefinitions for a sync table from the CLI.

Previously, we had the ability to explicitly fetch permissions for rows, but you had to give the rows in a json parameter. This mode of Sync will instead do the sync fetching, pipe those into GetPermissions, and then return the result.

Screencast: https://coda.d.pr/v/3H3gwW

### DIFF
--- a/cli/execute.ts
+++ b/cli/execute.ts
@@ -36,6 +36,7 @@ export async function handleExecute({
         'compatible with your platform. Try again but omitting the --vm flag.',
     );
   }
+
   const fullManifestPath = makeManifestFullPath(manifestPath);
   const {bundlePath, bundleSourceMapPath} = await compilePackBundle({
     manifestPath: fullManifestPath,
@@ -44,7 +45,9 @@ export async function handleExecute({
   });
   const manifest = await importManifest(bundlePath);
   if (manifest.networkDomains && manifest.networkDomains.length > 1 && !allowMultipleNetworkDomains) {
-    return printAndExit('Using multiple network domains requires approval from Coda. Visit https://coda.io/packs/build/latest/support#approvals to make a request. Disable this warning by including the flag: --allowMultipleNetworkDomains');
+    return printAndExit(
+      'Using multiple network domains requires approval from Coda. Visit https://coda.io/packs/build/latest/support#approvals to make a request. Disable this warning by including the flag: --allowMultipleNetworkDomains',
+    );
   }
   await executeFormulaOrSyncFromCLI({
     formulaName,
--- a/runtime/types.ts
+++ b/runtime/types.ts
@@ -97,6 +97,8 @@ export type FormulaSpecification =
   | PropertyAutocompleteFormulaSpecification
   | GetPermissionsFormulaSpecification;
 
+export type ChainedCommandFormulaSpecification = GetPermissionsFormulaSpecification;
+
 export type PackFunctionResponse<T extends FormulaSpecification> = T extends SyncFormulaSpecification
   ? GenericSyncFormulaResult
   : T extends SyncUpdateFormulaSpecification
--- a/test/execution_test.ts
+++ b/test/execution_test.ts
@@ -504,6 +504,67 @@ describe('Execution', () => {
           }
         });
 
+        it('sync with includePermissions works', async () => {
+          await executeFormulaOrSyncFromCLI({
+            vm,
+            formulaName: 'Students>permissions',
+            params: ['Cunningham'],
+            manifest: fakePack,
+            manifestPath: '',
+            bundleSourceMapPath,
+            bundlePath,
+            contextOptions: {useRealFetcher: false},
+          });
+          const result = mockPrintFull.args[0][0];
+
+          assert.deepEqual(
+            result,
+            [
+              {name: 'Albert'},
+              {name: 'Brenda'},
+              {name: 'Cory'},
+              {name: 'Dylan'},
+              {name: 'Ethan'},
+              {name: 'Fiona'},
+              {name: 'Gina'},
+              {name: 'Hank'},
+              {name: 'Ivy'},
+              {name: 'Jack'},
+              {name: 'Kyle'},
+              {name: 'Liam'},
+              {name: 'Mia'},
+              {name: 'Noah'},
+              {name: 'Olivia'},
+              {name: 'Pam'},
+              {name: 'Quinn'},
+              {name: 'Ryan'},
+              {name: 'Sam'},
+              {name: 'Tia'},
+              {name: 'Uma'},
+              {name: 'Vince'},
+              {name: 'Wendy'},
+              {name: 'Xavier'},
+              {name: 'Yara'},
+              {name: 'Zack'},
+              {name: 'Aaron'},
+              {name: 'Bella'},
+              {name: 'Charlie'},
+              {name: 'Diana'},
+              {name: 'Easton'},
+              {name: 'Frank'},
+              {name: 'Greg'},
+              {name: 'Hannah'},
+              {name: 'Ian'},
+              {name: 'Julia'},
+            ].map(student => {
+              return {
+                rowId: student.name,
+                permissions: [{permissionType: PermissionType.Direct, principal: {type: 'user', userId: 1}}],
+              };
+            }),
+          );
+        });
+
         it('sync update works', async () => {
           const syncUpdates: GenericSyncUpdate[] = [
             {previousValue: {name: 'Alice'}, newValue: {name: 'Alice Smith'}, updatedFields: ['name']},
@@ -632,7 +693,7 @@ describe('Execution', () => {
           assert.equal(exitValue, 1);
         });
 
-        it('CLI errors don\'t print stacktrace', async () => {
+        it("CLI errors don't print stacktrace", async () => {
           await executeFormulaOrSyncFromCLI({
             vm,
             formulaName: 'NotRealFormula',
--- a/testing/execution.ts
+++ b/testing/execution.ts
@@ -1,4 +1,5 @@
 import type {BasicPackDefinition} from '../types';
+import type {ChainedCommandFormulaSpecification} from '../runtime/types';
 import type {Credentials} from './auth_types';
 import type {ExecutionContext} from '../api_types';
 import type {FormulaSpecification} from '../runtime/types';
@@ -26,6 +27,7 @@ import type {SyncMetadataFormulaSpecification} from '../runtime/types';
 import type {SyncUpdateFormulaSpecification} from '../runtime/types';
 import type {TypedPackFormula} from '../api';
 import type {UpdateSyncExecutionContext} from '../api_types';
+import {chunkArray} from './helpers';
 import {coerceParams} from './coercion';
 import {ensureExists} from '../helpers/ensure';
 import {ensureUnreachable} from '../helpers/ensure';
@@ -66,7 +68,6 @@ export interface ContextOptions {
   useRealFetcher?: boolean;
   manifestPath?: string;
 }
-
 function resolveFormulaNameWithNamespace(formulaNameWithNamespace: string): string {
   const [namespace, name] = formulaNameWithNamespace.includes('::')
     ? formulaNameWithNamespace.split('::')
@@ -245,38 +246,20 @@ export async function executeFormulaOrSyncFromCLI({
       : newMockSyncExecutionContext();
     executionContext.sync.dynamicUrl = dynamicUrl || undefined;
 
-    const formulaSpecification = makeFormulaSpec(manifest, formulaName);
+    const {formulaSpecification, chainedCommand} = getFormulaSpecAndChainedCommand(manifest, formulaName);
 
     if (formulaSpecification.type === FormulaType.Sync) {
-      let result = [];
-      let iterations = 1;
-      do {
-        if (iterations > MaxSyncIterations) {
-          throw new Error(
-            `Sync is still running after ${MaxSyncIterations} iterations, this is likely due to an infinite loop.`,
-          );
-        }
-        const response: GenericSyncFormulaResult = vm
-          ? await executeFormulaOrSyncWithRawParamsInVM({
-              formulaSpecification,
-              params,
-              bundleSourceMapPath,
-              bundlePath,
-              executionContext,
-            })
-          : await executeFormulaOrSyncWithRawParams({formulaSpecification, params, manifest, executionContext});
-
-        if (response.permissionsContext && response.permissionsContext.length !== response.result.length) {
-          throw new Error(`Got ${response.result.length} results but only ${response.permissionsContext.length} passthrough items (on page ${iterations})`)
-        }
-        result.push(...response.result);
-        executionContext.sync.continuation = response.continuation;
-        iterations++;
-      } while (executionContext.sync.continuation && result.length < maxRows);
-      if (result.length > maxRows) {
-        result = result.slice(0, maxRows);
-      }
-      printFull(result);
+      const {result, chainedCommandResults} = await executeSyncFormulaWithContinuations({
+        formulaSpecification,
+        chainedCommand,
+        params,
+        manifest,
+        executionContext,
+        vm,
+        bundleSourceMapPath,
+        bundlePath,
+      });
+      printFull(chainedCommand ? chainedCommandResults : result);
     } else {
       const result = vm
         ? await executeFormulaOrSyncWithRawParamsInVM({
@@ -332,6 +315,42 @@ function invert<K extends string | number | symbol, V extends string | number | 
   obj: Record<K, V>,
 ): Record<V, K> {
   return Object.fromEntries(Object.entries(obj).map(([key, value]) => [value, key]));
+}
+
+/**
+ * Given a formula name with a > delimited chained command, returns the formula specification and the chained command.
+ *
+ * @param manifest The manifest of the pack.
+ * @param formulaNameInput The formula name with a chained command.
+ * @returns The formula specification and the chained command.
+ */
+function getFormulaSpecAndChainedCommand(
+  manifest: BasicPackDefinition,
+  formulaNameInput: string,
+): {
+  formulaSpecification: FormulaSpecification;
+  chainedCommand?: ChainedCommandFormulaSpecification;
+} {
+  const chainedCommands = formulaNameInput.split('>');
+  const formulaSpecification = makeFormulaSpec(manifest, chainedCommands[0]);
+
+  if (chainedCommands.length === 1) {
+    return {formulaSpecification};
+  }
+
+  if (formulaSpecification.type !== FormulaType.Sync) {
+    throw new Error(`Chained commands are only supported for sync formulas. Received: ${formulaSpecification.type}`);
+  }
+
+  const chainedCommand = makeFormulaSpec(manifest, [formulaSpecification.formulaName, chainedCommands[1]].join(':'));
+
+  if (chainedCommand.type !== FormulaType.GetPermissions) {
+    throw new Error(
+      `Chained commands are only supported for GetPermissions formulas. Received: ${chainedCommand.type}`,
+    );
+  }
+
+  return {formulaSpecification, chainedCommand};
 }
 
 // Exported for tests.
@@ -407,6 +426,7 @@ export function makeFormulaSpec(manifest: BasicPackDefinition, formulaNameInput:
         formulaName: formulaOrSyncName,
       };
     }
+
     const metadataFormulaType = invert(SyncMetadataFormulaTokens)[metadataFormulaTypeStr];
     if (!metadataFormulaType) {
       throw new Error(`Unrecognized metadata formula type "${metadataFormulaTypeStr}".`);
@@ -464,7 +484,12 @@ export async function executeFormulaOrSyncWithVM<T extends PackFormulaResult | G
   const ivmContext = await ivmHelper.setupIvmContext(bundlePath, executionContext);
 
   try {
-    return await executeThunk(ivmContext, {params, formulaSpec: formulaSpecification}, bundlePath, bundlePath + '.map') as T;
+    return (await executeThunk(
+      ivmContext,
+      {params, formulaSpec: formulaSpecification},
+      bundlePath,
+      bundlePath + '.map',
+    )) as T;
   } catch (err: any) {
     throw new DeveloperError(err);
   }
@@ -587,7 +612,6 @@ export async function executeFormulaOrSyncWithRawParams<T extends FormulaSpecifi
       ({params, permissionRequest} = parseGetPermissionRequest(manifest, formulaSpecification, rawParams));
       break;
     }
-
     default:
       ensureUnreachable(formulaSpecification);
   }
@@ -645,7 +669,7 @@ export async function executeSyncFormula(
     }
   }
   const result = [];
-  const permissionsContext = []
+  const permissionsContext = [];
   const deletedRowIds = [];
   let iterations = 1;
   do {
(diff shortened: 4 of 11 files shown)


testing/execution.ts
```diff
@@ -274,6 +275,38 @@ export async function executeFormulaOrSyncFromCLI({
         result = result.slice(0, maxRows);
       }
       printFull(result);
+    } else if (formulaSpecification.type === FormulaType.ExecuteGetPermissions) {
```
zack-codaio, line 278: this is mostly duplicative with the standard Sync loop, could possibly be deduped

testing/execution.ts
```diff
@@ -274,6 +275,38 @@ export async function executeFormulaOrSyncFromCLI({
...
+      let result = [];
+      let iterations = 1;
+      do {
+        if (iterations > MaxSyncIterations) {
+          throw new Error(
+            `Sync is still running after ${MaxSyncIterations} iterations, this is likely due to an infinite loop.`,
+          );
+        }
+        const response: any = vm
```
zack-codaio, line 287: need to revisit the type here, slapped an any on to run the tests

test/execution_test.ts
```diff
@@ -542,6 +542,90 @@ describe('Execution', () => {
           });
         });
 
+        it.only('read permissions for executed syncTable works -- chained, not actually what we want', async () => {
```
zack-codaio, line 545: won't actually keep this test case, just started with this to see how the tests run and to see wha the expected output would be

test/execution_test.ts
```diff
@@ -542,6 +542,90 @@ describe('Execution', () => {
...
+              },
+              {
+                rowId: 'Diana',
+                permissions: [{permissionType: PermissionType.Direct, principal: {type: 'user', userId: 1}}],
+              },
+            ],
+          });
+        });
+
+        it.only('read permissions for executed syncTable works -- merged, what we actually want', async () => {
```
zack-codaio, line 595: remove .only calls

runtime/types.ts
```diff
@@ -9,6 +9,7 @@ export enum FormulaType {
   Sync = 'Sync',
   SyncUpdate = 'SyncUpdate',
   GetPermissions = 'GetPermissions',
+  ExecuteGetPermissions = 'ExecuteGetPermissions',
```
zack-codaio, line 12: probably not the best name. Maybe SyncReturnPermissions or something? I kept it in a the same pattern as GetPermissions, rather than just using Sync with a separate mode, but can rework to whichever pattern is preferred

testing/execution.ts
jonathan-codaio: So I don't think I'd actually introduce a new formula type, I do think this really just is a modality of `execute`. Maybe start by adding a new optional property to `ContextOptions` like 

```ts
interface ContextOptions {
  syncOptions?: {
    includePermissions?: boolean;
  }
}
```

Then if that's set, in the existing sync loop above, can just make an extra call to get the permissions. Can ignore how that gets set for now but probably we just want to introduce a `--permissions` flag that populates it.

I'd probably also take the opportunity to split out the sync logic (lines 252-277) into a separate function since it'll get more complicated. For the final version of this, we'll want to respect the `maxPermissionBatchSize` and have an inner loop for getting permissions, no need for that in the next draft of the PR but will want that before merging and it will add enough lines that it'll be nice to have a standalone function for all this.

runtime/types.ts
jonathan-codaio, line 12: (I think things are still in-flight here and this is on your radar, but in the final version we'd want to revert adding a new FormulaType, and only rely on contextOptions for the new behavior.)

cli/coda.ts
```diff
@@ -72,6 +72,11 @@ if (require.main === module) {
           default: false,
           desc: 'Allow executing Packs that use multiple network domains. You must get approval from Coda before you can upload these Packs.',
         },
+        includePermissions: {
+          boolean: true,
+          default: false,
+          desc: 'Include permissions in the sync table response.',
```
zack-codaio, line 78: not entirely accurate here, this is really a toggle to show the row access definitions *instead* of the sync rows.

We could also instead return the rows AND the permissions, or have a mode of rows | permissions | rowsAndPermissions

testing/execution.ts
```diff
@@ -951,3 +941,134 @@ class DeveloperError extends Error {
     Object.setPrototypeOf(this, DeveloperError.prototype);
   }
 }
+
+/**
```
zack-codaio, line 945: needs jsdocs here and below

testing/execution.ts
```diff
@@ -951,3 +941,134 @@ class DeveloperError extends Error {
...
+
+  const response: any = vm
+    ? await executeFormulaOrSyncWithRawParamsInVM<GetPermissionsFormulaSpecification>({
+        formulaSpecification: getPermissionsFormulaSpecification,
+        params: getPermissionsParams,
+        bundleSourceMapPath,
+        bundlePath,
+        executionContext,
+      })
+    : await executeFormulaOrSyncWithRawParams<GetPermissionsFormulaSpecification>({
```
zack-codaio, line 1066: not sure if this is how we want to call this, or if we should be doing something like findSyncFormula > formula.executeGetPermissions like we do inside of thunk.ts

cli/coda.ts
patrick-codaio: Let's start with just one mode (I'm open minded about which one), but update the name so that we can eventually support all 3 of those modes.

jonathan-codaio: Yeah, maybe `permissionMode` or something then, with options being `none` (default)` and `include` to start or something.

cli/execute.ts
```diff
@@ -44,7 +47,9 @@ export async function handleExecute({
   });
   const manifest = await importManifest(bundlePath);
   if (manifest.networkDomains && manifest.networkDomains.length > 1 && !allowMultipleNetworkDomains) {
-    return printAndExit('Using multiple network domains requires approval from Coda. Visit https://coda.io/packs/build/latest/support#approvals to make a request. Disable this warning by including the flag: --allowMultipleNetworkDomains');
+    return printAndExit(
+      'Using multiple network domains requires approval from Coda. Visit https://coda.io/packs/build/latest/support#approvals to make a request. Disable this warning by including the flag: --allowMultipleNetworkDomains',
```
jonathan-codaio, line 51: Ah, I guess we have enforced standardized `prettier` formatting in the `packs` repo now, but still not here... For later.

testing/execution.ts
```diff
@@ -951,3 +941,151 @@ class DeveloperError extends Error {
...
+      );
+    }
+    executionContext.sync.continuation = response.continuation;
+    iterations++;
+  } while (executionContext.sync.continuation && result.length < maxRows);
+  if (result.length > maxRows) {
+    result = result.slice(0, maxRows);
+  }
+  if (rowAccessDefinitions.length > maxRows) {
+    rowAccessDefinitions = rowAccessDefinitions.slice(0, maxRows);
```
jonathan-codaio, line 1030: Hmm, I don't think I'd do this. The relationship of permissions to rows is many:1 so we could easily exceed the row count with permissions and I don't think we should arbitrarily truncate.

I think maxRows is just meant to be a safety stop for the sync loop, so I think we can ignore it here.

testing/execution.ts
jonathan-codaio: That's a good question. I think the way you have it is reasonable and possibly optimal.

testing/execution.ts
```diff
@@ -951,3 +941,151 @@ class DeveloperError extends Error {
...
+  bundlePath: string;
+}) {
+  const getPermissionsFormulaSpecification: GetPermissionsFormulaSpecification = {
+    ...formulaSpecification,
+    type: FormulaType.GetPermissions,
+  };
+  const mappedRows = rows.map(row => ({row}));
+  const getPermissionsParams = [...params, JSON.stringify({rows: mappedRows})];
+
+  const response: any = vm
```
jonathan-codaio, line 1075: Again no need to block this PR, but eventually this should do a loop where the rows are batched into arrays of size `maxPermissionBatchSize`.

jonathan-codaio (commented):
Awesome, looks really good.

testing/execution.ts
zack-codaio, line 1075: also an `any` i need to figure out an actual type for

cli/coda.ts
ekoleda-codaio: Is there some way to make this feel more akin to our existing method of running metadata functions and `executeUpdate`?

https://coda.io/packs/build/latest/guides/development/cli/#running-metadata-functions
https://coda.io/packs/build/latest/guides/development/cli/#two-way-sync

Perhaps something like:

`npx coda execute MySyncTable:permissions`

zack-codaio: So originally I had it as `MySyncTable:executeGetPermissions`. There actually already is `MySyncTable:permissions`, but that requires passing in the JSON set of rows as a parameter from the command line.

This new version is different in that it's actually running the Sync formula, but just adjusting the output that returns to be the permissions for those Sync rows, rather than just executing executeGetPermissions on an arbitrary set of input rows.

I don't feel strongly though, I'm still learning so happy to defer to whatever you, @jonathan-codaio and @patrick-codaio  think is best.

test/execution_test.ts
```diff
@@ -495,6 +495,40 @@ describe('Execution', () => {
...
+            params: ['Smith'],
+            manifest: fakePack,
+            manifestPath: '',
+            bundleSourceMapPath,
+            bundlePath,
+            contextOptions: {useRealFetcher: false, syncOptions: {includePermissions: true}},
+          });
+          const result = mockPrintFull.args[0][0];
+          assert.deepEqual(result, [
+            [
```
zack-codaio, line 511: this gets an extra array level due to chunking

cli/coda.ts
patrick-codaio: cc @drew-codaio since he's touching the inputs to GetPermissions

jonathan-codaio: Yeah as noted we already support the traditional way of `npx coda execute MySyncTable:getPermissions` and it's great if makers are able to make use of that when they want to exercise that in isolation.

In practice, it's so gnarly both to construct payloads and to express them on the CLI, that I think it's big barrier to actually developing this way. So I think it will be much more desirable to just be able to use `coda execute` which will produce real response payloads and let the SDK handle passing them into executeGetPermissions. I'm pretty confident this will become the preferred way to develop.

test/execution_test.ts
jonathan-codaio: Ah sorry, you want to chunk the inputs to executeGetPermissions, not the outputs. You want to chunk the sync result rows before passing them into executeGetPermissions, and call it multiple times, once per chunk.

testing/execution.ts
```diff
@@ -951,3 +943,153 @@ class DeveloperError extends Error {
...
+    }
+    executionContext.sync.continuation = response.continuation;
+    iterations++;
+  } while (executionContext.sync.continuation && result.length < maxRows);
+  if (result.length > maxRows) {
+    result = result.slice(0, maxRows);
+  }
+
+  const formula = findSyncFormula(manifest, formulaSpecification.formulaName, executionContext.authenticationName);
+  const maxPermissionBatchSize = formula?.maxPermissionBatchSize ?? Limits.PermissionsBatchSize;
```
jonathan-codaio, line 1027: This is horribly confusing but it looks like the default that's used in practice in 10, which is just hardcoded inline randomly in process_csb_ingestion.ts. So I think we should just hardcode 10 here as well and note that this matches what's used in `coda` at the moment.

(We can and _maybe_ should try to unify these, but ultimately I think we want to reserve the right to change the default via runtime config to tune performance.)

cli/coda.ts
ekoleda-codaio: Ya, I totally get the need, it just feels like the pattern for invoking the original and new one are quite different. If the result is a list of permissions, then I think the formula description should look more like `:permissions`. Some alt ideas:

- Have the user run `:permissions` and if they don't specify the request info on the command line, we default to running the sync and using that data.
- Have the user run `:permissions` and pass an optional flag like `--fromSync.`
- Extend the formula specification string to indicate that a chaining is happening, like `MySyncTable>permmissions`

test/execution_test.ts
zack-codaio: Ahhh okay my misunderstanding, I can do that

testing/execution.ts
zack-codaio: Sounds good, I went with `const maxPermissionBatchSize = formula?.maxPermissionBatchSize || 10;` to match that file, since it looks like we'll also try to read off the formula if we can.

Also a default of 10 instead of 5000 means I can actually write a reasonable test to exercise this code

cli/coda.ts
jonathan-codaio: Well, one confounding factor is that the nature of the inputs you are expected to pass are radically different.

If you run `:permissions`, you are expected to pass a JSON object containing an array of row data. If you then add the `--fromSync` flag, you are all of the sudden expected to pass regular shell args representing the parameter values to the sync formula (the same things you would pass to `coda execute`).

So the way you have to pass inputs really points that this is truly a modality of `coda execute`.

jonathan-codaio: The chaining approach is not unreasonable. If we have something that builds on the "primary" execution being the sync and that you are passing sync parameters. So I'd be cool with that instead of a flag.

I still don't have a good argument for what the right thing to return for the return value/printing. I'm not happy with anything I've proposed so far, and having the output vary in kind of odd ways depending on flags feels confusing. I also don't feel strongly.

Maybe this is a point in favor of chaining because the las thing in the chain then implies the output format. And so that makes it justified to _only_ print the permission result, and suppress the row data, which makes the output more comprehensible.

zack-codaio: I'll take a look at implementing a chaining syntax, that also feels like it lines up with code implementation.

Are there other things that we might plausibly chain after the Sync results?

testing/execution.ts
```diff
@@ -364,6 +364,15 @@ export function makeFormulaSpec(manifest: BasicPackDefinition, formulaNameInput:
   const formula = ensureExists(syncFormula || standardFormula);
 
   if (parts.length === 0) {
+    if (chainedCommands.length > 1) {
+      const chainedCommand = makeFormulaSpec(manifest, [formulaOrSyncName, chainedCommands[1]].join(':'));
```
zack-codaio, line 368: not sure if we want this to happen on Standard, or only on Sync formulas (probably just on Sync?)

zack-codaio, line 368: this currently only allows going down a single level of >

we could pass through the rest of the chained commands if we want more levels, but it would probably be tricky to implement since we are executing these things somewhat concurrently (within the continuation loop) rather than finishing an entire cycle (all pages of Sync) before passing to chained commands.

testing/execution.ts
```diff
@@ -322,7 +321,8 @@ function invert<K extends string | number | symbol, V extends string | number |
 
 // Exported for tests.
 export function makeFormulaSpec(manifest: BasicPackDefinition, formulaNameInput: string): FormulaSpecification {
-  const [formulaOrSyncName, ...parts] = formulaNameInput.split(':');
+  const chainedCommands = formulaNameInput.split('>');
```
zack-codaio, line 324: In actual CLI usage, I found that `>` needed to be escaped with `\>` to actually work. Had the same issue with pipe `|`, open to other symbols

testing/execution.ts
```diff
@@ -1038,7 +1045,7 @@ async function executeSyncFormulaWithContinuations({
  * @param bundlePath The bundle path
  * @returns Returns an array of the rowAccessDefinitions
  */
-async function getPermissionsForRows({
+async function chainCommandOnSyncResult({
```
zack-codaio, line 1048: made this more generic, though I only implemented a single method in practice.

Need to update jsdocs

cli/coda.ts
zack-codaio: I put up a chained syntax in this commit here: https://github.com/coda/packs-sdk/pull/3162/commits/81da58309a3a1e80083316ba9b2262078acb89e4

Couple of notes:
- > (and |) require escaping via `\>` when actually used in the terminal
- I only allowed a single level of chaining, rather than arbitrary length of chaining. I think arbitrary length would be easy if we were actually just fully executing the sync and passing the full set to the next command, but we're executing them in the same chunk loop so I think it would be pretty awkward to implement (but not impossible)
- I only implemented the GetPermissions formula type, but I set up the functions in a way that we could add others
- Only allowed it for cases where the first formula is Sync (or Standard? but I'm guessing we only want Sync)

runtime/types.ts
```diff
@@ -28,11 +28,13 @@ export enum MetadataFormulaType {
 export interface StandardFormulaSpecification {
   type: FormulaType.Standard;
   formulaName: string;
+  chainedCommand?: FormulaSpecification;
```
jonathan-codaio, line 31: So this file is actually part of the runtime that's used in production. I wouldn't want to change it just for the CLI simulator, I have some suggestions below.

test/execution_test.ts
```diff
@@ -495,6 +495,67 @@ describe('Execution', () => {
           }
         });
 
+        it('sync with includePermissions works', async () => {
+          await executeFormulaOrSyncFromCLI({
+            vm,
+            formulaName: 'Students>permissions',
```
jonathan-codaio, line 501: I don't think we should consider the chain to be part of the formula name, see below.

testing/execution.ts
```diff
@@ -65,6 +66,9 @@ export interface ExecuteOptions {
 export interface ContextOptions {
   useRealFetcher?: boolean;
   manifestPath?: string;
+  syncOptions?: {
+    includePermissions?: boolean;
+  };
```
jonathan-codaio, line 71: I think you can revert this now.

testing/execution.ts
```diff
@@ -248,32 +252,16 @@ export async function executeFormulaOrSyncFromCLI({
     const formulaSpecification = makeFormulaSpec(manifest, formulaName);
```
jonathan-codaio, line 252: I would probably just make some tweaks right here. Perhaps wrap `makeFormulaSpec` in a broader parser that separates the chain command from the rest of the formula string, and returns a composite object that includes the formula spec and optionally info about the chain command. This can validate that the chain command lines up with the formula, e.g. given that `permissions` is our only chain command that it only can be used when the formula spec is for a sync.

In theory we could change `makeFormulaSpec` to do all that but right now its return type is `FormulaSpecification` which is a type that's used in production, so I'd rather just do things in a way that don't touch production types or code at all and don't require us to extend those types in weird ways.

So once you do the parsing here, you'll have both the same formula spec you always had plus some indication there is a permission chain, and you can key your new loop behavior off of that bit. Basically keeping all the code you had before this latest commit, but just conditioning the new behavior on what you parse here instead of the contextOptions bit.

testing/execution.ts
jonathan-codaio: Ah yes, that is unfortunate. We should keep this ergonomic, @ekoleda-codaio do you have a favorite alternate symbol that doesn't require shell escaping?

testing/execution.ts
jonathan-codaio: Yeah no need to generalize more than what you have, I was just imagining that the `>` is purely syntactic and a way for makers to develop and intuition about how certain invocations behave like chaining and affect the final output, but in terms of the implementation, it's basically exactly what you had before, just a new syntax to invoke it. Not a general framework.

test/execution_test.ts
zack-codaio: I'm not sure I follow here, I modeled this off of the existing `Students:permissions` test case, and I thought this was basically adding a slightly different syntax (but keeping the overall structure).

Separating out the chaining out of `makeFormulaSpec()` makes sense, but I thought the `formulaName` here is trying to approximate the raw input you'd give through the cli?

testing/execution.ts
zack-codaio: Done, I pulled this out into a separate function which definitely simplifies things. Makes it easier to see that we only allow one level and that we only allow it if the first formula is a sync.

I had thought about doing this initially, but then didn't want to update all of the other calls of makeFormulaSpec, but I guess we only care about this type of parsing in this particular CLI flow, right?

test/execution_test.ts
jonathan-codaio: Ah sorry, this is all good. I didn't realize this was calling into the wrapper function that has extra orchestration logic before calling to do the raw pack execution.

testing/execution.ts
```diff
@@ -331,6 +316,30 @@ function invert<K extends string | number | symbol, V extends string | number |
...
+  formulaNameInput: string,
+): {
+  formulaSpecification: FormulaSpecification;
+  chainedCommand?: FormulaSpecification;
+} {
+  const chainedCommands = formulaNameInput.split('>');
+  const formulaSpecification = makeFormulaSpec(manifest, chainedCommands[0]);
+
+  if (chainedCommands.length > 1 && formulaSpecification.type === FormulaType.Sync) {
+    const chainedCommand = makeFormulaSpec(manifest, [formulaSpecification.formulaName, chainedCommands[1]].join(':'));
```
jonathan-codaio, line 337: I would go further and throw and error if:

* There is a chained command used with a non-sync formula spec.
* The chained command is anything other than for permissions.

testing/execution.ts
```diff
@@ -331,6 +316,30 @@ function invert<K extends string | number | symbol, V extends string | number |
...
+ * @param manifest The manifest of the pack.
+ * @param formulaNameInput The formula name with a chained command.
+ * @returns The formula specification and the chained command.
+ */
+function getFormulaSpecAndChainedCommand(
+  manifest: BasicPackDefinition,
+  formulaNameInput: string,
+): {
+  formulaSpecification: FormulaSpecification;
+  chainedCommand?: FormulaSpecification;
```
jonathan-codaio, line 331: Maybe have this return a new type:

```ts
type ChainedCommandFormulaSpecification = GetPermissionsFormulaSpecification;
```

Which is a union type that only has one member so far.

Down at line 337 if you check the `type` of the parsed formula spec I think TS will narrow this for your return value and you won't need any casting.

And then your `switch` in `chainCommandOnSyncResult` can include an `ensureUnreachable`.

testing/execution.ts
```diff
@@ -951,3 +962,168 @@ class DeveloperError extends Error {
...
+
+  switch (chainedCommand.type) {
+    case FormulaType.GetPermissions:
+      const mappedRows = denormalizedSyncResult.map((row: unknown) => ({row}));
+
+      // 10 hardcoded as fallback to match process_csb_ingestion.ts
+      const maxPermissionBatchSize = formula?.maxPermissionBatchSize || 10;
+      const chunks = chunkArray(mappedRows, maxPermissionBatchSize);
+
+      const chunkResponses = await Promise.all(
```
jonathan-codaio, line 1101: I would just do a `for (const chunk of chunks) {`, no need to do the chunks in parallel, the production doe does the chunks serially and I think it'll be easier to follow/debug as well.

jonathan-codaio (commented):
Great, all LGTM, just some minor touch-up suggestions.

testing/execution.ts
zack-codaio: done

testing/execution.ts
zack-codaio: done, thanks for the suggestion -- a narrower type definitely makes sense here

jonathan-codaio (approved):
🎉 

Looks like you'll have to deal with the merge conflict though.

testing/execution.ts
ekoleda-codaio: Sorry, I missed this callout originally. I just ran into this myself. Not ideal, but it seems like so many special characters are overloaded in bash that we might need to escape it no matter what. I find wrapping it in single quotes (`'Table>permissions'`) to be the simplest way to escape it.

PR #3165, opened 2025-02-27
[Brain Boost] Add Authority Boost Support in Packs

solomon-codaio:
## Summary
- Add `popularityNormProperty` as a renamed version of `popularityRankProperty`.
    - We keep `popularityRankProperty` for backwards compatibility. When all Brain packs have migrated to the new `popularityNormProperty` we will remove the old property.
- Add `authorityNormProperty` for enabling authority boost indexing as a ranking signal to enhance unstructured search on Pack data.
- This change required for downstream changes:
    - `coda` https://github.com/coda/coda/pull/124092
    - `packs` (Slack) https://github.com/coda/packs/pull/4817


randy-codaio:
While we're at it do we want to commit switching names to `fooNormProperty`? Would need to keep the old popularity one around until the packs are updated before fully deprecating.

solomon-codaio:
> While we're at it do we want to commit switching names to `fooNormProperty`? Would need to keep the old popularity one around until the packs are updated before fully deprecating.

If I rename and deploy this change on `packs-sdk`, what coordinating changes are needed to avoid breakage?

randy-codaio:
I think we'd just copy the existing popularity property with an updated name, but still keep the old name around as well to maintain backwards compatibility until we update all the packs that are using it. At that point we can delete the old one.

randy-codaio:
Curious about the change to -1 to 1 though, is this something that we're set on? I think at the very least it should be standard between popularity/authority (i.e., shouldn't have different ranges)

schema.ts
```diff
@@ -1125,7 +1126,16 @@ export interface IndexDefinition {
   contextProperties?: ContextProperties;
   /**
    * The name of the property within {@link ObjectSchemaDefinition.properties} that can be be interpreted as
-   * a number between 0.0 and 1.0 representing the popularity rank of this entity compared to all other entities.
+   * a number between -1.0 and 1.0 representing the authority rank of this entity compared to all other entities.
```
chris-codaio, line 1129: Seems like we should keep this range aligned with the existing popularity rank property one way or the other?

solomon-codaio: I can imagine ways in which a "baseline measure" of an item here is a state of being "not popular" in a way that is distinct from that item being "very unpopular". An example is thumbs-up / thumbs-down ratio.

solomon-codaio:
> @solomon-codaio I think we'd just copy the existing popularity property with an updated name, but still keep the old name around as well to maintain backwards compatibility until we update all the packs that are using it. At that point we can delete the old one.

Done! Can you review the latest diff? @randy-codaio

chris-codaio (approved):

ekoleda-codaio:
Chiming in late (I was at SKO) but I don't think adding `Norm` to these fields increases clarity. It may not be clear to developers what "Norm" is short for, and there isn't a need to distinguish it from non-normalized equivalent fields.

solomon-codaio:
> Chiming in late (I was at SKO) but I don't think adding `Norm` to these fields increases clarity. It may not be clear to developers what "Norm" is short for, and there isn't a need to distinguish it from non-normalized equivalent fields.

Thanks for the feedback. Would it be clearer if we called it something simpler like `PopularitySearchBoost`?

chris-codaio:
I wouldn't use the term `boost` since we may use these for other purposes (e.g., filtering).

FWIW, I like the inclusion of the `normalization` term in the param since it makes it clear these have to be globally normalized across the sync table, not just for one call or individually.  Might be other ways to make that clear, though.

PR #3170, opened 2025-03-03
Add a Google DWD error type

patrick-codaio:
https://codaproject.slack.com/archives/C06HG23CV08/p1741037646800209?thread_ts=1739904282.527969&cid=C06HG23CV08


api.ts
```diff
@@ -247,6 +247,22 @@ export class MissingScopesError extends Error {
   }
 }
 
+/**
+ * This error will be thrown by the fetcher if it fails to generate a valid DWD impersonation token.
+ */
+export class GoogleDwdError extends Error {
```
sam-codaio, line 253: Do we want to be more specific and call this an impersonation error?

patrick-codaio: Maybe? I wasn't sure if other things might go wrong too, and I didn't want to have to keep adding more special errors.

ebo-codaio: if we want to handle other types of DWD errors, do we just add `type` property to this error type?

sam-codaio: That is fine as well

ebo-codaio (approved):

PR #3171, opened 2025-03-04
Allow incremental sync from CLI

zack-codaio:
This PR adds support for calling an incremental sync from the CLI.

When calling MyTable>incremental, we will:
- Parse the call into a Sync formula and then another Sync formula inside of a chained command object.
- Run the first sync until it hits a completion (without a concurrent continuation)
- Run the second sync using the completion from the first sync in the executionContext
- Return the results from the second (incremental) sync

Some notes:
- We do not run the confirmation loop in an arbitrary length, we're essentially just grabbing the second "batch" defined by confirmations
- We do not currently allow MyTable:incremental to be called directly, but can add support for this later.
- The internal logic for > syntax was updated to allow some commands to be run in the continuation loop (interleaved), versus in the confirmation loop (subsequent)

### DIFF
--- a/runtime/types.ts
+++ b/runtime/types.ts
@@ -97,7 +97,7 @@ export type FormulaSpecification =
   | PropertyAutocompleteFormulaSpecification
   | GetPermissionsFormulaSpecification;
 
-export type ChainedCommandFormulaSpecification = GetPermissionsFormulaSpecification;
+export type ChainedCommandFormulaSpecification = GetPermissionsFormulaSpecification | SyncFormulaSpecification;
 
 export type PackFunctionResponse<T extends FormulaSpecification> = T extends SyncFormulaSpecification
   ? GenericSyncFormulaResult
--- a/runtime/types.ts
+++ b/runtime/types.ts
@@ -97,8 +97,6 @@ export type FormulaSpecification =
   | PropertyAutocompleteFormulaSpecification
   | GetPermissionsFormulaSpecification;
 
-export type ChainedCommandFormulaSpecification = GetPermissionsFormulaSpecification | SyncFormulaSpecification;
-
 export type PackFunctionResponse<T extends FormulaSpecification> = T extends SyncFormulaSpecification
   ? GenericSyncFormulaResult
   : T extends SyncUpdateFormulaSpecification
--- a/test/execution_test.ts
+++ b/test/execution_test.ts
@@ -584,9 +584,6 @@ describe('Execution', () => {
           });
           const result = mockPrintFull.args[0][0];
 
-          // What happens if we have multiple incremental syncs?
-          // Assuming we sync the first page in the normal sync, and then passing
-          // >incremental just gets the next page.
           assert.deepEqual(
             result,
             ['Ethan', 'Fiona', 'Gina', 'Hank'].map(name => (vm ? {name} : {Name: name})),
--- a/test/execution_test.ts
+++ b/test/execution_test.ts
@@ -128,12 +128,18 @@ describe('Execution', () => {
   });
 
   it('execute sync works and returns passthrough', async () => {
-    const results = await executeSyncFormula(fakePack, 'Students', ['Smith', 'true'], undefined, {}, {useRealFetcher: false});
-    assert.deepEqual(results.permissionsContext, [{userId : 42},{userId : 123},{userId : 53},{userId : 22}]);
+    const results = await executeSyncFormula(
+      fakePack,
+      'Students',
+      ['Smith', 'true'],
+      undefined,
+      {},
+      {useRealFetcher: false},
+    );
+    assert.deepEqual(results.permissionsContext, [{userId: 42}, {userId: 123}, {userId: 53}, {userId: 22}]);
     assert.deepEqual(results.result, [{Name: 'Alice'}, {Name: 'Bob'}, {Name: 'Chris'}, {Name: 'Diana'}]);
     assert.equal(results.permissionsContext?.length, results.result.length);
   });
-
 
   describe('execution errors', () => {
     it('not enough params', async () => {
@@ -565,6 +571,28 @@ describe('Execution', () => {
           );
         });
 
+        it('sync with incremental chaining works', async () => {
+          await executeFormulaOrSyncFromCLI({
+            vm,
+            formulaName: 'Students>incremental',
+            params: ['Mr. Incremental'],
+            manifest: fakePack,
+            manifestPath: '',
+            bundleSourceMapPath,
+            bundlePath,
+            contextOptions: {useRealFetcher: false},
+          });
+          const result = mockPrintFull.args[0][0];
+
+          // What happens if we have multiple incremental syncs?
+          // Assuming we sync the first page in the normal sync, and then passing
+          // >incremental just gets the next page.
+          assert.deepEqual(
+            result,
+            ['Ethan', 'Fiona', 'Gina', 'Hank'].map(name => (vm ? {name} : {Name: name})),
+          );
+        });
+
         it('sync update works', async () => {
           const syncUpdates: GenericSyncUpdate[] = [
             {previousValue: {name: 'Alice'}, newValue: {name: 'Alice Smith'}, updatedFields: ['name']},
@@ -616,9 +644,9 @@ describe('Execution', () => {
           const syncRows: GenericExecuteGetPermissionsRequest = {
             rows: [{row: {name: 'Alice'}}, {row: {name: 'Bob'}}],
             permissionsContext: [
-              {userId: 42},  // For first row
-              {userId: 123}    // For second row
-            ]
+              {userId: 42}, // For first row
+              {userId: 123}, // For second row
+            ],
           };
           await executeFormulaOrSyncFromCLI({
             vm,
--- a/testing/execution.ts
+++ b/testing/execution.ts
@@ -1,5 +1,5 @@
 import type {BasicPackDefinition} from '../types';
-import type {ChainedCommandFormulaSpecification} from '../runtime/types';
+import type {ChainedCommandFormulaSpecification} from './types';
 import type {Continuation} from '../api';
 import type {Credentials} from './auth_types';
 import type {ExecutionContext} from '../api_types';
@@ -11,6 +11,7 @@ import type {GenericSyncUpdate} from '../api';
 import type {GenericSyncUpdateResult} from '../api';
 import type {GetPermissionsFormulaSpecification} from '../runtime/types';
 import type {GetPermissionsResult} from '../api';
+import type {InterleavedChainedCommandFormulaSpecification} from './types';
 import type {MetadataContext} from '../api';
 import type {MetadataFormula} from '../api';
 import type {MetadataFormulaSpecification} from '../runtime/types';
@@ -22,6 +23,7 @@ import type {ParamDefs} from '../api_types';
 import type {ParamValues} from '../api_types';
 import type {PostSetupMetadataFormulaSpecification} from '../runtime/types';
 import type {StandardFormulaSpecification} from '../runtime/types';
+import type {SubsequentChainedCommandFormulaSpecification} from './types';
 import type {SyncCompletionMetadata} from '../api_types';
 import type {SyncExecutionContext} from '../api_types';
 import type {SyncFormulaSpecification} from '../runtime/types';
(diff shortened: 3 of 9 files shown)


runtime/types.ts
```diff
@@ -97,7 +97,7 @@ export type FormulaSpecification =
   | PropertyAutocompleteFormulaSpecification
   | GetPermissionsFormulaSpecification;
 
-export type ChainedCommandFormulaSpecification = GetPermissionsFormulaSpecification;
+export type ChainedCommandFormulaSpecification = GetPermissionsFormulaSpecification | SyncFormulaSpecification;
```
jonathan-codaio, line 100: Oh sorry I should have caught this last time, can you move this to somewhere under `testing/...`? Stuff in `runtime` is actually used in production, so we should keep the simulation stuff out of there.

test/execution_test.ts
```diff
@@ -565,6 +571,28 @@ describe('Execution', () => {
...
+            manifestPath: '',
+            bundleSourceMapPath,
+            bundlePath,
+            contextOptions: {useRealFetcher: false},
+          });
+          const result = mockPrintFull.args[0][0];
+
+          // What happens if we have multiple incremental syncs?
+          // Assuming we sync the first page in the normal sync, and then passing
+          // >incremental just gets the next page.
```
jonathan-codaio, line 589: It's a good question. I think I gravitate towards running a "complete" incremental sync, that is, kick off a sync loop that has a `previousCompletion` provided in the context, and run it like normal, in a loop until there is no more `continuation`. Print all of the results from the incremental loop.

testing/execution.ts
```diff
@@ -248,7 +250,7 @@ export async function executeFormulaOrSyncFromCLI({
 
     const {formulaSpecification, chainedCommand} = getFormulaSpecAndChainedCommand(manifest, formulaName);
 
-    if (formulaSpecification.type === FormulaType.Sync) {
+    if (formulaSpecification.type === FormulaType.Sync && chainedCommand?.type !== FormulaType.Sync) {
```
jonathan-codaio, line 253: Sounds like this is already on your cleanup list, but I'd try to have one `if` case here for `if (formulaSpecification.type === FormulaType.Sync)` and handle the sub-cases inside of it, ideally using an exhaustive `switch` for the chained command types.

testing/execution.ts
```diff
@@ -1060,6 +1099,10 @@ async function executeSyncFormulaWithContinuations({
       );
     }
     executionContext.sync.continuation = response.continuation;
+    if (response.completion) {
```
jonathan-codaio, line 1102: To your question, @patrick-codaio should confirm but I believe that a `continuation` and a `completion` are mutually exclusive, so we could potentially assert that we don't have both of them.

testing/execution.ts
```diff
@@ -1060,6 +1099,10 @@ async function executeSyncFormulaWithContinuations({
...
+      executionContext.sync.previousCompletion = response.completion as SyncCompletionMetadata<Continuation>;
```
jonathan-codaio, line 1103: I'm not sure I would mutate the existing execution context. I think I'd try to return the completion, and have the caller generate a new `context.sync` value and then just call this function again with that.

testing/execution.ts
```diff
@@ -1097,7 +1140,7 @@ async function chainCommandOnSyncResult({
 }: {
   rows: any[];
   formulaSpecification: SyncFormulaSpecification;
-  chainedCommand: ChainedCommandFormulaSpecification;
+  chainedCommand: GetPermissionsFormulaSpecification;
```
jonathan-codaio, line 1143: It's probably worth seeing if you can still accept the full type and have a case here that calls into `executeSyncFormulaWithContinuations()`, so doing mutual recursion. If it's clean that seems conceptually reasonable. If it's ugly we can stick with something closer to what you have but clean up the names and types so it's clearer that not all chain types are covered by this function.

testing/execution.ts
```diff
@@ -260,6 +262,30 @@ export async function executeFormulaOrSyncFromCLI({
         bundlePath,
       });
       printFull(chainedCommand ? chainedCommandResults : result);
+    } else if (formulaSpecification.type === FormulaType.Sync && chainedCommand?.type === FormulaType.Sync) {
```
jonathan-codaio, line 265: Ah hmm maybe we should move to a new enum for chained command types, it's not that intuitive that Sync chained with Sync means "incremental sync", at least to me.

jonathan-codaio (commented):
Overall this is reasonable

testing/execution.ts
patrick-codaio: We didn't make it an error, but we are clear in the documentation that we will ignore a `completion` if there's also a `continuation`:
Screenshot 2025-03-04 at 4 39 27 PM

jonathan-codaio: Ah yep, ok ignoring sounds like good behavior to have in the simulator too.

zack-codaio: sounds good, I can do that

testing/execution.ts
zack-codaio: I think the issue I was seeing was that this `chainCommandOnSyncResult()` function is actually being called on chunks defined by `continuations` rather than on `completions`

I'm leaning towards having `getFormulaSpecAndChainedCommand()` return some bit that specifies whether the chained command should be happening on each `continuation` page (like we do with permissions) or if it should happen after the full first sync up to a `completion` happens. If we do that, then I don't know if we want to put those effects into the same function, since they'll be called at different times and you'd have to know that to work within the function properly.

Type wise, `getFormulaSpecAndChainedCommand()` could still return `ChainedCommandFormulaSpecification`, but then maybe that would be a union type of something like `InterleavedChainedCommandFormulaSpecification` (GetPermissions for now) and `SubsequentChainedCommandFormulaSpecification` (Sync)

test/execution_test.ts
jonathan-codaio: Yep, good summary, we're on the same page.

testing/execution.ts
jonathan-codaio: Oh, yeah great observation. I hadn't noticed the difference that one does chaining by interleaving and one does chaining only at the end. That does complicate the conceptual model and reuse of the existing code.

I think you're on the right track to add a dimension that distinguishes interleaving from post-completion, and with dedicated handler functions for each case I think the opportunities for recursion should become clearer.

testing/execution.ts
```diff
@@ -426,6 +445,15 @@ export function makeFormulaSpec(manifest: BasicPackDefinition, formulaNameInput:
         formulaName: formulaOrSyncName,
       };
     }
+    if (metadataFormulaTypeStr === 'incremental') {
```
zack-codaio, line 448: I'm on the fence on if this should be here. I was originally hard coding this formulaSpec return in  `getFormulaSpecAndChainedCommands()`, but if I put the parsing here then it's colocated with the :permissions parsing and we can just always run `makeFormulaSpec()` a single time.

The negative is that doing an explicit `MyTableName:incremental` doesn't really do what you might expect, which in my mind would be to run the incremental sync with a pre-specified completion in the executionContext. What actually happens is that there's not currently a way to specify the completion from the CLI (AFAICT), and we'll just do a regular Sync, and the :incremental part is kind of a noop.

testing/execution.ts
zack-codaio: I ended up changing `getFormulaSpecAndChainedCommands()` to return an interleavedChainedCommand and a subsequentChainedCommand, and returning the command in the right slot based on the formula type.

Then, when those commands are provided, we can pass them through to the continuation loop or the completion loop. Also pulled all of that junk into its own function again `executeSyncFormulaWithOptionalChaining()` since it was getting a little unwieldy to keep in the main function, and pulling it out makes printing the right thing easier (just print whatever we choose to return).

testing/execution.ts
```diff
@@ -1044,7 +1149,8 @@ async function executeSyncFormulaWithContinuations({
       );
     }
     result.push(...response.result);
-    if (chainedCommand) {
+    // Skip Sync here, since we'll handle incremental syncs after this full sync loop happens
+    if (chainedCommand && chainedCommand.type !== FormulaType.Sync) {
```
jonathan-codaio, line 1153: If you type `chainedCommand` at line 1117 as `InterleavedChainedCommandFormulaSpecification` then you can just delete the extra check here since if you have a chained command here it'll be guaranteed to be applicable. This is already the case at runtime, just need to tighten up the types.

testing/execution.ts
jonathan-codaio: I think it's reasonable to have this colocated here. If you're able to detect that `MyTableName:incremental` was called directly and not via chaining, you can just have it throw an error for now and indicate that direct invocation is not yet supported, but add a TODO and that's something we can consider adding if useful. (You'd probably just have to pass a JSON string representing the completion on the CLI like we do for other complex formula types.)

testing/execution.ts
```diff
@@ -320,16 +326,23 @@ function invert<K extends string | number | symbol, V extends string | number |
...
  */
-function getFormulaSpecAndChainedCommand(
+function getFormulaSpecAndChainedCommands(
   manifest: BasicPackDefinition,
   formulaNameInput: string,
 ): {
   formulaSpecification: FormulaSpecification;
-  chainedCommand?: ChainedCommandFormulaSpecification;
+  interleavedChainedCommand?: InterleavedChainedCommandFormulaSpecification;
+  subsequentChainedCommand?: SubsequentChainedCommandFormulaSpecification;
```
jonathan-codaio, line 345: I think I would just beef up the types here. It'll be a little verbose but it will ultimately both model the behavior more directly and be more apparent from the types that incremental is involved.

There's a comment that says `We will only return at most a single command across these two types` which is generally signal to user a union type, which is more overhead, but it makes this self-enforcing. So we can have something like:

```ts
enum ChainedCommandType {
  Interleaved = 'Interleaved'
  Subsequent = 'Subsequent'
}

interface InterleavedChainedCommand {
  type: ChainedCommandType.Interleaved;
  // I think we can dispense with InterleavedChainedCommandFormulaSpecification now, for our
  // current purposes ChainedCommand defined below is our new union type, but if we add more
  // chained command in the future the formulaSpec here can expand to a union type.
  formulaSpec: GetPermissionsFormulaSpecification;
}

interface SubsequentChainedCommand {
  type: ChainedCommandType.Subsequent;
  // These two fields can later become part of another union type if there are more subsequent
  // commands in the future. The point here is that the command text 'incremental' is
  // bound together wit ha sync formula spec, so you can do business logic by looking
  // for the command called 'incremental' which is really clear and explicit, and still use
  // the existing formulaSpec to pass around to existing helpers.
  commandText: 'incremental';
  formulaSpec: SyncFormulaSpecification;
}

type ChainedCommand = InterleavedChainedCommand | SubsequentChainedCommand;
```

jonathan-codaio (commented):
Nice, overall looks good, I think mostly it's just about hardening the types but the implementation is looking nice.

testing/execution.ts
```diff
@@ -982,12 +1026,88 @@ class DeveloperError extends Error {
...
+  if (!chainedCommand) {
+    return result;
+  }
+
+  if (chainedCommand.type === ChainedCommandType.Interleaved) {
+    return chainedCommandResults;
+  }
+
+  switch (chainedCommand.formulaSpec.type) {
+    case FormulaType.Sync:
```
jonathan-codaio, line 1087: It's logically all the same but my preference is still to have a `commandName: 'incremental';` or whatever in the `SubsequentChainedCommand` interface, so that the switch here is `switch (chainedCommand.commandName)` and the case is `case 'incremental':` which makes it explicit when reading this that this case is about incremental sync.

The `formulaSpec` would be narrowed to `Sync` by that same switch statement so it should have the same effect for TS, it's just for the human reader that you're indicating "aha, this is where incremental sync gets handled".

If we had more subsequent chained commands we'd have to make it more apparent that `SubsequentChainedCommand` is a union of multiple command types with their own names, but for now we can just add a new commandName property and get everything we need.

jonathan-codaio (approved):
Nice, code looks great, this is a clean addition to the existing chaining.

testing/execution.ts
zack-codaio: Done, lmk if my latest commit matches what you had in mind. You're just asking for a separate, more explicit property on the chainedCommand, rather than relying on the underlying formulaSpec, right?

jonathan-codaio (approved):
Yep

PR #3174, opened 2025-03-11
Adding displayName to sync tables

farazyasrobi-coda:
This PR adds optional displayName property to definition of the sync tables.


patrick-codaio (approved):

api.ts
```diff
@@ -2185,11 +2189,19 @@ export interface SyncTableOptions<
   PermissionsContextT extends SyncPassthroughData,
 > {
   /**
-   * The name of the sync table. This is shown to users in the Coda UI.
+   * The name of the sync table. This is shown to users in the Coda UI if displayName is not present.
    * This should describe the entities being synced. For example, a sync table that syncs products
    * from an e-commerce platform should be called 'Products'. This name must not contain spaces.
+   *
+   * Important: This should not be updated after the first time a table is included in the pack version.
```
ekoleda-codaio, line 2196: Maybe better to reframe this as a breaking change.

"Important: This value acts as a unique ID for the table, and updating it later is a breaking change. If you want to change the value shown to the users set `displayName` instead.

api.ts
```diff
@@ -2185,11 +2189,19 @@ export interface SyncTableOptions<
...
    */
   name: string;
+
+  /**
+   * This is the name shown to users in the Coda UI. If not present, name will be used. This is safe to update.
```
ekoleda-codaio, line 2201: Nit: Perhaps use an `{@link}` to link to the name field, for better documentation.

ekoleda-codaio, line 2201: I think "This is safe to update." is not needed, as that's sort of the default assumption. It would be nice to document what happens when you do change the value. Specifically, will we go in and rename all instances of the table? Or only the instances that haven't already been renamed by the user? Or perhaps we don't rename any existing instances, and it only affects new ones?

ekoleda-codaio (changes requested):
Thanks for reaching out! Left a few small comments. Developers are going to love this feature!

api.ts
gary-codaio: Wondering if you have any leaning on what should happen when we change the value! Seems like a product decision; I personally think we should not rename existing instances

ekoleda-codaio: For sync table columns (top-level schema properties) we do rename them, but only if they haven't already been manually renamed by the user. We could take the same approach here, but I think renaming an existing table is a bit too magical and unexpected, and I would probably just have it affect new tables.

farazyasrobi-coda: Added this to the comment: "Changing this value will not affect existing tables and only affects newly created tables. "

farazyasrobi-coda:
Hi @ekoleda-codaio! I addressed your comments. Would you mind taking another look?

ekoleda-codaio (approved):

PR #3175, opened 2025-03-18
Update MkDocs and MkDocs Material

ekoleda-codaio:
Updating to the latest version. Also fixing some broken anchor links reported in the build process.


jonathan-codaio (approved):
Thanks!

PR #3176, opened 2025-03-28
Handle continuations when running executeGetPermissions from CLI

zack-codaio:
This PR allows executeGetPermissions to run in a continuation loop, both when run directly with `:permissions` and when run on a Sync result with `>permissions`

### DIFF
--- a/testing/execution.ts
+++ b/testing/execution.ts
@@ -264,6 +264,17 @@ export async function executeFormulaOrSyncFromCLI({
         bundlePath,
       });
       printFull(result);
+    } else if (formulaSpecification.type === FormulaType.GetPermissions) {
+      const result = await executeGetPermissionsFormulaWithContinuations({
+        formulaSpecification,
+        params,
+        manifest,
+        executionContext,
+        vm,
+        bundleSourceMapPath,
+        bundlePath,
+      });
+      printFull(result);
     } else {
       const result = vm
         ? await executeFormulaOrSyncWithRawParamsInVM({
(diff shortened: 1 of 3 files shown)


testing/execution.ts
```diff
@@ -1199,6 +1210,70 @@ async function executeSyncFormulaWithContinuations({
...
+
+    // XXX need to fix type inference in executeFormulaOrSyncWithRawParams
+    const castedResponse = response as GetPermissionsResult;
+
+    result.push(...castedResponse.rowAccessDefinitions);
+    executionContextCopy.sync.continuation = castedResponse.continuation;
+
+    iterations++;
+  } while (executionContextCopy.sync.continuation && result.length < maxRows);
+  if (result.length > maxRows) {
```
zack-codaio, line 1270: probably should not be using maxRows here, since I copied and pasted from the Sync loop. Unless we do want to keep this just as a fallback? Looks like it's not actually hooked up currently, so it's just going to hard cap things to 1000.

testing/execution.ts
```diff
@@ -1199,6 +1210,70 @@ async function executeSyncFormulaWithContinuations({
...
+async function executeGetPermissionsFormulaWithContinuations({
+  formulaSpecification,
+  params,
+  manifest,
+  executionContext,
+  bundleSourceMapPath,
+  bundlePath,
+  maxRows = DEFAULT_MAX_ROWS,
+  vm,
+}: {
```
zack-codaio, line 1222: may need to add support for permissionsContext (and update the `get permissions works with passthrough data` test case)

testing/execution.ts
```diff
@@ -1199,6 +1210,70 @@ async function executeSyncFormulaWithContinuations({
   return {result, chainedCommandResults, completion};
 }
 
+async function executeGetPermissionsFormulaWithContinuations({
+  formulaSpecification,
+  params,
+  manifest,
+  executionContext,
```
jonathan-codaio, line 1217: Super nit, I think it would be safer if you aliased this input, using `executionContext: itemsExecutionContext` or something. Then you can call your copy `executionContext`. With the default naming, if you forget to use the copy you'll accidentally pass through this source context and get the pollution issue you're worried about.

testing/execution.ts
jonathan-codaio: There is a flag for `--maxRows` and the value does it get passed into all this logic from execute.ts. I think you just need to plumb it into `executeSyncFormulaWithOptionalChaining()`, it probably got omitted in the last PR.

I think in a perfect world we'd have another flag for maxPermissionRows or maxChainedRow, but I'm not sure it's worth the added complexity. I'd say either:

* Don't use maxRows for chaining. The user can restrict the number of item rows via --maxRows, but given that item set, we'll return the full set of chained results.
* Reuse --maxRows even though it has two meanings. When chaining, it will mean "Fetch up to n items, and then up to n permissions for each item."

testing/execution.ts
```diff
@@ -1199,6 +1210,70 @@ async function executeSyncFormulaWithContinuations({
...
+  // We need to make a copy of the execution context so we don't pollute the continuation on the Sync formula
+  // if we are running inside of an actual Sync continuation loop.
+  const executionContextCopy: SyncExecutionContext = {
+    ...executionContext,
+    sync: {...executionContext.sync, continuation: undefined},
+  };
+  do {
+    if (iterations > MaxSyncIterations) {
+      throw new Error(
+        `Sync is still running after ${MaxSyncIterations} iterations, this is likely due to an infinite loop.`,
```
jonathan-codaio, line 1244: Nit, would just update the error message so it's clear that the infinite loop was when getting permissions.

testing/execution.ts
ekoleda-codaio: I've come around to the first option. Conceptually a row is what `execute` returns, not a permission, so I think it's the most reasonable use.

zack-codaio: Done, and added a test case to make sure I actually use the maxRows flag in the sync loop 🤦

testing/execution.ts
```diff
@@ -1199,6 +1210,70 @@ async function executeSyncFormulaWithContinuations({
...
+          executionContext: executionContextCopy,
+        })
+      : await executeFormulaOrSyncWithRawParams({
+          formulaSpecification,
+          params,
+          manifest,
+          executionContext: executionContextCopy,
+        });
+
+    // XXX need to fix type inference in executeFormulaOrSyncWithRawParams
```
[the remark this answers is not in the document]
zack-codaio: I didn't seem to be able to get all casting to go away, but I was able to push it down so we have a consistent mapping of formulaSpecifications to result types, and then cast the results in `executeFormulaOrSyncWithRawParams` back to the same mapping that we've already defined. Wasn't able to find a way for typescript to properly narrow things down.

jonathan-codaio (approved):

PR #3182, opened 2025-04-11
Remove AWS user creds from CircleCI

chris-codaio:
PTAL @alexd-codaio @neal-codaio @jonathan-codaio

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  aws-cli: circleci/aws-cli@4.1.3
+  aws-cli: circleci/aws-cli@3.2.0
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  aws-cli: circleci/aws-cli@4.1.3
```
neal-codaio, line 3: Downgrade?

neal-codaio (approved):

.circleci/config.yml
chris-codaio: Yeah, just getting everything to be the same consistent version across repos for now and using a version I know works via testing.

jonathan-codaio (approved):

PR #3184, opened 2025-04-21
Minor doc updates

ekoleda-codaio:
- Adding information about timeouts and retries for the fetcher.
- Adding a video to the videos page.

### DIFF
--- a/docs/guides/basics/fetcher.md
+++ b/docs/guides/basics/fetcher.md
@@ -400,6 +400,10 @@ try {
 }
 ```
 
+!!! info "429 responses automatically retried"
+
+    Many APIs have internal rate limits that they enforce on their clients, and will return a 429 response code when the rate limit is exceeded. When the fetcher receives a 429 response it will automatically retry the request until it succeeds or fails.
+
 
 ### Headers {: #headers}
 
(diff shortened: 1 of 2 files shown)


docs/guides/basics/fetcher.md
```diff
@@ -439,6 +443,11 @@ let response = await context.fetcher.fetch({
 For performance reasons responses for HTTP `GET` requests are cached by default. See the [caching guide][caching] for more information.
 
 
+## Timeouts
+
+The fetcher will wait for a response for at most 50 seconds, after which it will fail with an error like `DEADLINE_EXCEEDED: Deadline exceeded after 50s`. Fifty seconds was chosen so that the request would fail before the 60 second Pack execution timeout would be triggered, giving your code a chance to catch and handle the exception.
```
jonathan-codaio, line 448: And FWIW as we kick off some efforts around rate limiting improvements for Brain, I intend to have us finally address the static 50-second timeout and instead have fetcher deadlines be, say, 1 second less than however much time is left on the lambda execution, so you can always respond prior to lambda getting turned down no longer how far into the execution the request begins.

jonathan-codaio (approved):

PR #3185, opened 2025-04-24
New filterableProperties field on index definition

drew-codaio:
https://coda.io/d/Brain-Bullpen-go-brainbull_d138vLd6I-E/Brain-Bullpen-go-brainbull_sudNmSXU#Upcoming-Bullpen-Topics_tuHlW6Z6/r58&view=center

Setting up this new property with the rules discussed above.

### DIFF
--- a/schema.ts
+++ b/schema.ts
@@ -1091,12 +1091,12 @@ export type BasicIndexedProperty = PropertyIdentifier<string>;
 /**
  * A property that will be used to filter the results of a search.
  * 
- * For example, if you are trying to query for all emails last modified in the last 30 days,
- * you would set the filterableProperties to the property that represents the last modified date.
+ * For example, if you want to allow users to query for all emails last modified in the last 30 days,
+ * you would add the property that represents the last modified date to filterableProperties.
  * 
- * Filterable properties must be one of type 
- * {@link ValueHintType.SelectList} {@link ValueType.Boolean} {@link ValueHintType.DateTime} or 
- * {@link ValueType.Number} (Floats will be supported in the future, today they will be truncated to an integer)
+ * Filterable properties must be one of the following types: 
+ * {@link ValueHintType.SelectList}, {@link ValueType.Boolean}, {@link ValueHintType.DateTime} or 
+ * {@link ValueType.Number}. For filtering purposes, number values will be rounded down to the nearest integer.
  * @hidden
  */
 export type FilterableProperty = PropertyIdentifier<string>;
--- a/dist/bundle.d.ts
+++ b/dist/bundle.d.ts
@@ -2011,12 +2011,12 @@ export type BasicIndexedProperty = PropertyIdentifier<string>;
 /**
  * A property that will be used to filter the results of a search.
  *
- * For example, if you are trying to query for all emails last modified in the last 30 days,
- * you would set the filterableProperties to the property that represents the last modified date.
+ * For example, if you want to allow users to query for all emails last modified in the last 30 days,
+ * you would add the property that represents the last modified date to filterableProperties.
  *
- * Filterable properties must be one of type
- * {@link ValueHintType.SelectList} {@link ValueType.Boolean} {@link ValueHintType.DateTime} or
- * {@link ValueType.Number} (Floats will be supported in the future, today they will be truncated to an integer)
+ * Filterable properties must be one of the following types:
+ * {@link ValueHintType.SelectList}, {@link ValueType.Boolean}, {@link ValueHintType.DateTime} or
+ * {@link ValueType.Number}. For filtering purposes, number values will be rounded down to the nearest integer.
  * @hidden
  */
 export type FilterableProperty = PropertyIdentifier<string>;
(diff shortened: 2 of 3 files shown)


schema.ts
```diff
@@ -1088,6 +1088,19 @@ export type ContextProperties = Array<PropertyIdentifier<string>>;
...
 
+/**
+ * A property that will be used to filter the results of a search.
+ * 
+ * For example, if you are trying to query for all emails last modified in the last 30 days,
+ * you would set the filterableProperties to the property that represents the last modified date.
+ * 
+ * Filterable properties must be one of type 
+ * {@link ValueHintType.SelectList} {@link ValueType.Boolean} {@link ValueHintType.DateTime} or 
+ * {@link ValueType.Number} (the number must be an integer)
```
jonathan-codaio, line 1099: Nit, could say that floats will be supported in the near future, but until then, floats will be truncated to ints for filtering.

schema.ts
```diff
@@ -1117,6 +1130,12 @@ export interface IndexDefinition {
    */
   properties: IndexedProperty[];
 
+  /**
+   * A list of properties from within {@link ObjectSchemaDefinition.properties} 
+   * that should be used to filter the results of a search.
```
jonathan-codaio, line 1135: Nit, maybe "that will be made available to filter"

schema.ts
```diff
@@ -1117,6 +1130,12 @@ export interface IndexDefinition {
    */
   properties: IndexedProperty[];
 
+  /**
+   * A list of properties from within {@link ObjectSchemaDefinition.properties}
```
jonathan-codaio, line 1134: Maybe also note something like "Only up to 5 properties can be filterable, so these should be the properties most likely to be useful as filters."

testing/upload_validation.ts
```diff
@@ -1769,6 +1771,44 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
             'contextProperties',
           ]);
         }
+
+        if (filterableProperties) {
+          if (filterableProperties.length > 5) {
```
jonathan-codaio, line 1776: Nit, we have an object in this file called `Limits` where we stash constants for all the limits.

testing/upload_validation.ts
```diff
@@ -1377,6 +1377,7 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
     contextProperties: contextPropertiesSchema.optional(),
     authorityNormProperty: propertySchema.optional(),
     popularityNormProperty: propertySchema.optional(),
+    filterableProperties: z.array(propertySchema).optional(),
```
jonathan-codaio, line 1380: I think you can put the length limit check right here, using `max()`?

jonathan-codaio (commented):
Nice, that was fast :)

Make sure to also update the prerelease version in package.json.

jonathan-codaio (approved):
LGTM, but probably good to give @ekoleda-codaio a chance to weigh in as well.

schema.ts
```diff
@@ -1088,6 +1088,19 @@ export type ContextProperties = Array<PropertyIdentifier<string>>;
 
 export type BasicIndexedProperty = PropertyIdentifier<string>;
 
+/**
+ * A property that will be used to filter the results of a search.
+ * 
+ * For example, if you are trying to query for all emails last modified in the last 30 days,
```
ekoleda-codaio, line 1094:
```suggestion
 * For example, if you want to allow users to query for all emails last modified in the last 30 days,
```

schema.ts
```diff
@@ -1088,6 +1088,19 @@ export type ContextProperties = Array<PropertyIdentifier<string>>;
...
+ * you would set the filterableProperties to the property that represents the last modified date.
```
ekoleda-codaio, line 1095:
```suggestion
 * you would add the property that represents the last modified date to filterableProperties.
```

schema.ts
```diff
@@ -1088,6 +1088,19 @@ export type ContextProperties = Array<PropertyIdentifier<string>>;
...
+ * 
+ * Filterable properties must be one of type 
+ * {@link ValueHintType.SelectList} {@link ValueType.Boolean} {@link ValueHintType.DateTime} or 
+ * {@link ValueType.Number} (Floats will be supported in the future, today they will be truncated to an integer)
```
ekoleda-codaio, line 1099:
```suggestion
 * Filterable properties must be one of the following types: 
 * {@link ValueHintType.SelectList}, {@link ValueType.Boolean}, {@link ValueHintType.DateTime} or 
 * {@link ValueType.Number}. For filtering purposes, number values will be rounded down to the nearest integer.
```

Do we know if it's round or round down?

ekoleda-codaio (approved):
Added some suggestion to the documentation, otherwise LGTM.

schema.ts
drew-codaio: I assume since we're going to clamp it will be a round down.

jonathan-codaio: It's up to us, it's whatever logic we use when we load into TP, so we can do whatever your preference is. Round down sounds reasonable to me (keeping only the integer part as our casting logic).

PR #3187, opened 2025-04-28
Add index categories field to IndexProperty

drew-codaio:
https://grammarly.atlassian.net/browse/BRN-1058 https://grammarly.slack.com/archives/C06QGU1AAN9/p1745434068083699
Adding some category property to SyncTableDef to allow pack makers to define type of tables. This will allow the backend to optimize and be smarter about each type of table.


api_types.ts
```diff
@@ -828,6 +828,23 @@ export enum PermissionSyncMode {
...
+/**
+ * The type of content being indexed, which determines how the property is processed and queried.
+ * Messaging: Chat or instant messaging content
+ * Document: General document content
+ * Email: Email message content
+ * Comment: User comments or feedback
+ * Help: Help documentation or support content
+ * @hidden
+ */
+export enum IndexCategory {
```
drew-codaio, line 840: @chris-codaio Let me know if this list is good. I can add or remove any.

jonathan-codaio, line 840: Nit, I might call this `ContentCategory` (and ditto for the property) to be a little bit more flexible in how this can be used in the future. "index" is a little implementation-specific. Generally in the SDK we like to frame things as "describe your data abstractly, and we'll use that info in ever-changing ways in the platform to make best use of your pack".

jonathan-codaio (commented):
FYI @ekoleda-codaio. It came up that we need content hints like this for search quality purposes.

I know this goes against what we did for top-level pack categories, which got moved _out_ of the SDK and into the pack listing, but the signals here need to exist and the sync table level, and right now all building-block-level metadata lives inside the pack/SDK.

api_types.ts
```diff
@@ -828,6 +828,23 @@ export enum PermissionSyncMode {
   PermissionAware = 'PermissionAware',
 }
 
+/**
+ * The type of content being indexed, which determines how the property is processed and queried.
+ * Messaging: Chat or instant messaging content
```
ekoleda-codaio, line 833: Move these description to tsdoc comments on each enum value individually. It makes it harder to miss one in the future, and still shows up nice in the generated docs.

api_types.ts
```diff
@@ -828,6 +828,23 @@ export enum PermissionSyncMode {
...
+ * Comment: User comments or feedback
+ * Help: Help documentation or support content
+ * @hidden
+ */
+export enum ContentCategory {
+  Messaging = 'Messaging',
+  Document = 'Document',
+  Email = 'Email',
+  Comment = 'Comment',
+  Help = 'Help',
```
ekoleda-codaio, line 845: This one feels different from the others, where it's more about the meaning of the content than the form. For example, help content could be delivered in a doc or an email. Is it worth thinking through if there are multiple dimensions? Or should we allow multiple categories?

api.ts
```diff
@@ -322,6 +323,10 @@ export interface SyncTableDef<
 
   /** See {@link SyncTableOptions.description} */
   description?: string;
+  /** See {@link SyncTableOptions.contentCategory}
+   * @hidden
+   */
+  contentCategory?: ContentCategory;
```
ekoleda-codaio, line 329: Is there a logic behind putting this on the sync table vs in the schema? It would seem that this is describing the data, and not the table that syncs it. I don't feel too strongly though.

jonathan-codaio: Yeah, good point. We are not always consistent, e.g. identityName is very much about the schema but syntactically people preferred it on the table since it was required. I'm happy to see it moved to the schema, it does feel like it's annotating the data in a similar way to other semantic properties.

One pressure test though, would we imagine that we also categorize other building blocks like action formulas, in order to indicate that e.g. they are using for sending emails? If so, that would be labeling the formula, would it feel inconsistent if sync tables were labeled on the schema?

ekoleda-codaio: Maybe? As I understand it, this category is a targeted feature for helping Brain understand Pack output, and not part of some larger plan to apply taxonomy to building blocks, so that hypothetical feels unlikely. I'm fine with either approach, but I think it's good to have a rule of thumb on when to use one vs the other.

api_types.ts
chris-codaio: + @abhishek-codaio - any other categories you think we should consider short term to help with ranking / quality?

api_types.ts
drew-codaio: I was thinking of the Help content as similar to our help docs https://help.coda.io/en/. Some CMS driven system for support documentation. I can also remove this but I was thinking about it in that very specific form. Could also do a CMS content type for that sort of thing as well.

abhishek-codaio:
> ```ts
> index: {
>   properties: ['body'],
>   contextProperties: ['subject'],
>   contentCategorization: {
>     type: ContentCategorizationType.Email,
>     fromProperty: 'from',
>     toProperty: 'to',
>     ...
>   },
> }
> ```

I think those set of properties might be disjoint. So you either provide:
```ts
{
  properties: ['body'],
  contextProperties: ['subject'],
}
```

Or you provide:
```ts
{
  contentCategorization: {
    type: ContentCategorizationType.Email,
    fromProperty: 'from',
    toProperty: 'to',
    ...
  },
}
```

You'd take the former approach if you want custom control over the context properties, latter if you want to follow our recommendations for a given type of content.

But this location seems right to me.

schema.ts
```diff
@@ -1136,7 +1138,11 @@ export interface IndexDefinition {
    * so these should be the properties most likely to be useful as filters.
    */
   filterableProperties?: FilterableProperty[];
-  
+  /**
+   * The category of the content to be indexed. Used to determine how to support the indexing
+   * and querying for the text. Must be one of {@link ContentCategorization}.
+   */
+  contentCategorization?: ContentCategorization;
```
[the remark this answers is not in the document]
drew-codaio: Ah I see. I don't think everything will be mutually exclusive though. I can imagine that we'd still want some of the authorityNorm and popularityNorm values for boosting during query. It seems like when we have the contentCategorization we would not define `properties` and `contextProperties`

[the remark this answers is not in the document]
jonathan-codaio: Ok I buy that. So I think we maybe want:

```ts
interface BaseIndexDefinition {
  authorityNormProperty?: ...
  popularityNormProperty?: ...
}

interface CustomIndexDefinition extends BaseIndexDefinition {
  properties: ...
  contextProperties?: ...
}

interface CategorizedIndexDefinition extends BaseIndexDefinition {
  contentCategorization: ...
}

export type IndexDefinition = CustomIndexDefinition | CategorizedIndexDefinition;
```

Sound ok to folks?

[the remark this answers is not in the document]
drew-codaio: Would we have any issues with backwards compatibility?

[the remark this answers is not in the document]
jonathan-codaio: This definition should be backwards-compatible (please shout at me if it's not).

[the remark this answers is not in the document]
drew-codaio: It seemed to work with the tests so that's good enough for me...

schema.ts
```diff
@@ -1160,8 +1142,105 @@ export interface IndexDefinition {
...
+   * and querying for the text. Must be one of {@link ContentCategorization}.
+   */
+  contentCategorization: ContentCategorization;
+}
+
+/**
+ * Defines how to index objects for use with full-text indexing.
+ * @hidden
+ */
+export type IndexDefinition = CustomIndexDefinition | CategorizationIndexDefinition;
```
drew-codaio, line 1242: Not sure if we want to add a type here to make it a discriminated union as well.

jonathan-codaio: Discriminated unions are definitely preferable! For non-SDK code that would definitely be the way to go. For a developer-friendly SDK I'm on the fence if it's better, it's certainly clearer typing but it's more verbose.

We'd have to go update existing packs to add a new discriminant, if we like it better that way we can go back and do that but no need to do that right now.

schema.ts
```diff
@@ -2081,10 +2160,77 @@ function normalizeIndexProperty(value: IndexedProperty, normalizedProperties: Ob
...
+      const {
+        type: emailType,
+        toProperty,
+        fromProperty,
+        subjectProperty,
+        htmlBodyProperty,
+        plainTextBodyProperty,
+        ...emailRest
+      } = value;
+      ensureNever<keyof typeof emailRest>();
```
jonathan-codaio, line 2184: Nice, thank you.

schema.ts
```diff
@@ -2081,10 +2160,77 @@ function normalizeIndexProperty(value: IndexedProperty, normalizedProperties: Ob
...
+      return {type};
+    case ContentCategorizationType.Email: {
+      const {
+        type: emailType,
+        toProperty,
+        fromProperty,
+        subjectProperty,
+        htmlBodyProperty,
+        plainTextBodyProperty,
+        ...emailRest
```
jonathan-codaio, line 2182: Just FYI if you wanted to reuse the `rest` name here, I think it would work if you put curly brackets up above in the case on line 2170, the brackets create a nested namespace frame.

jonathan-codaio (approved):

PR #3188, opened 2025-04-29
[wip/rfc] draft of validateParameters

gary-codaio:
Some notes
- the context object that will be accessible to pack makers contains `sync?.permissionSyncMode` which may be relevant for pack makers deciding when parameters are valid


api.ts
```diff
@@ -1049,7 +1050,7 @@ export type TypedPackFormula = Formula | GenericSyncFormula;
 
 export type TypedObjectPackFormula = ObjectPackFormula<ParamDefs, Schema>;
 /** @hidden */
-export type PackFormulaMetadata = Omit<TypedPackFormula, 'execute' | 'executeUpdate' | 'executeGetPermissions'>;
+export type PackFormulaMetadata = Omit<TypedPackFormula, 'execute' | 'executeUpdate' | 'executeGetPermissions' | 'executeValidateParameters'>;
```
gary-codaio, line 1053: I am unclear why we do this to be honest.

api.ts
```diff
@@ -2193,14 +2206,14 @@ export interface SyncTableOptions<
    * This should describe the entities being synced. For example, a sync table that syncs products
    * from an e-commerce platform should be called 'Products'. This name must not contain spaces.
    *
-   * Important: This value acts as a unique ID for the table, and updating it later is a breaking change. 
+   * Important: This value acts as a unique ID for the table, and updating it later is a breaking change.
```
gary-codaio, line 2209: This _was_ a change made by Faraz for the display name change - I guess we have different trailing space settings in our IDEs

ekoleda-codaio:
Some initial thoughts:
1. I have a slight preference against using the `execute*` pattern for this function. It's been primarily used for the "heavy" workloads that do the more core functions of the building block. This feature feels a bit lighter, and more akin to the `propertyOptions` function on dynamic sync tables. It's true that there aren't many examples outside of dynamic sync tables, but I don't know if that's a reason to avoid it.

1. Is it feasible to add this for all formulas, vs just sync tables? It would also be useful to have something like this for Action formulas with the builder UI.

1. Would it make sense to add a new error type that would allow the developer to provide per-parameter error messages? That way the UI could show the errors closer to the actual value. Perahps something like:

    ```ts
    // The first argument is the top-level message, and the second is an optional map from parameter
    // name to error message.
    throw new coda.InvalidParametersError("You must specify either a product code or name.", {
      price: "The price must be a positive number.",
      email: "Invalid email address",
    });
    ```

api_types.ts
```diff
@@ -546,6 +546,20 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+   * This function takes in parameters and a context containing permissions mode information,
+   * and validates the parameters.
+   *
+   * @returns true if the parameters are valid
+   * @throws {@link ParameterValidationError} if the parameters are invalid
+   */
+  validateParameters?(
+    params: ParamValues<ParamDefs>,
+    context: ExecutionContext | SyncExecutionContext,
+  ): Promise<true>;
```
ekoleda-codaio, line 562: Opinion: Only being able to return `true` is a bit unexpected. I'd either expect `boolean` or `void`.

gary-codaio: Okay sounds good. Then I prefer void where returning without an error thrown means valid.

gary-codaio:
@ekoleda-codaio I played around with making this a metadata formula and ran into one point of friction. The typing for metadata formulas seems to make it hard for us to get the parameters of the parent formula if it's not a sync.

gary-codaio:
I'm going to start working on the coda side changes with this sha but still open to changes to this and will come back and fix the tests

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 - Remove `popularityNormProperty` and `authorityNormProperty` from top-level `schema` since they reside primarily in `schema.index`.
 - Remove deprecated `popularityRankProperty` from `schema`.
+- Added `validateParameters` to CommonPackFormulaDef
```
jonathan-codaio, line 11: Just note you'll have a merge conflict because Drew just released 1.9.5.

api.ts
```diff
@@ -99,6 +99,21 @@ export class UserVisibleError extends Error {
...
 }
 
+/**
+ * An error that occurs when validating parameters.
+ */
+export class ParameterValidationError extends UserVisibleError {
+  /**
+   * The parameters that were invalid, alongside a message describing the error for the parameter.
+   */
+  invalidParameters: Record<string, string>;
```
jonathan-codaio, line 109: Nit, I think it's a little more flexible and consistent to have this be an array of error objects. That allows for enriching the error data, for example by having enumerated error codes/types. And it's also kind of an established pattern, e.g. zod does this, and our own upload_validation.ts ultimately gives an array of errors, with each having a message and a path. So something like:

```ts
readonly errors: ParameterError[];
...
export interface ParameterError {
  message: string;
  parameterName: string;
}
```

api.ts
jonathan-codaio: I guess we fixed the linter for the `packs` repo to enforce a common `prettier` style via the CI, but didn't do that work for the SDK. We should do that at some point.

requirements.txt
```diff
@@ -1,29 +1,31 @@
 # Autogenerated by 'make bs'
 # If the format has changed, update MIN_PIPENV_VERSION in Makefile
 -i https://pypi.org/simple
-babel==2.16.0; python_version >= '3.8'
+babel==2.17.0; python_version >= '3.8'
```
jonathan-codaio, line 4: Do you know why any of this changed? Maybe will go away after a rebase.

api_types.ts
```diff
@@ -546,6 +546,20 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+   *
+   * This function takes in parameters and a context containing permissions mode information,
+   * and validates the parameters.
+   *
+   * @returns if the parameters are valid
+   * throws {@link ParameterValidationError} if the parameters are invalid.
+   */
+  validateParameters?(
+    params: ParamValues<ParamDefs>,
+    context: ExecutionContext | SyncExecutionContext,
```
jonathan-codaio, line 561: Could maybe follow up with this because the parameterization is annoying, but I think Oleg has successfully made our usage of the *Context types correct via generics and it would be nice not to regress that. That is, this isn't actually a union type, it's exactly one or the other, depending on whether this is a sync or not. So in other places you have usage like `MetadataFormula<ContextT extends ExecutionContext = ExecutionContext>`

api_types.ts
```diff
@@ -546,6 +546,20 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+  /**
+   * The JavaScript function that implements parameter validation
+   *
+   * This function takes in parameters and a context containing permissions mode information,
+   * and validates the parameters.
+   *
+   * @returns if the parameters are valid
+   * throws {@link ParameterValidationError} if the parameters are invalid.
+   */
+  validateParameters?(
```
jonathan-codaio, line 559: I feel like this is kind of a creating a third pattern for executable code:

* execute() and executeUpdate(): the heavy work, which take in parameter arrays. These live in the Formula object.
* metadata formulas: have a common interface for small side work. These live outside the Formula, or at least not at the top level (getSchema, listDynamicUrls, etc live at the SyncTable level; autocomplete lives inside the param def).
* validateParameters(): Does small side work, but lives inside the Formula, but has a unique interface.

And this is more of an implementation detail, but all metadata formulas thus far have been actual Formulas, i.e. makeMetadataFormula() turns a function into a Formula object which has metadata.

I don't love metadata formulas, I think the common interface has outlived its usefulness and I'm not sure if the Formula object wrapping has added value or not. But I'm not sure how I would describe the 3 different ways we have for executing code. @ekoleda-codaio thoughts?

runtime/thunk/thunk.ts
```diff
@@ -146,6 +147,24 @@ async function doFindAndExecutePackFunction<T extends FormulaSpecification>({
       );
       return response;
     }
+    case FormulaType.ValidateParameters: {
```
jonathan-codaio, line 150: Does implementation actually require a new top-level FormulaType for this? I was hoping we could use FormulaType.Metadata and just have a new MetadataFormulaType for this, but I don't know if that creates issues if the interface diverges from other metadata formulas.

api_types.ts
jonathan-codaio, line 559: You'll need to update something or other in metadata.ts. We generate a pure-JSON metadata file from the full pack def, which is what the client and various parts of the app use to know the contents of the pack without looking needing the executable code. So we need bits in the metadata JSON to know whether a given formula has a validator or not. If we wind up making this a metadata formula, you can reuse `compileMetadataFormulaMetadata()`,  otherwise you'll maybe want to do something similar to executeUpdate where we convert that to a `supportUpdates` boolean.

ekoleda-codaio: Regarding where it lives, I think it makes sense that it lives near the `parameters` themselves. Plus this makes it easy to use for both sync and non-sync formulas.

In [this comment](https://github.com/coda/packs-sdk/pull/3188#issuecomment-2840394793) I suggested making this a metadata formula, but [here](https://github.com/coda/packs-sdk/pull/3188#issuecomment-2847880845) @gary-codaio said he ran into some issues implementing that. I agree that the common metadata formula isn't perfect, but I'd prefer using it here if possible just to avoid introducing another pattern.

gary-codaio: From my attempt, I think it's possible to add the parameters of the attached formula if an attached formula exists to all metadata formulas. The current typing reflects the fact that metadata formulas like getConnectionName for example are not attached to a formula.

gary-codaio:
@jonathan-codaio @ekoleda-codaio made the change to a metadata formula

api.ts
```diff
@@ -1737,7 +1762,7 @@ export type MetadataContext = Record<string, any> & {__brand: 'MetadataContext'}
 /**
  * The type of values that can be returned from a {@link MetadataFormula}.
  */
-export type MetadataFormulaResultType = string | number | MetadataFormulaObjectResultType;
+export type MetadataFormulaResultType = string | number | MetadataFormulaObjectResultType | boolean;
```
gary-codaio, line 1765: The intention is that validateParameters will return a boolean

test/execution_test.ts
```diff
@@ -776,6 +776,67 @@ describe('Execution', () => {
           });
         });
 
+        it('context includes sync for validateParameters on sync formula', async () => {
```
gary-codaio, line 779: Tests are failing locally so looking into that, but wanted to get the shift over to metadata formula reviewed

testing/upload_validation.ts
```diff
@@ -1388,7 +1388,7 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
     zodCompleteStrictObject<CommentContentCategorization>({
       type: z.literal(ContentCategorizationType.Comment),
     }),
-  ]).refine(data => 
+  ]).refine(data =>
```
gary-codaio, line 1391: more trailing whitespace :/

api_types.ts
```diff
@@ -546,6 +546,17 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
    * @hidden
    */
   readonly allowedAuthenticationNames?: string[];
+
+  /**
+   * The JavaScript function that implements parameter validation
+   *
+   * This function takes in parameters and a context containing permissions mode information,
```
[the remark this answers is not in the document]
gary-codaio: On a sync execution context we have an attribute called `permissionSyncMode` which is what I'm referring to here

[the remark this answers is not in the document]
ekoleda-codaio: Ah, I see. Is the underlying thought here that devs may want to have different validation criteria for regular vs meta syncs? If so, we may want to say something like that more explicitly.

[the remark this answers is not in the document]
gary-codaio: Yes for our first party packs, we are already planning on disallowing personal ingestions from having certain parameters blank for example. I'll draft a more detailed explanation

api.ts
jonathan-codaio, line 1765: So my fear that I mentioned last week is that merely adding to the return type here makes metadata formula typing worse across the board. It's not the case that other metadata formulas can return booleans, only this one can.

We have this problem already, for example getConnectionName should only be able to return a string but it's typed as being able to return arrays and objects.

So I was hoping we would stem the bleeding here and allow for return type parameterization, so that this new metadata formula is typed only as returning a boolean and we don't introduce boolean as a new return type for existing metadata formulas.

If it's too gross, I suppose we could introduce boolean here as-is and do the parameterization cleanup separately.

jonathan-codaio: @gary-codaio Is this sufficient to allow declaring the new formula with a parameterized type of only `boolean` and have it behave properly? https://github.com/coda/packs-sdk/pull/new/jlg-metadata-parameterizatin

gary-codaio: I think that works for me to use. Want to merge or should I just add that in this change?

jonathan-codaio: Would just add it to your change.

api_types.ts
```diff
@@ -546,6 +546,17 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+   * and validates the parameters.
+   *
+   * @returns if the parameters are valid
+   * throws {@link ParameterValidationError} if the parameters are invalid.
+   */
+  validateParameters?: MetadataFormula;
```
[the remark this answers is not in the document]
gary-codaio: Used MetadataFunction

runtime/thunk/thunk.ts
```diff
@@ -307,7 +308,13 @@ async function doFindAndExecutePackFunction<T extends FormulaSpecification>({
           }
 
           break;
-
+        case MetadataFormulaType.ValidateParameters: {
```
gary-codaio, line 311: Before running a formula/sync, should we run the validateParameters metadata formula in here? Or only do that in the coda repo

api.ts
```diff
@@ -1847,31 +1879,46 @@ export type PropertyOptionsMetadataFormula<SchemaT extends Schema> = ObjectPackF
...
 /**
  * @hidden
  */
-export declare type GenericMetadataFormulaMetadata = Omit<GenericMetadataFormula, 'execute'>;
+export declare type GenericMetadataFormulaMetadata = Omit<GenericMetadataFormula, 'execute' | 'validateParameters'>;
+
+/**
+ * Historically, metadata formulas could return a variety of types.
+ * This type is used to represent the legacy return types.
+ * Metadata formulas now are generic and can return any value.
```
jonathan-codaio, line 1892: Nit, maybe "Metadata formulas now work with generics and in the future will be updated to declare a more precise return type for each formula."

api_types.ts
```diff
@@ -546,6 +547,35 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+   *   let errors = [];
+   *   if (quantity < 0) {
+   *     errors.push({message: "Must be a positive number.", propertyName: "quantity"});
+   *   }
+   *   if (!isValidSku(context, sku)) {
+   *     errors.push({message: `Product SKU not found.`, propertyName: "sku"});
+   *   }
+   *   if (errors.length > 0) {
+   *     throw new ParameterValidationError("Invalid parameter values.", errors);
+   *   }
```
jonathan-codaio, line 574: This example doesn't return a boolean, do you want to return `true` at the end, or do you prefer if we just make this a void function?

api_types.ts
```diff
@@ -546,6 +547,35 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+   * This function takes in parameters and a context containing a {@link PermissionSyncMode}
+   * and validates the parameters. A formula may want to validate parameters differently
+   * for permissionSyncMode 'PermissionAware' vs 'Personal' vs undefined (which represents a formula).
+   *
+   * @returns if the parameters are valid
+   * @throws {@link ParameterValidationError} if the parameters are invalid.
+   *
+   * @example
+   * ```
+   * valiateParameters: async function (context, _, formulaContext) {
```
jonathan-codaio, line 563: validate

api_types.ts
```diff
@@ -546,6 +547,35 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+   *   if (!isValidSku(context, sku)) {
+   *     errors.push({message: `Product SKU not found.`, propertyName: "sku"});
+   *   }
+   *   if (errors.length > 0) {
+   *     throw new ParameterValidationError("Invalid parameter values.", errors);
+   *   }
+   * },
+   * ```
+   */
+  validateParameters?: MetadataFunction<ExecutionContext, boolean>;
```
jonathan-codaio, line 578: Hmm, if we want to keep this consistent with other metadata formulas, wouldn't this be `MetadataFormulaDef`, and then we'd auto-wrap this with `makeMetadataFormula` if we see the function version like we do for other metadata formulas?

Right now it looks like you're wrapping only when making the pack metadata, but for other metadata formulas we wrap and convert to a formula object at import time. I agree that the object representation is kind of obsolete at this point but I think we might as well be consistent.

requirements.txt
jonathan-codaio: Were you able to figure out why these diffs are showing?

runtime/thunk/thunk.ts
jonathan-codaio: I don't think we should auto-run it in the SDK, let's let the `coda` side orchestrate that, gives us more control.

runtime/thunk/thunk.ts
```diff
@@ -166,12 +166,13 @@ async function doFindAndExecutePackFunction<T extends FormulaSpecification>({
             return selectedAuthentication.getConnectionUserId.execute(params as [string, string], executionContext);
           }
           break;
-        case MetadataFormulaType.ParameterAutocomplete:
-          const parentFormula = findParentFormula(manifest, formulaSpec);
```
jonathan-codaio, line 170: Oh geez, was this mis-named before?

jonathan-codaio (commented):
Nice

runtime/thunk/thunk.ts
gary-codaio: Previously findParentFormula was only used for ParameterAutocomplete. For ValidateParameters, we use much of the same logic so this was a side effect of extracting out findParentFormula to be used in both helpers.

requirements.txt
gary-codaio: Oh I had reverted the changes to this file and actually thought these changes were not re-introduced

api_types.ts
gary-codaio: `wrapMetadataFunction` seems to be what I needed, didn't know that existed before

helpers/metadata.ts
```diff
@@ -49,19 +50,24 @@ function compileFormatsMetadata(formats: Format[]): PackFormatMetadata[] {
...
+export function compileFormulasMetadata(formulas: Formula[]): PackFormulaMetadata[] {
   return formulas.map(compileFormulaMetadata);
 }
 
-function compileFormulaMetadata(formula: TypedPackFormula): PackFormulaMetadata {
-  const {execute, ...rest} = formula;
-  return rest;
+export function compileFormulaMetadata(formula: TypedPackFormula): PackFormulaMetadata {
+  const {execute, validateParameters, ...rest} = formula;
+  const validateParametersAsFormula = wrapMetadataFunction(validateParameters);
```
jonathan-codaio, line 59: Hmm not sure why wrapMetadataFunction is spread around, for all the other metadata formulas it's done only once in api.ts. So would imagine we would only do the wrapping once, in `makeFormula()` probably. Does that work or is there some complication?

gary-codaio: Looked at getSchema for example; on SyncTableDef, it is typed as MetadataFormula. It's only MetadataFunctionDef on DynamicOptions which is what gets used in the builder. I guess I need to change what the builder/legacy helpers takes as an argument and what CommonPackFormulaDef has to be more similar. Trying that now

api.ts
```diff
@@ -1660,15 +1683,15 @@ export interface BaseFormulaDef<ParamDefsT extends ParamDefs, ResultT extends st
 export type StringFormulaDef<ParamDefsT extends ParamDefs> = BaseFormulaDef<ParamDefsT, string> & {
   resultType: ValueType.String;
   execute(params: ParamValues<ParamDefsT>, context: ExecutionContext): Promise<string> | string;
-} & ({schema?: StringSchema} | {codaType?: StringHintTypes});
+} & {schema?: StringSchema; codaType?: StringHintTypes};
```
jonathan-codaio, line 1686: How come these are changing? I think these were intentionally mutually-exclusive before.

api_types.ts
```diff
@@ -576,7 +575,7 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
    * },
    * ```
    */
-  validateParameters?: MetadataFormulaDef<ExecutionContext, boolean>;
+  validateParameters?: MetadataFormula;
```
jonathan-codaio, line 578: Can this be parameterized with `boolean`?

api.ts
```diff
@@ -1426,9 +1458,12 @@ export type SyncFormula<
  * @param definition The definition of a formula that returns a number.
  */
 export function makeNumericFormula<ParamDefsT extends ParamDefs>(
-  definition: PackFormulaDef<ParamDefsT, number>,
+  definition: FormulaOptions<ParamDefsT, PackFormulaDef<ParamDefsT, number>>,
 ): NumericPackFormula<ParamDefsT> {
-  return Object.assign({}, definition, {resultType: Type.number as Type.number}) as NumericPackFormula<ParamDefsT>;
+  return Object.assign({}, definition, {
+    resultType: Type.number as Type.number,
+    validateParameters: wrapMetadataFunction(definition.validateParameters),
```
jonathan-codaio, line 1465: If it helps, I'm definitely ok if validateParameters isn't available in the deprecated legacy makeXYZFormula() wrappers and is only offered in the builder + makeFormula. I don't think we need to continue adding new functionality to those, we can migrate 1PP code to makeFormula if that functionality is needed, and 3PPs don't use these anyway.

jonathan-codaio (commented):
Sorry this was such a pain, this was my fear for adding a new metadata formula, but hopefully if we get through this we can rinse-and-repeat for future ones.

api.ts
gary-codaio: I'm not sure if I'm not writing this well or if TS is not able to recognize something that a human is able to recognize. When I use FormulaOptions<ParamDefsT, StringFormulaDef<...>>, TS makes codaType and schema unknown even if we check 'codaType' in fullDefinition. I've reverted this now and have added some types to cast to that make TS happy.

api_types.ts
gary-codaio: Needed to make MetadataFormula support a second generic parameter. I added that with any as the default and applied it here but leaving remaining usage untouched

api.ts
gary-codaio: Changes from this round of review did not necessitate anything extra to make the make... wrapper work and since I've gone ahead and added support already, will leave it

gary-codaio: I spoke too soon, it does require some, let me see what's easier

api.ts
```diff
@@ -1683,15 +1686,15 @@ export interface BaseFormulaDef<ParamDefsT extends ParamDefs, ResultT extends st
 export type StringFormulaDef<ParamDefsT extends ParamDefs> = BaseFormulaDef<ParamDefsT, string> & {
   resultType: ValueType.String;
   execute(params: ParamValues<ParamDefsT>, context: ExecutionContext): Promise<string> | string;
-} & {schema?: StringSchema; codaType?: StringHintTypes};
```
gary-codaio: moved these to FormulaOptions. The only other usage of StringFormulaDef and NumericFormulaDef is:
```
const def: StringFormulaDef<ParamDefsT> & {codaType?: StringHintTypes; formulaSchema?: StringSchema} = {
        ...fullDefinition,
        codaType: 'codaType' in fullDefinition ? fullDefinition.codaType : undefined,
        formulaSchema: 'schema' in fullDefinition ? fullDefinition.schema : undefined,
        validateParameters: wrapMetadataFunction(fullDefinition.validateParameters),
      };
      const {onError: _, resultType: unused, codaType, formulaSchema, ...rest} = def;
```
which means we union codaType and do not need schema there anyways

api.ts
```diff
@@ -1920,7 +2011,7 @@ export function makeMetadataFormula<ContextT extends ExecutionContext>(
     ],
     examples: [],
     connectionRequirement: options?.connectionRequirement || ConnectionRequirement.Optional,
-  });
+  }) as MetadataFormula<ContextT, ReturnT>;
```
gary-codaio, line 2014: This was something that I resorted to due to TS complaining and not being able to resolve it another way. Left a comment above

helpers/sample_utils.ts
```diff
@@ -22,8 +25,12 @@ export function fakeDefinitionToMetadata(def: FakePackDefinition): PackMetadata
     ...packMetadata
   } = def;
   const formulas = originalFormulas!.map(formula => {
-    const {execute, ...formulaMetadata} = formula;
-    return formulaMetadata;
+    const {execute, validateParameters, ...formulaMetadata} = formula;
+    const validateParametersAsFormula = wrapMetadataFunction<ExecutionContext, boolean>(validateParameters);
```
jonathan-codaio, line 29: How come we never needed to use wrapMetadataFunction here for the existing metadata formulas? Is it just because this file kinda ignores sync tables or is sloppy about them?

index.ts
```diff
@@ -130,6 +131,9 @@ export {Type} from './api_types';
 export type {TypedPackFormula} from './api';
 export type {Formula} from './api';
 export {UserVisibleError} from './api';
+export {ParameterValidationError} from './api';
+export {ParameterError} from './api';
+export {LegacyDefaultMetadataReturnType} from './api';
```
jonathan-codaio, line 136: Do we need to export this in the package index? Generally preferable not to export intermediate internal types.

testing/validation.ts
```diff
@@ -67,9 +68,10 @@ export function validateResult<ResultT extends any>(formula: TypedPackFormula, r
     throw ResultValidationException.fromErrors(formula.name, [maybeError]);
   }
 
-  if (isObjectPackFormula(formula)) {
```
jonathan-codaio, line 70: This was just buggy before?

api_types.ts
```diff
@@ -546,6 +546,36 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
...
+   *     errors.push({message: `Product SKU not found.`, propertyName: "sku"});
+   *   }
+   *   if (errors.length > 0) {
+   *     throw new ParameterValidationError("Invalid parameter values.", errors);
+   *   }
+   * return true;
+   * },
+   * ```
+   */
+  validateParameters?: MetadataFormula<ExecutionContext, boolean>;
```
jonathan-codaio, line 578: For a sync table definition, do we need to override `ExecutionContext` with `SyncExecutionContext`?

api.ts
jonathan-codaio: Did you get things working with the legacy wrappers in the end?

api.ts
jonathan-codaio: All makes sense, can you just add a comment above this line to remove the cast when `makeObjectFormula()` is removed?

jonathan-codaio (commented):
Looking good.

Thanks for trying this out on `coda`, can you also try it out on `packs`? I'm actually more concerned about the effects there. Left a couple comments on the `coda` draft.

api.ts
```diff
@@ -1767,9 +1831,12 @@ export type MetadataFormulaResultType = string | number | MetadataFormulaObjectR
  * values of the others. This is dictionary mapping the names of each parameter to its
  * current value.
  */
-export type MetadataFormula<ContextT extends ExecutionContext = ExecutionContext> = BaseFormula<
+export type MetadataFormula<
+  ContextT extends ExecutionContext = ExecutionContext,
+  ResultT extends PackFormulaResult = LegacyDefaultMetadataReturnType,
```
gary-codaio, line 1836: This change was causing some downstream changes, as TS is not always able to determine if a return type is compatible with LegacyDeafultMetadataReturnType. Thinking it's still preferred to have this be the default and eliminate a usage of `any`; this just means we have to explicitly say what the return type is for some metadata formulas, the one i've seen most often is getConnectionName

helpers/sample_utils.ts
gary-codaio: We handle getConnectionName by stripping out its execute attribute. For the sync table metadata formulas, the input type for sync tables doesn't seem to accept a MetadataFunction, only MetadataFormulas already

testing/validation.ts
gary-codaio: This change is unnecessary. Reverting

api.ts
gary-codaio: Yes validateParameters is still compatible with them.

gary-codaio:
In the packs repo: https://github.com/coda/packs/pull/4975

api_types.ts
jonathan-codaio: Yeah for precision is what I meant, it's actually important because our main use case is validating sync table and we probably need the extra bits from `SyncExecutionContext` to do some of our validation and we wouldn't want the packs to have to do a cast.

jonathan-codaio (approved):
`packs` change was nice and clean!

So just confirming, if a pack uses modern builder syntax, this is fully backwards-compatible, right?

gary-codaio:
@jonathan-codaio there was a getConnectionName in the packs repo that I kept trying to resolve today. I could only trace it back to some autocomplete type mismatching on a getConnectionName when using a setUserAuthentication. I wanted to time box looking into it for now; just going to unrefine the definition of the getConnectionName MetadataFormula for now which was an unrelated change.

PR #3192, opened 2025-05-13
include validateParameters in zod commonPackFormulaSchema otherwise it gets stripped out on upload

gary-codaio:
This is used in `validatePackVersionMetadata` which is used on pack upload to strip out extraneous fields. I do not want validateParameters to be stripped out

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -30,6 +30,7 @@ import type {EmailContentCategorization} from '../schema';
 import {EmailDisplayType} from '../schema';
 import type {EmailSchema} from '../schema';
 import {FeatureSet} from '../types';
+import type {FormulaOptions} from '../api';
 import type {GenericObjectSchema} from '../schema';
 import type {GenericSyncFormula} from '..';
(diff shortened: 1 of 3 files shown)


dweitzman-codaio (approved):

testing/upload_validation.ts
```diff
@@ -999,9 +1000,11 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
     isSystem: z.boolean().optional(),
     extraOAuthScopes: z.array(z.string()).optional(),
     allowedAuthenticationNames: z.array(z.string()).optional(),
+    // Has to be any to avoid circular dependency.
+    validateParameters: z.any().optional(),
```
jonathan-codaio, line 1004: Hmm we should play with this a bit after you're unblocked. I think we should separate out a base schema that doesn't include `validateParameters`, so we can properly validate this. With this change we're also actually allowing you to recursively put validateParameters on validateParameters, which I don't think makes sense either.

jonathan-codaio (approved):

PR #3193, opened 2025-05-14
metadata context should just be params

gary-codaio:


api_types.ts
```diff
@@ -559,8 +559,8 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
    *
    * @example
    * ```
-   * validateParameters: async function (context, _, formulaContext) {
-   *   let [quantity, sku] = formulaContext?.params;
+   * validateParameters: async function (context, _, params) {
+   *   let {quantity, sku} = ensureExists(params);
```
gary-codaio, line 563: this ensureExists is because metadataContext may be undefined for metadata formulas even though it should never be for validateParameters

runtime/thunk/thunk.ts
```diff
@@ -313,7 +313,7 @@ async function doFindAndExecutePackFunction<T extends FormulaSpecification>({
           const validateParametersFormula =
             wrapMetadataFunction<ExecutionContext, boolean>(findValidateParametersFormula(manifest, formulaSpec));
           if (validateParametersFormula) {
-            return validateParametersFormula.execute(['', JSON.stringify({params: params as any, __brand: 'MetadataContext'})], executionContext);
+            return validateParametersFormula.execute(params as any, executionContext);
```
gary-codaio, line 316: this much more closely matches what we do with other params and in `executeFormulaOrSyncWithRawParams` we have this note which explains why this execute doesn't need three arguments when a metadataformula defines 3 arguments


```
case FormulaType.Metadata: {
      // Interestingly we don't need special handling for the formula context dict (the optional second arg
      // to an autocomplete metadata formula), because at execution time it gets passed as a serialized
      // JSON string anyway which is already parsed by the compiled pack definition.
      params = rawParams as ParamValues<ParamDefs>;
      // Default the search string (first arg) to an empty string.
      params[0] ??= '';
      break;
    }
```

test/execution_test.ts
```diff
@@ -781,7 +781,7 @@ describe('Execution', () => {
           await executeFormulaOrSyncFromCLI({
             vm,
             formulaName: 'Students:validateParameters',
-            params: ['Personal'],
+            params: ['', '{"teacher": "Personal"}'],
```
gary-codaio, line 784: second argument is the serialized json metadata context string

api_types.ts
jonathan-codaio: Not for this PR but can you check whether we're always passing a metadata context into all metadata formulas in practice? Likely just `{}`? If so we could just update the type to not be optional and this would be much more ergonomic. Or I guess we could just make it so. Otherwise we should update the ContextT here somehow so that in this case it's not optional. This isn't very ergonomic otherwise.

test/execution_test.ts
```diff
@@ -781,7 +781,7 @@ describe('Execution', () => {
           await executeFormulaOrSyncFromCLI({
             vm,
             formulaName: 'Students:validateParameters',
-            params: ['Personal'],
+            params: ['', '{"teacher": "Personal", "__brand": "MetadataContext"}'],
```
jonathan-codaio, line 784: Can you remove the `__brand` here (casting as necessary)? The idea of a brand is that it exists only in the type system, not at runtime.

jonathan-codaio (approved):

api_types.ts
gary-codaio: Noting it down to look into this

PR #3194, opened 2025-05-15
Update AWS AssumeRole auth to include the packId in externalId

chris-codaio:
This is blocked on figuring out how to plumb the packId through all of the test execution code in here.  Hopefully someone on the packs team can help out @jonathan-codaio ?

### DIFF
--- a/docs/guides/basics/authentication/aws.md
+++ b/docs/guides/basics/authentication/aws.md
@@ -9,7 +9,6 @@ APIs that are part of Amazon Web Services (AWS) use a proprietary authentication
 
 The SDK supports two different ways of authenticating with an AWS service: access keys and by assuming a role.
 
-
 ## Access key
 
 The simplest method of authentication requires that the Coda user provides an access key and secret associated with an AWS user. Coda will use these values to sign outgoing requests. Simply set the authentication type to [`AWSAccessKey`][sdk_awsaccesskey] and specify the AWS service you are connecting to.
@@ -17,7 +16,7 @@ The simplest method of authentication requires that the Coda user provides an ac
 ```ts
 pack.setUserAuthentication({
   type: coda.AuthenticationType.AWSAccessKey,
-  service: "s3",
+  service: 's3',
 });
 ```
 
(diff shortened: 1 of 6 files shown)


jonathan-codaio (commented):
I just pushed up a commit to plumb the pack id, might testing out if it works end to end?

The pack id needs to be optional because part of point of the CLI's `execute` command is that you can run the pack before you officially create it. So the typing is updated to reflect this.

I'd suggest perhaps throwing an error in the fetcher if you don't get a pack id and prompt the user to do `coda create` to get a real pack id first so they can properly simulate the fetch (though maybe it's pointless if you're not fetching from Coda's AWS in the simulator?)

chris-codaio:
Wonder if this is even feasible for most folks to run.  They need to be hooked up to AWS auth already in their local environment, so might not be worth it until someone asks?

jonathan-codaio:
> Wonder if this is even feasible for most folks to run. They need to be hooked up to AWS auth already in their local environment, so might not be worth it until someone asks?

Yeah, fine if it's unlikely to work in practice, I just think we should do something very minimal to set expectations for the user. If we think this is hopeless, we can just fail in the fetcher with an error that says this is not supported, though I guess in theory they can try to configure things to work so I would just do something like this:

```
ensure(packId !== undefined, 'You must run `coda create` to be assigned a Pack ID before using AWSAssumeRole locally because the Pack ID is part of the ExternalId.');
```

docs/guides/basics/authentication/aws.md
```diff
@@ -9,15 +9,14 @@ APIs that are part of Amazon Web Services (AWS) use a proprietary authentication
...
-
 ## Access key
 
 The simplest method of authentication requires that the Coda user provides an access key and secret associated with an AWS user. Coda will use these values to sign outgoing requests. Simply set the authentication type to [`AWSAccessKey`][sdk_awsaccesskey] and specify the AWS service you are connecting to.
 
 ```ts
 pack.setUserAuthentication({
   type: coda.AuthenticationType.AWSAccessKey,
-  service: "s3",
+  service: 's3',
```
ekoleda-codaio, line 19: Drive by: The double quotes are correct here, as sample code has a different style guide:

https://coda.io/packs/build/latest/support/contributing/style/#use-double-quotes

I assume that your IDE did this automatically though?

PR #3195, opened 2025-05-19
For validateParameters, add search string to params from cli so first argument can just be the metadata context string

gary-codaio:
This is so we can do
```
npx coda execute manifest.ts SomeFormula:validateParameters '{"blah":"blah"}'
```
instead of
```
npx coda execute manifest.ts SomeFormula:validateParameters '' '{"blah":"blah"}'
```
in the cli. Let me know if this is actually undesired

### DIFF
--- a/dist/testing/execution.js
+++ b/dist/testing/execution.js
@@ -444,6 +444,10 @@ async function executeFormulaOrSyncWithRawParamsInVM({ formulaSpecification, par
             params = rawParams;
             // Default the search string (first arg) to an empty string.
             (_a = params[0]) !== null && _a !== void 0 ? _a : (params[0] = '');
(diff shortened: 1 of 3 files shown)


jonathan-codaio (commented):
Yeah unfortunately we can't do this without a larger change because of the inconsistency with other metadata formulas. It's not that it's a convention, it's that the type defs enforce it, if you look at `MetadataFunction`, it declares that the params are `(ExecutionContext, searchString, MetadataContext)`. So this would violate the typing.

Overall I kinda wonder about the usefulness of MetadataFormula/MetadataFunction as types, vs just having specific types for each actual occurrence of the formula. But given the effort involved just getting this far, I think we should punt that for another day.

gary-codaio:
Ack, abandoning this

PR #3198, opened 2025-05-23
Generate llms.txt for docs

ekoleda-codaio:
This PR enables the mkdocs plugin [mkdocs-llmstxt](https://github.com/pawamoy/mkdocs-llmstxt), to generate an llms.txt file for the docs. [llms.txt](https://llmstxt.org/) is a new pseudo-standard to make it easier for LLMs to access your site's content, primarily by publishing a markdown version of all the content in the site. 

Here's the file it generates as of this PR: [llms.txt](https://github.com/user-attachments/files/20402397/llms.txt)

It also generates `llms-full.txt`, which is a convention where you publish the full content of your site as a single markdown file.

This PR was inspired by some comments I saw in a DevRel Slack group about making your content accessible to LLMs.


mkdocs.yml
```diff
@@ -57,6 +57,26 @@ plugins:
...
         filename: .nav.yml
+    - llmstxt:
+        markdown_description: > 
+          A Pack is an extension that adds new powers to your doc.
+          What other platforms call plugins, add-ons, or extensions, we call "Packs". 
+          A Pack extends the core building blocks of Coda, allowing users to build more powerful and integrated docs. 
+          Anyone can build a Pack, and they can use it in their own docs, share it with their team, or publish it to the world.
+          Packs are built using JavaScript or TypeScript, either on your local machine or in the browser, and run on Coda's own servers. 
+          We have a gallery where users can discover and install Packs, as well as a robust set of admin controls.
+        full_output: llms-full.txt
```
jonathan-codaio, line 68: Is this output file meant to be checked in, or does it only get generated by the release tooling?

jonathan-codaio (commented):
Cool

mkdocs.yml
ekoleda-codaio: Yes, the file needs a preamble about what the documentation is about. This config file is hand-written.

jonathan-codaio: Sorry I just meant, this line indicates that there is a file called llms-full.txt but there isn't one in this PR (nor an llms.txt actually) so I was just wondering if you need to `git add` those or if the release tooling generates them such that they don't need to be in the repo.

ekoleda-codaio: Ah, I see. Both of those files are generated by this MkDocs plugin during the doc build process. Like the other doc artifacts, we currently don't check these into version control, but generate them on-the-fly as part of the build and deploy process.

PR #3203, opened 2025-06-06
add brain feature set id to sync context

gary-codaio:
TBD whether this will be useful in syncs, but I'm reasonably confident this will be useful for validateParameters


gary-codaio:
Decided to leave billing information out of the sdk, handle within the system

PR #3205, opened 2025-06-07
Add email parameter type

himanshu-codaio:
Adding the parameter type so we can start using it in various functions etc as per the discussion [here](https://staging.coda.io/d/No-more-formulas_dZaNYO0JsEi/Packs-opportunities_su5wN2D3#Packs-opportunities_tup7hnyx/r10&view=center)


jonathan-codaio:
Code looks good but in the linked thread @ekoleda-codaio had mentioned the idea of using `ParameterType.Person` which would provide a `{name, email}` object, so you can kind of get the best of both worlds if these params are populated from people table references. Is that the intended use here?

ekoleda-codaio:
@himanshu-codaio - Sorry for not catching it sooner, but can you also add an entry to the [CHANGELOG](https://github.com/coda/packs-sdk/blob/main/CHANGELOG.md) for this new addition?

PR #3206, opened 2025-06-10
upload validation only sync tables can have validateParameters

gary-codaio:
initially had hoped to support this in all formulas but coda support needs more refinement so will start with only sync tables

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -1936,7 +1936,9 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
         }
         paramNames.add(param.name);
       }
-    });
+    }).refine(val => {
+      return !val.validateParameters;
+    }, {message: 'validateParameters is only allowed on sync formulas.'});
 
   const formatMetadataSchema = zodCompleteObject<PackFormatMetadata>({
     name: z.string().max(Limits.BuildingBlockName),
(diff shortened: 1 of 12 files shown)


testing/upload_validation.ts
```diff
@@ -1918,7 +1924,9 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
     // TODO(jonathan): See if we should really allow this. The SDK right now explicitly tolerates an undefined
     // schema for objects, but that doesn't seem like a use case we actually want to support.
     schema: z.union([genericObjectSchema, arrayPropertySchema]).optional(),
-  });
+  }).refine(val => {
+    return !val.validateParameters;
+  }, {message: 'validateParameters is only allowed on sync formulas.'});
 
   const formulaMetadataSchema = z
```
jonathan-codaio: Can you just do the check here instead of in the separate union members?

gshen437: Yes done

ekoleda-codaio:
Is it possible to remove `validateParameters` in regular formulas if it isn't supported yet? Or at least update the docs to mention that it's only supported in sync formulas?

gary-codaio:
@ekoleda-codaio Updated docs. I prefer to keep it on regular formulas since we have all the plumbing set up for it to run but we are choosing not to in order to limit scope and invocation cost at this time.

ekoleda-codaio (approved):

PR #3213, opened 2025-06-17
easy packs-sdk validate parameters feedback

gary-codaio:
- Remove Error from naming to avoid confusion that you do not throw a ParameterError.
- make errors optional, will update coda support to treat this by just showing the `message` which is required

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -102,7 +102,7 @@ export class UserVisibleError extends Error {
 /**
  * An error that occurs when validating parameters.
  */
-export interface ParameterError {
+export interface ParameterValidationDetail {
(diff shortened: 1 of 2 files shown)


jonathan-codaio (approved):
You'll need to bump the prerelease semantic version too.

gary-codaio:
@jonathan-codaio @ekoleda-codaio Do we want to release this as a named version so it's easier to use now? Or are we okay to wait given this is so minor?

jonathan-codaio (commented):
I don't think there's a hurry, I'm sure we'll wind up doing a release in the next 1-2 weeks anyway.

PR #3247, opened 2025-07-02
Change support for filterable properties

sam-codaio:
* Add arbitrary string support
* Remove arbitrary select list support

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -1866,25 +1866,9 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
               filterableProperty,
               'filterableProperty',
               filterablePropertySchema => {
-                if (filterablePropertySchema.type === ValueType.Boolean) {
-                  return true;
-                }
-                if (filterablePropertySchema.type === ValueType.Number) {
-                  return true;
-                }
-                if (!('codaType' in filterablePropertySchema && filterablePropertySchema.codaType)) {
-                  return false;
-                }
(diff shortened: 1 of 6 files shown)


wesley-codaio (commented):
LGTM, will defer to someone closer to the packs SDK code

package.json
```diff
@@ -103,6 +100,7 @@
     "eslint-plugin-local": "1.0.0",
     "eslint-plugin-prefer-let": "3.0.1",
     "glob": "10.3.10",
+    "isolated-vm": "5.0.1",
```
jonathan-codaio: Make sure to revert this, this is the way it is because it breaks a lot of external developers on various platforms.

testing/upload_validation.ts
```diff
@@ -1872,19 +1879,12 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
-                  return false;
-                }
-                switch (filterablePropertySchema.codaType) {
-                  case ValueHintType.DateTime:
-                  case ValueHintType.Date:
-                  case ValueHintType.SelectList:
-                    return true;
-                  default:
-                    return false;
+                if (filterablePropertySchema.type === ValueType.String) {
```
jonathan-codaio: Nit, this is silly if we're not doing a switch, can just do `return [ValueType.String, ValueType.Number, ValueType.Boolean].contains(filterablePropertySchema.type);`

jonathan-codaio (approved):

package.json
sam-codaio: This did not resolve the CI issues anyways

PR #3253, opened 2025-07-02
Use and enforce prettier on packs-sdk

jonathan-codaio:
We see a lot of whitespace/style thrashing across PRs. Despite there being a .prettierrc.js I guess without prettier actually being installed in the repo certain developer setups were not using it.

So this PR:
- Makes prettier a devDependency so it should always be installed.
- Adds Make commands to both autoformat and check formatting (copied from `packs`).
- Runs prettier on git-staged files on precommit.
- Enforces prettier formatting on the CI.

This should bring us to parity with other repos.

PTAL @dweitzman-codaio @coda/orchestra-platform

### DIFF
--- a/bundles/thunk_bundle.js
+++ b/bundles/thunk_bundle.js
@@ -6481,15 +6481,7 @@ module.exports = (() => {
         return { type };
       }
       case "Email" /* Email */: {
-        const {
-          type,
-          toProperty,
-          fromProperty,
-          subjectProperty,
-          htmlBodyProperty,
-          plainTextBodyProperty,
-          ...rest
-        } = value;
+        const { type, toProperty, fromProperty, subjectProperty, htmlBodyProperty, plainTextBodyProperty, ...rest } = value;
         ensureNever();
         return {
           type,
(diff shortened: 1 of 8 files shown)


schema.ts
```diff
@@ -2173,15 +2172,7 @@ function normalizeContentCategorization(
...
-      const {
-        type,
-        toProperty,
-        fromProperty,
-        subjectProperty,
-        htmlBodyProperty,
-        plainTextBodyProperty,
-        ...rest
-      } = value;
+      const {type, toProperty, fromProperty, subjectProperty, htmlBodyProperty, plainTextBodyProperty, ...rest} = value;
```
dweitzman-codaio, line 2175: For some of these I kind of like the old formatting of putting each value on a new line for easier diffs and blame history as these change, but the benefits of using prettier everywhere still seem bigger

dweitzman-codaio (approved):

schema.ts
jonathan-codaio: I'm supportive of that, to my knowledge that might be an eslint rule that controls that, I'd be happy to have it ported over here.

PR #3261, opened 2025-07-14
Proposed structure for account

sam-codaio:
* Proposes a structure for account. Main details that should change are the exact properties in the account object


schema.ts
```diff
@@ -177,6 +177,33 @@ export enum ValueHintType {
...
+   *   id: 'id',
+   *   displayProperty: 'name',
+   *   properties: {
+   *     id: {type: ValueType.String, required: true},
+   *     email: {type: ValueType.String, codaType: ValueHintType.Email, required: false},
+   *     name: {type: ValueType.String, required: true},
+   *   },
+   * });
+   * ```
+   *
```
sam-codaio: The exact properties here should be finalized. There are a few options on if we want to re-use these id properties or have specific properties used to identify this such as userEmailProperty, userIdProperty, etc.

schema.ts
```diff
@@ -177,6 +177,33 @@ export enum ValueHintType {
...
+   * account, and there is an (optional) single column with ValueHintType.Email that represents
+   * the email for the account in the external system.
+   *
+   *
+   * @example
+   * ```
+   * makeObjectSchema({
+   *   type: ValueType.Object,
+   *   codaType: ValueHintType.Account,
+   *   id: 'id',
```
ekoleda-codaio: Shouldn't this be `idProperty: 'id'`?

schema.ts
```diff
@@ -177,6 +177,33 @@ export enum ValueHintType {
    * Indicates to render a value as a select list.
    */
   SelectList = 'selectList',
+
+  /**
+   * Indicates that this field should be interpreted as an account in
+   * an external system. The provided value should be an object where the
+   * 'id' is the ID in the external system, the displayName is the user name for the
+   * account, and there is an (optional) single column with ValueHintType.Email that represents
```
ekoleda-codaio: "column" => "property"

schema.ts
```diff
@@ -177,6 +177,33 @@ export enum ValueHintType {
...
+   * @example
+   * ```
+   * makeObjectSchema({
+   *   type: ValueType.Object,
+   *   codaType: ValueHintType.Account,
+   *   id: 'id',
+   *   displayProperty: 'name',
+   *   properties: {
+   *     id: {type: ValueType.String, required: true},
+   *     email: {type: ValueType.String, codaType: ValueHintType.Email, required: false},
```
ekoleda-codaio: I think you can omit `required: false`, it's the default and implied.

schema.ts
sam-codaio: displayProperty might not be guaranteed in every system so we may want to mark it not required.

helpers/migration.ts
```diff
@@ -111,6 +111,10 @@ class ObjectSchemaHelper<T extends ObjectSchemaDefinition<string, string>> {
   get titleProperty() {
     return this._schema.titleProperty ?? this._schema.displayProperty;
   }
+
+  get displayProperty() {
```
jonathan-codaio: Why not just use `primary`?

schema.ts
```diff
@@ -177,6 +177,33 @@ export enum ValueHintType {
    * Indicates to render a value as a select list.
    */
   SelectList = 'selectList',
+
+  /**
+   * Indicates that this field should be interpreted as an account in
+   * an external system. The provided value should be an object where the
+   * 'id' is the ID in the external system, the displayName (optional) is the user name for the
```
jonathan-codaio: s/displayName/displayProperty

jonathan-codaio (commented):
Overall LGTM

helpers/migration.ts
sam-codaio: We can use primary instead if that has the same functionality.

schema.ts
sam-codaio: Will do

helpers/migration.ts
jonathan-codaio: Yeah, `displayProperty` is just a rename of `primary`. The schemaHelper exists so that business logic doesn't have to think about the name change.

It's maybe confusing that the helper calls it `primary` but IIRC we did that to minimize diffs in `coda`, but could choose to rename it at any point.

PR #3262, opened 2025-07-16
AR-59 Added new screen annotation tool definition for vBars support

anishhGR:
## Overview
This PR introduces support for defining default screen annotation tools through a pack configuration, instead of relying on hardcoded values in the Agent Runtime's system prompt. This change enables more flexibility and scalability in how screen annotations (e.g., vBars) are managed and displayed.

## Key Changes
- Pack-defined annotation tools: Default screen annotations (e.g., vBars) can now be specified declaratively via pack definitions, avoiding hardcoded logic.

- Opt-in annotation control: Agent developers can choose whether to show screen annotations and define custom trigger conditions for when these annotations should appear.

- The new structure is extensible and designed to support additional annotation types in the future (e.g., underlines or custom overlays).

Screenshot 2025-07-16 at 10 50 30 AM


types.ts
```diff
@@ -1274,6 +1278,46 @@ export interface KnowledgeTool extends BaseTool<ToolType.Knowledge> {
...
+ * @hidden
+ */
+type ScreenAnnotationToolSource = SuggestionsScreenAnnotationToolSource;
+
+/**
+ * Tool that provides access to screen annotation capabilities.
+ * @hidden
+ */
+export interface ScreenAnnotationTool extends BaseTool<ToolType.ScreenAnnotation> {
+  source: ScreenAnnotationToolSource;
```
richard-codaio: nit: Wonder if this should be `source` or something like `annotation` where we then specify the type of annotation that we're making.

anishhGR: Good call, that makes more sense.

PR #3265, opened 2025-07-18
Add skill entrypoints

richard-codaio:
[AR-77]
We're planning on letting pack developers specify which skills get invoked based on entrypoints. The initial version involves specifying the skill that gets triggered when clicking on an agent for the first time in the bench.

[AR-77]: https://grammarly.atlassian.net/browse/AR-77?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -1354,7 +1354,7 @@ export interface Skill {
 export interface SkillEntrypoints {
   /** Skill to be invoked when the agent is clicked on in the bench for the first time. */
   benchInitialization?: {
-    skillName: string
+    skillName: string;
   };
 }
 
--- a/dist/testing/upload_validation.js
+++ b/dist/testing/upload_validation.js
@@ -1750,9 +1750,11 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
         }),
         skillEntrypoints: z
             .object({
-            benchInitialization: z.object({
+            benchInitialization: z
+                .object({
                 skillName: z.string(),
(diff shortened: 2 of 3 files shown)


types.ts
```diff
@@ -1351,6 +1351,11 @@ export interface Skill {
   tools: Tool[];
 }
 
+export interface SkillEntrypoints {
+  /** Skill to be invoked when the agent is clicked on in the bench for the first time. */
+  benchInitialization?: string;
```
jonathan-codaio: It feels like we could have additional configuration for each entrypoint later, so I wonder if a safer shape is:

`benchInitialization?: {skillName: string};`

(I think it's also clearer to the developer what string they need to provide.)

sam-codaio: +1 on this lets keep an extra layer here for additional configuration.

types.ts
```diff
@@ -1351,6 +1351,13 @@ export interface Skill {
   tools: Tool[];
 }
 
+export interface SkillEntrypoints {
+  /** Skill to be invoked when the agent is clicked on in the bench for the first time. */
+  benchInitialization?: {
+    skillName: string
+  };
```
jonathan-codaio: I think the next time you add an entrypoint you'll want to pull out this object shape into something named to reuse, but can wait until that's needed.

jonathan-codaio (approved):

PR #3268, opened 2025-07-29
Add a new entrypoint to allow developers to specify the default chat skill

richard-codaio:
[AR-86]
Adds a new entrypoint that marks the skill to be used when chatting with the agent.

[AR-86]: https://grammarly.atlassian.net/browse/AR-86?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/testing/upload_validation.ts
+++ b/testing/upload_validation.ts
@@ -94,6 +94,7 @@ import type {SetEndpoint} from '../types';
 import {SimpleStringHintValueTypes} from '../schema';
 import type {SimpleStringSchema} from '../schema';
 import type {Skill} from '../types';
+import type {SkillEntrypointConfig} from '../types';
 import type {SkillEntrypoints} from '../types';
 import type {SliderSchema} from '../schema';
 import type {StringDateSchema} from '../schema';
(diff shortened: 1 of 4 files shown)


types.ts
```diff
@@ -1351,11 +1351,16 @@ export interface Skill {
   tools: Tool[];
 }
 
+interface SkillEntrypointConfig {
```
sam-codaio: Hidden

types.ts
```diff
@@ -1351,11 +1351,16 @@ export interface Skill {
...
+  /** The name of the skill to be invoked. */
+  skillName: string;
+}
+
 export interface SkillEntrypoints {
```
sam-codaio, line 1367: Hidden

sam-codaio (commented):
Seems like a reasonable start to me

sam-codaio (approved):
Would also get Jonathan to take a look as well

testing/upload_validation.ts
```diff
@@ -2269,6 +2269,11 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
             skillName: z.string(),
           })
           .optional(),
+        defaultChat: z
+          .object({
+            skillName: z.string(),
+          })
```
jonathan-codaio: Can you split this out into a shared validator, and that uses `zodCompleteObject<SkillEntrypointConfig>` so that we get TS enforcement that new fields in the config object require an update to the validator.

testing/upload_validation.ts
```diff
@@ -2146,6 +2147,19 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
     tools: z.array(toolSchema),
   });
 
+  const skillEntrypointsSchema = zodCompleteObject<SkillEntrypoints>({
```
jonathan-codaio: Ah sorry I should have said `zodCompletelyStrictObject`.

The difference is that "strict" also validates that there are no extraneous keys. I think all but like 2-3 cases in this file should be using strict, there's some bad old usage in there to be cleaned up.

testing/upload_validation.ts
```diff
@@ -2146,6 +2147,19 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+    benchInitialization: z
+      .object({
+        skillName: z.string(),
+      })
```
jonathan-codaio: This shape should be separated out into its own validator because you use the same shape for each entrypoint.

PR #3275, opened 2025-08-01
packs-sdk: release v1.9.13

denyszavhorodnii:
release v1.9.13

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.9.13-prerelease.1",
+  "version": "1.9.13",
   "license": "MIT",
   "workspaces": [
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.9.13-prerelease.1",
+  "version": "1.9.13",
```
neal-codaio: Do you want/need this to be a public release?  If not, this should be: `1.9.13-prerelease.2`

denyszavhorodnii: switched to prerelease

neal-codaio (approved):

neal-codaio:
You should update the title since you are doing a pre-release now.

PR #3277, opened 2025-08-05
Add "Guide" annotation type

denyszavhorodnii:
We need a new annotation type (as generic as possible) to reflect UI suggestions which does not contain rewrites. 
The idea is to rename `Suggestions` into `Rewrite` because it better reflects the nature of that kind of annotation, and to intrcoduce new annotation type "Guide". 
Guide – is annotation that does not contain any rewrites but might contain some actions (Ack, Dismiss etc)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 - Add `skillEntrypoints` to pack definition
 - Add `AssistantMessage` tool
+- Rename `Suggestions` tool to `Rewrite`
 
 ## [1.9.12] - 2025-07-25
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.9.13-prerelease.2",
+  "version": "1.9.13-prerelease.3",
   "license": "MIT",
   "workspaces": [
     "dev/eslint"


types.ts
```diff
@@ -1307,7 +1311,7 @@ interface BaseScreenAnnotation<T extends ScreenAnnotationType> {
  * Annotation for screen suggestions.
  * @hidden
  */
-interface SuggestionsScreenAnnotation extends BaseScreenAnnotation<ScreenAnnotationType.Suggestions> {}
+interface SuggestionsScreenAnnotation extends BaseScreenAnnotation<ScreenAnnotationType> {}
```
sam-codaio: Does not really matter for now but should we not instead also add `GuideScreenAnnotation extends BaseScreenAnnotation<ScreenAnnotationType.Guide>` and `RewriteScreenAnnotation extends BaseScreenAnnotation<ScreenAnnotationType.Rewrite>`. Just allows us to add parameters in the future here

denyszavhorodnii: Makes sense. I changed it to `RewriteScreenAnnotation` and `GuideScreenAnnotation`

types.ts
```diff
@@ -1290,9 +1290,13 @@ export interface KnowledgeTool extends BaseTool<ToolType.Knowledge> {
  */
 export enum ScreenAnnotationType {
   /**
-   * Use screen annotation suggestions.
+   * Use screen annotation for text rewriting.
```
sam-codaio, line 1293: As discussed can we keep Suggestions for a few weeks (just mark with `@deprecated`) and then remove once communicated to partners? 
Otherwise LGTM

denyszavhorodnii: I thought we agreed on removing it right now. Not to overcomplicate it. See https://grammarly.slack.com/archives/C090L3BSAA1/p1754407828566309?thread_ts=1754395173.055869&cid=C090L3BSAA1

sam-codaio: Thread is a little unclear to me on the ultimate conclusion. If you support it server side than I am generally fine with either approach, not sure what is more clear to partners

sam-codaio (approved):

sam-codaio (commented):
Mind updating the change log here?

types.ts
denyszavhorodnii: Yeah, that's the idea https://github.com/coda/coda/pull/131854 here is PR

PR #3282, opened 2025-08-11
Fix Cats sample Pack

ekoleda-codaio:
The cataas.com API changed a while back and broke this sample code. This PR fixes the sample in a few ways:

- Returns the full URL from the API.
- Removes filter options that have been removed.
- Uses the new query parameter names for font color and size.

The sync table was also broken, but even after fixing it I noticed it worked poorly since the API serves images so slowly. I decided to delete the sync table completely to avoid confusion.

That sync table was also serving the purpose of showing a simple sync table with a parameter, so I replaced it with a new sample that lists Emojis.

I also made some small stylistic changes:

- Renamed `CatImage()` to `CatPhoto()`, since they are all photos.
- Use array destructuring to unpack the formula parameter values, since it is easier to read.

This sample is also broken in the new Pack dialog in the Pack Studio, but this fix will eventually flow into that UI once that repo is using an updated version of the SDK.


patrick-codaio (approved):
cats

PR #3288, opened 2025-08-21
Upgrade packages to avoid security issues

neal-codaio:
I noticed some packages that are old or deprecated.  It would be good to keep more of these up to date.  Probably drop some off where possible.  `release-it` and `@julian_cataldo/remark-lint-frontmatter-schema`

PTAL @jonathan-codaio @patrick-codaio @ekoleda-codaio

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,13 +38,13 @@
     "handlebars": "4.7.8",
     "jsonpath-plus": "10.3.0",
     "mersenne-twister": "1.1.0",
-    "node-fetch": "2.6.11",
+    "node-fetch": "2.6.13",
     "open": "8.4.2",
     "pascalcase": "1.0.0",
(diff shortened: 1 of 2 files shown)


ekoleda-codaio:
I don't see any changes to `release-it` or `@julian_cataldo/remark-lint-frontmatter-schema`, but both of those are still in use.

neal-codaio:
> I don't see any changes to `release-it` or `@julian_cataldo/remark-lint-frontmatter-schema`, but both of those are still in use.

Yup, I didn't update those.  I only fixed the security problems in this PR.  Those 2 deps should be updated as well as others that are out of date.  I didn't check others.  release-it can be upgrade to 19.0.4 IIRC and remark is deprecated and I wonder if it would be removed.

PR #3303, opened 2025-09-18
[Packs SDK] Add Summarizer Agents Tool

wesley-codaio:
Supporting https://github.com/coda/coda/pull/134027

Based on https://github.com/coda/packs-sdk/pull/3270


types.ts
```diff
@@ -1191,10 +1191,13 @@ export enum ToolType {
   ScreenAnnotation = 'ScreenAnnotation',
   /**
    * Allows assistant messages to be used as tools.
-   * This tool is for internal usage only.
-   * @hidden
    */
   AssistantMessage = 'AssistantMessage',
+  /**
+   * Allows summarization to be used as a tool.
```
chris-codaio: Could we say more about what specifically this is doing?  Guessing this comment won't be sufficient for someone reading the SDK documentation.

types.ts
```diff
@@ -1191,10 +1191,13 @@ export enum ToolType {
   ScreenAnnotation = 'ScreenAnnotation',
   /**
    * Allows assistant messages to be used as tools.
-   * This tool is for internal usage only.
-   * @hidden
```
anishhGR, line 1195: I noticed the hidden annotation was removed, was this intentional?

types.ts
sam-codaio: Yeah it is a bit unclear to me why you need this as an explicit tool vs just something that you can ask the model to do?

types.ts
wesley-codaio: Good call

PR #3304, opened 2025-09-18
Add MCP provider support to Packs SDK

abhishek-codaio:
## Summary
Add ability to wire MCP (Model Context Protocol) providers into packs via `pack.addMcpProvider()` method, similar to the existing skills functionality from PR #3202.

## Changes
- **Types**: Added `McpProvider` interface with `endpoint`, `agentName`, and `agentVersion` fields
- **Builder**: Implemented `pack.addMcpProvider()` method in `PackDefinitionBuilder` class
- **Validation**: Added endpoint uniqueness validation and 10 provider limit
- **Metadata**: Include MCP providers in pack metadata compilation
- **Exports**: Export `McpProvider` type for external use

## API Usage
```typescript
pack.addMcpProvider({
  endpoint: 'http://localhost:3000/mcp',
  agentName: 'MyAgent',
  agentVersion: '1.0.0'
});
```

## Test plan
- [x] TypeScript compilation passes
- [x] Basic functionality test with multiple providers
- [x] Validation schema includes proper URL validation
- [x] Endpoint uniqueness enforced
- [ ] Integration test with actual MCP server
- [ ] Documentation updates

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/builder.ts
+++ b/builder.ts
@@ -16,6 +16,7 @@ import type {PackVersionDefinition} from './types';
 import type {ParamDefs} from './api_types';
 import type {Schema} from './schema';
 import type {Skill} from './types';
+import type {McpProvider} from './types';
 import type {SkillEntrypoints} from './types';
 import type {SyncExecutionContext} from './api_types';
 import type {SyncPassthroughData} from './api';
@@ -71,6 +72,11 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
    */
   skills: Skill[];
   /**
+   * See {@link PackVersionDefinition.mcpProviders}.
+   * @hidden
+   */
+  mcpProviders: McpProvider[];
+  /**
    * See {@link PackVersionDefinition.skillEntrypoints}.
    * @hidden
    */
@@ -114,6 +120,7 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
       formats,
       syncTables,
       skills,
+      mcpProviders,
       networkDomains,
       defaultAuthentication,
       systemConnectionAuthentication,
@@ -125,6 +132,7 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
     this.formats = formats || [];
     this.syncTables = syncTables || [];
     this.skills = skills || [];
+    this.mcpProviders = mcpProviders || [];
     this.skillEntrypoints = skillEntrypoints;
     this.networkDomains = networkDomains || [];
     this.defaultAuthentication = defaultAuthentication;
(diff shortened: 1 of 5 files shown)


types.ts
```diff
@@ -1369,6 +1369,20 @@ export interface Skill {
...
+/**
+ * Definition of an MCP (Model Context Protocol) provider for this pack.
+ * MCP providers allow packs to integrate with external AI agents and services.
+ * @hidden
+ */
+export interface McpProvider {
+  /** The MCP endpoint URL to connect to. */
+  endpoint: string;
+  /** The name of the agent that will use this MCP provider. */
+  agentName: string;
```
sam-codaio, line 1381: I am not sure if agentName makes sense here? Would rather just have mcpName and then the agent can choose this as a tool potentially? That feels more consistent.

Also I believe this is called skillName elsewhere not agentName?

Or is the concept that this becomes broadly available to any skill?

builder.ts
```diff
@@ -281,6 +289,27 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
     return this;
   }
 
+  /**
+   * Adds an MCP provider definition to this pack.
+   *
+   * MCP providers allow packs to integrate with external AI agents and services
```
chris-codaio, line 295: Wording seems weird here since you're not connecting to an external agent - you're connecting to an external service.

testing/upload_validation.ts
```diff
@@ -167,6 +168,7 @@ export const Limits = {
   BuildingBlockDescription: 1000,
   ColumnMatcherRegex: 300,
   MaxSkillCount: 15,
+  MaxMcpProviderCount: 10,
```
sam-codaio, line 171: Lets just allow 1 per pack to start?

types.ts
chris-codaio: Agreed with your first point.

testing/upload_validation.ts
```diff
@@ -2160,6 +2162,12 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
     tools: z.array(toolSchema),
   });
 
+  const mcpProviderSchema = zodCompleteObject<McpProvider>({
+    endpoint: z.string().min(1).url('MCP provider endpoint must be a valid URL'),
```
chris-codaio, line 2166: Won't this have to be tied into the connection object's endpoint URL so we can handle dynamic domain situations?

testing/upload_validation.ts
```diff
@@ -2160,6 +2162,12 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+    agentName: z.string().min(1).max(Limits.BuildingBlockName),
+    agentVersion: z.string().min(1).max(50),
+  });
```
chris-codaio, line 2169: I think this assumes the API OAuth endpoint is the same as the MCP OAuth endpoint - is that always true?  MCP requires PKCE I think, but our packs don't mandate it.  Might need to do more alignment work here.

Also - do we have to worry about dynamic client registration?  https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization#dynamic-client-registration

types.ts
ekoleda-codaio: +1. The protocol requires that the client identify itself with a name and version, but I think we should just pull that automatically from the Pack's name and version, and not require the developer to specify that separately.

testing/upload_validation.ts
sam-codaio: This is frequently not true from my experience. They often have different OAuth endpoints and thus might need separate connections

PR #3306, opened 2025-09-25
Make makeObjectSchema() be strict about flagging extraneous properties.

jonathan-codaio:
@alexd-codaio noticed that we had a bunch of incorrect property names being used in pack defs like updatedAtProperty instead of modifiedAtProperty that weren't being caught by TS. It's because those schemas had all of the required properties correctly specified but TS was permitting extraneous properties beyond the required once. This updates things to be strict about the allowed properties.

Will need to merge this in coordination with the fixes here: https://github.com/coda/packs/compare/jlg-make-object-schema?expand=1

PTAL @sam-codaio @patrick-codaio @coda/orchestra-platform


schema.ts
```diff
@@ -2010,9 +2011,9 @@ export function makeSchema<T extends Schema>(schema: T): T {
 export function makeObjectSchema<
   K extends string,
   L extends string,
-  const T extends Omit<ObjectSchemaDefinition<K, L>, 'type'>,
+  const T extends Omit<ObjectSchemaDefinition<K, L>, 'type'> & ObjectSchemaProperty,
 >(
-  schemaDef: T & {type?: ValueType.Object},
+  schemaDef: Exact<T, Omit<ObjectSchemaDefinition<K, L>, 'type'> & ObjectSchemaProperty & {type?: ValueType.Object}>,
```
jonathan-codaio, line 2016: This is basically the fix. All the rest is fallout.

helpers/object_utils.ts
```diff
@@ -36,3 +36,9 @@ export function deepCopy<T>(obj: T): T {
 export function isPromise<T>(obj: any): obj is Promise<T> {
   return obj && typeof obj === 'object' && 'then' in obj;
 }
+
+/**
+ * Type helper to ensure that a given type can only contain keys from a another given type.
```
patrick-codaio: extra 'a'

helpers/object_utils.ts
```diff
@@ -36,3 +36,9 @@ export function deepCopy<T>(obj: T): T {
...
+ * Prevents extraneous keys from being allowable.
+ */
+export type Exact<T, U> = T & Record<Exclude<keyof T, keyof U>, never>;
```
patrick-codaio, line 44: I find this confusing but Claude agrees with you, so 👍 :

```
> It still feels weird to have 2 different generics... can't you make a type that says "U and never any key that isn't in keyof U"?

⏺ You're absolutely right! That would be much cleaner. The issue is TypeScript needs to know which properties are "allowed" vs "extra".  The closest you can get with one generic is:

  type NoExtraProperties<T> = {
    [K in keyof T]: T[K]
  } & Record<PropertyKey, never>;

  But this still doesn't work because TypeScript can't distinguish between "intended" and "extra" properties from a single type.

  The two-generic approach exists because:
  - U defines the "allowed" shape
  - T is the actual input type being constrained
  - The type system needs both to compute which properties are "extra"

  You're right that it feels weird - it's a limitation of how TypeScript's type system works. The type needs a reference point (U) to know what constitutes an "extra" property.
```

patrick-codaio (approved):

helpers/object_utils.ts
jonathan-codaio: Yeah I find it confusing too. I think it comes about because we are trying (elsewhere) to use type inference to do dynamic-y type enforcement of the specific schema properties in the current schema, so that's why we wind up with an extra generic for `T extends ObjectSchemaDefinition`. If we weren't trying to do that extra inference I think we would just type the function as accepting `ObjectSchemaDefinition` at existing literal checking would cover it.

PR #3316, opened 2025-10-03
[Packs SDK] LLM-Facing Descriptions

wesley-codaio:
Add LLM facing description fields for pack formulas and skills. 

Full discussion: https://grammarly.slack.com/archives/C090L3BSAA1/p1759444251490569?thread_ts=1758746247.248289&cid=C090L3BSAA1

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -1226,6 +1226,8 @@ export interface PackTool extends BaseTool<ToolType.Pack> {
   formulas?: Array<{
     /** The name of the formula to use as a tool. */
     formulaName: string;
+    /** Instructions for LLMs to use the formula. */
+    instructions?: string;
   }>;
 }
 
--- a/dist/bundle.d.ts
+++ b/dist/bundle.d.ts
@@ -5371,6 +5371,8 @@ export interface PackTool extends BaseTool<ToolType.Pack> {
 	formulas?: Array<{
 		/** The name of the formula to use as a tool. */
 		formulaName: string;
+		/** Instructions for LLMs to use the formula. */
+		instructions?: string;
 	}>;
 }
 /**
--- a/dist/types.d.ts
+++ b/dist/types.d.ts
@@ -1095,6 +1095,8 @@ export interface PackTool extends BaseTool<ToolType.Pack> {
     formulas?: Array<{
         /** The name of the formula to use as a tool. */
         formulaName: string;
+        /** Instructions for LLMs to use the formula. */
+        instructions?: string;
     }>;
 }
 /**


types.ts
```diff
@@ -1226,6 +1226,8 @@ export interface PackTool extends BaseTool<ToolType.Pack> {
   formulas?: Array<{
     /** The name of the formula to use as a tool. */
     formulaName: string;
+    /** Instructions for LLMs to use the formula. */
+    modelInstructions?: string;
```
wesley-codaio: Suggestions welcome on naming and documentation. Want this to be used for the LLM

types.ts
```diff
@@ -1370,6 +1372,8 @@ export interface Skill {
   displayName: string;
   /** Description of what this skill does. */
   description: string;
+  /** Instructions for LLMs to use the skill. */
+  modelInstructions?: string;
```
wesley-codaio: Suggestions welcome on naming and documentation. Want this to be used for the LLM

types.ts
ekoleda-codaio: I like `model` more than `agent`, but this skill information is only visible to agents anyway, so I think just `instructions` would be fine. I don't feel strongly.

ekoleda-codaio: Are formula level instructions enough? Or do you also need parameter-level instructions to get good results?

types.ts
ekoleda-codaio: Isn't this duplicating of the `prompt` field?

types.ts
wesley-codaio: Good point, since skill information is visible only to agents, `instructions` would be fine. Updating

wesley-codaio: Good question, ideally we have parameter-level instructions.

One approach could be adding another array<parameter, instructions>? 

ex:
```
export interface PackTool extends BaseTool<ToolType.Pack> {
  /**
   * The ID of the pack to use as a tool, if referencing a different pack.
   * Omit this to reference the current pack.
   */
  packId?: number;
  /** The specific formulas to make available as tools. */
  formulas?: Array<{
    /** The name of the formula to use as a tool. */
    formulaName: string;
    /** Instructions for LLMs to use the formula. */
    formulaInstructions?: string;
    parameters?: Array<{
      /** The name of the parameter. */
      parameterName: string;
      /** Instructions for LLMs to use the parameter. */
      parameterInstructions?: string;
    }>;
  }>;
}
```

types.ts
wesley-codaio: Good call, we could fold this into the existing `prompt` field. Removing this

types.ts
wesley-codaio: Happy to leave parameters out of this change for now

abhishek-codaio (approved):

PR #3317, opened 2025-10-06
Add parallel flag to register bundles

sam-codaio:


sam-codaio:
This does not work as the pack bundle depends on the thunk being pre-loaded in the environment

PR #3319, opened 2025-10-07
Minify thunk bundle

sam-codaio:


drew-codaio (approved):
Can we test this out by uploading a lambda and invoking it?

sam-codaio:
> > Can we test this out by uploading a lambda and invoking it?
> 
> Yeah, will certainly do pre-release although this should be tested as part of calling registerBundles in the IVM I believe

Have validated that this works in IVM

sam-codaio:
> The thing to test is to throw an error inside of a pack and see if the stack trace is still correct.

```
{ category: 'Internal', type: 'Unknown error', message: 'sam', details: 'Runtime error', stack: 'Error: sam\n' + ' at Object.execute (code.ts:18:11)\n' + ' at doFindAndExecutePackFunction (/Users/samuelsloate/src/coda/data/AWS/lambda/2000000001_4_n22v2/code/thunk.js:13:20888)\n' + ' at Object.findAndExecutePackFunction (/Users/samuelsloate/src/coda/data/AWS/lambda/2000000001_4_n22v2/code/thunk.js:13:20251)\n' + ' at <unknown> (<isolated-vm>:1:13)\n' }
```

PR #3512, opened 2026-03-25
[DO NOT MERGE] Add WebFetch ToolType

solomon-codaio:
Prototype currently functions in coda/coda as an Extension Tool. If we want to enable 3rd party agent builder to declare the tool, we'll need to merge this.
(Ok, what's happening if orchestra agents flatten tools? I guess this isn't needed for 3rd party development?)

Adds a new `ToolType` to allow us to test `WebFetch` as a tool for Go. Right now, our `web_search` tool calls OpenAI's web search on queries. If a website or URL is requested by the user, we can instead use `web_fetch` to get the content (webpage or PDF) directly using Claude's `web_fetch` (OpenAI does not have this tool).

Potential use cases:
- Allow Go to dive deeper into content from web sources cited from a `web_search` result
- Allow users to refer Go to specific online sources (web pages or PDFs) for questioning and analysis
- Allow agents to load web content from predefined URLs (in prompts or tool descriptions):
  - Example: Store additional LLM guidance for agents to load dynamically from the web when needed. We could re-design our AboutMe agent with source-of-truth instructions contained in links organized by section instead of having engineers maintain and update this full content. This also saves a lot of context space.

Refer to draft implementation in this `coda` PR: https://github.com/coda/coda/pull/143926


types.ts
```diff
@@ -1477,6 +1481,16 @@ export interface WebSearchTool extends BaseTool<ToolType.WebSearch> {
   allowedDomains?: string[];
 }
 
+export interface WebFetchTool extends BaseTool<ToolType.WebFetch> {
```
ekoleda-codaio, line 1476: Add a doc block comment for this interface, and include `@hidden` so it doesn't get documented until we are ready to release it to external devs.

types.ts
```diff
@@ -1261,6 +1261,10 @@ export enum ToolType {
    * @internal
    */
   CodaDocsAndTables = 'CodaDocsAndTables',
+  /**
+   * Tool that enables fetching the full content of a specific web page or document at a URL.
```
ekoleda-codaio: Include `@hidden` in the comment so that it doesn't get documented.

ekoleda-codaio (commented):
Are the `WebSearch` and `WebFetch` tools sematically similar enough that they should be the same in the SDK? Having model-dependant tools sound problematic.

PR #3523, opened 2026-04-12
Validate formula names in the chat and bench initialization skills

ekoleda-codaio:
Formula name validation was added in https://github.com/coda/packs-sdk/pull/3433, but it didn't include `chatSkill` or `benchInitializationSkill`. This PR extends the validation to those skills as well.

Claude Code was used to generate the code changes, and I reviewed it.


daviddriscoll-codaio (approved):
thank you!

PR #3529, opened 2026-04-22
Add hidden agentSettings to PackVersionDefinition

vivekvenkatram:
## Summary

Proofreader-style customizable agent settings, on the SDK side. Lets pack authors declare boolean / single-select / multi-select fields (with optional grouping) that the agent builder UI can render and persist on each agent instance.

Hidden behind `@hidden In development` for now while the Coda-side runtime + UI catch up. This PR is the future-state demo so we can migrate proofreader-style packs onto SDK-declared schemas without changing the wire shape.

What's in here:
- `AgentSettingFieldType` enum + `AgentSettingField` discriminated union (`Boolean`, `SingleSelect`, `MultiSelect`) in `types.ts`
- `agentSettings?: AgentSettingField[]` on `PackVersionDefinition`
- `pack.addAgentSetting(...)` builder method
- Upload validation: dup name detection, option uniqueness, default values must be in the option list, per-pack and per-field caps
- Builder + upload-validation tests

## Notes

- Filed as draft. This is for review / pattern-fit feedback.
- Enum values use SDK PascalCase convention (`Boolean`, `SingleSelect`, `MultiSelect`). Coda's existing internal registry uses lowercase camelCase; the migration RFC will cover the values mapping (or a one-time conversion).

## Test plan

- [x] `make test-file FILE=test/builder_test.ts` (35 passing, 3 new)
- [x] `make test-file FILE=test/upload_validation_test.ts` (347 passing, 5 new)
- [x] `npx tsc --noEmit`
- [x] `make lint-code` + `make lint-changelog`


ekoleda-codaio:
Have you considered reusing the existing parameters system for these settings? While semantically a bit different, there is a ton of feature overlap and we already have existing code that can render parameters as UI.

PR #3544, opened 2026-05-12
Add WebAccess ToolType with Search & Fetch domain filtering

solomon-codaio:
## Summary

Introduces `ToolType.WebAccess`, a new tool type that unifies web search and web fetch capabilities under a single, extensible configuration. This replaces the need to configure web capabilities separately and lays the groundwork for future capabilities like dynamic browsing.

- New `ToolType.WebAccess` enum value, types, and validation schema
- Per-capability config (`search`, `fetch`) with independent enable/disable and domain filtering
- Domain filtering via discriminated union (`AllowList` / `BlockList` mode)
- `ToolType.WebSearch` remains unchanged for backward compatibility (will be deprecated later)

## Design rationale

**Why a new ToolType instead of extending WebSearch?**

`ToolType.WebSearch` is already consumed by `src/coda`'s `ToolResolverManager`, agent builder UI, and custom agent creator schemas. Changing its shape or semantics would break compatibility across both codebases. `WebAccess` is a new ToolType that `src/coda` can resolve into the existing `WebSearchToolResolver` and `WebFetchToolResolver` independently, while `WebSearch` continues to work as-is.

**Per-capability config objects with `enabled` field (not `boolean | object`):**

We considered `search?: boolean | DomainFilter` for conciseness but rejected it for three reasons:

1. *No precedent in the codebase.* No existing tool config uses a `boolean | object` union. Existing patterns use optional nested config objects with optional sub-fields (e.g., `WebBasicAuthentication.uxConfig`).
2. *Backend persistence.* With `enabled` as a separate field, the admin panel can store `{ enabled: false, domainFilter: { mode: "AllowList", domains: [...] } }`. Toggling a capability off preserves the domain config; toggling it back on restores it. A `boolean` replacement would wipe the config.
3. *Extensibility.* `WebAccessSearchConfig` and `WebAccessFetchConfig` are independent interfaces. Adding search-specific settings (e.g., `maxResults`) or fetch-specific settings (e.g., `maxContentTokens`) is a non-breaking addition to the respective interface.

**Discriminated union for domain filtering (not separate `allowedDomains` / `blockedDomains` fields):**

Having both fields simultaneously on a capability is semantically contradictory — an allowlist implies everything else is blocked, making a blocklist redundant. Rather than validating mutual exclusivity at runtime, the `WebAccessDomainFilterMode` discriminant makes invalid states unrepresentable at the type level. A single `domains` field is interpreted by the `mode`, and the zod schema uses `z.discriminatedUnion('mode', [...])` with no `.superRefine()` needed.

### Schema shape

```typescript
enum WebAccessDomainFilterMode { AllowList, BlockList }

interface WebAccessAllowListFilter {
  mode: WebAccessDomainFilterMode.AllowList;
  domains: string[];  // 1-100 entries
}

interface WebAccessBlockListFilter {
  mode: WebAccessDomainFilterMode.BlockList;
  domains: string[];  // 1-100 entries
}

type WebAccessDomainFilter = WebAccessAllowListFilter | WebAccessBlockListFilter;

interface WebAccessSearchConfig {
  enabled?: boolean;
  domainFilter?: WebAccessDomainFilter;
}

interface WebAccessFetchConfig {
  enabled?: boolean;
  domainFilter?: WebAccessDomainFilter;
}

interface WebAccessTool extends BaseTool<ToolType.WebAccess> {
  search?: WebAccessSearchConfig;
  fetch?: WebAccessFetchConfig;
}
```

### Usage examples

```typescript
// Everything enabled, no restrictions
{ type: ToolType.WebAccess }

// Search with allowlist, fetch disabled (config preserved for re-enable)
{
  type: ToolType.WebAccess,
  search: {
    domainFilter: { mode: WebAccessDomainFilterMode.AllowList, domains: ["stackoverflow.com"] },
  },
  fetch: {
    enabled: false,
    domainFilter: { mode: WebAccessDomainFilterMode.BlockList, domains: ["evil.com"] },
  },
}
```

## Follow-up work

- `src/coda`: Add `WebAccessToolResolver` that fans out to `WebSearchToolResolver` and `WebFetchToolResolver`, promoting `WebFetchToolResolver` from its current extension-tool pattern
- `src/coda`: Update `custom_agent_builder.ts` to push `WebAccess` instead of separate WebSearch + WebFetch injection
- `src/coda`: Admin panel UI for configuring WebAccess capabilities


ekoleda-codaio:
I don't feel strongly, but having the tool have two distinct sub-sections, each with its own `enabled` flag, sort of smells like this should just be two tools. I assume you considered this approach as well?

The two features are so semantically similar, and perhaps the 80% case will be that a developer wants both enabled. And if you squint, it's sort of like the `formulas` filter on the `Pack` tool, where the developer is narrowing the scope of the tool.

types.ts
```diff
@@ -1463,6 +1469,125 @@ export interface WebSearchTool extends BaseTool<ToolType.WebSearch> {
...
+ */
+export enum WebAccessDomainFilterMode {
+  /**
+   * Only allow access to the specified domains.
+   */
+  AllowList = 'AllowList',
+  /**
+   * Block access to the specified domains, allow everything else.
+   */
+  BlockList = 'BlockList',
```
ekoleda-codaio, line 1483: What's the use case for a block list? It's not useful for blocking categories of domains (spam, gambling, etc) given that there are too many and they move too fast. It seems like it would only be useful if, for some reason, there was a very specific set of domains you wanted blocked, but I'm struggling to think of when that would come up.

solomon-codaio:
> I don't feel strongly, but having the tool have two distinct sub-sections, each with its own `enabled` flag, sort of smells like this should just be two tools. I assume you considered this approach as well?
> 
> The two features are so semantically similar, and perhaps the 80% case will be that a developer wants both enabled. And if you squint, it's sort of like the `formulas` filter on the `Pack` tool, where the developer is narrowing the scope of the tool.

Yes, I considered separate initially but I was thinking these tools will likely always be packaged together in any agent builder. Most people think of web access as being able to use a search bar for either search or URL lookup. Also I think the admin UIs would likely exposed this under the same "Web access" toggle, with individual settings if needed.

types.ts
solomon-codaio: Good point. For search, one possible use is wanting to block known "false positive" results for particular topics / keywords. Another use could be blocking common "answer key" websites to promote more authoritative or high-quality sources for a topic.

One practical example: you're building an FAQ agent to allow web search for topics related to your company and products, but you don't want answers using content from your competitors.

Fortunately, both APIs we're using for Search and Fetch have native support for `blockedDomains` so there isn't logic we need to maintain in either case.

ekoleda-codaio: Thanks, those are some good examples!

PR #3551, opened 2026-05-19
Reject "account" as a pack parameter name

patrick-codaio:
## Summary
- A parameter named `account` collides with how Coda surfaces connection accounts in the formula UI, leading to user confusion about whether the input is a connection param.
- Adds a new `ReservedParameterNames` enum (currently just `account`) in `api_types.ts` and wires it into the existing `paramDefValidator` in `testing/upload_validation.ts` so the check applies to formulas, actions, and sync tables (both `parameters` and `varargParameters`).
- Matching is case-insensitive, so `Account` / `ACCOUNT` are also rejected. The error message is generated from the enum values, so adding future reserved names won't require updating the message.

## Test plan
- [x] `make test-file FILE=test/upload_validation_test.ts` — all 348 tests pass, including two new ones covering `account` and `Account`.
- [x] `tsc --noEmit` clean.


ekoleda-codaio:
Thanks for catching this! Would this break any existing Packs?

Also, please add an entry to the `CHANGELOG`.

patrick-codaio:
@ekoleda-codaio oh darn it looks like 3 packs exist: https://coda.io/d/All-Pack-metadata_dE-Ht5Bpw9L/All-Packs-building-blocks_suv7sLYh#Packs_tu5-Pack/account-params_vuW4va9n

they're all published, but not heavily used. the most widely used one is in 45 docs: https://admin.coda.io/docPacks?packId=11976

one is from simpladocs, though, and it's the only sync table they have in the pack... do we have the kind of relationship with them where we can ask them how this is working for them? for me, a sync table with this param is completely unconfigurable/unsyncable - there's no account picker or param input in the settings panel:
Screenshot 2026-05-19 at 2 18 09 PM

ekoleda-codaio:
I set up the SimplaDocs Meow Pack in a doc, and it rendered the account picker and "Account" parameter just fine.

image

But for a new test Pack I am seeing the same broken UI you are. 🤷 

The two other Packs don't actually use auth, so there wouldn't be a conflict. Should we only apply this rule to Packs that use user auth? If so, then it's really just one Pack affected, and I can reach out to Scott and have him update the Pack.

patrick-codaio:
@ekoleda-codaio looks like the difference is that Meow capitalizes "Account", I can repro that this fixes the issue. 

The other packs use lower case, but like you said, no auth.

So, a narrow constraint that blocks the lower-case param only, and only when there's auth, seems fully safe to launch... if perhaps a little confusing, and leaving sharp edges for cases like later adding auth. I'll move forward with this unless you have reservations.

ekoleda-codaio:
SGTM.

PR #3556, opened 2026-05-25
Pin gitpython to 3.1.50 to fix CVE-2026-42284

ovoroniak-grm:
## Summary

  CVE-2026-42284 reports a vulnerability in gitpython versions before 3.1.47.
  We were pulling it in transitively via mkdocs-rss-plugin at 3.1.46.

  ## Detail

  - Add a top-level `gitpython = "==3.1.50"` pin in `Pipfile` to force resolution past the fix (3.1.47 is the minimum; 3.1.50 matches what coda/coda uses).
  - Regenerated lockfile entry via `pipenv upgrade gitpython==3.1.50`. smmap (a transitive dep of gitpython) bumped 5.0.2 → 5.0.3 as part of the upgrade.

  ## Other notes

  - Lockfile was generated with Python 3.13 locally (Pipfile requires 3.12). The diff is minimal because I used `pipenv upgrade` rather than a full `pipenv lock`. A maintainer with Python 3.12 may want to re-lock cleanly before merge.
  - Once this merges and a new packs-sdk version is cut, coda/coda will pick up the fix on its next packs-sdk bump.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,6 +20,9 @@ mkdocs-simple-hooks = "==0.1.5"
 mkdocs-site-urls = "==0.3.1"
 pillow = "==12.1.1"
 urllib3 = "==2.6.0"
+# Pulled in transitively (by mkdocs-rss-plugin). Pinned to a patched version
+# to address CVE-2026-42284 (fixed in 3.1.47).
+gitpython = "==3.1.50"
 
 [dev-packages]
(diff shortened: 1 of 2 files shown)


ovoroniak-grm:
Closing in favor https://github.com/coda/packs-sdk/pull/3542

PR #3563, opened 2026-06-15
Add resource field for OAuth2 auth

ekoleda-codaio:
Adds support for the OAuth2 `resource` parameter (RFC 8707), letting Packs specify the target service/resource they're requesting access to. This is required by the MCP specification, and some MCP providers are starting to require it.

- Added an optional `resource` field to `BaseOAuthAuthentication`
- Added validation on Pack upload
- Added support to the CLI for local testing
- Added test cases

While the spec requires absolute URLs, I've allowed for relative URLs when the `requiredEndpointUrl` option is enabled. In this case, the server join the connection-specific endpoint URL with the relative URL of the resource. This will allows us to support cases where there are customer-specific MCP endpoints (`https://acme.example.com/mcp`, etc).

I used Claude to create this PR, with lots of human review.

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -538,10 +538,10 @@ export interface BaseOAuthAuthentication extends BaseAuthentication {
    */
   tokenQueryParam?: string;
   /**
-   * One or more absolute URIs identifying the target services or resources to which access is being requested.
+   * A URI identifying the target service or resource to which access is being requested.
    * See https://datatracker.ietf.org/doc/html/rfc8707 for more details.
    */
-  resource?: string | string[];
+  resource?: string;
 }
 
 /**
--- a/types.ts
+++ b/types.ts
@@ -537,6 +537,11 @@ export interface BaseOAuthAuthentication extends BaseAuthentication {
    * that should contain the token.
    */
   tokenQueryParam?: string;
+  /**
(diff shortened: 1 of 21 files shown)


types.ts
```diff
@@ -537,6 +537,12 @@ export interface BaseOAuthAuthentication extends BaseAuthentication {
    * that should contain the token.
    */
   tokenQueryParam?: string;
+  /**
+   * A URI identifying the target service or resource to which access is being requested.
+   * See https://datatracker.ietf.org/doc/html/rfc8707 for more details.
+   * @hidden Under development
+   */
+  resource?: string;
```
daviddriscoll-codaio, line 545: could be `string | string[]` but i think this is a better starting point

daviddriscoll-codaio (approved):

types.ts
ekoleda-codaio: I started out with that in a first cut of this PR, but there were some challenges with the implementation. Namely that you need to pass multiple resources with multiple copies of the URL parameter (ex: `?resource=abc&resource=def`), and the JS object -> query string logic we had throughout the codebase wasn't well setup for that. So punting on that for now.

PR #3564, opened 2026-06-16
Allow relative endpointUrl for MCP servers

kh-daniil:
## Summary

Lets a pack point an MCP server at a per-account host without hardcoding it: `MCPServer.endpointUrl`
now accepts a root-relative path that is resolved against the user's connection endpoint when the
pack's authentication uses `requiresEndpointUrl`. Absolute `https` URLs keep working unchanged.

This is what an integration like Databricks needs — every customer has their own workspace host
(`https://<workspace>/api/2.0/mcp/...`), collected via `requiresEndpointUrl` — but it applies to any
MCP server whose host is per-account.

## Details

- No new fields. `endpointUrl` is overloaded the same way `authorizationUrl` / `tokenUrl` already are:
  - when the auth sets `requiresEndpointUrl` (and not `endpointKey`) → `endpointUrl` must be a
    relative path (e.g. `/mcp`);
  - otherwise → an absolute `https` URL is required.
- Validation (`upload_validation.ts`): `endpointUrl` uses `validateUrlParsesIfAbsolute`, and a
  top-level refinement enforces the relative-vs-absolute rule against the auth config. The existing
  "MCP server domain must be declared in networkDomains" check is unchanged — a relative URL has no
  host, so it's skipped (the host is validated at request time against the connection endpoint).
- An earlier revision of this PR added separate `endpointPath` / `endpointDomain` fields; per review
  that was dropped in favor of overloading `endpointUrl`, to stay consistent with the existing
  OAuth URL convention and avoid widening the public surface.

> Note: the runtime that resolves the relative URL against the decrypted connection endpoint lives in
> the agent-runtime (separate coda PR); this PR is the SDK type + upload validation only.

## Test plan

- Added `upload_validation` cases: relative URL passes with `requiresEndpointUrl`; relative URL fails
  without it; absolute URL fails when `requiresEndpointUrl` is set; existing absolute/HTTPS cases still hold.
- `make test`, `make lint`, `make validate-no-changes` green.


types.ts
```diff
@@ -1532,9 +1532,22 @@ export interface EmbeddedContentTool extends BaseTool<ToolType.EmbeddedContent>
...
    */
-  endpointUrl: string;
+  endpointUrl?: string;
+  /**
+   * A root-relative MCP path (e.g. "/my/mcp"). The host is taken from
+   * {@link endpointDomain}, or — when that is omitted — from the user's connection endpoint, which
+   * requires the pack's authentication to set `requiresEndpointUrl`. Provide either this or
+   * {@link endpointUrl}, not both.
+   */
+  endpointPath?: string;
```
ekoleda-codaio: Is it really necessary to create another field for relative paths? In other places in the SDK (authorizationUrl, tokenUrl) we accept relative paths in the same field when `requiredEndpointUrl` is set on the authentication. Could we do the same in `endpointUrl` above?

types.ts
```diff
@@ -1532,9 +1532,22 @@ export interface EmbeddedContentTool extends BaseTool<ToolType.EmbeddedContent>
...
+  /**
+   * An optional fixed host (e.g. "example.com", with no scheme or path) used together with
+   * {@link endpointPath}. Omit it to resolve the host from the user's connection endpoint.
+   */
+  endpointDomain?: string;
```
ekoleda-codaio: When would you use this? If you speicfy both an endpointPath and an endpointDomain, why not just specify endpointUrl instead?

types.ts
kh-daniil: I didn’t know that - thanks for the suggestion!

types.ts
```diff
@@ -1532,7 +1532,10 @@ export interface EmbeddedContentTool extends BaseTool<ToolType.EmbeddedContent>
  */
 export interface MCPServer {
   /**
-   * The MCP endpoint URL (e.g. https://example.com/mcp).
+   * The MCP server URL — typically an absolute URL (e.g. https://example.com/mcp). When the pack's
+   * authentication collects a per-account endpoint via `requiresEndpointUrl`, provide a relative
+   * path (e.g. "/mcp") instead; it is resolved against the user's endpoint, the same convention as
+   * `authorizationUrl` and `tokenUrl`.
```
ekoleda-codaio: I don't think we need to mention the convention here. Those are OAuth2 fields, and this is an MCP field. Ditto for the CHANGELOG entry.

test/upload_validation_test.ts
```diff
@@ -6325,8 +6321,56 @@ describe('Pack metadata Validation', async () => {
...
+      const err = await validateJsonAndAssertFails(metadata);
+      assert.deepEqual(err.validationErrors, [
+        {
+          path: 'mcpServers[0].endpointUrl',
+          message: 'MCP server endpointUrl must be an absolute URL when requiresEndpointUrl is not true',
+        },
+      ]);
+    });
+
+    it('fails when MCP server endpointUrl is absolute but auth sets requiresEndpointUrl', async () => {
```
ekoleda-codaio: I'm not sure we should enforce this. For fetcher requests, we allow absolute URLs even when there is endpoint URL set. Some apps require that you hit global URLs for some data and customer-specific URLs for other data.

ekoleda-codaio: We are enforcing this for the auth and token URL currently, but I'm not sure it makes sense in that case either.

types.ts
```diff
@@ -1532,7 +1532,9 @@ export interface EmbeddedContentTool extends BaseTool<ToolType.EmbeddedContent>
  */
 export interface MCPServer {
   /**
-   * The MCP endpoint URL (e.g. https://example.com/mcp).
+   * The MCP server URL — typically an absolute URL (e.g. https://example.com/mcp). When the pack's
+   * authentication collects a per-account endpoint via `requiresEndpointUrl`, provide a relative
+   * path (e.g. "/mcp") instead; it is resolved against the user's endpoint.
```
ekoleda-codaio: I think we typically mark new SDK fields with `@hidden` until the implemention is live, then un-hide them in a separate PR.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+### Changed
+
+- `MCPServer.endpointUrl` now accepts a root-relative path (e.g. `/mcp`) that is resolved against the user's connection endpoint when the pack's authentication sets `requiresEndpointUrl`. Absolute `https` URLs continue to work unchanged.
```
ekoleda-codaio: Hold off on adding it to the changelog until the feature is un-hidden (see comment below).

ekoleda-codaio (changes requested):

test/upload_validation_test.ts
kh-daniil: Good call — I relaxed this. I removed the rule that rejects an absolute endpointUrl when requiresEndpointUrl is set, so a pack can point an MCP server at a fixed or global host while still collecting a per-account endpoint for other calls — same as fetcher requests. The only check I kept is the sensible one: a relative endpointUrl requires requiresEndpointUrl (and no endpointKey), since otherwise there's nothing to resolve it against and the agent-runtime resolver would throw. I left the existing authorizationUrl/tokenUrl behavior alone — I agree it's worth revisiting, but it’s out of scope here.

test/upload_validation_test.ts
```diff
@@ -6325,12 +6321,97 @@ describe('Pack metadata Validation', async () => {
       const err = await validateJsonAndAssertFails(metadata);
       assert.deepEqual(err.validationErrors, [
         {
-          path: 'mcpServers.mcpServers',
-          message: 'MCP server endpointUrl must be HTTPS URLs only.',
+          path: 'mcpServers[0].endpointUrl',
+          message: 'MCP server endpointUrl must be an absolute https URL or a root-relative path (e.g. "/mcp").',
```
ekoleda-codaio: Shouldn't the error message here still mention that HTTPS is required?

test/upload_validation_test.ts
```diff
@@ -6325,12 +6321,97 @@ describe('Pack metadata Validation', async () => {
...
+          requiresEndpointUrl: true,
+          endpointDomain: 'attacker.com',
+        },
+        mcpServers: [{name: 'Example', endpointUrl: 'http://attacker.com/mcp'}],
+      });
+      const err = await validateJsonAndAssertFails(metadata);
+      assert.deepEqual(err.validationErrors, [
+        {
+          path: 'mcpServers[0].endpointUrl',
+          message: 'MCP server endpointUrl must be an absolute https URL or a root-relative path (e.g. "/mcp").',
```
ekoleda-codaio: Ditto

testing/upload_validation.ts
```diff
@@ -3018,6 +3001,43 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+        return;
+      }
+      // endpointUrl may be an absolute https URL or a root-relative path. A relative path is only
+      // valid when the authentication collects a per-account endpoint to resolve it against
+      // (requiresEndpointUrl set, and no endpointKey extracting one); without one there is nothing
+      // to resolve against. Absolute https URLs are always allowed.
+      const auth = data.defaultAuthentication;
+      const requiresEndpointUrl = auth && 'requiresEndpointUrl' in auth ? auth.requiresEndpointUrl : undefined;
+      const endpointKey = auth && 'endpointKey' in auth ? auth.endpointKey : undefined;
+      const canResolveRelativeUrl = Boolean(requiresEndpointUrl && !endpointKey);
```
ekoleda-codaio: I don't think the `endpointKey` should matter for this logic. That field is specific to OAuth2, and provides a way to automatically pull the endpoint from the OAuth token response. Whether or not it's set, we should allow relative URLs either way, I believe.

ekoleda-codaio (changes requested):

testing/upload_validation.ts
vivekvenkatram: Bit worried `startsWith('/')` also treats protocol-relative URLs like `//evil.com/mcp` as relative, which then skips the networkDomains check and (if the runtime resolves via `new URL(path, base)`) resolves to a different host than the connection endpoint. Wdyt about tightening it?

```ts
const isRelativeUrl = (url: string) => url.startsWith('/') && !url.startsWith('//');
```

testing/upload_validation.ts
```diff
@@ -3018,6 +3001,43 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
         }
       }
     })
+    .superRefine((untypedData, context) => {
```
vivekvenkatram, line 3009: This relative-vs-absolute block is now the third near-identical copy (authorizationUrl and tokenUrl have the same `expectsRelativeUrl` / `isRelativeUrl` / addIssue dance). Could we pull it into a shared helper so the next per-account field doesn't become copy #4?

```ts
validateRelativeOrAbsoluteUrl(url, {requiresEndpointUrl, endpointKey}, 'endpointUrl', context);
```

PR #3568, opened 2026-06-22
Update dependency tmp@&lt;0.2.4 to v0.2.6 [SECURITY]

krypton-build-user:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [tmp@<0.2.4](https://redirect.github.com/raszi/node-tmp) | pnpm.overrides | patch | [`0.2.4` -> `0.2.6`](https://renovatebot.com/diffs/npm/tmp@&lt;0.2.4/0.2.4/0.2.6) |

### Summary

`tmp@0.2.3` is vulnerable to an Arbitrary temporary file / directory write via symbolic link `dir` parameter.

### Details

According to the documentation there are some conditions that must be held:

```
// https://github.com/raszi/node-tmp/blob/v0.2.3/README.md?plain=1#L41-L50

Other breaking changes, i.e.

- template must be relative to tmpdir
- name must be relative to tmpdir
- dir option must be relative to tmpdir //<-- this assumption can be bypassed using symlinks

are still in place.

In order to override the system's tmpdir, you will have to use the newly
introduced tmpdir option.

// https://github.com/raszi/node-tmp/blob/v0.2.3/README.md?plain=1#L375
* `dir`: the optional temporary directory that must be relative to the system's default temporary directory.
     absolute paths are fine as long as they point to a location under the system's default temporary directory.
     Any directories along the so specified path must exist, otherwise a ENOENT error will be thrown upon access, 
     as tmp will not check the availability of the path, nor will it establish the requested path for you.
```

Related issue: [https://github.com/raszi/node-tmp/issues/207](https://redirect.github.com/raszi/node-tmp/issues/207).

The issue occurs because `_resolvePath` does not properly handle symbolic link when resolving paths:
```js
// https://github.com/raszi/node-tmp/blob/v0.2.3/lib/tmp.js#L573-L579
function _resolvePath(name, tmpDir) {
  if (name.startsWith(tmpDir)) {
    return path.resolve(name);
  } else {
    return path.resolve(path.join(tmpDir, name));
  }
}
```

If the `dir` parameter points to a symlink that resolves to a folder outside the `tmpDir`, it's possible to bypass the `_assertIsRelative` check used in `_assertAndSanitizeOptions`:
```js
// https://github.com/raszi/node-tmp/blob/v0.2.3/lib/tmp.js#L590-L609
function _assertIsRelative(name, option, tmpDir) {
  if (option === 'name') {
    // assert that name is not absolute and does not contain a path
[...]

PR #3567, opened 2026-06-22
Update dependency serialize-javascript@&lt;&#x3D;7.0.2 to v7.0.5 [SECURITY]

krypton-build-user:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [serialize-javascript@<=7.0.2](https://redirect.github.com/yahoo/serialize-javascript) | pnpm.overrides | patch | [`7.0.4` -> `7.0.5`](https://renovatebot.com/diffs/npm/serialize-javascript@&lt;&#x3D;7.0.2/7.0.4/7.0.5) |

### Impact

The serialize-javascript npm package (versions <= 7.0.2) contains a code injection vulnerability. It is an incomplete fix for CVE-2020-7660.

While `RegExp.source` is sanitized, `RegExp.flags` is interpolated directly into the generated output without escaping. A similar issue exists in `Date.prototype.toISOString()`.

If an attacker can control the input object passed to `serialize()`, they can inject malicious JavaScript via the flags property of a RegExp object. When the serialized string is later evaluated (via `eval`, `new Function`, or `<script>` tags), the injected code executes.

```javascript
const serialize = require('serialize-javascript');
// Create an object that passes instanceof RegExp with a spoofed .flags
const fakeRegex = Object.create(RegExp.prototype);
Object.defineProperty(fakeRegex, 'source', { get: () => 'x' });
Object.defineProperty(fakeRegex, 'flags', {
  get: () => '"+(global.PWNED="CODE_INJECTION_VIA_FLAGS")+"'
});
fakeRegex.toJSON = function() { return '@&#8203;placeholder'; };
const output = serialize({ re: fakeRegex });
// Output: {"re":new RegExp("x", ""+(global.PWNED="CODE_INJECTION_VIA_FLAGS")+"")}
let obj;
eval('obj = ' + output);
console.log(global.PWNED); // "CODE_INJECTION_VIA_FLAGS" — injected code executed!

#h2. PoC 2: Code Injection via Date.toISOString()
```

```javascript
const serialize = require('serialize-javascript');
const fakeDate = Object.create(Date.prototype);
fakeDate.toISOString = function() { return '"+(global.DATE_PWNED="DATE_INJECTION")+"'; };
fakeDate.toJSON = function() { return '2024-01-01'; };
const output = serialize({ d: fakeDate });
// Output: {"d":new Date(""+(global.DATE_PWNED="DATE_INJECTION")+"")}
eval('obj = ' + output);
[...]

PR #3574, opened 2026-06-22
Bump tmp to 0.2.6 to fix CVE-2026-44705

ovoroniak-grm:
Path traversal in the tmp package via unsanitized prefix/postfix/dir (GHSA-ph9p-34f9-6g65), fixed in 0.2.6. tmp is a dev-only transitive dep (via external-editor). Widens the existing pnpm override to <0.2.6 and adds the matching npm overrides entry so package-lock.json no longer resolves the vulnerable tmp@0.0.33.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13419,16 +13419,6 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "node_modules/os-tmpdir": {
-      "version": "1.0.2",
-      "resolved": "https://artifactory.grammarly.io/artifactory/api/npm/common-npm/os-tmpdir/-/os-tmpdir-1.0.2.tgz",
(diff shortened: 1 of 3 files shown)


daviddriscoll-codaio (approved):
Approving — reviewed the full security batch (#3567–#3574).

This is the correct and complete fix for the `tmp` path-traversal (CVE-2026-44705; reaching 0.2.6 also covers the 0.2.4 symlink fix, CVE-2025-54798):

- Adds a top-level npm `overrides` entry so `package-lock.json` stops resolving the vulnerable `tmp@0.0.33`. npm does not read `pnpm.overrides`, so the pnpm-only override alone left the npm lockfile vulnerable.
- Widens the pnpm override key from `tmp@<0.2.4` to `tmp@<0.2.6` so 0.2.4/0.2.5 are also caught.
- Drops the now-orphaned `os-tmpdir`. Version bumped to `1.15.0-prerelease.2`; all checks green.

This supersedes #3568 (which only updated the pnpm override + `pnpm-lock.yaml` and left `tmp@0.0.33` in `package-lock.json`) — I've left a note there recommending it be closed.

— Reviewed with Claude Code

PR #3575, opened 2026-06-22
DW-4967: add qa and pp as valid domains

rvlaskovski:
Extend list of valid domains to include qa and preprod

### DIFF
--- a/test/upload_validation_test.ts
+++ b/test/upload_validation_test.ts
@@ -4758,7 +4758,7 @@ describe('Pack metadata Validation', async () => {
           type: AuthenticationType.CodaApiHeaderBearerToken,
           networkDomain: 'coda.io',
         },
-        networkDomains: ['coda.io', 'notetaker.superhuman.com'],
+        networkDomains: ['coda.io', 'foo.superhuman.com'],
       });
       await validateJson(metadata);
     });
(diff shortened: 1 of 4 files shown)


test/upload_validation_test.ts
```diff
@@ -4763,6 +4763,17 @@ describe('Pack metadata Validation', async () => {
       await validateJson(metadata);
     });
 
+    it('CodaApiHeaderBearerToken, valid Superhuman docs env subdomains in auth networkDomain', async () => {
+      const metadata = createFakePackVersionMetadata({
+        defaultAuthentication: {
+          type: AuthenticationType.CodaApiHeaderBearerToken,
+          networkDomain: ['coda.io', 'notetaker.superhuman.com', 'notetaker.pp-sh.io', 'notetaker.qa-sh.io'],
```
ekoleda-codaio: Would it be better here to use a dummy subdomain? Like `foo.superhuman.com`, etc?

zack-codaio (approved):

testing/upload_validation.ts
```diff
@@ -2608,7 +2608,7 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
             return;
```
vivekvenkatram: Pre-existing, not from this PR, but since you're in here: should this be `continue` instead of `return`? If a pack ever has a non-matching auth before a matching `CodaApiHeaderBearerToken` one, the `return` bails on the whole loop and skips validating the matching auth.

testing/upload_validation.ts
```diff
@@ -2608,7 +2608,7 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
           }
 
-          const codaDomains = ['coda.io', 'localhost', 'superhuman.com'];
+          const codaDomains = ['coda.io', 'localhost', 'superhuman.com', 'pp-sh.io', 'qa-sh.io'];
```
vivekvenkatram: Tiny naming thing: `codaDomains` (and the "can only be used for coda.io domains" message below) now holds superhuman.com/pp-sh.io/qa-sh.io too. Not blocking since the drift started with superhuman.com, but `allowedFirstPartyDomains` would read truer if you touch this again.

test/upload_validation_test.ts
vivekvenkatram: Seconding this, a dummy subdomain like foo.pp-sh.io reads better and avoids implying a real service.

rvlaskovski: Good point, fixed

testing/upload_validation.ts
rvlaskovski: makes, sense, fixed

testing/upload_validation.ts
rvlaskovski: fixed

vivekvenkatram (approved):

PR #3581, opened 2026-06-30
Add DefaultIngestionConfiguration field

peizhaomei-codaio:
## Summary

Lets a pack opt in to simplified agent setup in Superhuman Go via a new pack-level `defaultIngestionConfiguration`. Packs that don't declare it aren't enrolled automatically. No new values: which tables are included still comes from `indexing.default`, and default filters from `ingestionSuggestedValue` — this just marks them as the authoritative default and optionally narrows the table set.

## Details

- New optional `PackVersionDefinition.defaultIngestionConfiguration`, set via `pack.setDefaultIngestionConfiguration({})`. Optional `syncTables: string[]` narrows the default set; omitted, it's every table not excluded via `indexing.default`.
- Validation: top-level refinement requires any `syncTables` entry to name a table defined in the pack.
- `@hidden`, consistent with other new agent primitives; documented in `agents/indexing/default-configuration.md`.

> Note: the platform side that reads this field lives in coda (separate PR); this is SDK type + validation only.

## Test plan

- `make test`, `make lint` green.


peizhaomei-codaio:
Ticket canceled

PR #3589, opened 2026-07-14
Allow relative MCP server endpointUrl resolved against the connection's endpoint

kh-daniil:
## Summary
MCPServer.endpointUrl now accepts a root-relative path (e.g. `/mcp`) that's resolved against the connection's endpoint when the pack's authentication provides one — via `requiresEndpointUrl` or `endpointKey` — mirroring how `authorizationUrl`/`tokenUrl` already work. This also applies to packs using `setSystemAuthentication` instead of a per-user default authentication. Absolute `https` URLs continue to work unchanged.

## Details
Extracts a shared relative-vs-absolute URL validator so both `endpointUrl` and the OAuth2 `authorizationUrl`/`tokenUrl` conventions resolve consistently, and adds the `systemConnectionAuthentication`/`endpointKey` fallback so system-authenticated packs get the same relative-URL support.

## Test plan
`make build && make lint` pass. Added/updated tests cover the relative-URL validator and `endpointUrl` resolution for both `defaultAuthentication` and `systemConnectionAuthentication` packs.


ekoleda-codaio (commented):
The docs and behavior LGTM. Leaving it to an Ecosystem engineer to approve the implementation.

testing/upload_validation.ts
```diff
@@ -3028,6 +3008,48 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+      // valid when the connection has a per-account endpoint to resolve it against; without one
+      // there is nothing to resolve against. Absolute HTTPS URLs are always allowed.
+      //
+      // A connection gets an endpoint either from `requiresEndpointUrl` (the user enters it) or from
+      // `endpointKey` (extracted from the OAuth token response and saved on the account). Unlike
+      // authorizationUrl/tokenUrl — which run during auth, before any endpointKey value exists — an
+      // MCP request happens at runtime when the endpoint is already resolved, so either source works.
+      // Packs that use system authentication instead of a per-user default authentication resolve
+      // their connection endpoint from `systemConnectionAuthentication` instead.
+      const auth = data.defaultAuthentication ?? data.systemConnectionAuthentication;
```
daviddriscoll-codaio: The runtime precedence is the opposite of this selection: [`resolveMcpConnectionWithSystemFallback`](https://github.com/coda/coda/blob/f4e37be10d81813322b8d8c3e150882e4f42572c/modules/server/agent-runtime/tools/mcp/resolve_mcp_system_connection.ts#L42-L68) returns a configured system connection first and only falls back to the per-user connection. The MCP resolver then [loads the endpoint from that selected connection and resolves the server URL against it](https://github.com/coda/coda/blob/f4e37be10d81813322b8d8c3e150882e4f42572c/modules/server/agent-runtime/tools/mcp_tool_resolver.ts#L605-L615). With both auths present, this can accept a pack whose relative MCP URL later fails against the selected system connection, or reject one the system connection could resolve. Please mirror the runtime precedence here and add coverage where both auths are declared.

— GPT-5.6 SOL

testing/upload_validation.ts
```diff
@@ -3028,6 +3008,48 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+      const requiresEndpointUrl = auth && 'requiresEndpointUrl' in auth ? auth.requiresEndpointUrl : undefined;
+      const endpointKey = auth && 'endpointKey' in auth ? auth.endpointKey : undefined;
+      const canResolveRelativeUrl = Boolean(requiresEndpointUrl || endpointKey);
```
daviddriscoll-codaio: `canResolveRelativeUrl` overlooks `postSetup: [{type: PostSetupType.SetEndpoint, ...}]`. The post-setup UI [submits the selected value as the connection endpoint](https://github.com/coda/coda/blob/14b9397f272855354d4de4614a5ad8fb69a462b7/modules/browser/packs/postload/private/packs_auth_post_setup_page.tsx#L102-L117), and the server explicitly treats an endpoint as coming from either [`requiresEndpointUrl` or post-setup connection options](https://github.com/coda/coda/blob/14b9397f272855354d4de4614a5ad8fb69a462b7/modules/server/packs/external_connection_manager.ts#L1293-L1296) before [encrypting and persisting it](https://github.com/coda/coda/blob/14b9397f272855354d4de4614a5ad8fb69a462b7/modules/server/packs/external_connection_manager.ts#L1461-L1484). MCP runtime later [decrypts that same endpoint](https://github.com/coda/coda/blob/f4e37be10d81813322b8d8c3e150882e4f42572c/modules/server/packs/external_connection_manager.ts#L2161-L2171) without distinguishing its source. As written, a runtime-resolvable relative MCP URL is rejected at upload. Please include `SetEndpoint` in this capability check and add a test.

— GPT-5.6 SOL

testing/upload_validation.ts
```diff
@@ -3028,6 +3008,48 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+      data.mcpServers.forEach((server, i) => {
+        if (!server.endpointUrl || isAbsoluteUrl(server.endpointUrl)) {
```
daviddriscoll-codaio, line 3028: `endpointUrl: ""` now passes validation. The field-level `validateUrlParsesIfAbsolute` accepts every non-`https://` string, this guard skips the empty string, and the network-domain refinement skips it too. Main rejected this via `z.string().url()`, so this is a strictness regression. Please add a non-empty constraint (with a regression test); `.min(1, ...)` on the field schema would be the clearest fix.

— GPT-5.6 SOL

daviddriscoll-codaio (commented):
Review findings from GPT-5.6 SOL.

testing/upload_validation.ts
```diff
@@ -3028,6 +3008,48 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
         }
       }
     })
+    .superRefine((untypedData, context) => {
+      const data = untypedData as PackVersionMetadata;
+      if (!data.mcpServers?.length) {
+        return;
+      }
+      // endpointUrl may be an absolute https URL or a root-relative path. A relative path is only
```
daviddriscoll-codaio: maybe shorten this comment?

daviddriscoll-codaio (approved):
Re-reviewed the latest changes. The authentication-precedence fix now matches the MCP runtime, SetEndpoint is correctly recognized as an endpoint source, the empty-string regression is covered, and the updated tests exercise both precedence directions. CI is green. Approved.\n\n— GPT-5.6 SOL

PR #3590, opened 2026-07-14
Allow relative tokenUrl for OAuth2ClientCredentials authentication

kh-daniil:
## Summary
`OAuth2ClientCredentialsAuthentication.tokenUrl` now accepts a root-relative path (e.g. `/token`) that's resolved against the connection's endpoint when the pack's authentication sets `requiresEndpointUrl`, mirroring the OAuth2 authorization-code flow's `tokenUrl` convention added in the parent PR. Absolute `https` URLs continue to work unchanged.

## Details
Reuses the shared relative-vs-absolute URL validator from the parent PR rather than duplicating the resolution logic.

## Test plan
`make build && make lint` pass. Added test coverage for relative `tokenUrl` resolution.

### DIFF
--- a/test/upload_validation_test.ts
+++ b/test/upload_validation_test.ts
@@ -5188,6 +5188,66 @@ describe('Pack metadata Validation', async () => {
       await validateJson(metadata);
     });
 
+    it('OAuth2ClientCredentials, requiresEndpointUrl requires relative tokenUrl', async () => {
+      const metadata = createFakePackVersionMetadata({
+        defaultAuthentication: {
+          type: AuthenticationType.OAuth2ClientCredentials,
+          tokenUrl: 'https://example.com/tokenUrl',
+          requiresEndpointUrl: true,
+        },
+      });
+      const err = await validateJsonAndAssertFails(metadata);
+      assert.deepEqual(err.validationErrors, [
+        {
+          message: 'tokenUrl must be a relative URL when requiresEndpointUrl is true',
+          path: 'defaultAuthentication.tokenUrl',
+        },
+      ]);
+
+      assertCondition(metadata.defaultAuthentication?.type === AuthenticationType.OAuth2ClientCredentials);
+      metadata.defaultAuthentication.tokenUrl = '/tokenUrl';
+      await validateJson(metadata);
+    });
+
+    it('OAuth2ClientCredentials, no requiresEndpointUrl requires absolute tokenUrl', async () => {
+      const metadata = createFakePackVersionMetadata({
+        defaultAuthentication: {
+          type: AuthenticationType.OAuth2ClientCredentials,
+          tokenUrl: '/tokenUrl',
+        },
+      });
+      const err = await validateJsonAndAssertFails(metadata);
+      assert.deepEqual(err.validationErrors, [
+        {
+          message: 'tokenUrl must be an absolute URL when requiresEndpointUrl is not true',
+          path: 'defaultAuthentication.tokenUrl',
+        },
+      ]);
+
+      assertCondition(metadata.defaultAuthentication?.type === AuthenticationType.OAuth2ClientCredentials);
+      metadata.defaultAuthentication.tokenUrl = 'https://example.com/tokenUrl';
+      await validateJson(metadata);
+    });
+
+    it('OAuth2ClientCredentials, requiresEndpointUrl rejects protocol-relative tokenUrl', async () => {
+      // A protocol-relative URL like "//evil.com/tokenUrl" starts with "/" but resolves to a
+      // different host than the connection endpoint, so it must not be treated as relative.
+      const metadata = createFakePackVersionMetadata({
+        defaultAuthentication: {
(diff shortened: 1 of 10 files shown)


ekoleda-codaio:
I don't think we should enforce that the token URL __must__ be relative when `requiresEndpointUrl` is set. I understand you are mirroring the existing behavior for `OAuth2` auth, but here is some history:

- `requiredEndpointUrl` was originally intended to capture the endpoint of the API.
- There are APIs where the auth is always at the same endpoint, but requests need to be directed to customer-specific endpoints (Salesforce, Gong, etc). In these cases we need absolute auth URLs and relative fetcher URLs.
- We support the case above for `OAuth2` auth by using `endpointKey`, when the token response includes the API endpoint to use.
- Historically we didn't support the user manually entering the endpoint for `OAuth2` authentication, even though we did support it for other auth types. 
- We later added support for manual endpoints for OAuth2, but behind an allowlist. The decision was made at that time to enforce relative auth URLs in this case, but I believe that was the wrong decision, as it limits the APIs we can support unnecessarily.

Given that we already support manual endpoints in `OAuth2ClientCredentials`, it wouldn't be backwards compatible to change the rules now to enforce relative token URLs. We should allow either absolute or relative token URLs.

ekoleda-codaio (commented):
Docs and behavior LGTM. Leaving to an Ecosystem engineer to approve the implementation.

testing/upload_validation.ts
```diff
@@ -847,17 +847,29 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
     }),
     [AuthenticationType.OAuth2ClientCredentials]: zodCompleteStrictObject<OAuth2ClientCredentialsAuthentication>({
       type: zodDiscriminant(AuthenticationType.OAuth2ClientCredentials),
-      tokenUrl: z.string().url().refine(validateUrlParsesIfAbsolute),
+      /**
+       * May be relative or absolute when requiresEndpointUrl is true (enforced in the superRefine
+       * below) — packs already support a manually-entered endpoint alongside a fixed absolute
+       * tokenUrl, so a relative tokenUrl is offered as an additional option, not a requirement.
+       */
+      tokenUrl: z.string().refine(validateUrlParsesIfAbsolute),
```
daviddriscoll-codaio: This subtly narrows the set of previously accepted absolute URLs. The old `z.string().url()` accepted `http://...`, and `validateUrlParsesIfAbsolute()` returned true for those because it only parses strings beginning with `https://`. After this change, the super-refinement uses `isAbsoluteUrl()`, which only accepts `https://`, so existing HTTP token URLs now fail upload validation. Is that intentional? If not, please preserve the previous absolute-URL semantics; if it is intentional, please call out the breaking change and add coverage for it. — GPT-5.6 SOL

daviddriscoll-codaio:
Could we also update the public JSDoc for `OAuth2ClientCredentialsAuthentication.tokenUrl` in `types.ts`? The generated API reference currently only says “The URL that the platform will hit…” and its example only shows an absolute URL, so users won’t discover the new relative-path behavior from the primary API docs. Please document that a root-relative path is supported when `requiresEndpointUrl` is true, while absolute URLs remain supported. — GPT-5.6 SOL

test/upload_validation_test.ts
```diff
@@ -5188,6 +5188,64 @@ describe('Pack metadata Validation', async () => {
       await validateJson(metadata);
     });
 
+    it('OAuth2ClientCredentials, requiresEndpointUrl allows either a relative or absolute tokenUrl', async () => {
+      // Packs already support a manually-entered endpoint alongside a fixed absolute tokenUrl (auth
```
daviddriscoll-codaio: could we shorten all comments to max 2 lines and ideally eliminate any comments we don't need?

testing/upload_validation.ts
```diff
@@ -3452,21 +3464,35 @@ function validateUrlParsesIfAbsolute(url: string) {
...
+ * one), so enforcing relative-only here would be a backwards-incompatible break for those packs.
  */
 function validateUrlIsRelativeOrAbsolute(
   fieldName: string,
   url: string,
-  {requiresEndpointUrl, endpointKey}: {requiresEndpointUrl?: boolean; endpointKey?: string},
+  {
+    requiresEndpointUrl,
+    endpointKey,
+    allowEitherWhenEndpointRequired,
```
daviddriscoll-codaio: i see how you got here, but this variable name and the resulting logic is kinda confusing. GPT5.6 suggests the following refactor:

```
A clearer abstraction would express URL policy directly, for example:

relativeUrlPolicy: 'forbidden' | 'allowed' | 'required'

For this PR, client credentials would use allowed when requiresEndpointUrl is set. That would eliminate the confusing allowEitherWhenEndpointRequired exception and make the helper’s behavior explicit.
```

but I think such a refactor is out of scope.

daviddriscoll-codaio (approved):

PR #3601, opened 2026-07-22
[CLI] Open the Superhuman token page during registration

wesley-codaio:
## Summary

- Make `coda register` and `packs register` open the Superhuman Docs account page when they use the default API endpoint, instead of the retired Coda account domain.
- Preserve custom and single-tenant API endpoints so their registration flow continues to open the matching account host.
- Cover default, trailing-slash, and custom-endpoint URL generation with focused unit tests.
- Advance the SDK prerelease version to `1.17.1-prerelease.2`, as required for manual changes in this repository.

| Registration endpoint | Before | After |
| --- | --- | --- |
| Default (`https://coda.io`) | Opens `coda.io/account` | Opens `docs.superhuman.com/account` |
| Custom endpoint | Opens that endpoint's account page | Unchanged |

For example, running `coda register` with the default endpoint now opens `https://docs.superhuman.com/account?openDialog=CREATE_API_TOKEN&scopeType=pack#apiSettings`.

## Test plan

- [x] `./scripts/dev/check-version-updated.sh`
- [x] `make test-file FILE=test/cli_test.ts` (6 passing)
- [x] `make build`
- [ ] GitHub CI passes on the final commit

## Other notes

- The generated CLI JavaScript and declarations are updated alongside the TypeScript source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+### Fixed
+
+- The `packs register` and `coda register` commands now open the Superhuman Docs account page when creating a Pack-scoped API token.
+
 ### Changed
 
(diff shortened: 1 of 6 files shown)


wesley-codaio (commented):
**AI-assisted review** — has notes.

Found **1** small cleanup:
I left 1 inline suggestion:
- 1 consistency issue

<details>
<summary>1 finding not tied to a changed line</summary>

- `cli/register.ts:29` — The confirmation prompt still says it will “visit Coda,” but accepting now opens Superhuman Docs, giving users a misleading description of the external destination.

</details>

ekoleda-codaio (approved):

PR #3609, opened 2026-08-05
Add replacesMcpTools formula option

solomon-codaio:
## Summary
Add an `@internal` `replacesMcpTools` field to `CommonPackFormulaDef` so a Superhuman-authored pack formula can declare the agent-facing MCP tool name(s) it replaces. When the formula and the listed MCP tool(s) both resolve for an agent, the agent runtime hides the MCP tool(s) in favor of the purpose-built formula.

## Detail
- Field sits beside the existing agent-facing `instructions`/`purpose`; flows to `PackFormulaMetadata` via the existing `Omit`-based pass-through (no serialization changes).
- Values are the plain tool name the MCP server publishes (e.g. `slack_send_message`); the consuming agent runtime resolves the internal namespacing, so pack authors don't need to know it.
- Adds the matching `commonPackFormulaSchema` entry in `upload_validation.ts` (the `ZodCompleteShape` exhaustiveness guard requires it) and regenerated `dist/`.

## Test plan
CI (`make compile-ts` clean locally). Consuming logic is validated in the front-end-web-monorepo agent-runtime branch.

### DIFF
--- a/api_types.ts
+++ b/api_types.ts
@@ -613,6 +613,16 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
   readonly purpose?: FormulaPurpose;
 
   /**
+   * Names of MCP tools, exposed by this same pack's MCP server(s), that this formula replaces.
+   * When both this formula and the listed MCP tool(s) resolve for an agent, the MCP tool(s) are
+   * hidden in favor of this purpose-built formula. Use the plain tool name your MCP server
+   * publishes (e.g. `slack_send_message`) — the agent runtime maps it to the internal namespaced
+   * name for you. Used by agents; not for use by packs that are not Superhuman-authored.
+   * @internal
+   */
+  readonly replacesMcpTools?: readonly string[];
+
+  /**
    * OAuth scopes that the formula needs that weren't requested in the pack's overall authentication
    * config. For example, a Slack pack can have one formula that needs admin privileges, but non-admins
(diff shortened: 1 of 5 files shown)


api_types.ts
```diff
@@ -612,6 +612,16 @@ export interface CommonPackFormulaDef<T extends ParamDefs> {
    */
   readonly purpose?: FormulaPurpose;
 
+  /**
+   * Names of MCP tools, exposed by this same pack's MCP server(s), that this formula replaces.
+   * When both this formula and the listed MCP tool(s) resolve for an agent, the MCP tool(s) are
+   * hidden in favor of this purpose-built formula. Use the plain tool name your MCP server
```
ekoleda-codaio, line 618: The SDK shape allows for adding multiple MCP servers to a Pack, although by policy we currently limit it to only one. You may want to namespace these to the MCP server in case we ever relax that restriction. Maybe something like:

```
replacesMcpTools: [
  {serverName: "Todoist", toolName: "add_task"},
],
```

ekoleda-codaio: Ah, I just saw this in your PR description:

> Values are the agent-facing (server-name-prefixed) tool names, e.g. Slack_slack_send_message.

I'd recommend against having the SDK refer to these using the generated tool name, as it's an implementation detail and we may change that logic later on.

PR #3610, opened 2026-08-05
Add Go/Docs behavior differences to the documentation

ekoleda-codaio:
Certain connector features work differently in Go and Docs, and this PR adds those caveats to the documentation. New screenshots and videos were also added to certain pages to demonstrate the equivalent feature in Go.

These differences are called using some new conventions, codified in the style guide page. This includes new callout box (admonition) styles, with their look and feel added to the CSS.


docs/guides/blocks/sync-tables/index.md
```diff
@@ -5,23 +5,66 @@ description: Create tables of data that automatically sync in records from an ex
...
 ## Using sync tables
 
-Sync tables are added directly to the document, usually by dragging them in from the side panel. Navigate to {{ custom.pack_panel_clicks }} and drag the table into the canvas.
+Before a sync table can pull in data it must be added and configured.
+
+=== ":superhuman-go: Go"
+
+    Sync tables are configured when installing the connector, during which time the user can also set parameters. Once the connector is installed, the platform will begin indexing the data into the knowledge layer. See the [Indexing guide][indexing] for more information.
+
+    <img src="site:images/sync_table_usage_go.png" srcset="site:images/sync_table_usage_go_2x.png 2x" class="screenshot" alt="Sync table setup in Go.">
```
vivekvenkatram, line 26: Pretty sure these are the wrong asset: `sync_table_usage_go.png` and `sync_table_usage_go_2x.png` are byte-identical to each other and to `parameter_go_sync_2x.png` (all blob `1f269ce0`), so there's no real 2x here and it's likely showing the params-on-install shot rather than the sync-table setup screen. Could we re-export the actual Go sync-table setup screen for both the 1x and 2x?

schema.ts
```diff
@@ -148,6 +150,8 @@ export enum ValueHintType {
    *   },
    * });
    * ```
+   *
+   * Only supported in Superhuman Docs.
```
vivekvenkatram: This says references are Docs-only, but the schemas and sync-tables guides in this same PR frame them as "Limited in Go" (Go still sees the display value + ID, it just doesn't resolve the full row). Should the caveat match that, e.g.

```ts
 * Only partially supported in Superhuman Go, where the reference isn't resolved to the full row.
```

ekoleda-codaio: Fixed.

docs/guides/blocks/sync-tables/index.md
ekoleda-codaio: Good catch, I forgot to scale down the 1x version, fixed now. It is the same image as the parameter one, but I made a copy in case I want to change it in the future.

vivekvenkatram (approved):

PR #3618, opened 2026-08-18
Support defining a custom agent with SDK

peizhaomei-codaio:
## Description

Adds `sdk.newAgent()`, a builder exposing only `setInstructions` and `setVersion`. The agent compiles into a new top-level `agent` field on the pack metadata, so it rides the bundle `packs upload` already ships. An agent is still a pack — same commands, same upload path. No CLI changes, and `newPack` is untouched.

```ts
export const pack = sdk.newAgent();
pack.setInstructions('You help a team run async standups. Keep replies short.');

{
  "agent": {"instructions": "You help a team run async standups. Keep replies short."},
  "version": "1.0.0",
  "formats": [],
  "formulas": [],
  "syncTables": []
}
```
## Future

Tools and triggers, the connector surface, export const agent / newConnector.


builder.ts
```diff
@@ -48,8 +63,28 @@ import {wrapMetadataFunction} from './api';
...
+ *
+ * `isCustomAgent` must be the literal `true` for the agent authoring methods to resolve. If you
+ * are generating agents in bulk from a dynamic flag, declare it `as const` or cast the result.
+ *
+ * @internal
+ * @hidden
+ */
+export interface NewCustomAgentOptions extends Partial<PackVersionDefinition> {
+  isCustomAgent: true;
 }
```
vivekvenkatram: Bit worried this lets `newPack({isCustomAgent: true, formulas: [...]})` type-check and copy those fields through `super()`. Wdyt about narrowing it to just the agent bits?

```ts
export interface NewCustomAgentOptions {
  isCustomAgent: true;
  version?: string;
  customAgent?: CustomAgentConfig;
}
```

builder.ts
```diff
@@ -613,3 +657,129 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
...
+ * the agent's own surface: tools and triggers are not authorable yet and will be added here.
+ *
+ * The methods return `CustomAgentDefinitionBuilder` rather than `this` on purpose. A derived
+ * type (`Omit<PackDefinitionBuilder, ...>`) would resolve `this` back to the full builder, so
+ * the withheld methods would reappear after the first chained call.
+ *
+ * @internal
+ * @hidden
+ */
+export interface CustomAgentDefinitionBuilder extends CustomAgentDefinition {
```
vivekvenkatram: Looks good on the methods, but `extends CustomAgentDefinition` still leaves `formulas` / `syncTables` writable, so `pack.formulas.push(...)` type-checks. If withholding the connector surface is the invariant, maybe this interface shouldn't inherit the full pack shape?

```ts
export interface CustomAgentDefinitionBuilder {
  customAgent: CustomAgentConfig;
  version?: string;
  setInstructions(instructions: string): CustomAgentDefinitionBuilder;
  setVersion(version: string): CustomAgentDefinitionBuilder;
}
```

types.ts
```diff
@@ -1729,6 +1729,22 @@ export interface SuggestedPrompt {
...
+ * Display fields (name, description, icon) are not part of this; they come from the pack's
+ * listing. Tools and triggers are not authorable yet; both will be added as optional fields,
+ * so neither is a breaking change for agents written against this shape.
+ *
+ * @internal
+ * @hidden
+ */
+export interface CustomAgentConfig {
+  /** The agent's instructions, i.e. its prompt. */
+  instructions?: string;
```
vivekvenkatram, line 1742: Am I missing a mapping on the server PR, or are we intentionally calling this `instructions`? The persisted custom-agent blob is still `{skill: {prompt, tools}}`, so I want to make sure the names don't fork by accident.

```ts
prompt?: string; // match Skill / CustomAgentBuilderDefinition
```

types.ts
```diff
@@ -1833,6 +1849,30 @@ export interface PackVersionDefinition {
...
+  /**
+   * The custom agent defined by this pack, if it defines one.
+   *
+   * Authored via `sdk.newPack({isCustomAgent: true})`. Its presence is what marks this pack
+   * as a custom agent, both for local validation and for the server on upload.
+   *
+   * @internal
+   * @hidden
+   */
+  customAgent?: CustomAgentConfig;
```
vivekvenkatram: If presence of this field is the classification signal, `newPack({customAgent: {instructions: '...'}})` without the flag is also an agent. Is that the contract (field = source of truth, flag = DX only)?

testing/upload_validation.ts
```diff
@@ -2526,6 +2531,7 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
       }),
     chatSkill: chatSkillSchema.optional(),
     benchInitializationSkill: chatSkillSchema.optional(),
+    customAgent: customAgentSchema.optional(),
```
vivekvenkatram: Little concerned we accept `customAgent` next to formulas/sync tables/MCP. If agents aren't supposed to ship a fetcher, could we reject that mix here so the throwing builder isn't the only gate?

```ts
.superRefine((data, context) => {
  if (!data.customAgent) {
    return;
  }
  if (data.formulas?.length) {
    context.addIssue({
      code: 'custom',
      path: ['formulas'],
      message: 'A custom agent cannot also define connector building blocks.',
    });
  }
});
```

testing/upload_validation.ts
```diff
@@ -2372,6 +2373,10 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
   });
   const chatSkillSchema = skillSchema.partial();
 
+  const customAgentSchema = zodCompleteStrictObject<CustomAgentConfig>({
+    instructions: z.string().min(1).max(Limits.PromptLength),
+  });
```
vivekvenkatram: Could we add a couple `validateJson` cases while we're here? Happy path with `instructions`, fail on `customAgent: {}`, and a `compilePackMetadata` round-trip so the field doesn't get stripped.

vivekvenkatram:
Slight preference for separate factories instead of a flag on `newPack`:

```ts
export const agent = sdk.newAgent();
export const connector = sdk.newConnector();
```

Why do we need `isCustomAgent: true`? The thing we actually need is "the author meant this to be an agent," including before they call `setInstructions`. A different function does that without a boolean, and TypeScript gets a real return type so we don't have to fight `this` / `Omit` / overload order.

Introspecting the finished object (presence of `customAgent`) is still the right *upload* signal. It just isn't available at `newPack()` time. If we wait until `setInstructions` to decide, a forgotten call looks like an empty connector, which is the bug the empty `{}` is trying to catch. Looking at the export name (`pack` vs `agent`) is a heuristic too, and it breaks the moment someone writes `export default` or `export const pack = newAgent()`.

Compatibility: I wouldn't rename `newPack` in this PR. Leave it as the connector path (every existing pack uses `export const pack = newPack()`), add `newAgent()`, and let `newConnector` be an alias later if we want the naming. `importManifest` today is `module.pack || module.manifest`, so `export const agent = newAgent()` would need `module.agent` added there, or authors keep `export const pack = newAgent()` and the factory is the only signal. CLI/upload otherwise stay the same: still a pack version, `customAgent` on the metadata is what the server classifies.

The in-app builder blob can stay `{skill: {prompt, tools}}`. Pack-authored agents are a new path going forward; the server maps `customAgent.instructions` into whatever we persist.

builder.ts
```diff
@@ -52,6 +54,32 @@ export function newPack(definition?: Partial<PackVersionDefinition>): PackDefini
...
+ * here rather than inferring it from what the author eventually calls means a forgotten
+ * `setInstructions` is caught as an incomplete agent instead of passing as an empty pack.
+ *
+ * It takes no seed definition on purpose. Accepting one would let connector fields — formulas,
+ * sync tables, authentication — reach the builder without going through the methods that
+ * withhold them.
+ *
+ * @example
+ * ```
+ * export const pack = newAgent();
```
vivekvenkatram, line 61: Fwiw I'd keep this as `pack` rather than `export const agent`. `importManifest` is still `module.pack || module.manifest`, so a different export name wouldn't load until we add another lookup.

```ts
export const pack = sdk.newAgent();
pack.setInstructions('You help a team run async standups. Keep replies short.');
```

index.ts
```diff
@@ -25,6 +25,7 @@ export type {PostSetupDef} from './types';
 export {PostSetupType} from './types';
 export type {SuggestedPrompt} from './types';
 export {newPack} from './builder';
+export {newAgent} from './builder';
```
vivekvenkatram, line 28: Wdyt about leaving `newPack` as-is rather than renaming it to `newConnector` in this PR? Every live pack is `export const pack = newPack()`, so a rename would be a breaking change we don't need here.

```ts
export {newPack} from './builder';
export {newAgent} from './builder';
```

builder.ts
```diff
@@ -613,3 +650,134 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
     return this;
   }
 }
+
+/**
+ * A builder for a custom agent. Use `sdk.newPack({isCustomAgent: true})` to create one.
```
vivekvenkatram: This still says `newPack({isCustomAgent: true})`. Could we point it at `newAgent()` instead?

```ts
 * A builder for a custom agent. Use {@link newAgent} to create one.
```

builder.ts
```diff
@@ -613,3 +650,134 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
...
+/**
+ * The runtime implementation behind {@link CustomAgentDefinitionBuilder}.
+ *
+ * `CustomAgentDefinitionBuilder` is what hides the connector-authoring methods; a subclass
+ * cannot, because TypeScript does not allow a subclass to narrow the visibility of an
+ * inherited member. The overrides here exist so that callers who reach past the types — a
+ * cast, or plain JavaScript — fail loudly at build time instead of silently uploading a
+ * definition the server will not honor.
+ *
+ * Not exported: `newPack` is the only way to construct one.
```
vivekvenkatram: Pretty sure this is leftover from the flag version. `newAgent()` is the only caller now.

```ts
 * Not exported: {@link newAgent} is the only way to construct one.
```

types.ts
vivekvenkatram: Field presence is still the upload signal; `newAgent()` is just how we stamp the empty `{}` at construction. The JSDoc still names `newPack({isCustomAgent: true})` though, should we point it at `newAgent()`?

index.ts
peizhaomei-codaio: I support keeping `newPack` for now

testing/upload_validation.ts
peizhaomei-codaio: Tests added

builder.ts
```diff
@@ -52,6 +53,32 @@ export function newPack(definition?: Partial<PackVersionDefinition>): PackDefini
   return new PackDefinitionBuilder(definition);
 }
 
+/**
+ * Creates a new skeleton custom agent definition that can be added to.
+ *
+ * An agent is still a pack: it is uploaded and released through the same commands, and the
```
ekoleda-codaio: All of the descriptions in this PR read like AI slop. Too much detail, and including internal-only information. Please trim them way down, to just the basics that an external developer would need to know.

builder.ts
ekoleda-codaio: I think it would be nice to support `export const agent` and `export const connector` eventually, but it can be done later.

builder.ts
```diff
@@ -105,6 +132,13 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
    */
   mcpServers: MCPServer[];
 
+  /**
+   * See {@link PackVersionDefinition.customAgent}.
+   * @internal
+   * @hidden
+   */
+  customAgent?: CustomAgentConfig;
```
ekoleda-codaio: I know we call these "Custom agents" internally, but I don't think the "custom" part is really needed in the SDK. There is no non-custom agent to compare it to. I'd suggest dropping custom from field names, interface names, etc.

builder.ts
```diff
@@ -613,3 +649,142 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
     return this;
   }
 }
+
+/**
+ * A builder for a custom agent. Use {@link newAgent} to create one.
```
ekoleda-codaio: Ditto, about this description reading like slop.

builder.ts
```diff
@@ -613,3 +649,142 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
...
+ * inherited member. The overrides here exist so that callers who reach past the types — a
+ * cast, or plain JavaScript — fail loudly at build time instead of silently uploading a
+ * definition the server will not honor.
+ *
+ * Not exported: {@link newAgent} is the only way to construct one.
+ *
+ * @internal
+ * @hidden
+ */
+class CustomAgentDefinitionBuilderImpl extends PackDefinitionBuilder {
```
ekoleda-codaio: (A bit out of my depth with TypeScript) Rather than extend the existing building and having to override all of the methods with errors, would it make more sense to extract the common methods into a BaseDefinitionBuilder that both inherit from?

index.ts
ekoleda-codaio: I do think we should eventually move to `newConnector()`, but that can be done later.

types.ts
```diff
@@ -1833,6 +1846,13 @@ export interface PackVersionDefinition {
    * Definitions of MCP servers that this pack can connect to.
    */
   mcpServers?: MCPServer[];
+  /**
+   * The agent defined by this pack, if it defines one. Authored via `sdk.newAgent()`.
+   *
+   * @internal
+   * @hidden
+   */
+  agent?: AgentConfig;
```
xicheng-codaio: how do you plan to serialize the new field? 
I don't know if the BE will support this field easily

peizhaomei-codaio: we can build a full `Skill` object from `metadata.agent.skill` in `insertPackVersion`, filling the missing fields from the pack listing, and write it into custom_agent_definitions, if this is what you are asking about

types.ts
```diff
@@ -1729,6 +1729,19 @@ export interface SuggestedPrompt {
...
+ * The definition of an agent.
+ *
+ * @internal
+ * @hidden
+ */
+export interface AgentConfig {
+  /**
+   * The agent's behavior. Its name and description come from the pack's listing, not here.
+   */
+  skill?: Partial<Skill>;
```
vivekvenkatram: Wdyt about keeping this agent-shaped (`prompt` / `tools`) instead of `Partial<Skill>`? The wrap-as-skill step can live on the server when we write the definition table.

```ts
export interface AgentConfig {
  prompt?: string;
  tools?: Tool[];
}
```

vivekvenkatram: Nit: `AgentDefinition` would match `AgentDefinitionBuilder` / `PackVersionDefinition` better than `Config` (that's knobs like `SkillModelConfiguration` here). Field can stay `agent`.

```ts
export interface AgentDefinition {
  prompt?: string;
  tools?: Tool[];
}

agent?: AgentDefinition;
```

builder.ts
vivekvenkatram: Yeah, `PackDefinitionBuilder`'s generated page now lists `Extends BaseDefinitionBuilder`. Could we keep that class unexported so it stays out of the public docs?

builder.ts
vivekvenkatram: +1, "custom" doesn't belong in the SDK. Leftover naming I'd still change: `AgentConfig` → `AgentDefinition`, so it matches `AgentDefinitionBuilder`.

index.ts
vivekvenkatram: +1, later is fine. I'd keep `newPack` as the connector path until we're ready to alias `newConnector`.

types.ts
vivekvenkatram: The field just rides along as JSON (`compilePackMetadata` rest-spreads leftover keys). I'd keep `agent` as `{prompt, tools}` and do the Skill wrap when we publish the definition row, not in `insertPackVersion` (that would make `coda upload` live).

builder.ts
vivekvenkatram: +1, later is fine. Load is still `module.pack || module.manifest`, so `export const pack = newAgent()` is the path that works today.

builder.ts
peizhaomei-codaio: > Yeah, `PackDefinitionBuilder`'s generated page now lists `Extends BaseDefinitionBuilder`. Could we keep that class unexported so it stays out of the public docs?

Tried dropping the export, didn't remove the`Extends BaseDefinitionBuilder`. The export only controls whether it gets its own page, and it already doesn't have one. The only way to remove this is for `PackDefinitionBuilder` not to extend the base.

types.ts
peizhaomei-codaio: Both updated

types.ts
```diff
@@ -1833,6 +1850,13 @@ export interface PackVersionDefinition {
    * Definitions of MCP servers that this pack can connect to.
    */
   mcpServers?: MCPServer[];
+  /**
+   * The agent defined by this pack, if it defines one. Authored via `sdk.newAgent()`.
+   *
+   * @internal
+   * @hidden
+   */
+  agent?: AgentDefinition;
```
sam-codaio, line 1859: Why are there not  mutually exclusive types that are then unioned if we have the validation that prevents having this and other fields? Would be clearer to enforce via TS

types.ts
```diff
@@ -1729,6 +1729,23 @@ export interface SuggestedPrompt {
   prompt: string;
 }
 
+/**
+ * The definition of an agent.
+ *
+ * @internal
+ * @hidden
+ */
+export interface AgentDefinition {
```
sam-codaio, line 1738: What are "skills" if we have thisd new object? When would someone use them, etc

vivekvenkatram: We're leaving existing skills as is for now. We will revisit skills as a noun in a separate conversation

types.ts
peizhaomei-codaio: `PackVersionDefinition` is depended on heavily through `Partial<>` and `Omit<>`, so making it a union would need to rewrite the upload pipeline and break the public type

vivekvenkatram: Peizhao's right that unioning `PackVersionDefinition` is expensive (`Omit`/`Partial` on a union only keep shared keys, and `BasicPackDefinition` / `PackVersionMetadata` / `newPack` all sit on that). Wdyt about dropping `agent` off the connector seed instead, so only `newAgent()` can stamp it?

```ts
newPack(definition?: Partial<Omit<PackVersionDefinition, 'agent'>>)
```

builder.ts
```diff
@@ -671,7 +671,7 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
    * ```
    */
   setInstructions(instructions: string): this {
-    this.agent = {...this.agent, prompt: instructions};
+    this.agent.prompt = instructions;
```
vivekvenkatram: Can we call this instructions instead of prompt?

ekoleda-codaio (commented):
LGTM, leaving approval to someone in the eng team to confirm the implementation.

vivekvenkatram (approved):

vivekvenkatram (changes requested):
requesting changes

types.ts
```diff
@@ -1833,6 +1850,13 @@ export interface PackVersionDefinition {
    * Definitions of MCP servers that this pack can connect to.
    */
   mcpServers?: MCPServer[];
+  /**
+   * The agent defined by this pack, if it defines one. Authored via `sdk.newAgent()`.
```
vivekvenkatram, line 1854: Bit worried `agent.instructions` lands on the public pack-metadata CDN. Skills already strip `prompt`, but `BasePackVersionMetadata` still keeps `agent` and the server rest-spreads leftover keys, so I'd omit it (or at least `instructions`) from the external type:

```ts
export type AgentMetadata = Omit<AgentDefinition, 'instructions'>;
```

testing/upload_validation.ts
```diff
@@ -3172,6 +3200,40 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
         }
       });
+    })
+    .superRefine((data, context) => {
+      // Agents don't ship connector building blocks. The agent builder can't add them, but a
+      // hand-written manifest can, so check here too.
+      if (!(data as PackVersionMetadata).agent) {
+        return;
+      }
+      const connectorBuildingBlocks: Array<[keyof PackVersionMetadata, string]> = [
```
vivekvenkatram: A hand-written mix of `chatSkill` / `skillEntrypoints` / `suggestedPrompts` still uploads, since those aren't in this list. Could we reject them off presence (they're not arrays)?

```ts
for (const field of ['chatSkill', 'benchInitializationSkill', 'skillEntrypoints'] as const) {
  if ((data as PackVersionMetadata)[field]) {
    context.addIssue({
      code: 'custom',
      path: [field],
      message: `An agent cannot also define ${field}.`,
    });
  }
}
```

builder.ts
```diff
@@ -48,14 +49,61 @@ import {wrapMetadataFunction} from './api';
  * pack.setUserAuthentication({type: AuthenticationType.HeaderBearerToken});
  * ```
  */
-export function newPack(definition?: Partial<PackVersionDefinition>): PackDefinitionBuilder {
+export function newPack(definition?: Partial<Omit<PackVersionDefinition, 'agent'>>): PackDefinitionBuilder {
```
vivekvenkatram, line 52: `newPack` drops `agent` here, but `PackDefinitionBuilder`'s constructor is still `Partial<PackVersionDefinition>`, so `new PackDefinitionBuilder({agent})` typechecks and then silently drops it. Wdyt about the same Omit on the constructor?

```ts
constructor(definition?: Partial<Omit<PackVersionDefinition, 'agent'>>) {
```

testing/upload_validation.ts
```diff
@@ -2372,6 +2373,32 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
   });
   const chatSkillSchema = skillSchema.partial();
 
+  // The skill's display fields come from the pack listing, so only the prompt is required here.
```
vivekvenkatram: This still says prompt. Could we point it at instructions?

```ts
// Listing owns the display fields, so only instructions are required here.
```

test/upload_validation_test.ts
```diff
@@ -8069,6 +8069,89 @@ describe('Pack metadata Validation', async () => {
...
+      const err = await validateJsonAndAssertFails(
+        createFakeAgentMetadata({
+          agent: {instructions: 'Do a thing.', tools: [{type: 'NotARealTool'} as any]},
+        }),
+      );
+      assert.isNotEmpty(err.validationErrors);
+    });
+
+    // The agent builder can't produce this, but a hand-written manifest can.
+    it('rejects connector building blocks alongside an agent', async () => {
```
vivekvenkatram, line 8120: Could we add a `chatSkill` / `adminAuthentications` XOR case next to this, plus one that extra keys like `name` fail the strict object? If we strip `instructions` from external metadata, an assert there would lock it in.

types.ts
```diff
@@ -1729,6 +1729,23 @@ export interface SuggestedPrompt {
   prompt: string;
 }
 
+/**
+ * The definition of an agent.
```
vivekvenkatram, line 1733: Wdyt about keeping name/description off this? Listing already owns those, and putting them here would be a third copy next to the definition table.

```ts
export interface AgentDefinition {
  instructions?: string;
  tools?: Tool[];
}
```

builder.ts
```diff
@@ -613,3 +640,29 @@ export class PackDefinitionBuilder implements BasicPackDefinition {
...
+ * A class that assists in constructing an agent definition. Use {@link newAgent} to create one.
+ *
+ * @internal
+ * @hidden
+ */
+export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
+  /**
+   * See {@link PackVersionDefinition.agent}.
+   */
+  agent: AgentDefinition = {};
```
vivekvenkatram, line 654: Empty `{}` is the upload discriminant, but `classifyPack` on the server still ignores `agent`, so these will list as Connector. Could we have the docs PR treat `metadata.agent` as CustomAgent before this ships?

```ts
if (metadata.agent) {
  return PackClassification.CustomAgent;
}
```

builder.ts
```diff
@@ -48,14 +49,61 @@ import {wrapMetadataFunction} from './api';
...
+ * @example
+ * ```
+ * export const pack = newAgent();
+ * pack.setInstructions('You help a team run async standups. Keep replies short.');
+ * ```
+ *
+ * @internal
+ * @hidden
+ */
+export function newAgent(): AgentDefinitionBuilder {
```
vivekvenkatram, line 68: Pretty sure GOECO-510 will 400 the next `coda upload` once the listing is CustomAgent, because it freezes pack-code edits. Docs needs to allow versions that have `agent`, or authors get stuck after the first upload.

```ts
if (version.agent) {
  // skip assertPackCodeCanBeModified
}
```

testing/upload_validation.ts
vivekvenkatram: Same list should include `adminAuthentications` (that's an array, so `.length` works).

```ts
['skills', 'mcpServers', 'networkDomains', 'adminAuthentications'],
```

types.ts
vivekvenkatram: Fwiw Latest should read `metadata.agent` off the uploaded pack version (that fetch already exists). CAD wrap stays on release so `coda upload` doesn't go live.

types.ts
vivekvenkatram: Keeping `instructions` on the pack file. Server maps it to CAD `skill.prompt` on release; Latest reads `metadata.agent.instructions` and shouldn't look for `.prompt` or `.skill`.

types.ts
peizhaomei-codaio: Keeping name/description off sounds good

vivekvenkatram (approved):
lgtm

PR #3620, opened 2026-08-20
Add an MCP tutorial

ekoleda-codaio:
A lot of partners are building MCP-based connectors, so wanted to add a tutorial that walked through the process end-to-end.

This PR also removes some getting started tutorials for GitHub Codespaces and Replit. I don't think they added a ton of value, and with AI-assisted coding I suspect that local development will be the norm. Likewise, I've also moved the CLI getting started tutorial to the first position in the nav.


docs/images/tutorial_mcp_search_2x.png
tesserachin-codaio, line 1: mostly out of curiosity - how do we keep screenshots in these docs accurate? is there a process to get them updated when we update the UI?

tesserachin-codaio (approved):

docs/images/tutorial_mcp_search_2x.png
ekoleda-codaio: It's currently just best effort, and in practice, they do fall out of date over time. For this reason, I tend to use screenshots infrequently. There are more advanced solutions (like automating the screenshot generation, using the actual React component instead of a screenshot), but those are much more involved to set up, and I don't think they are warranted given the scale of our developer audience and the value of screenshots for the audience.

PR #3624, opened 2026-08-24
Require an agent's instructions in the type

peizhaomei-codaio:
`AgentDefinition.instructions` is required now, so a definition without one is a compile error rather than something validation catches later.

The builder holds a `Partial<AgentDefinition>`, since it starts as an empty object and fills in as the author calls methods — same split as `version`, which `PackVersionDefinition` requires and `BaseDefinitionBuilder` doesn't. The empty object is still what marks a pack as an agent.

That also drops the hand-written check for missing instructions: zod covers it now, so the error is the standard "Missing required field".

### DIFF
--- a/builder.ts
+++ b/builder.ts
@@ -651,7 +651,7 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
   /**
    * See {@link PackVersionDefinition.agent}.
    */
-  agent: AgentDefinition = {};
+  agent: Partial<AgentDefinition> = {};
 
   /**
    * Sets this agent's instructions.
--- a/test/upload_validation_test.ts
+++ b/test/upload_validation_test.ts
@@ -8087,12 +8087,24 @@ describe('Pack metadata Validation', async () => {
       assert.deepEqual(result.agent, {instructions: 'You help a team run async standups.'});
     });
 
-    // The builder can't produce this — `instructions` is required on `AgentDefinition` — but a
-    // hand-written manifest can.
+    // Neither is reachable through the builder, but a hand-written manifest can produce both.
     it('rejects an agent that declared itself but never set instructions', async () => {
       const err = await validateJsonAndAssertFails(createFakeAgentMetadata({agent: {} as any}));
       assert.deepEqual(err.validationErrors, [
-        {path: 'agent.instructions', message: 'Missing required field agent.instructions.'},
+        {
+          path: 'agent.instructions',
+          message: 'An agent must have instructions. Call setInstructions() on the agent.',
+        },
+      ]);
+    });
+
+    it('rejects an agent whose instructions are empty', async () => {
+      const err = await validateJsonAndAssertFails(createFakeAgentMetadata({agent: {instructions: ''}}));
+      assert.deepEqual(err.validationErrors, [
+        {
+          path: 'agent.instructions',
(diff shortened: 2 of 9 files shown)


types.ts
```diff
@@ -1739,7 +1739,7 @@ export interface AgentDefinition {
   /**
    * What the agent is told to do.
    */
-  instructions?: string;
+  instructions: string;
```
vivekvenkatram, line 1742: Love making this required on the compiled contract. One JSDoc line might help, though: `newAgent()` still holds a `Partial`, so `tsc` will not catch a missing `setInstructions()` the way a raw `AgentDefinition` object literal will.

builder.ts
```diff
@@ -651,7 +651,7 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
   /**
    * See {@link PackVersionDefinition.agent}.
    */
-  agent: AgentDefinition = {};
+  agent: Partial<AgentDefinition> = {};
```
vivekvenkatram, line 654: The empty `{}` still marking an agent is the right split (same idea as `version`). Just flagging that this is still a validation-time check for the real authoring path, not a compile error, since this field stays `Partial` until upload.

testing/upload_validation.ts
```diff
@@ -2375,7 +2375,7 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
 
   // The listing owns the display fields, so only the instructions are required here.
   const agentSchema = zodCompleteStrictObject<AgentDefinition>({
-    instructions: z.string().min(1).max(Limits.PromptLength).optional(),
+    instructions: z.string().min(1).max(Limits.PromptLength),
```
vivekvenkatram: Bit worried we lost the "Call setInstructions() on the agent" copy. `{}` and `""` also hit different Zod codes now (missing vs `min(1)`), so authors get two generic messages for the same mistake. Wdyt about keeping one explicit message?

```ts
instructions: z
  .string({error: 'An agent must have instructions. Call setInstructions() on the agent.'})
  .min(1, 'An agent must have instructions. Call setInstructions() on the agent.')
  .max(Limits.PromptLength),
```

test/upload_validation_test.ts
```diff
@@ -8087,23 +8087,12 @@ describe('Pack metadata Validation', async () => {
       assert.deepEqual(result.agent, {instructions: 'You help a team run async standups.'});
     });
 
+    // The builder can't produce this — `instructions` is required on `AgentDefinition` — but a
+    // hand-written manifest can.
     it('rejects an agent that declared itself but never set instructions', async () => {
-      const err = await validateJsonAndAssertFails(createFakeAgentMetadata({agent: {}}));
+      const err = await validateJsonAndAssertFails(createFakeAgentMetadata({agent: {} as any}));
```
vivekvenkatram, line 8092: Nice that this still covers a hand-written `agent: {}`. Could we also assert `instructions: ''`? That's a different Zod failure than missing, and the old `if (!data.instructions)` treated them the same.

```ts
it('rejects an agent with empty instructions', async () => {
  const err = await validateJsonAndAssertFails(
    createFakeAgentMetadata({agent: {instructions: ''}}),
  );
  assert.isNotEmpty(err.validationErrors);
});
```

vivekvenkatram (approved):

PR #3626, opened 2026-08-25
Turn on Connectors, Docs, Mail, and web search for Agent

peizhaomei-codaio:
## Description

Adds setTools to the agent builder, for the three tools and connectors the builder offers, not including knowledge and MCP.
```
pack.setTools({
  docs: true,
  mail: true,
  webSearch: {allowedDomains: ['status.example.com']},
  connectors: [{packId: 1234, formulas: [{formulaName: 'CreateTask'}]}],
});
```
Serializes to the same tool list the in-app builder writes. tools is required now, and empty rather than absent when an agent has none.


types.ts
```diff
@@ -1741,9 +1741,46 @@ export interface AgentDefinition {
    */
   instructions: string;
   /**
-   * The tools the agent may use.
+   * The tools the agent may use. An empty list means the agent has no tools, and is never a
+   * request for a default set — agents have none.
    */
-  tools?: Tool[];
+  tools: Tool[];
```
vivekvenkatram: Pretty sure `Tool[]` lets ContactResolution / Knowledge / MCP type-check on an agent, and only Zod catches them. Wdyt about a dedicated union so the stored type matches the allowlist?

```ts
export type AgentTool =
  | PackTool
  | WebSearchTool
  | CodaDocsAndTablesTool
  | MailAndCalendarTool;

tools: AgentTool[];
```

types.ts
```diff
@@ -1741,9 +1741,46 @@ export interface AgentDefinition {
...
+ * {@link AgentDefinition.tools}.
+ *
+ * @internal
+ * @hidden
+ */
+export interface AgentToolsDef {
+  /**
+   * Read and write Superhuman Docs documents and tables.
+   */
+  docs?: boolean;
```
vivekvenkatram, line 1776: Little concerned a boolean can't carry the read/write consent the in-app builder already stores, so copying a real builder blob onto `metadata.agent` would fail strict validate. Could `docs` grow the same way `webSearch` did, with `setTools` still omitting the modes when they're the defaults?

```ts
docs?: boolean | {
  readApprovalMode?: 'auto' | 'alwaysAsk';
  writeApprovalMode?: 'auto' | 'alwaysAsk';
};
```

types.ts
```diff
@@ -1741,9 +1741,46 @@ export interface AgentDefinition {
...
+  /**
+   * Read and send Superhuman Mail email, and read the calendar.
+   */
+  mail?: boolean;
```
vivekvenkatram, line 1780: Fwiw `mail: true` is the right shorthand, but a pack that sends on a schedule has no way to set write to auto, and the runtime already honors `writeApprovalMode` when it's present. Same object escape hatch as docs would cover that without a second schema change.

```ts
mail?: boolean | {
  readApprovalMode?: 'auto' | 'alwaysAsk';
  writeApprovalMode?: 'auto' | 'alwaysAsk';
};
```

types.ts
```diff
@@ -1741,9 +1741,46 @@ export interface AgentDefinition {
...
+  /**
+   * Search the public internet. Pass `allowedDomains` to restrict results to those domains;
+   * subdomains are included, and the scheme is omitted.
+   */
+  webSearch?: boolean | {allowedDomains?: string[]};
```
vivekvenkatram, line 1784: `true` vs `{allowedDomains}` is a nice author shape. Could we leave room for `approvalMode` too? Unknown-URL fetch already reads that field, and omitting it still defaults to always-ask.

```ts
webSearch?: boolean | {allowedDomains?: string[]; approvalMode?: 'auto' | 'alwaysAsk'};
```

types.ts
```diff
@@ -1741,9 +1741,46 @@ export interface AgentDefinition {
...
+  /**
+   * Connector packs whose formulas this agent can call. Omit `formulas` to offer all of them.
+   */
+  connectors?: Array<{
```
vivekvenkatram, line 1788: Bit worried "connectors" here only means Pack formulas, while the in-app builder's connector is one `packId` that can also emit MCP (and later knowledge). If we keep those as keys on this same slot, we don't have to rename the API when they land.

```ts
connectors?: Array<{
  packId: number;
  formulas?: Array<{formulaName: string; enabled?: boolean; approvalMode?: 'auto' | 'alwaysAsk'}>;
  // mcp / knowledge later, same packId
}>;
```

builder.ts
```diff
@@ -665,4 +668,32 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
...
+   * ```
+   * pack.setTools({docs: true, mail: true, webSearch: {allowedDomains: ['docs.example.com']}});
+   * pack.setTools({connectors: [{packId: 1234, formulas: [{formulaName: 'CreateTask'}]}]});
+   * ```
+   */
+  setTools({docs, mail, webSearch, connectors}: AgentToolsDef): this {
+    const tools: Tool[] = [];
+    if (webSearch) {
+      const allowedDomains = typeof webSearch === 'object' ? webSearch.allowedDomains : undefined;
+      tools.push({type: ToolType.WebSearch, ...(allowedDomains?.length ? {allowedDomains} : {})});
```
vivekvenkatram: Am I missing something, or does dropping an empty `allowedDomains` turn a mistake into unrestricted search, which validate would have rejected if the array stayed? Wdyt about keeping the empty list so the author actually sees that error?

```ts
tools.push({type: ToolType.WebSearch, ...(allowedDomains ? {allowedDomains} : {})});
```

builder.ts
```diff
@@ -665,4 +668,32 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
...
+    }
+    if (docs) {
+      tools.push({type: ToolType.CodaDocsAndTables});
+    }
+    if (mail) {
+      tools.push({type: ToolType.MailAndCalendar});
+    }
+    for (const connector of connectors || []) {
+      tools.push({type: ToolType.Pack, ...connector});
```
vivekvenkatram: Little concerned the spread will copy whatever we add to the connector object later (mcp, knowledge) onto a Pack tool. Could we pick the Pack fields explicitly?

```ts
tools.push({
  type: ToolType.Pack,
  packId: connector.packId,
  ...(connector.formulas ? {formulas: connector.formulas} : {}),
});
```

testing/upload_validation.ts
```diff
@@ -2376,20 +2376,48 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+
   const agentSchema = zodCompleteStrictObject<AgentDefinition>({
     instructions: z.string({error: MissingInstructions}).min(1, MissingInstructions).max(Limits.PromptLength),
     tools: z
       .array(toolSchema)
       .optional()
+      .default([])
       .superRefine((tools, context) => {
-        for (const duplicate of findDuplicateTools((tools || []) as Tool[])) {
+        for (const duplicate of findDuplicateTools(tools as Tool[])) {
```
vivekvenkatram: `findDuplicateTools` only flags deep-equal copies, so two WebSearch entries with different domain lists both pass, and two Pack tools with the same `packId` but different formula lists do too. Could we uniqueness-check the singletons by `type` and Pack by `packId`?

```ts
const singletonTypes = new Set([ToolType.WebSearch, ToolType.CodaDocsAndTables, ToolType.MailAndCalendar]);
```

test/upload_validation_test.ts
```diff
@@ -8219,6 +8219,68 @@ describe('Pack metadata Validation', async () => {
...
+      const err = await validateJsonAndAssertFails(
+        createFakeAgentMetadata({agent: {instructions: 'Do a thing.', tools: [{type: ToolType.Pack}]}}),
+      );
+      assert.deepEqual(err.validationErrors, [
+        {path: 'agent.tools[0].packId', message: 'An agent must name the connector pack it uses.'},
+      ]);
+    });
+
+    // The in-app builder writes `enabled` and `approvalMode` per formula; PackTool has neither.
+    it('rejects a connector formula carrying fields the SDK has no room for', async () => {
```
vivekvenkatram: This test is fighting the PR description a bit: the in-app builder does write `enabled` / `approvalMode`, and the runtime already defaults them when they're missing. Could we accept those as optional instead of treating a builder blob as invalid?

```ts
formulas: z
  .array(
    z.object({
      formulaName: z.string().min(1),
      enabled: z.boolean().optional(),
      approvalMode: z.enum(['auto', 'alwaysAsk']).optional(),
    }),
  )
  .optional(),
```

test/builder_test.ts
```diff
@@ -667,6 +757,8 @@ describe('Agent builder', () => {
         builder.formulas.push();
         // @ts-expect-error an agent takes no seed definition
         newAgent({formulas: []});
+        // @ts-expect-error an agent cannot use every tool type
+        builder.setTools({mcp: true});
```
vivekvenkatram, line 760: Nit: `mcp: true` isn't the shape we'd want anyway (MCP is per `packId`, same slot as formulas). Fine as an unknown-key check, just wouldn't want someone to copy it as the future API.

```ts
// @ts-expect-error MCP belongs on a connector packId, not as a top-level boolean
builder.setTools({mcp: true});
```

types.ts
```diff
@@ -1741,9 +1779,55 @@ export interface AgentDefinition {
...
+  tools: AgentTool[];
+}
+
+/**
+ * A tool an agent can use.
+ *
+ * @internal
+ * @hidden
+ */
+export type AgentTool =
```
vivekvenkatram, line 1760: Looks good, this is the lock I wanted: ContactResolution / Knowledge / MCP no longer type-check onto an agent. Grow this union only when `setTools` actually emits those tools, so we don't get a second allowlist.

```ts
export type AgentTool =
  | CodaDocsAndTablesTool
  | MailAndCalendarTool
  | WebSearchTool
  | (Omit<PackTool, 'packId'> & {packId: number});
```

types.ts
```diff
@@ -1741,9 +1779,55 @@ export interface AgentDefinition {
...
+ *
+ * @internal
+ * @hidden
+ */
+export interface AgentToolsDef {
+  /**
+   * Read and write Superhuman Docs documents and tables.
    */
-  tools?: Tool[];
+  docs?: boolean | {readApprovalMode?: ToolConsentMode; writeApprovalMode?: ToolConsentMode};
```
vivekvenkatram: Looks good that docs/mail/webSearch take the same optional consent the in-app builder stores, and formulas got `enabled` / `approvalMode` too. That's one dialect instead of a pack-only subset that would fail a builder blob on `packs validate`.

```ts
docs?: boolean | {readApprovalMode?: ToolConsentMode; writeApprovalMode?: ToolConsentMode};
formulas?: Array<{formulaName: string; enabled?: boolean; approvalMode?: ToolConsentMode}>;
```

types.ts
```diff
@@ -1741,9 +1779,55 @@ export interface AgentDefinition {
...
+   */
+  mail?: boolean | {readApprovalMode?: ToolConsentMode; writeApprovalMode?: ToolConsentMode};
+  /**
+   * Search the public internet, optionally restricted to `allowedDomains`.
+   */
+  webSearch?: boolean | {allowedDomains?: string[]; approvalMode?: ToolConsentMode};
+  /**
+   * Connector packs this agent can call, one entry per pack.
+   */
+  connectors?: Array<{
```
vivekvenkatram, line 1788: This is the leftover: `connectors` is still Pack-formulas only. Could we reserve `mcp` / `knowledge` on the same `packId` even if this slice doesn't emit them, so the next PR doesn't have to widen every `setTools` call?

```ts
connectors?: Array<{
  packId: number;
  formulas?: Array<{formulaName: string; enabled?: boolean; approvalMode?: ToolConsentMode}>;
  mcp?: {serverNames?: string[]; tools?: Array<{toolName: string; enabled?: boolean; approvalMode?: ToolConsentMode}>};
  knowledge?: boolean | {organizational?: boolean};
}>;
```

builder.ts
```diff
@@ -665,4 +668,50 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
...
+   * pack.setTools({connectors: [{packId: 1234, formulas: [{formulaName: 'CreateTask'}]}]});
+   * ```
+   */
+  setTools({docs, mail, webSearch, connectors}: AgentToolsDef): this {
+    const tools: AgentTool[] = [];
+    if (webSearch) {
+      const options = typeof webSearch === 'object' ? webSearch : {};
+      tools.push({
+        type: ToolType.WebSearch,
+        ...(options.allowedDomains ? {allowedDomains: options.allowedDomains} : {}),
```
vivekvenkatram: Looks good that an empty `allowedDomains` stays on the tool so `packs validate` can fail, instead of rewriting it to unrestricted search. The new test name matches.

```ts
...(options.allowedDomains ? {allowedDomains: options.allowedDomains} : {}),
```

types.ts
```diff
@@ -1234,6 +1234,18 @@ export interface RateLimits {
  */
 export type BasicPackDefinition = Omit<PackVersionDefinition, 'version'>;
 
+/**
+ * Whether a tool asks the user before it acts.
+ *
+ * @hidden In development
+ */
+export enum ToolConsentMode {
```
vivekvenkatram: Fwiw this is a Go runtime policy enum (we already have it on the server overlay), not something `pack.ts` should import. If upload validate has to tolerate builder JSON, keep the keys `@hidden` on the stored objects and off `AgentToolsDef` / `setTools`.

```ts
/** @hidden */
readApprovalMode?: ToolConsentMode;
```

types.ts
```diff
@@ -1312,6 +1324,16 @@ export interface PackTool extends BaseTool<ToolType.Pack> {
...
+    /**
+     * Whether the formula is offered. Defaults to true.
+     * @hidden In development
+     */
+    enabled?: boolean;
+    /**
+     * Whether this formula asks before it runs.
+     * @hidden In development
+     */
+    approvalMode?: ToolConsentMode;
```
vivekvenkatram: `enabled` is a grant (this formula is off). `approvalMode` is not, and the resolver already defaults writes/actions to ask when it's missing. Could we keep `enabled` and drop `approvalMode` from the public PackTool shape?

```ts
formulas?: Array<{formulaName: string; enabled?: boolean}>;
```

types.ts
```diff
@@ -1450,13 +1472,23 @@ export interface ContactResolutionTool extends BaseTool<ToolType.ContactResoluti
  * Tool that provides access to Superhuman Docs documents and tables capabilities.
  * @internal
  */
-export interface CodaDocsAndTablesTool extends BaseTool<ToolType.CodaDocsAndTables> {}
+export interface CodaDocsAndTablesTool extends BaseTool<ToolType.CodaDocsAndTables> {
+  /** Whether reading asks first. */
+  readApprovalMode?: ToolConsentMode;
```
vivekvenkatram: The SDK grant is presence of `{type: CodaDocsAndTables}` (and the same for Mail). Read/write consent is instance policy, and omitting it already gets reads-auto / writes-ask in the resolver, so a published pack shouldn't ship `writeApprovalMode: auto` to every installer.

```ts
export interface CodaDocsAndTablesTool extends BaseTool<ToolType.CodaDocsAndTables> {}
```

types.ts
```diff
@@ -1482,6 +1514,12 @@ export interface WebSearchTool extends BaseTool<ToolType.WebSearch> {
    * ```
    */
   allowedDomains?: string[];
+  /**
+   * Whether an unknown URL asks first.
+   *
+   * @hidden In development
+   */
+  approvalMode?: ToolConsentMode;
```
vivekvenkatram: `allowedDomains` belongs in the SDK, it's which hosts the tool may search. `approvalMode` is the unknown-URL fetch gate, which Go already defaults to always-ask, so I'd leave it off the public tool type.

```ts
export interface WebSearchTool extends BaseTool<ToolType.WebSearch> {
  allowedDomains?: string[];
}
```

builder.ts
```diff
@@ -665,4 +668,50 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
...
+        ...(options.approvalMode ? {approvalMode: options.approvalMode} : {}),
```
vivekvenkatram: If modes stay out of `AgentToolsDef`, this spread shouldn't copy them either. `setTools` only needs grant fields: `allowedDomains`, `packId`, formula names / `enabled`.

```ts
tools.push({
  type: ToolType.WebSearch,
  ...(options.allowedDomains ? {allowedDomains: options.allowedDomains} : {}),
});
```

types.ts
vivekvenkatram: Taking a second look: approval modes shouldn't be in the author API. Presence is the grant; ask-vs-auto is Go policy (runtime defaults, chat Allow, pack-policy ceiling), and a published pack that ships `writeApprovalMode: auto` would bake send-without-asking into every install.

```ts
docs?: boolean;
mail?: boolean;
webSearch?: boolean | {allowedDomains?: string[]};
connectors?: Array<{packId: number; formulas?: Array<{formulaName: string; enabled?: boolean}>}>;
```

types.ts
peizhaomei-codaio: Do we want to promote `MCPTool.tools` and `PackKnowledgeToolSource.organizational` from the server overlay into the public SDK?

types.ts
peizhaomei-codaio: Deleted

types.ts
```diff
@@ -1741,9 +1746,55 @@ export interface AgentDefinition {
...
+   */
+  connectors?: Array<{
+    /**
+     * The id of the connector pack.
+     */
+    packId: number;
+    /**
+     * The formulas to offer, if not all of them.
+     */
+    formulas?: Array<{formulaName: string; enabled?: boolean}>;
```
vivekvenkatram: Listing the formula name is the grant, same as `docs: true` vs omitting it. `enabled: false` is a builder-toggle leftover; an SDK author who doesn't want CreateTask just leaves it out. Keep `enabled?` `@hidden` on stored `PackTool` so builder blobs still validate; drop it from `AgentToolsDef` (and the setTools test that passes it through):

```ts
    formulas?: Array<{formulaName: string}>;
```

types.ts
vivekvenkatram: No, don't promote `organizational` or per-tool MCP lists into the public SDK. Presence of the connector `packId` is the grant; those fields are Go overlays, same split as approval mode. We can add `mcp` / `knowledge` keys on this slot later, when `setTools` actually emits them.

types.ts
vivekvenkatram: Taking a second look: drop `enabled` from `AgentToolsDef` / `setTools` too. Presence of the formula name is the grant; keep `enabled?` `@hidden` on stored `PackTool` only so a builder blob still validates.

vivekvenkatram (approved):
Pls address comments before merging

PR #3632, opened 2026-09-02
Add default While Writing trigger authoring to agent builder

peizhaomei-codaio:
## Description

Pack authors can now declare a default contextual ("while writing") trigger for their agent via `pack.setDefaultWhileWritingTrigger()`. It's stored on the pack version as `defaultTriggers.contextualTrigger`, sibling to agent. The server should copy it onto new agent instances at install time, and users can edit it afterward.

The type mirrors the stored contract from front-end-web-monorepo's StoredContextualTrigger: condition, enabled, assistMode, surfaces, blockedDomains, contextRules (signal register/task, domain, domain category), and evaluationConfig (REGEX/SEMANTIC_CENTROID/LLM). evaluationConfig is authored as a discriminated union on strategy.

``` ts
interface WhileWritingTriggerDefinition {
  condition: string;
  enabled: boolean;
  assistMode?: ContextualTriggerAssistMode;
  suggestionColor?: ContextualTriggerSuggestionColor;
  decorationStyle?: ContextualTriggerDecorationStyle;
  surfaces?: ContextualTriggerSurface[];
  blockedDomains?: string[];
}
```
evaluationConfig, contextRules and semanticCentroid are intentionally not authorable. Those are runtime shapes the server's trigger compiler produces from the natural-language condition, not values a human picks — and the signal-task / domain-category taxonomies behind them would drift from StoredContextualTrigger the first time docs adds an entry. The schema is strict, so a manifest that tries to declare them is rejected.

Example
```ts
agent.setDefaultWhileWritingTrigger({
  condition: 'Offer a citation when the user asserts a statistic',
  enabled: true,
});
```

Compiles to:
```json
{
  "defaultTriggers": [
    {"kind": "whileWriting", "condition": "Offer a citation when the user asserts a statistic", "enabled": true}
  ]
}
```
With the optional fields:
```ts
agent.setDefaultWhileWritingTrigger({
  condition: 'Offer a citation when the user asserts a statistic',
  enabled: true,
  assistMode: ContextualTriggerAssistMode.OnDemand,
  surfaces: [ContextualTriggerSurface.Docs, ContextualTriggerSurface.Email],
  blockedDomains: ['internal.example.com'],
});
```

```json
{
  "defaultTriggers": [
    {
      "kind": "whileWriting",
      "condition": "Offer a citation when the user asserts a statistic",
      "enabled": true,
      "assistMode": "on_demand",
      "surfaces": ["docs", "email"],
      "blockedDomains": ["internal.example.com"]
    }
  ]
}
```


types.ts
```diff
@@ -1908,6 +2154,13 @@ export interface PackVersionDefinition {
    * @hidden
    */
   agent?: AgentDefinition;
+  /**
+   * The triggers this pack's agent ships with, if it defines any. Authored via `sdk.newAgent()`.
+   *
+   * @internal
+   * @hidden
+   */
+  defaultTriggers?: DefaultTriggersDefinition;
```
xicheng-codaio: should this be an array?

types.ts
```diff
@@ -1797,6 +1797,252 @@ export interface AgentToolsDef {
...
+ * The triggers an agent ships with. Each one is a default the user can edit after installing.
+ *
+ * @internal
+ * @hidden
+ */
+export interface DefaultTriggersDefinition {
+  /**
+   * The while-writing trigger the agent runs on.
+   */
+  contextualTrigger?: WhileWritingTriggerDefinition;
```
xicheng-codaio: i expect this would be a type union in the future?

types.ts
```diff
@@ -1797,6 +1797,267 @@ export interface AgentToolsDef {
...
+  /** Whether the trigger starts on for new agent instances. */
+  enabled: boolean;
+  assistMode?: ContextualTriggerAssistMode;
+  suggestionColor?: ContextualTriggerSuggestionColor;
+  decorationStyle?: ContextualTriggerDecorationStyle;
+  surfaces?: ContextualTriggerSurface[];
+  /** Domains this trigger will not activate on. Max 50. */
+  blockedDomains?: string[];
+  contextRules?: ContextualTriggerContextRules;
+  evaluationConfig?: ContextualTriggerEvaluationConfig;
```
vivekvenkatram: Fwiw I'd keep the authoring type to what a pack maker actually writes (`condition`, `enabled`, `surfaces`, `blockedDomains`, `assistMode`) and drop `evaluationConfig` / `contextRules` / `SemanticCentroid` from this SDK. Those are runtime/DB shapes, and a centroid in pack.ts will also ride to the public pack JSON unless the server strips `defaultTriggers` the way it already strips `agent.instructions`.

```ts
export interface WhileWritingTriggerDefinition {
  condition: string;
  enabled: boolean;
  assistMode?: ContextualTriggerAssistMode;
  suggestionColor?: ContextualTriggerSuggestionColor;
  decorationStyle?: ContextualTriggerDecorationStyle;
  surfaces?: ContextualTriggerSurface[];
  blockedDomains?: string[];
}
```

index.ts
```diff
@@ -316,6 +316,21 @@ export type {WebSearchTool} from './types';
...
+export {ContextualTriggerSuggestionColor} from './types';
+export {ContextualTriggerDecorationStyle} from './types';
+export {ContextualTriggerSurface} from './types';
+export {ContextualTriggerGranularity} from './types';
+export {ContextualTriggerStrategy} from './types';
+export {ContextualTriggerSignalRegister} from './types';
+export {ContextualTriggerSignalTask} from './types';
+export {ContextualTriggerDomainCategory} from './types';
+export type {ContextualTriggerContextRules} from './types';
+export type {ContextualTriggerEvaluationConfig} from './types';
```
vivekvenkatram: Do we need to re-export the signal-task / domain-category / strategy enums from the public SDK? They'll drift from `StoredContextualTrigger` the first time someone adds a task, and pack authors shouldn't be picking calibration taxonomies. Wdyt about exporting only `WhileWritingTriggerDefinition`, `DefaultTriggerKind`, and the small authoring enums (`AssistMode`, `Surface`, colors)?

compiled_types.ts
```diff
@@ -169,6 +169,7 @@ type BasePackVersionMetadata = Omit<
   | 'syncTables'
   | 'skills'
   | 'agent'
+  | 'defaultTriggers'
```
vivekvenkatram, line 172: Nice that the browser type drops `defaultTriggers`. FEWM `getExternalDefinition` still rest-spreads leftover keys though, so when we pin this we'll need to pull `defaultTriggers` out of `rest` the same way as `agent.instructions`, or the condition/regex/centroid lands on the public pack JSON.

types.ts
```diff
@@ -1908,6 +2169,13 @@ export interface PackVersionDefinition {
    * @hidden
    */
   agent?: AgentDefinition;
+  /**
+   * The triggers this pack's agent ships with, if it defines any. Authored via `sdk.newAgent()`.
+   *
+   * @internal
+   * @hidden
+   */
+  defaultTriggers?: DefaultTriggersDefinition;
```
vivekvenkatram: Like that this sits next to `agent` rather than inside it. Instance-editable defaults shouldn't go through the CAD converter that only reads `metadata.agent`.

types.ts
vivekvenkatram: Yeah, array-of-kind is the right shape now that schedule/event will join. Uniqueness per kind also matches what we store (one `defaultContextualTriggers` row).

types.ts
vivekvenkatram: Discriminated union on `kind` landed, same pattern as `agent.tools`. I would still keep `WhileWritingTriggerDefinition` itself small so the union does not grow a copy of the evaluator config.

types.ts
```diff
@@ -1797,6 +1797,267 @@ export interface AgentToolsDef {
...
+ */
+export type DefaultTriggerDefinition = {kind: DefaultTriggerKind.WhileWriting} & WhileWritingTriggerDefinition;
+
+/**
+ * The triggers an agent ships with. Each one is a default the user can edit after installing.
+ *
+ * @internal
+ * @hidden
+ */
+export type DefaultTriggersDefinition = DefaultTriggerDefinition[];
```
xicheng-codaio: do we need this alias 🤔 
most other places just use array directly, e.g.
`mcpServers?: MCPServer[];`

types.ts
xicheng-codaio: agree that we need to audit and sanitize some of the existing fields.

types.ts
```diff
@@ -1797,6 +1797,98 @@ export interface AgentToolsDef {
...
+  Purple = 'purple',
+  Red = 'red',
+  Yellow = 'yellow',
+}
+
+/**
+ * @internal
+ * @hidden
+ */
+export enum ContextualTriggerDecorationStyle {
```
xicheng-codaio, line 1836: non-blocking, can we also get the comments from our server files? e.g. 
`// How a 1p agent's contextual suggestion renders. Absent resolves to Auto (let the client decide).`
this would make the consumptions easier

builder.ts
```diff
@@ -655,6 +660,10 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
    * See {@link PackVersionDefinition.agent}.
    */
   agent: Partial<AgentDefinition> = {tools: []};
+  /**
+   * See {@link PackVersionDefinition.defaultTriggers}.
+   */
+  defaultTriggers?: DefaultTriggerDefinition[];
```
xicheng-codaio, line 666: Should this be nested under AgentDefinition? the default triggers are part of agents i think?

types.ts
```diff
@@ -1797,6 +1797,98 @@ export interface AgentToolsDef {
...
+}
+
+/**
+ * A single default trigger a pack's agent ships with, tagged with its kind so more kinds can
+ * join this union later.
+ *
+ * @internal
+ * @hidden
+ */
+export type DefaultTriggerDefinition = {kind: DefaultTriggerKind.WhileWriting} & WhileWritingTriggerDefinition;
```
xicheng-codaio: a bit confused here: can we just use a type union over types`WhileWritingTriggerDefinition` ... , curious if we need the `enum DefaultTriggerKind ` at all?

builder.ts
peizhaomei-codaio: I would prefer this way so that instance-editable defaults don't go through the CAD converter. @vivekvenkatram

types.ts
peizhaomei-codaio: i'm fine with either way, just following precedent in the project. enum seems more organized to me, and kind is easier for the server to read?

peizhaomei-codaio:
@vivekvenkatram do we want this field `enabled: boolean;`? feels like pack author shouldn't be able to set this

builder.ts
vivekvenkatram: Agree with Peizhao here. Default triggers are instance-editable (status, rrule, condition), and the CAD converter only reads `metadata.agent`. Nesting them under `AgentDefinition` would either ship them through CAD/CDN or force a special-case strip, same as `agent.instructions`. Sibling of `agent` on `PackVersionDefinition` is the right home.

types.ts
vivekvenkatram: I'd keep the enum. Same pattern as `ToolType` plus `z.literal(DefaultTriggerKind.Schedule)`, and the uniqueness refine keys off `trigger.kind`. A string-literal union works too, just a style call. Not a blocker.

vivekvenkatram:
Drop it. On/off is instance state, same as schedule `status` (install always copies paused, so an author-set status would be ignored). Copy the trigger with storage's default (absent `enabled` = on) and let the owner toggle it. If we later want packs to ship an opt-in trigger that starts off, we can add the field then.

vivekvenkatram (approved):

PR #3633, opened 2026-09-03
Add default schedule trigger authoring to the agent builder

peizhaomei-codaio:
## Description

Lets an agent ship a default schedule, alongside the while-writing trigger the base branch added:

```ts
pack.setDefaultScheduleTrigger({
  rruleString: 'DTSTART;TZID=America/New_York:20260101T090000\nRRULE:FREQ=WEEKLY;INTERVAL=2;BYDAY=MO,WE',
});
```

Authors declare an RFC 5545 recurrence rule rather than the builder form's `interval`/`unit`/`hour` picker, so a pack can express any legal rule the form omits (`BYMINUTE` other than 0/30, `BYSETPOS`, `COUNT`).

`DTSTART` is a separate RFC 5545 property, and in the long run the start date and timezone belong to the install rather than to the author. Until the server anchors one, a pack has to carry its own: `parseScheduleTriggerFromRRule` reads `origOptions.dtstart` and throws when it is absent, and `get_custom_agent_for_user` calls it without a catch, so a rule with no `DTSTART` 500s the agent detail page. The field doc says to include one for that reason, and the example above does.

Once install fills in a missing `DTSTART` and `TZID`, the bare `RRULE:FREQ=WEEKLY;...` form becomes the natural thing for an author to write and the doc can relax. A `DTSTART;TZID=...` stays meaningful even then, for a schedule genuinely pinned to one zone (US market hours, wherever the adopter sits).

The validator accepts every legal `RRULE` part rather than narrowing to what the builder form can
draw. That leaves a read-back gap on the server: `ScheduleTriggerInput` has no field for `BYMONTH`,
`BYSETPOS`, `COUNT` or `WKST`, and `BYDAY=-1FR`, `BYHOUR=9,17`, `BYMONTHDAY=1,15`, `BYMINUTE=17` and
`FREQ=YEARLY` all lose something through `parseScheduleTriggerFromRRule`. None of it affects firing —
install stores the string verbatim and the sweeper evaluates it raw — so the fix is for
`get_custom_agent_for_user` to skip that parse for pack defaults, or return the raw string, rather
than for this validator to turn the parts away.

## Where to look

`testing/rrule_validation.ts` is the only real logic; everything else is wiring.

- **Line splitting** rejects anything that would parse as an `RRuleSet` (a second `RRULE`, `EXDATE`). The server's `computeNextRunAt` returns `null` for a set, and callers read that as "retire this trigger" — so a set stored here would silently never fire.
- **`RulePartNames`** is a whitelist; unknown parts (`BYSECOND`) are rejected.
- **Frequency** crosses `BYHOUR` with `BYMINUTE` into minutes-past-midnight and requires every adjacent gap to be >= 60.

The server's `validateMinimumFrequency` can't be reused: it takes the structured `ScheduleTriggerInput` and only checks `freq > HOURLY`, which is sound because `buildRRuleString` is the sole writer there. Pack install is a different writer, so `FREQ=DAILY;BYHOUR=0..23;BYMINUTE=0,15,30,45` would pass that check while running 96 times a day. Hand-rolled rather than adding the `rrule` dependency, since `upload_validation` reaches a browser bundle via Coda web's CJS import.

## Notes

- No `status` field. On server, every path that copies a pack's default onto an instance forces `status: 'paused'` (`create_agent_instance_from_custom_definition_v6`), so an author-declared status would have no effect


testing/rrule_validation.ts
```diff
@@ -0,0 +1,141 @@
...
+    return minutes.length > 1 ? TooFrequent : undefined;
+  }
+
+  // A day's occurrences are BYHOUR crossed with BYMINUTE, as minutes past midnight.
+  const runTimes = [...new Set(hours.flatMap(hour => minutes.map(minute => hour * MinutesPerHour + minute)))].sort(
+    (a, b) => a - b,
+  );
+  const gaps = runTimes.slice(1).map((runTime, index) => runTime - runTimes[index]);
+
+  return gaps.some(gap => gap < MinutesPerHour) ? TooFrequent : undefined;
```
vivekvenkatram, line 135: This misses the wrap from last occurrence back to first+1440. `FREQ=DAILY;BYHOUR=23,0;BYMINUTE=30,0` (23:30 and 00:00) is a 30-minute gap overnight, but `gaps` only sees ~23.5h so this returns valid. Wdyt about `const wrapGap = runTimes[0] + 24 * 60 - runTimes[runTimes.length - 1]` and failing when `runTimes.length >= 2 && wrapGap < MinutesPerHour`? Full-day `BYHOUR=0..23` still wraps at 60 and stays legal.

testing/rrule_validation.ts
```diff
@@ -0,0 +1,141 @@
+// Validation for the RFC 5545 recurrence a pack declares in `defaultTriggers.scheduleTrigger`.
+
+const Frequencies = ['HOURLY', 'DAILY', 'WEEKLY', 'MONTHLY', 'YEARLY'];
```
vivekvenkatram: YEARLY will upload, but `parseScheduleTriggerFromRRule` maps unknown freq to `CadenceUnit.Day` (no Year on the builder). A pack that ships `FREQ=YEARLY` then gets shown/edited as daily on the agent detail page. If install copies the string through that parser, I'd drop YEARLY here until the server round-trips it, or skip parse and persist `rruleString` as-is.

testing/rrule_validation.ts
```diff
@@ -0,0 +1,141 @@
...
+  'BYMONTHDAY',
+  'BYSETPOS',
+];
+
+// The ranges each numeric rule part takes, as [min, max].
+const NumericRulePartRanges: Array<[string, number, number]> = [
+  ['BYHOUR', 0, 23],
+  ['BYMINUTE', 0, 59],
+  ['BYMONTH', 1, 12],
+  ['BYMONTHDAY', -31, 31],
```
vivekvenkatram: RFC 5545 forbids 0 on BYMONTHDAY and BYSETPOS (`[-31,-1] ∪ [1,31]`, same idea for BYSETPOS). `parseNumberList` currently accepts 0 because the range is a single closed interval. Easy reject: after parse, fail if the key is one of those two and the list includes 0.

testing/rrule_validation.ts
```diff
@@ -0,0 +1,141 @@
+// Validation for the RFC 5545 recurrence a pack declares in `defaultTriggers.scheduleTrigger`.
```
vivekvenkatram: Stale path: this is now an array entry with `kind: schedule`, not `defaultTriggers.scheduleTrigger`.

types.ts
```diff
@@ -1875,13 +1875,28 @@ export interface WhileWritingTriggerDefinition {
...
- * enum as they become authorable.
+ * A default schedule trigger a pack's agent ships with. Adopters receive it paused.
+ *
+ * @internal
+ * @hidden
+ */
+export interface ScheduleTriggerDefinition {
+  /**
+   * The recurrence, as an RFC 5545 `RRULE`. An optional `DTSTART` line pins the start, and its
+   * `TZID` pins the timezone; install supplies whichever of the two the rule leaves out.
```
vivekvenkatram: Worth calling out that a rule with no `DTSTART` is legal here but `parseScheduleTriggerFromRRule` throws when `origOptions.dtstart` is missing, and `get_custom_agent_for_user` has no catch. Install has to write a DTSTART (or a TZID) before that page is safe, otherwise a pack like `FREQ=WEEKLY;BYDAY=MO;BYHOUR=9` 500s the detail view.

testing/rrule_validation.ts
peizhaomei-codaio: `ScheduleTriggerInput` also has no field for BYMONTH, BYSETPOS, COUNT or WKST, and BYDAY=-1FR, BYHOUR=9,17, BYMONTHDAY=1,15, BYMINUTE=17 all get truncated on read.
I think install stores the string verbatim and the sweeper runs it raw. Only the read API coerces, so the detail page misrepresents the schedule.
Do we want to drop all of these until the server is good?

testing/rrule_validation.ts
peizhaomei-codaio: `BY*` is a cross product, so `BYHOUR=23,0;BYMINUTE=30,0` expands to `23:00/23:30/00:00/00:30` and the `00:00–00:30` pair trips the within-day check; with a single `BYMINUTE` the wrap is always ≥ 60, so the extra check would be unreachable.

vivekvenkatram: Yep, I undershot that. Two `BYMINUTE`s in the same hour are always < 60 apart, so the within-day check already catches the cross product, and a single `BYMINUTE` wraps at exactly 60 when 0 and 23 are both present. No extra wrap check.

testing/rrule_validation.ts
vivekvenkatram: Don't shrink the SDK to `ScheduleTriggerInput`. Sweeper already runs the string raw, so COUNT/BYSETPOS/BYMONTH are fine if install copies `rruleString` verbatim and writes a `DTSTART`/`TZID` when the pack omitted them.

`parseScheduleTriggerFromRRule` is builder-only (throws with no `dtstart`, truncates `BYMINUTE=17` / `BYHOUR=9,17`). `get_custom_agent_for_user` should skip that parse for pack defaults (or return the raw string) rather than us dropping legal RRULE parts here.

vivekvenkatram (approved):

PR #3637, opened 2026-09-04
Add default event trigger authoring to the agent builder

peizhaomei-codaio:
## Description

Lets an agent ship default event triggers, alongside the schedule trigger the base branch added:

```ts
pack.addDefaultMailEventTrigger({
  mailEventType: MailEventType.MessageReceived,
  filters: {
    conditions: [{field: MailFilterField.From, operator: FilterOperator.TextContains, value: '@customers.example.com'}],
  },
});
```

One method per product — mail, Slack, Notetaker — following the per-kind setters already on the builder. Each fills in `kind` and `type`, so an author writes only the event and its filters. They append rather than replace, so the per-kind uniqueness check in `defaultTriggersSchema` skips the event kind, and the array is capped at the server's `MAX_TRIGGERS_PER_AGENT` of 20.

| Event | Author sets | Adopter binds |
| --- | --- | --- |
| mail `message_sent` / `message_received` | `filters` on `from`/`to` (≤320 chars, exact or substring) and `subject`/`body` (≤998, substring) | seat, `enabled` |
| slack `message_keyword` | `keywords`, `monitorThreadFollowUps` | `teamId`, `channelIds`, `audience`, `enabled` |
| notetaker `meeting.summary.completed` | `filters` on participant, project tag, meeting type, participant count, duration, external attendees, recurring; `keywords` | `enabled` |

An author declares what the agent reacts to; the adopter supplies where it listens, who may use it, and whether it runs. That rule keeps out `audience`, `recurringEventId`, the docs events, Slack's `agent_mentioned`, and mail's `label_added` — each either names something per-account or, once the per-account parts are gone, carries nothing but the event.

Field names follow the server's (`mailEventType`, and `eventType` for both Slack and Notetaker), since install copies them onto the instance.

## Where to look

`types.ts` is the shape; `testing/upload_validation.ts` is the same tree in zod.

- **Operators are per field family, bounds are per field.** `subject`/`body` take `textContains` only, `meetingType` `textEquals` only, `durationMinutes` the numeric three — and a participant holds 320 characters where a project tag or meeting type holds 128. The operator half is in the types, so `{field: 'subject', operator: 'textEquals'}` fails to compile.
- **`DistributiveOmit`** — `Omit` over a union collapses to the shared keys, reducing the shared `_addDefaultEventTrigger` parameter to `{type}`.

## Notes

- Nothing consumes `defaultTriggers` yet, here or on main: no reference exists in front-end-web-monorepo or coda. The install half of GOECOS-165/166/168 is open.
- Slack is the one trigger install cannot complete from the pack alone, since `teamId` and `channelIds` are required on the stored request. Nothing expresses that partial today.
- No `enabled`. Author-declared triggers should arrive off, the way a schedule trigger is forced to `status: 'paused'` — the server's to implement, and what makes an unbound Slack trigger safe rather than broken.
- The ticket names the field `defaultTriggers.eventTriggers`; the landed shape is the flat kind-tagged array from #3632.


builder.ts
```diff
@@ -730,6 +732,32 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
...
+   *         field: sdk.MailFilterField.From,
+   *         operator: sdk.FilterOperator.TextContains,
+   *         value: '@customers.example.com',
+   *       },
+   *     ],
+   *   },
+   * });
+   * ```
+   */
+  addDefaultEventTrigger(eventTrigger: DistributiveOmit<EventTriggerDefinition, 'kind'>): this {
```
vivekvenkatram: wdyt about splitting this by product instead of one `addDefaultEventTrigger` union? Mail/Slack/Docs/Notetaker are separate stored request schemas, and a single method plus four nested discriminants is a lot to review and a lot for pack.ts to get wrong.

```ts
addDefaultMailEventTrigger(trigger: DistributiveOmit<MailEventTriggerDefinition, 'kind' | 'type'>): this
addDefaultSlackEventTrigger(...): this
```

Each still pushes `{kind: Event, type: Mail, ...}`. Fine to land mail in this PR and Slack/Docs/Notetaker as follow-ups if you want smaller diffs.

testing/upload_validation.ts
```diff
@@ -2451,13 +2484,163 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+  const docsEventTriggerSchema = zodCompleteStrictObject<DocsEventTriggerDefinition>({
+    kind: eventTriggerKindSchema,
+    type: z.literal(EventTriggerType.Docs),
+    docEventType: z.nativeEnum(DocsEventType),
+  });
+
+  const notetakerAddressFilterConditionSchema = zodCompleteStrictObject<NotetakerAddressFilterCondition>({
+    field: z.enum([NotetakerFilterField.Participant, NotetakerFilterField.ProjectTag]),
+    operator: addressFilterOperatorSchema,
+    value: z.string().min(1).max(Limits.NotetakerParticipantValue),
```
vivekvenkatram: `Limits.NotetakerTagValue` (128) is unused, and `projectTag` / `meetingType` are sharing the participant (320) and recurring-event (1024) caps. Shared package is 128 for those two. wdyt about splitting the address/id schemas so tag fields use `NotetakerTagValue`?

```ts
field: z.literal(NotetakerFilterField.ProjectTag),
value: z.string().min(1).max(Limits.NotetakerTagValue),
```

types.ts
```diff
@@ -1909,13 +1910,461 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+ */
+export enum NotetakerFilterField {
+  DurationMinutes = 'durationMinutes',
+  HasExternalAttendees = 'hasExternalAttendees',
+  IsRecurring = 'isRecurring',
+  MeetingType = 'meetingType',
+  Participant = 'participant',
+  ParticipantCount = 'participantCount',
+  ProjectTag = 'projectTag',
+  RecurringEventId = 'recurringEventId',
```
vivekvenkatram: `recurringEventId` is a calendar instance id, same bucket as `channelIds` / `docId`. A pack that ships one will not match the adopter's calendar. wdyt about dropping this field from the pack union (keep participant / duration / tags)?

types.ts
```diff
@@ -1909,13 +1910,461 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+}
+
+/**
+ * A default trigger on a label being added. The labels it watches are bound at install, so it
+ * fires on every label until an adopter narrows it.
+ *
+ * @internal
+ * @hidden
+ */
+export type MailLabelAddedTriggerDefinition = BaseMailEventTrigger<MailEventType.LabelAdded>;
```
vivekvenkatram: Same as your note: with no portable label filter this fires on every label until install binds IDs. wdyt about omitting `label_added` here, same as docs `agent_mentioned`, until labels can be named without provider IDs?

testing/upload_validation.ts
```diff
@@ -2451,13 +2478,137 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+    type: z.literal(EventTriggerType.Mail),
+    mailEventType: z.nativeEnum(MailEventType),
+    filters: mailEventFiltersSchema.optional(),
+  });
+
+  const slackEventTriggerSchema = zodCompleteStrictObject<SlackEventTriggerDefinition>({
+    kind: eventTriggerKindSchema,
+    type: z.literal(EventTriggerType.Slack),
+    eventType: z.nativeEnum(SlackEventType),
+    keywords: z.array(keywordSchema).max(Limits.MaxKeywords).optional(),
```
vivekvenkatram, line 2529: Empty `keywords` matches every Slack message once the adopter binds a channel (same class as the triggers we dropped). wdyt about requiring at least one here and on the type:

```ts
keywords: z.array(keywordSchema).min(1).max(Limits.MaxKeywords),
```

```ts
keywords: [string, ...string[]];
```

types.ts
```diff
@@ -1909,13 +1910,368 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+ * at install.
+ *
+ * @internal
+ * @hidden
+ */
+export interface SlackEventTriggerDefinition extends BaseEventTrigger<EventTriggerType.Slack> {
+  /** The Slack event that fires the trigger. */
+  eventType: SlackEventType;
+  /** Keywords to match, case insensitively. Up to 50, each up to 200 characters. */
+  keywords?: string[];
```
vivekvenkatram, line 2118: Runtime treats `keywords: []` as every message in the bound channels. wdyt about making this required (`[string, ...string[]]`) so a pack default always ships a portable filter?

test/upload_validation_test.ts
```diff
@@ -8483,6 +8491,408 @@ describe('Pack metadata Validation', async () => {
...
+      };
+      const metadata = createFakeAgentMetadata({
+        agent: {instructions: 'Do a thing.', tools: []},
+        defaultTriggers: [slackTrigger],
+      });
+      const result = await validateJson(metadata);
+      assert.deepEqual(result.defaultTriggers, [slackTrigger]);
+    });
+
+    it('rejects a slack channel, which is bound at install', async () => {
```
vivekvenkatram, line 8630: This is the right pattern for install-only fields. wdyt about the same for `enabled`, `teamId`, and a mail seat (`superhumanUserMailId`) so a future passthrough cannot turn a pack default on before the adopter auths?

vivekvenkatram (approved):
lgtm, address comments before merging

types.ts
```diff
@@ -1909,13 +1910,368 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+}
+
+/**
+ * A default Slack event trigger a pack's agent ships with. The workspace and channels are bound
+ * at install.
+ *
+ * @internal
+ * @hidden
+ */
+export interface SlackEventTriggerDefinition extends BaseEventTrigger<EventTriggerType.Slack> {
```
xicheng-codaio, line 2114: not blocking: do we imagine that we need to split the trigger interfaces for "partial" vs "complete" triggers? e.g. for Slack Event trigger, there are two valid forms: 
1. only has the type and maybe sub-type such as Agent mentioned/Message Received
2. has the type, the subtype, the workspace, the channel ....

The SDK only allow developers to define type 1 (partial triggers), since type 2 is only required during the app-like install (after connecting to Slack etc)

types.ts
```diff
@@ -1909,13 +1910,368 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+ * @internal
+ * @hidden
+ */
+export type NumericFilterOperator =
+  | FilterOperator.NumberAtLeast
+  | FilterOperator.NumberAtMost
+  | FilterOperator.NumberEquals;
+
+/**
+ * How a filter's conditions combine. Absent means `And`.
```
xicheng-codaio, line 1992: maybe `FilterCombinator` should not be optional?  can we make it required instead?

types.ts
```diff
@@ -1909,13 +1910,368 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+  AddressFilterOperator
+>;
+
+/**
+ * A condition on an identifier field.
+ *
+ * @internal
+ * @hidden
+ */
+export type NotetakerIdFilterCondition = BaseNotetakerFilterCondition<
```
xicheng-codaio, line 2182: who defines the list of possible meeting types 🤔 , can we add some unit tests on how this would be used?

types.ts
```diff
@@ -1909,13 +1910,368 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+}
+
+/**
+ * A condition on a list-valued field.
+ *
+ * @internal
+ * @hidden
+ */
+export type NotetakerAddressFilterCondition = BaseNotetakerFilterCondition<
+  NotetakerFilterField.Participant | NotetakerFilterField.ProjectTag,
```
xicheng-codaio, line 2172: How can the developer specify the possible project tag values? can we add some unit tests?

types.ts
peizhaomei-codaio: A combinator has no meaning with zero or one condition, not sure if we want to force authors to state it then

types.ts
```diff
@@ -1916,13 +1917,368 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+/**
+ * A default mail event trigger a pack's agent ships with. The mailbox is bound at install.
+ *
+ * @internal
+ * @hidden
+ */
+export interface MailEventTriggerDefinition extends BaseEventTrigger<EventTriggerType.Mail> {
+  /** The mail event that fires the trigger. */
+  mailEventType: MailEventType;
+  /** Which messages fire the trigger. Absent fires on every one. */
```
vivekvenkatram: This is the same firehose we just banned on Slack (`keywords: []`) and used to drop docs / `label_added`: the pack ships only an event name, and install would listen to every message until the adopter narrows it. wdyt about requiring `filters` the way Slack requires `keywords`?

```ts
filters: MailEventFilters;
```

types.ts
```diff
@@ -1916,13 +1917,368 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+/**
+ * A default notetaker event trigger a pack's agent ships with.
+ *
+ * @internal
+ * @hidden
+ */
+export interface NotetakerEventTriggerDefinition extends BaseEventTrigger<EventTriggerType.Notetaker> {
+  /** The notetaker event that fires the trigger. */
+  eventType: NotetakerEventType;
+  /** Which meetings fire the trigger. Absent fires on every one. */
```
vivekvenkatram: Same as mail: `filters` optional means every meeting summary fires. The `filters` object already requires a condition or a keyword once present, so the hole is omitting the object. wdyt about making `filters` required here too?

testing/upload_validation.ts
```diff
@@ -191,9 +210,17 @@ export const Limits = {
   BuildingBlockDescription: 1000,
   ColumnMatcherRegex: 300,
   ConditionLength: 2000,
+  KeywordLength: 200,
+  MailFilterAddressValue: 320,
+  MailFilterTextValue: 998,
   MaxBlockedDomains: 50,
+  MaxDefaultTriggers: 20,
```
vivekvenkatram: `MAX_TRIGGERS_PER_AGENT` on the server is 20 *event* triggers. This 20 is the whole `defaultTriggers` array, so schedule + while-writing eat two event slots (or 20 events then you cannot declare a schedule). wdyt about capping only the event kind at 20 and keeping the existing one-apiece check for the other kinds?

builder.ts
```diff
@@ -730,6 +736,72 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
     return this;
   }
 
+  /**
+   * Adds a mail event trigger this agent runs on. Call it once per event.
```
vivekvenkatram: `Call it once per event` is not what the code does: `_addDefaultEventTrigger` appends, and two `MessageReceived` rows with different `from` filters is a real default. wdyt about dropping that sentence, or enforcing uniqueness on `mailEventType` if you actually want one?

types.ts
```diff
@@ -1916,13 +1917,368 @@ export interface WhileWritingTriggerDefinition extends BaseDefaultTrigger<Defaul
...
+ *
+ * @internal
+ * @hidden
+ */
+export interface SlackEventTriggerDefinition extends BaseEventTrigger<EventTriggerType.Slack> {
+  /** The Slack event that fires the trigger. */
+  eventType: SlackEventType;
+  /** Keywords to match, case insensitively. One to 50, each up to 200 characters. */
+  keywords: [string, ...string[]];
+  /** Whether a firing records thread participation, so replies resume its chat. */
```
vivekvenkatram: `audience` left the pack because it is the adopter's workspace. `monitorThreadFollowUps` also changes how the agent behaves in *their* Slack once a keyword hits. wdyt about leaving it off the pack shape too, and defaulting it server-side on install?

vivekvenkatram (changes requested):

vivekvenkatram (approved):

PR #3642, opened 2026-09-09
Harden isolated-vm transfer options and stub installation

saikirangoud-grammarly:
Hardens three boundaries in the Pack runtime bootstrap.

- **Transfer options.** `isolated-vm` resolves option objects with a plain property get, so any
  option it reads that is not an own property resolves up the prototype chain. Every option it
  reads is now declared explicitly (`timeout`/`arguments`/`result`, and
  `copy`/`externalCopy`/`reference`/`promise` on the nested objects), using object-literal syntax
  so the properties are defined rather than assigned. An explicit `undefined` is equivalent to
  omitting the option, so behaviour is unchanged.

- **Stub installation.** Host stubs are installed with a property definition rather than an
  assignment, so installation cannot be intercepted by a setter inherited from the prototype chain.

- **Marshaling transform paths.** `applyTransform` now requires each transform path segment to be
  an own property of a plain object or array, and rejects anything else.

Adds unit coverage for the transform-path check. No public API change.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codahq/packs-sdk",
-  "version": "1.17.5",
+  "version": "1.17.6",
```
patrick-codaio: let's make this 1.17.6.prerelease-1 and allow the SDK to go through it's normal release process.

patrick-codaio (approved):

PR #3643, opened 2026-09-09
Tighten schedule trigger rrule validation

peizhaomei-codaio:
[GOECOS-373](https://linear.app/superhuman/issue/GOECOS-373/sdk-cli-validate-scheduled-trigger-rrule-during-serialization-reject)

## Validation

Changed:

- `Limits.RRuleStringLength` 1000 → 512, matching `rrule_string varchar(512)`.
- `DTSTART`/`UNTIL` must name a real date. `20260231` used to pass.
- `BYDAY`'s occurrence ranged at 1–53. `0MO`, `00MO`, `99MO` used to pass.

Added:

- `TZID` must resolve. `TZID=Mars/Olympus` used to pass.
- `UNTIL` must not precede `DTSTART`, compared by date.

Still permitted: `FREQ=YEARLY`, `BYSETPOS`, `BYDAY=-1FR`, negative `BYMONTHDAY`, `INTERVAL>30`, multi-value `BYHOUR`, `BYMINUTE` outside 0/30, `COUNT` with `UNTIL`, lone `BYSETPOS`, ordinal `BYDAY` under `FREQ=WEEKLY`, `BYMONTHDAY` under `FREQ=WEEKLY`.

`make test`: 843 passing, 1 pending.

## For the server side

- Install must fill in `DTSTART`+`TZID` (→ GOECOS-204). Without it `computeNextRunAt` anchors to parse time, and `parseScheduleTriggerFromRRule` throws `DTSTART missing; stored rrule is corrupt` — 500 on `getCustomAgentForUser`, null on the list surfaces.
- Bare `FREQ=YEARLY` reads back as `unit: day` (`_unitFromFrequency`'s `default`), and `AgentBuilderProvider` sends `toScheduleTriggerInput(...)` on every save, so an unrelated edit rewrites the stored rule to `FREQ=DAILY`.
- The rejection-class gate should be a lossless round-trip into the picker's value domain, not whether `rrulestr` throws, and it has to cover the save path.
- `describeScheduleTrigger` renders a negative `dayOfMonth` as `-1st of every month`.
- `schedule-triggers.md`: the UI table points at `agent_builder_trigger_by_time.tsx`, now `libs/go-agent-builder/src/triggers/AgentBuilderTriggerByTime.tsx`; it says the timezone picker hides when unit = Hour, but that row renders unconditionally.


zack-codaio (approved):

testing/rrule_validation.ts
```diff
@@ -135,6 +149,36 @@ export function validateRRuleString(rruleString: string): string | undefined {
...
+function isDateTime(value: string): boolean {
+  const parts = value.split('T');
+  const date = parts.length <= 2 ? DatePattern.exec(parts[0]) : null;
+  if (!date) {
+    return false;
+  }
+  const [year, month, day] = [Number(date[1]), Number(date[2]), Number(date[3])];
+  const parsed = new Date(Date.UTC(year, month - 1, day));
+  if (parsed.getUTCMonth() !== month - 1 || parsed.getUTCDate() !== day) {
+    return false;
```
vivekvenkatram, line 178: Date.UTC maps year 0-99 onto 1900-1999 and we never check getUTCFullYear, so `DTSTART:00040101` would pass as 1904. wdyt about:

```ts
if (
  parsed.getUTCFullYear() !== year ||
  parsed.getUTCMonth() !== month - 1 ||
  parsed.getUTCDate() !== day
) {
  return false;
}
```

testing/rrule_validation.ts
```diff
@@ -95,9 +105,13 @@ export function validateRRuleString(rruleString: string): string | undefined {
     }
   }
   const until = parts.get('UNTIL');
-  if (until !== undefined && !DateTimePattern.test(until)) {
+  if (until !== undefined && !isDateTime(until)) {
     return 'A schedule trigger has an invalid UNTIL.';
   }
+  // Compared by date alone: a DTSTART carrying a TZID is a wall clock, while UNTIL is UTC.
+  if (until !== undefined && dtstart !== undefined && until.slice(0, 8) < dtstart.slice(0, 8)) {
+    return 'A schedule trigger must not end before it starts.';
```
vivekvenkatram, line 128: `DTSTART;TZID=Pacific/Auckland:20260102T010000` with `UNTIL=20260101T120000Z` is the same instant, but `20260101 < 20260102` still rejects. wdyt about skipping this compare when a TZID is present, same as the same-day case we already leave to the runtime?

testing/rrule_validation.ts
```diff
@@ -60,9 +66,13 @@ export function validateRRuleString(rruleString: string): string | undefined {
   if (rule === undefined) {
     return 'A schedule trigger must have an RRULE.';
   }
-  if (dtstart !== undefined && !DateTimePattern.test(dtstart)) {
+  if (dtstart !== undefined && !isDateTime(dtstart)) {
     return 'A schedule trigger has an invalid DTSTART.';
   }
+  const timezone = dtstartParams.find(param => param.toUpperCase().startsWith(TimezoneParam));
+  if (timezone !== undefined && !isTimezone(timezone.slice(TimezoneParam.length))) {
+    return 'A schedule trigger has an invalid DTSTART timezone.';
```
vivekvenkatram: `TZID="America/New_York"` is legal in RFC 5545 and lands in `isTimezone` with the quotes still on. wdyt about:

```ts
const id = timezone.slice(timezone.indexOf('=') + 1).replace(/^"(.*)"$/, '$1');
if (!isTimezone(id)) {
  return 'A schedule trigger has an invalid DTSTART timezone.';
}
```

testing/rrule_validation.ts
```diff
@@ -60,9 +66,13 @@ export function validateRRuleString(rruleString: string): string | undefined {
   if (rule === undefined) {
     return 'A schedule trigger must have an RRULE.';
   }
-  if (dtstart !== undefined && !DateTimePattern.test(dtstart)) {
+  if (dtstart !== undefined && !isDateTime(dtstart)) {
     return 'A schedule trigger has an invalid DTSTART.';
   }
+  const timezone = dtstartParams.find(param => param.toUpperCase().startsWith(TimezoneParam));
```
vivekvenkatram, line 72: `DTSTART;TZID=America/New_York:20260101T090000Z` sets both a TZID and a Zulu time, which RFC 5545 forbids and which `rrulestr` is picky about. wdyt about rejecting when the value ends with `Z` and a TZID is set?

testing/rrule_validation.ts
```diff
@@ -135,6 +149,36 @@ export function validateRRuleString(rruleString: string): string | undefined {
...
+  const parsed = new Date(Date.UTC(year, month - 1, day));
+  if (parsed.getUTCMonth() !== month - 1 || parsed.getUTCDate() !== day) {
+    return false;
+  }
+  if (parts.length === 1) {
+    return true;
+  }
+  const time = TimePattern.exec(parts[1]);
+  // RFC 5545 leaves room for a leap second.
+  return time !== null && Number(time[1]) <= 23 && Number(time[2]) <= 59 && Number(time[3]) <= 60;
```
vivekvenkatram: JS Date and `rrule` don't actually have leap seconds, so `235960` can still blow up in `computeNextRunAt` after we let it through. If the runtime rolls or throws, wdyt about capping seconds at 59?

testing/rrule_validation.ts
```diff
@@ -28,8 +28,11 @@ const NumericRulePartRanges: Array<[string, ...Array<[number, number]>]> = [
   ['BYSETPOS', [-366, -1], [1, 366]],
 ];
 
-const DateTimePattern = /^\d{8}(T\d{6}Z?)?$/;
-const WeekdayPattern = new RegExp(`^[+-]?\\d{0,2}(${Weekdays.join('|')})$`);
+const DatePattern = /^(\d{4})(\d{2})(\d{2})$/;
+const TimePattern = /^(\d{2})(\d{2})(\d{2})Z?$/;
```
vivekvenkatram, line 32: DTSTART isn't uppercased (UNTIL is, via `parts.set`), so `T090000z` fails here while the same UNTIL would pass. wdyt about `z?` on TimePattern, or uppercasing the DTSTART value before `isDateTime`?

testing/upload_validation.ts
```diff
@@ -198,7 +198,7 @@ export const Limits = {
   NetworkDomainUrl: 253,
   PermissionsBatchSize: 5000,
   PromptLength: 20000,
-  RRuleStringLength: 1000,
+  RRuleStringLength: 512,
```
vivekvenkatram, line 201: Matching `varchar(512)` is the right call, but I don't see a test that 513 chars now fail upload. wdyt about one next to the existing MINUTELY `rruleString` case that pads a valid rule to `Limits.RRuleStringLength + 1`?

test/rrule_validation_test.ts
```diff
@@ -89,6 +102,81 @@ describe('validateRRuleString', () => {
...
+      assert.equal(
+        validateRRuleString('DTSTART:20260101\nRRULE:FREQ=DAILY;UNTIL=20250101T000000Z'),
+        'A schedule trigger must not end before it starts.',
+      );
+    });
+
+    it('takes an UNTIL on or after the DTSTART', () => {
+      assertValid('DTSTART:20260101\nRRULE:FREQ=DAILY;UNTIL=20270101T000000Z');
+      // Same day: the two are in different zones, so leave the boundary to the runtime.
+      assertValid('DTSTART;TZID=America/New_York:20260101T090000\nRRULE:FREQ=DAILY;UNTIL=20260101T000000Z');
```
vivekvenkatram, line 166: This same-day NY vs UTC case is the false-accept we chose on purpose. The false-reject twin is east-of-UTC (`Pacific/Auckland` next calendar day, UNTIL still that UTC morning). If we keep the date compare, wdyt about a test that documents which side we picked?

testing/rrule_validation.ts
xicheng-codaio, line 43: A non-blocking comment: can we make this method more modular, would it be possible to reuse this function on the backend code (which already depends on the SDK). The backend needs all the validation already on the CLI, plus more validations based on capabilities.
some thoughts:
1. maybe throw an exception instead of return a message?
2. update the documentation to describe what was allowed/rejected
3. can we define a shared "default" value if the input is invalid?

peizhaomei-codaio: i think the backend can already use this func. for your thoughts:

1. `upload_validation` is calling `rrule_validation` for error message, probably we don't want to do try & catch there
2. will do
3. my opinion is that, on cli, an invalid input should be rejected and the author should be notified, instead of giving a default value. If it is for existing invalid input living in db, we should do this at be server?

vivekvenkatram (approved):
lgtm, tests are failing

dist/testing/rrule_validation.js
```diff
@@ -124,6 +138,35 @@ function validateRRuleString(rruleString) {
...
+    const parsed = new Date(Date.UTC(year, month - 1, day));
+    if (parsed.getUTCMonth() !== month - 1 || parsed.getUTCDate() !== day) {
+        return false;
+    }
+    if (parts.length === 1) {
+        return true;
+    }
+    const time = TimePattern.exec(parts[1]);
+    // RFC 5545 leaves room for a leap second.
+    return time !== null && Number(time[1]) <= 23 && Number(time[2]) <= 59 && Number(time[3]) <= 60;
```
vivekvenkatram: This JS still has the pre-review validator (seconds `<= 60`, no quote strip, date compare even with a TZID). wdyt about rebuilding dist so the published package matches `testing/rrule_validation.ts`?

testing/rrule_validation.ts
```diff
@@ -60,9 +66,20 @@ export function validateRRuleString(rruleString: string): string | undefined {
...
     return 'A schedule trigger must have an RRULE.';
   }
-  if (dtstart !== undefined && !DateTimePattern.test(dtstart)) {
+  if (dtstart !== undefined && !isDateTime(dtstart)) {
     return 'A schedule trigger has an invalid DTSTART.';
   }
+  const timezone = dtstartParams.find(param => param.toUpperCase().startsWith(TimezoneParam));
+  if (timezone !== undefined) {
+    // RFC 5545 lets a param value carry quotes.
+    if (!isTimezone(timezone.slice(TimezoneParam.length).replace(/^"(.*)"$/, '$1'))) {
```
vivekvenkatram: `slice(TimezoneParam.length)` only works because `TZID=` is always 5 chars. wdyt about taking everything after the first `=`:

```ts
const id = timezone.slice(timezone.indexOf('=') + 1).replace(/^"(.*)"$/, '$1');
if (!isTimezone(id)) {
  return 'A schedule trigger has an invalid DTSTART timezone.';
}
```

dist/testing/rrule_validation.js
peizhaomei-codaio: Done in 41a827a6. `dist/testing/rrule_validation.js` now matches the source, and `dist/bundle.d.ts`/`dist/types.d.ts` picked up the new `rruleString` doc. That was also the `validate-no-changes` failure.

testing/rrule_validation.ts
peizhaomei-codaio: Done in 41a827a6 — reading from the first `=` now.

testing/rrule_validation.ts
peizhaomei-codaio: Documented in 41a827a6 (thought 2): the `rruleString` doc now lists the parts it takes, notes that picker-unrepresentable rules upload and render read-only, and lists what gets rejected.

PR #3645, opened 2026-09-15
Add a graceful parsing option for default triggers

peizhaomei-codaio:
Server-side parsing handles payloads it did not author, where a trigger it cannot read is worth skipping rather than discarding the ones it can. This adds a `bestEffortDefaultTriggers` option to `validatePackVersionMetadata` for that caller.

With the option on, `defaultTriggers` keeps every element that parses and drops the rest — including a duplicate kind, where the first one wins — without reporting an issue. The per-trigger schemas are shared between both modes, so a new trigger kind is still defined once. Everything else in the pack stays strict.

Default is off, so `packs upload` and every existing caller are unchanged: one malformed trigger still rejects the upload with an error pointing at `defaultTriggers[i]`.

Also pins the strict behavior with two regression tests, since the reported symptom was that a bad trigger silently emptied the array. It does not — the errors are already per-element and the upload is refused outright.


testing/upload_validation.ts
```diff
@@ -2453,21 +2456,38 @@ ${endpointKey ? 'endpointKey is set' : `requiresEndpointUrl is ${requiresEndpoin
...
+        });
+      }
+      seen.add(trigger.kind);
     });
+  });
+
+  // Keeps the triggers that read cleanly and drops the rest, so a payload this parser only partly
+  // understands still yields the triggers it does. Never reports an issue: a caller that wants to
+  // hear about a bad trigger wants the strict schema.
+  const bestEffortDefaultTriggersSchema = z.array(z.unknown()).transform(items => {
```
xicheng-codaio, line 2475: do we also need the "one trigger per kind" validation even in the graceful parsing?

xicheng-codaio (approved):

PR #3646, opened 2026-09-15
Fix packs init on npm 9+ and stop it from installing into the parent project

peizhaomei-codaio:
Fixes [GOECOS-192](https://linear.app/superhuman/issue/GOECOS-192/coda-init-crashes-with-enoent).

`packs init` / `coda init` failed two ways:

- `addPatches()` ran `npm set-script`, removed in npm 9, so every modern npm printed `Unknown command: "set-script"`. The patch it injected dates to Node 16; `patches/` and `patch-package` no longer exist in this repo, and it wrote a `postinstall: npx patch-package` hook into the user's new pack. Removed, along with the now-obsolete npm < 7 check.
- It read `node_modules/@codahq/packs-examples` relative to cwd, but npm installs into the nearest ancestor containing a `package.json`. Run from a subdirectory, init added `@codahq/packs-examples` to the parent project and then died with a bare `ENOENT`.

For the second one this PR makes the requirement explicit rather than resolving paths around it. init has no package.json of its own to install into — the template ships `pack.ts`, `tsconfig.json` and `.gitignore` only — so a Pack created in `apps/my-pack` would have `packs-sdk` plus the examples' 23 `devDependencies` added to the parent app. It now exits with the npm root, what would happen, and the two ways forward (`npm init -y` here, or run init there) before touching anything.

Everything that can fail now reports why: missing npm or git, the examples download, the dev dependency install, the SDK install. The examples directory is resolved with `require.resolve` so a hoisted install is found instead of reconstructed, and a failed cleanup of the examples warns rather than exiting, since the Pack is already usable by then.

## Testing

`test/init_test.ts` stubs `spawnProcess` and covers the refusal (asserting no npm call is made), missing npm, a failed examples install, and unresolvable examples.

Manually, against `dist/`:

- Subdirectory of an npm project: exits 1 with the message; the parent `package.json` is byte-identical and no `node_modules` appears.
- Same directory after `npm init -y`: exits 0, files land in cwd, parent stays clean.
- Project root: exits 0.
- `/tmp/...`, where `cwd` and its realpath differ: exits 0, i.e. the symlink does not read as "outside the npm root".


cli/init.ts
```diff
@@ -3,26 +3,22 @@ import path from 'path';
...
-function addPatches() {
-  spawnProcess(`npm set-script postinstall "npx patch-package"`);
-
-  updateMoldSourceMap();
-
-  spawnProcess(`npx patch-package --exclude 'nothing' mold-source-map`);
+// npm installs into the nearest ancestor directory containing a package.json, not into the current
+// directory, so every node_modules path has to be resolved from there.
+function getNpmRoot(): string {
+  const {status, stdout} = spawnProcess('npm prefix', {stdio: 'pipe'});
```
vivekvenkatram, line 12: iiuc `npm prefix` is the project root, not the directory npm actually installed into. In a workspace (or with hoisting) that is often `$workspace/node_modules`, so `path.join(prefix, 'node_modules/@codahq/packs-examples')` can still ENOENT after a successful install. wdyt about `npm root` or `npm ls --parseable @codahq/packs-examples` so we follow the real install location instead of reconstructing it?

cli/init.ts
```diff
@@ -35,12 +31,7 @@ function escapeShellCmd(cmd: string): string {
 }
 
 export async function handleInit() {
-  // stdout looks like `8.1.2\n`.
-  const npmVersion = parseInt(spawnProcess('npm -v', {stdio: 'pipe'}).stdout.toString().trim().split('.', 1)[0], 10);
-  if (npmVersion < 7) {
-    // need npm 7 to support "npm set-script"
-    throw new Error(`Your npm version is older than 7. Please upgrade npm to at least 7 with "npm install -g npm@7"`);
-  }
+  const packsExamplesDirectory = path.join(getNpmRoot(), 'node_modules/@codahq/packs-examples');
```
vivekvenkatram: This is the product fork I keep circling. Init from a subdirectory now succeeds, but every `npm install` / `npm uninstall` still mutates the ancestor `package.json`, while the template is copied into cwd. So a user who ran `packs init` in `apps/my-pack` just added packs-sdk plus the examples' entire `devDependencies` to the parent app. If the intent is a new pack in that folder, should we `npm init` (or `--prefix` cwd) first, or refuse when cwd !== npm root with a message to `cd` there? The old ENOENT at least didn't poison the parent.

cli/init.ts
```diff
@@ -58,10 +49,15 @@ export async function handleInit() {
       );
     }
     const installCommand = `npm install https://github.com/coda/packs-examples.git`;
-    spawnProcess(installCommand);
+    if (spawnProcess(installCommand).status !== 0) {
```
vivekvenkatram: Nice that this one is checked now. The `npm install --save-dev …`, `npm install --save @codahq/packs-sdk`, and `npm uninstall @codahq/packs-examples` calls below still ignore status though. If those fail we copy a template that won't build (or leave examples in the parent deps) and exit 0. Same `status !== 0` → `printAndExit` shape would close the rest of the ticket's "raw ENOENT / silent failure" class.

cli/init.ts
```diff
@@ -58,10 +49,15 @@ export async function handleInit() {
...
+      return printAndExit(
+        'The packs init command could not install the Pack examples. ' +
+          'Check that you can reach https://github.com/coda/packs-examples and try again.',
+      );
+    }
   }
 
-  const packageJson = JSON.parse(fs.readFileSync(path.join(PacksExamplesDirectory, 'package.json'), 'utf-8'));
+  const packageJson = JSON.parse(fs.readFileSync(path.join(packsExamplesDirectory, 'package.json'), 'utf-8'));
```
vivekvenkatram, line 75: The install status check still doesn't cover this read. `npm list` can be 0 (or the git install can be 0) while this path is missing, and we fall through to the same bare ENOENT the ticket was about. wdyt about wrapping this (and the `copySync` below) with the same `printAndExit` shape, including the resolved path in the message so it's diagnosable?

cli/init.ts
```diff
@@ -71,10 +67,7 @@ export async function handleInit() {
     spawnProcess('npm install --save @codahq/packs-sdk');
   }
 
-  // developers may run in NodeJs 16 where some packages need to be patched to avoid warnings.
-  addPatches();
-
-  fs.copySync(`${PacksExamplesDirectory}/examples/template`, process.cwd());
+  fs.copySync(path.join(packsExamplesDirectory, 'examples/template'), process.cwd());
```
vivekvenkatram: nit: after we copy into cwd we still `npm uninstall` the examples from the npm root with no status check. That's especially sharp now that init-from-subdir is a supported path. Longer-term wdyt about skipping the install/uninstall dance and reading the template from a path the SDK already ships, so we never write this package into the user's tree?

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 - CLI commands now fail immediately in non-interactive environments instead of waiting on confirmation prompts. Use `--yes` to skip confirmations on `clone`, `link`, `register`, and `release`. `register` requires `--apiToken` (or a pasted token on a TTY). `release` accepts `--use-latest` when the manifest has no version. `--help` for each command includes examples ready to run.
 
+### Fixed
+
+- `packs init` no longer fails with `ENOENT` when run from a subdirectory of an npm project, and no longer runs the obsolete `npm set-script` patching step that was removed in npm 9. A missing npm or a failed download of the Pack examples now prints an actionable message instead of a stack trace.
```
vivekvenkatram: The PR body has three solid repros (subdir, npm missing, git ls-remote fail) and none of them are in the repo. `handleInit` is a pile of `spawnProcess` so this will regress the next time someone touches paths. Even a test that stubs `spawnProcess` and asserts we resolve examples from the mocked prefix (and that we `printAndExit` on non-zero install) would lock the ticket. Happy to be wrong if CLI tests are deliberately not a thing here.

cli/init.ts
peizhaomei-codaio: will open a separate ticket

cli/init.ts
```diff
@@ -31,24 +39,22 @@ function isGitAvailable(): boolean {
...
   }
 
-  let isPacksExamplesInstalled: boolean;
-  try {
-    const listNpmPackages = spawnProcess('npm list @codahq/packs-examples');
-    isPacksExamplesInstalled = listNpmPackages.status === 0;
-  } catch (error: any) {
-    isPacksExamplesInstalled = false;
-  }
+  const isPacksExamplesInstalled = spawnProcess('npm list @codahq/packs-examples').status === 0;
```
vivekvenkatram: `npm list` exits 1 for any tree problem, not only a missing package, and this call inherits stdio so a first run prints npm's own error before we install. A project that already has the examples plus an unmet peer would take the install branch every time. wdyt about `npm ls @codahq/packs-examples --json --depth=0` with `stdio: 'pipe'` and treating it as installed when `dependencies['@codahq/packs-examples']` is set, ignoring the exit code?

cli/init.ts
```diff
@@ -57,31 +63,42 @@ export async function handleInit() {
...
+  const packageJson = JSON.parse(fs.readFileSync(path.join(packsExamplesDirectory, 'package.json'), 'utf-8'));
   const devDependencies = packageJson.devDependencies;
   const devDependencyPackages = Object.keys(devDependencies)
     .map(dependency => `${dependency}@${devDependencies[dependency]}`)
     .join(' ');
-  spawnProcess(escapeShellCmd(`npm install --save-dev ${devDependencyPackages}`));
-  if (spawnProcess('npm list @codahq/packs-sdk --depth=0').status !== 0) {
-    spawnProcess('npm install --save @codahq/packs-sdk');
+  if (spawnProcess(escapeShellCmd(`npm install --save-dev ${devDependencyPackages}`)).status !== 0) {
+    return printAndExit('The packs init command could not install the Pack development dependencies.');
```
vivekvenkatram: These exits (and the template one below) skip the uninstall, so a failed dev-dep install leaves the git dependency on `@codahq/packs-examples` in the package.json we just agreed not to touch by accident. wdyt about uninstalling before `printAndExit` on these paths, same as the success cleanup?

cli/init.ts
```diff
@@ -1,28 +1,36 @@
...
-
-  spawnProcess(`npx patch-package --exclude 'nothing' mold-source-map`);
+// npm may hoist the examples above the npm root, so ask Node where they actually landed instead of
+// assuming they are in the nearest node_modules.
+function getPacksExamplesDirectory(): string {
+  try {
+    return path.dirname(require.resolve('@codahq/packs-examples/package.json', {paths: [process.cwd()]}));
+  } catch (error: any) {
+    return printAndExit(
+      `The packs init command could not find @codahq/packs-examples from ${process.cwd()}. ` +
```
vivekvenkatram: This runs after `npm install` already returned 0, so "check that you can reach GitHub" is the wrong hint (bad exports, a hoist Node can't see, a missing `package.json` subpath). wdyt about including `error.code` and saying the examples installed but could not be resolved from `process.cwd()`?

test/init_test.ts
```diff
@@ -0,0 +1,87 @@
...
+      npmRoot: fs.realpathSync(process.cwd()),
+      failures: ['npm list @codahq/packs-examples', 'npm install https://github.com/coda/packs-examples.git'],
+    });
+
+    await runInit();
+
+    assert.include(exitMessage!, 'could not install the Pack examples');
+  });
+
+  it('exits with a message when the examples cannot be found after installing', async () => {
```
vivekvenkatram: Nothing covers a failed `--save-dev` install or a missing template, and `failures` matches the whole command string, so the escaped dev-dep line can never be listed. The last test also calls the real `require.resolve` and would `copySync` into this repo if that package ever resolves. wdyt about stubbing `fs` and asserting those two exits happen before `copySync`?

cli/init.ts
```diff
@@ -57,31 +63,38 @@ export async function handleInit() {
...
+  const packsExamplesDirectory = getPacksExamplesDirectory();
+  const packageJson = JSON.parse(fs.readFileSync(path.join(packsExamplesDirectory, 'package.json'), 'utf-8'));
   const devDependencies = packageJson.devDependencies;
   const devDependencyPackages = Object.keys(devDependencies)
     .map(dependency => `${dependency}@${devDependencies[dependency]}`)
     .join(' ');
-  spawnProcess(escapeShellCmd(`npm install --save-dev ${devDependencyPackages}`));
-  if (spawnProcess('npm list @codahq/packs-sdk --depth=0').status !== 0) {
-    spawnProcess('npm install --save @codahq/packs-sdk');
+  if (spawnProcess(escapeShellCmd(`npm install --save-dev ${devDependencyPackages}`)).status !== 0) {
```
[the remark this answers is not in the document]
vivekvenkatram: The global replace fixes the second `>`/`<`, but `spawnProcess` still runs `shell: true`, and packs-examples is installed from git HEAD, so a version like `1.0.0 && echo pwned` in that repo still executes. Current devDeps are plain versions so this is latent. Quoting each spec (`"pkg@version"`) would close the shell metacharacters the `<`/`>` escape misses.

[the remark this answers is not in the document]
vivekvenkatram: +1

peizhaomei-codaio:
migrated to monorepo

PR #3647, opened 2026-09-15
Add a SuggestionProducer tool type and the formula factory that makes its contract unmissable

daviddriscoll-codaio:
## Summary

**Problem:** A connector that already computes its own suggestions has no way to hand them back. Its formula returns data, an LLM re-reads that data and re-authors it as suggestions, and the platform pays for a model turn to restate an answer it was given. There is also nowhere for a pack to say "this formula's output is already finished suggestions."

**Fix:** The pack-side half of that contract — a tool type, a formula factory, a result shape, the builder method an agent declares the tool with, and validation that the named formula matches what the runtime will call. No runtime behavior here; `front-end-web-monorepo` consumes these in [superhumans-org/front-end-web-monorepo#21317](https://github.com/superhumans-org/front-end-web-monorepo/pull/21317).

Pack-side plumbing for [Suggestions from connectors (with importance)](https://docs.superhuman.com/d/go-ECO-Go-Ecosystem_d7gj39Whce5/Suggestions-from-connectors-with-importance_sufMAlcn), and the SDK dependency of @abhishek-amit's `/executeAgent`-shaped-interface proposal on that doc.

Verified end to end against a local stack: a real pack declaring one of these tools uploads, releases, installs, and has its formula called directly with the LLM loop skipped.

## The contract

Two packs. The connector holds the formula; the agent names it and is the thing the runtime actually runs.

```ts
// pack.ts, in the connector — the half that talks to the checker.
const pack = sdk.newPack();
pack.addNetworkDomain('radicalcandor.com');
pack.setUserAuthentication({type: sdk.AuthenticationType.HeaderBearerToken});

pack.addSuggestionFormula({
  name: 'CheckSuggestions',
  description: 'Flags feedback that is too hedged to land.',
  execute: async (text, context) => ({
    suggestions: [
      {
        startOffset: 21,
        endOffset: 42,
        original: text.slice(21, 42),
        title: 'Ruinous empathy',
        explanation: 'The praise cancels the critique.',
        replacement: 'rushed — slides 4-7 had no data behind them',
        importance: 1,
      },
    ],
  }),
});
```

```ts
// pack.ts, in the agent — the half the runtime runs. 1234 is the connector's pack id.
const pack = sdk.newAgent();
pack.setTools({suggestions: {packId: 1234, formulaName: 'CheckSuggestions'}});
pack.setDefaultWhileWritingTrigger({
  condition: 'The writer is giving someone feedback',
  surfaces: [sdk.ContextualTriggerSurface.Docs, sdk.ContextualTriggerSurface.Email],
});

// Reached only when the producer cannot be called, so it describes the producer's job rather
// than a different one — a run that falls back should not change the subject.
pack.setInstructions('Flag feedback that is too hedged to land, with a concrete rewrite.');
```

| | Before | After |
| --- | --- | --- |
| Declaring a suggestion producer | nothing; the runtime would have to match a magic formula name | `ToolType.SuggestionProducer`, holding the connector's `packId` and one `formulaName` |
| Declaring the formula | — | `pack.addSuggestionFormula()`, which fixes the parameter, the result schema and the result type |
| Declaring the producer from an agent | impossible — `setTools` knew only `{docs, mail, webSearch, connectors}` | `setTools({suggestions: {packId, formulaName}})` |
| Result shape | each pack hand-rolls one, runtime translates | `makeSuggestionResultSchema()` — `{suggestions: [...]}`, typed by `SuggestionResult` |
| Locating a finding | — | `startOffset`/`endOffset` into the submitted text, with `original` as a checksum on them |
| Checking the named formula | — | upload rejects it unless its parameter is a string **named `text`**, the rest are optional, it is not an action, and it returns the required shape |

Everything new is `@internal @hidden`, so `make typedoc` produces no docs diff.

## Why each part is shaped this way

**It's a tool type, not a field on `PackTool`.** Exactly one producer per agent is legal, and a tool type makes that the natural statement — `agentSchema.tools` already keys non-Pack types on `type` alone, so the agent path enforces "only once" with no new code, and skills need one four-line check. A boolean or a formula name on `PackTool` could express N producers when one is legal.

A field on `Skill` or `AgentDefinition` was considered and rejected: the runtime reads `customAgentDefinition.skill.tools`, which `agentDefinitionToRootSkill` assembles from `agent.tools`, so `tools[]` is the only channel that reaches it. A field would mean touching `Skill` anyway — the same deprecation cost — or blocking on plumbing that does not exist. The exhaustiveness maps that now need a `SuggestionProducer` entry are doing their job: each one is a call site that had to be decided.

**It takes two packs, and the SDK forces that.** Neither builder can hold both halves: `AgentDefinitionBuilder` has no `addFormula`, and `newPack()` is typed `Partial<Omit<PackVersionDefinition, 'agent' | 'defaultTriggers'>>`, so a connector cannot carry the while-writing trigger that makes a producer reachable at all. The two are joined by the connector's pack id — which is why `agentToolSchema` requires `packId` on a producer while `skillSchema` leaves it optional.

An earlier revision of this description showed one pack calling both `addSuggestionFormula` and `addSkill`. That cannot work, and the reason is worth stating rather than quietly fixing: `/executeAgent` builds its agent from `agentInstance.customAgentDefinition.skill`, and that skill comes from `metadata.agent` via `agentDefinitionToRootSkill` — `prompt` from `agent.instructions`, `tools` from `agent.tools`. A pack's own `skills` array is read on other paths (`pack_agent_builder.ts`, `pack_skill_agent_builder.ts`) but never on this one, so a producer declared through `addSkill` would be silently ignored while looking supported — the companion's upload validation even defaults its `packId`.

**So `AgentToolsDef.suggestions` closes the gap that left behind.** `agentToolSchema` already accepted a producer on `agent.tools`, but `setTools` understood only `{docs, mail, webSearch, connectors}`, which made the one tool an agent most needs the one tool it could not declare — and left `addSkill` as the only thing an example could show. `setTools({suggestions: {packId, formulaName}})` emits it, with `packId` required to match the schema, because an agent holds no formulas of its own. With connector/agent bundling now decided against, this split is the shape rather than a stopgap, so the builder should be able to say it.

The agent's `instructions` are not dead weight on this path, which is the other thing that revision got wrong. Every producer failure in the companion — `no-tool`, `off-contract`, `threw`, `unparseable` — falls back to the LLM loop, and that loop runs on `skill.prompt`, i.e. on `setInstructions`. A producer-backed agent still needs instructions that describe the producer's own job.

**The producer returns findings, not operations.** The runtime reconciles a checker's answer against the highlights already on screen: it reuses an existing id so a finding updates in place, deletes a highlight the checker no longer reports, and drops one the user already dismissed. That only works if the formula hands back findings. Emitting `operations` would ask every third-party checker to describe an edit to editor state it cannot see.

**A span is located by offset, not by quoting the text around it.** The shape first carried `contextBefore`/`contextAfter` and no offsets, because it was derived from the *model's* output schema — a model cannot count characters, so it has to quote landmarks. A producer computed its findings and already knows where they are. Two things came out of checking that assumption:

- Nothing read those fields. CAPI's `OperationDto.HighlightDto` never parsed them, and `PositionResolver` recovers the position with `indexOf(original)` — documenting that it anchors arbitrarily when a span repeats. So the fields were serialized, sent, and dropped.
- The first real first-party producer had the offsets already. AI Rewriter reports exact `start`/`end` and was spending them synthesizing 24-character context strings.

`original` stays, as a checksum rather than the anchor: the runtime drops a finding where `text.slice(startOffset, endOffset) !== original`, which catches a stale offset (the checker scanned different text), a paraphrased span, and a miscount. That is strictly stronger than the `includes()` check it replaces. Offsets are **UTF-16 code units** — JavaScript `String` indices, so `slice` agrees with them — documented explicitly because it is the field a non-JS partner is most likely to get wrong.

The companion does not yet forward offsets to CAPI, which has no field for them; it validates with them and emits `original`. Two occurrences of one phrase therefore still collapse to a single card, and the runtime counts that rather than losing it silently.

**`id` is gone.** It was documented as "stable id, so a later run can update or delete this suggestion" — but the id namespace is CAPI's: `currentHighlights` arrives in the `/executeAgent` request body, CAPI mints `"h1"`, `"h2"` in `LlmHighlight`, and `DEFAULT_SYSTEM_PROMPT` tells the model to echo an existing one back. A pack has never seen those ids, so any it minted was meaningless — which is why the runtime correctly preferred the on-screen id over the pack's. Both demo packs had written polynomial hash functions that bought nothing. If a producer ever needs stable identity, that wants a separate `externalId` so the two namespaces never share a key.

**Optional fields are omitted, not nulled.** `id` and `replacement` were previously `.nullable()` without `.optional()` — a required key with a nullable value — to match OpenAI's strict structured-output mode, which only drops a field from JSON Schema's `required` when it is wrapped in `.optional()`. That is a constraint on the *model's* schema and it had no business in the pack contract, where it forced every partner to write `replacement: null`. It also could not have stayed: a Coda object schema has no null, so `makeSuggestionResultSchema()` cannot express one, and `SuggestionResult` was not assignable to the schema's own type until this changed. Absence is the only thing the schema can say.

**`addSuggestionFormula()` exists because every way to get this wrong is silent.** A producer the runtime cannot call is not rejected at runtime — it is skipped, the run falls back to the LLM loop, and a plausible answer still reaches the user. The only tell is the `reasoning` line. So the factory fixes the three things that fail that way:

- **the parameter name.** Arguments reach a formula by parameter *name* (`executePackFormula` takes a `Record` keyed by name), so a first parameter called `content` receives nothing. Upload validation now checks the name, not just the position and type — the previous check validated the wrong thing.
- **the result schema**, so the runtime's parser recognizes it.
- **the result type**, so `execute` returning the wrong shape is a compile error.

It is a builder method rather than a bare `makeFormula` wrapper because `addFormula` types `execute`'s return against `ObjectSchemaType<>` of the schema beside it, and `makeSuggestionResultSchema()` is a `GenericObjectSchema` whose properties are erased. Absorbing that inside the builder — which already pushes `formula as any` — keeps the cast out of every pack.

**There is no `focus` parameter, and no versioned JSON parameter to hold one.** An earlier revision passed `{version, focus}` as a stringly-typed second argument carrying the spans the writer had just edited. Two reasons it is gone: it cannot bound a producer's work, because the runtime deletes any highlight the checker does not re-report, so a checker that narrowed its scan to the edited region would delete every finding outside it — leaving it a ranking hint, when bounding was the part worth paying for. And the one real first-party producer never declared the parameter at all, while the runtime kept sending it. A stringly-typed JSON envelope behind a hand-rolled version number is also a pattern that appears nowhere else in the SDK, and it made every partner reimplement the same 25 lines of defensive parsing. When reconciliation supports a partial answer — `reconcileCompleteAnswer`'s docstring sketches the `accept()`/`finish()` pair — `focus` becomes a real bound and earns a typed `ParameterType.StringArray` parameter. Adding an optional parameter later is backwards-compatible, because the factory owns the list.

**`ExecutionContext.suggestions` is gone too.** `SuggestionRun`, `SuggestionExecutionContext`, `newMockSuggestionExecutionContext` and the bootstrap injection were a typed surface with no consumer anywhere — I grepped the companion repo for all four names, zero hits. The runtime suppresses dismissed and accepted findings itself, which is where that policy belongs: a checker should not have to implement it. If a checker later wants to see the lists to skip work, that comes back with the plumbing rather than ahead of it.

## zod is out of the pack-execution path

`api_types.ts` had no value imports at all — only `import type`. Adding `import {z} from 'zod'` to it, so the agent runtime could reference `SuggestionHighlightSchema.shape.<field>`, pulled all of zod into the isolate:

| | Before | After |
| --- | --- | --- |
| `bundles/thunk_bundle.js` | 419,116 bytes | **98,077** |
| `dist/bundle.js` | 411,169 bytes | **91,873** |
| zod symbols in the thunk bundle | 488 | **0** |

`thunk_bundle.js` is `context.eval`'d into isolated-vm on every pack execution (`registerBundles` → `getThunkPath`), so that was a 4.3× parse-and-compile regression on every pack invocation, paid for a server-side consumer's convenience.

The zod schemas now live in `suggestion_schemas.ts`, which nothing under `runtime/` or `index.ts` reaches. Consumers deep-import `@codahq/packs-sdk/dist/suggestion_schemas` — `apps/docs` already deep-imports 1,585 other `dist/*` paths, including value-level ones like `validateCrawlHierarchy` from `dist/testing/upload_validation`, and packs-sdk publishes no `exports` map.

Drift between the schema and the type is caught by asserting assignability in **both** directions with the `Assert<>` helper `api_types.ts` already uses for enum completeness. Verified it bites: adding a field to either side alone fails the build. (The alternative the companion already uses two files away — `satisfies StrictZodShape<T>` — would also have worked at zero runtime cost; a separate module was chosen so the schema keeps its `.min(0).max(1)` refinements.)

## Also

`SuggestionProducerTool` had to be exported from `index.ts` — it was declared in `types.ts` and reachable through `ToolMap`, but a consumer naming the tool type had no way to type it.

The new tool interface had been inserted between `ToolMap`'s doc comment and `ToolMap` itself, orphaning it. Restored.

`ToolType.SuggestionProducer` rather than `ToolType.Suggestions` on purpose: partners already send the literal string `'Suggestions'` as a `ScreenAnnotationType`, and two enums sharing a wire value would be a bad thing to debug.

**One gap left open on purpose.** `validateSuggestionProducerFormula` runs only for a producer naming a formula in *this* pack. A producer naming another pack's formula cannot be checked from a manifest that does not contain it — and that is now the only shape there is, since the formula and the tool entry live in different packs by construction (above). The companion closes this at the point where the resolved formula's real metadata is in hand, which works for any pack; the comment here says so rather than leaving the asymmetry unexplained.

Two adjacent gaps this does **not** touch, both pre-existing:

- The `forcedFormula` existence check iterates `metadata.skills` only, so a `forcedFormula` declared through `setChatSkill` is never verified at upload. `validateSkillTools` already walks `chatSkill` and is the obvious place to fix it.
- `packToolSchema`'s nested `formulas[]` entry is typed with a hand-written inline generic that lists a `description` field `PackTool.formulas` does not have. Repointing it at `NonNullable<PackTool['formulas']>[number]` would force future fields to be validated, but it would also stop accepting `description`, which is a compat call that doesn't belong here.

## Test plan

| Assertion | Proof |
| --- | --- |
| A skill may use one producer tool | `test/upload_validation_test.ts` — "validates a skill with a suggestion producer tool" |
| Two producer tools are rejected | same file — "fails for a skill with two suggestion producer tools" |
| A local producer must name a real formula | same file — "fails when a suggestion producer names no formula in this pack" |
| A conforming formula passes | same file — "validates a suggestion producer whose formula matches the contract" |
| A wrong result shape is rejected | same file — "fails when a suggestion producer formula returns the wrong shape" |
| A non-string first parameter is rejected | same file — "fails when a suggestion producer formula does not take the text first" |
| **A first parameter named anything but `text` is rejected** | same file — "fails when a suggestion producer formula names its text parameter something else" |
| **What `addSuggestionFormula` builds passes, through the real builder** | same file — "accepts the formula addSuggestionFormula builds", via `newPack()` + `compilePackMetadata()` so it exercises `addFormula`'s `ValueType`→`Type` conversion and schema normalization |
| An agent may name a connector's producer | same file — "takes a suggestion producer naming a connector formula" |
| **What `setTools({suggestions})` builds passes agent validation** | same file — "accepts the producer setTools builds", so the builder and `agentToolSchema` cannot drift |
| `setTools` emits the producer, alone and beside the tools a fallback run uses | `test/builder_test.ts` — "names a connector formula that produces the suggestions itself", "takes a suggestion producer alongside the tools the fallback run uses" |
| `ToolType` exhaustiveness still holds | same file — "Exhaustiveness checking of ToolType works" |
| The factory and the validator cannot drift | breaking the factory's parameter name fails "accepts the formula addSuggestionFormula builds" — checked by hand |
| The schema and the type cannot drift | adding a field to `SuggestionHighlightSchema` or to `SuggestionHighlight` alone fails `tsc` — checked by hand, both directions |
| No regressions | `make test-file` over `upload_validation`, `builder`, `schema`, `api` — 438 + 74 + 17 + 32 passing, 0 failing |
| Typecheck and lint | `tsc --noEmit` clean; `eslint` clean over every changed file |
| Checked-in artifacts match source | `make compile-ts` + `dts-bundle-generator` + `make typedoc`; no typedoc diff |

`make test` also reports failures in `compile` / `Execution` / `Marshaling`, all of them `getIvm()` returning null. The cause is environmental: `isolated-vm`'s prebuilt binary fails to `dlopen` against the Node running the tests, `ivm_wrapper.ts` swallows that and sets `ivm = null`, and `getIvm()`'s `ensureExists` then throws. A type-and-schema change cannot affect a native module load, and every suite that does not need the isolate passes.

Visual: none — type and validation surface only.


api.ts
```diff
@@ -3132,3 +3135,116 @@ function moveJsPropertyOptionsFunctionsToFormulas({
...
+ *
+ * // Used only when the producer cannot be called, so it describes the producer's job rather
+ * // than a different one -- a run that falls back should not change the subject.
+ * pack.setInstructions('Flag feedback that is too hedged to land, with a concrete rewrite.');
+ * ```
+ *
+ * @internal
+ * @hidden
+ */
+export function makeSuggestionFormula<ContextT extends ExecutionContext = ExecutionContext>({
```
vivekvenkatram: This PR carries the tool type, the factory, the offset contract, purpose, and the importance tier, and a few of those shapes already got replaced in the history. wdyt about splitting so the factory and `SuggestionProducer` land first, with the tier enum as its own change?

index.ts
```diff
@@ -248,6 +250,10 @@ export {makeAttributionNode} from './schema';
 export {makeObjectSchema} from './schema';
 export {makeReferenceSchemaFromObjectSchema} from './schema';
 export {makeSchema} from './schema';
+export {makeSuggestionFormula} from './api';
```
vivekvenkatram: These land on the package entry while the contract is still moving. wdyt about keeping them `@hidden` (`ToolType.SuggestionProducer` and `FormulaPurpose.Suggestions` are missing it) and not exporting from `index` until a pack outside this repo is supposed to call them?

types.ts
```diff
@@ -1567,6 +1573,21 @@ export interface MCPServer {
...
 }
 
+/**
+ * Tool that produces suggestions from one formula without an LLM turn. At most one per agent.
+ *
+ * @internal
+ * @hidden
+ */
+export interface SuggestionProducerTool extends BaseTool<ToolType.SuggestionProducer> {
+  /** Pack containing the formula. Required for agents; omission refers to the current pack. */
```
vivekvenkatram: The comment says omitting `packId` means the current pack, but skills reject this tool and agents require `packId`. wdyt about making `packId` required and dropping the current-pack case?

schema.ts
```diff
@@ -2548,3 +2548,74 @@ export function throwOnDynamicSchemaWithJsOptionsFunction(dynamicSchema: any, pa
...
+            explanation: {
+              type: ValueType.String,
+              required: true,
+              description: 'One or two sentences on what to change about the span and why.',
+            },
+            replacement: {
+              type: ValueType.String,
+              description: 'A concrete rewrite of the span. Absent when there is nothing to swap in.',
+            },
+            importance: {
```
vivekvenkatram: The zod schema rejects a number or `"HIGH"`, but this Coda property is still a free string, so `coda validate` accepts both. wdyt about an options list of the four tiers here so upload fails before the runtime drops the finding?

types.ts
```diff
@@ -1795,6 +1817,15 @@ export interface AgentToolsDef {
      */
     formulas?: Array<{formulaName: string}>;
   }>;
+  /**
+   * Connector formula that returns suggestions without an LLM turn. Instructions remain fallback.
+   */
+  suggestions?: {
```
vivekvenkatram: `docs` and `mail` are booleans, and `connectors` is already how an agent names a formula in another pack, so `suggestions` is a second pointer that reads like another switch. wdyt about `suggestionProducer`, matching `ToolType.SuggestionProducer`, since this must not be a Pack tool the model can call?

builder.ts
```diff
@@ -692,9 +692,10 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
    * ```
    * pack.setTools({docs: true, mail: true, webSearch: {allowedDomains: ['docs.example.com']}});
    * pack.setTools({connectors: [{packId: 1234, formulas: [{formulaName: 'CreateTask'}]}]});
+   * pack.setTools({suggestions: {packId: 1234, formulaName: 'CheckSuggestions'}});
```
vivekvenkatram: `setTools` assigns `agent.tools`, so these three examples wipe each other out and only the last one sticks. wdyt about one call that sets docs, connectors, and suggestions together?

api.ts
```diff
@@ -3132,3 +3135,116 @@ function moveJsPropertyOptionsFunctionsToFormulas({
...
+export interface SuggestionFormulaDef<ContextT extends ExecutionContext = ExecutionContext> {
+  /** Formula name referenced by {@link core.SuggestionProducerTool}. */
+  name: string;
+  /** User-facing description of what the formula checks. */
+  description: string;
+  /**
+   * Runs the check. Offsets use UTF-16 code units, and `original` must equal the corresponding
+   * `text.slice(startOffset, endOffset)`.
+   */
+  execute: (text: string, context: ContextT) => Promise<SuggestionResult> | SuggestionResult;
```
vivekvenkatram: Other formula `execute` functions take the parameter tuple, and `addFormula` already runs the result through `makeFormula`. This one takes the bare string, so `SuggestionFormulaDef` is not a `PackFormulaDef`. wdyt about keeping the tuple so authors only learn one execute shape?

api_types.ts
```diff
@@ -1215,6 +1221,78 @@ export interface InvocationLocation {
...
+  Low = 'low',
+}
+
+/**
+ * A suggestion-producing formula's finding about a span of its input text.
+ *
+ * @internal
+ * @hidden
+ */
+export interface SuggestionHighlight {
```
vivekvenkatram: The schema text calls each item a finding, the array is `suggestions`, and the type is `SuggestionHighlight`, which is not a word this SDK uses anywhere else. wdyt about `Suggestion`, so the array is `suggestions: Suggestion[]`?

api_types.ts
```diff
@@ -1215,6 +1221,78 @@ export interface InvocationLocation {
...
+ * across reruns. If your formula asks an LLM, have it choose one of these, with each level defined
+ * in your domain's own terms -- generic wording borrowed from another checker is the largest
+ * single source of miscalibration. Order findings within a level by their position in
+ * {@link SuggestionResult}, which is most important first.
+ *
+ * @internal
+ * @hidden
+ */
+export enum SuggestionImportance {
+  /** The reader will be misled, or will not act on something consequential, unless this is fixed. */
```
vivekvenkatram: `FormulaPurpose` and `ToolType` comments are one line. These member docs are a generic rubric, and the comment above says borrowed generic wording is what miscalibrates a checker. wdyt about bare level names here, and letting each pack define what they mean?

api_types.ts
```diff
@@ -1215,6 +1221,78 @@ export interface InvocationLocation {
...
+      /** Every finding for the submitted text, most important first. */
+      suggestions: SuggestionHighlight[];
+      /** A successful result cannot also report a formula-level error. */
+      error?: never;
+    }
+  | {
+      /** An error result cannot carry findings that the runtime might partially apply. */
+      suggestions: [];
+      /** Why the check could not run. */
+      error: string;
```
vivekvenkatram: Formulas in this SDK fail by throwing, and a throw is a failed invocation. An `error` string on a normal return is a new channel, and the runtime keeps existing highlights instead of falling back to the model. wdyt about `unavailable` so it does not look like the formula itself failed?

suggestion_schemas.ts
```diff
@@ -0,0 +1,45 @@
...
+import type {SuggestionResult} from './api_types';
+import {z} from 'zod';
+
+/**
+ * Runtime schema for {@link SuggestionHighlight}.
+ *
+ * @internal
+ * @hidden
+ */
+export const SuggestionHighlightSchema = z.object({
```
vivekvenkatram: Result shapes in this SDK live in `schema.ts` as Coda schemas, and zod here is for upload metadata under `testing/`. This file is a second contract for the same result, which is why the runtime imports it from `dist/`. wdyt about keeping the parse in the runtime, or exporting this from `index` if the SDK owns it?

types.ts
```diff
@@ -1567,6 +1573,21 @@ export interface MCPServer {
   name: string;
 }
 
+/**
+ * Tool that produces suggestions from one formula without an LLM turn. At most one per agent.
```
vivekvenkatram: This says at most one per agent, but upload validation only rejects an identical tool via `findDuplicateTools`. Two producers with different formula names still upload. wdyt about a refine on the agent tool array that rejects a second `SuggestionProducer`?

builder.ts
```diff
@@ -713,6 +714,13 @@ export class AgentDefinitionBuilder extends BaseDefinitionBuilder {
         ...(connector.formulas ? {formulas: connector.formulas} : {}),
       });
     }
+    if (suggestions) {
```
vivekvenkatram: docs and mail are tools the model calls, and the runtime pulls this one back out so the model never sees it. wdyt about putting the producer on `setDefaultWhileWritingTrigger` (color, surfaces, and the condition already live there) and leaving the formula on the connector?

vivekvenkatram: The formula belongs on the connector that produces it, and the trigger should only say when it shows (condition, surfaces), not name a packId. `PackVersionDefinition` already holds formulas and defaultTriggers together, so wdyt about one pack, with `FormulaPurpose.Suggestions` on the formula, and a cross-pack pointer only when a different agent reuses the checker?

api.ts
daviddriscoll-codaio: Done — split into a stack: this PR is the formula contract (factory, purpose, result shape), and the tier is #3661 on top. I also squashed the history, so the replaced shapes are gone.

index.ts
daviddriscoll-codaio: Agreed, since none of it ships yet: everything is `@hidden` now, including `FormulaPurpose.Suggestions`, and nothing new is exported from `index`. The runtime and the demo packs import from `dist/` for now.

types.ts
daviddriscoll-codaio: Moot now — `SuggestionProducerTool` is gone. An agent lists the connector under `connectors`, and the runtime finds the formula by its `FormulaPurpose.Suggestions` tag.

schema.ts
daviddriscoll-codaio: Checked: upload validation never compares result values with schema options (the only check on `examples[].result` is non-null), so a `SelectList` would document the tiers without failing anything. #3661 instead rejects a `Suggestions`-purpose formula whose examples report a number or an off-vocabulary tier. Live results can only be caught by the runtime, which drops the finding and counts it.

types.ts
daviddriscoll-codaio: Removed rather than renamed: with the formula found by its purpose, the existing `connectors` entry is the only pointer, so there's no second switch.

builder.ts
daviddriscoll-codaio: This PR no longer touches `setTools`, so the `suggestions` example is gone. You're right that the two examples left on `main` overwrite each other too; I'll fix that separately rather than widen this PR.

api.ts
daviddriscoll-codaio: Done — `execute` takes `([text], context)` like every other formula. `makeSuggestionFormula` now only fixes the parameter, schema and purpose.

api_types.ts
daviddriscoll-codaio: Done — `Suggestion`, so the result is `suggestions: Suggestion[]`. The runtime keeps its own `ProactiveHighlight` name for the model-facing shape.

api_types.ts
daviddriscoll-codaio: Agreed, good catch. In #3661 the levels are bare names with one line saying each pack defines them. The generic rubric stays only in the runtime's own LLM prompt.

api_types.ts
daviddriscoll-codaio: Done — `unavailable: string`: the check ran but couldn't assess the text, so what's on screen is kept. A failure throws. With several producers, a throw from one keeps the others' findings and suppresses deletes; the run falls back to the model only when none answers.

suggestion_schemas.ts
daviddriscoll-codaio: Agreed — `suggestion_schemas.ts` is gone. The SDK owns the TypeScript type and the Coda schema. The parse lives in the runtime, with a compile-time check that it matches `Suggestion` in both directions, so there's no import from `dist/suggestion_schemas` any more.

types.ts
daviddriscoll-codaio: Moot now — the tool type is gone. The runtime runs every `Suggestions`-purpose formula on the agent's connectors in parallel, merges them through one reconcile pass, and tags each highlight with its connector's pack id so ranking across them can calibrate per source.

builder.ts
daviddriscoll-codaio: Took most of this: the formula's `FormulaPurpose.Suggestions` tag is now the declaration, the trigger still only says when it shows, and the only cross-pack pointer is the ordinary `connectors` entry. One pack isn't possible yet — `newAgent()` has no `addFormula` and `newPack()` has no `agent`/`defaultTriggers` — so I'd like to make allowing formulas on agent packs a follow-up. Once that exists, an agent's own formulas simply join the inferred set.

api.ts
```diff
@@ -3132,3 +3135,100 @@ function moveJsPropertyOptionsFunctionsToFormulas({
...
+ * a producer the runtime cannot call is skipped, not rejected:
+ *
+ * - the parameter list: one required string named `text`, because the runtime passes arguments by
+ *   parameter name;
+ * - the result schema: {@link makeSuggestionResultSchema};
+ * - {@link FormulaPurpose.Suggestions}, which is what marks the formula as a producer in the
+ *   uploaded metadata.
+ *
+ * The formula lives in a connector, because only a connector holds formulas, authentication and
+ * network domains. An agent that lists the connector under `connectors` in `setTools` runs every
```
vivekvenkatram: `connectors` means formulas the model may call, and this also runs every `Suggestions` formula on that pack and skips the model. A pack with `CreateTask` and `CheckSuggestions` then does both, and two suggestion formulas both run with no merge rule. wdyt about calling only the suggestion formulas, withholding them from the model, and only on a while-writing check?

daviddriscoll-codaio: Agreed on all three, and most of it was already the runtime's behavior — the doc just didn't say so. Suggestion formulas run only on a while-writing check, and they're withheld from the model there; `CreateTask` stays a model tool, used only if every producer fails. Several producers do have a merge rule: findings are reconciled in one pass, the same title on the same span collapses (first listed wins), an incomplete answer from any producer suppresses deletes, and each highlight names its connector. The gap you found was chat: nothing withheld `CheckSuggestions` there. The pack resolver now leaves `Suggestions`-purpose formulas out on every path unless the builder calls them directly (superhumans-org/front-end-web-monorepo#21317, 44087b5), and the doc here now says so in two lines (d4aea40).

daviddriscoll-codaio:
Moved to the monorepo per the libs/packs-sdk migration: superhumans-org/front-end-web-monorepo#23777 (suggestion formulas), stacked under the rest of the feature (stack #23783). This PR keeps the review history; the monorepo copy has the same source as its last commit here.

PR #3652, opened 2026-09-16
SV-138435: build an isolated-vm native for the Node 24 pack runtime

saikirangoud-grammarly:
## Why

CVE-2024-7971 is a V8 type confusion reachable from the Pack sandbox. It gives **native code execution in the production Pack worker** — see SV-138435, where it was reproduced in production.

It was fixed upstream in **V8 12.8**. Node pins one V8 branch per major line and patches it in place rather than rebasing, so there is no Node 22 patch to wait for. Measured across the whole line:

| Node | V8 |
|---|---|
| 22.0.0 | 12.4.254.14-node.10 |
| 22.22.0 | 12.4.254.21-node.33 |
| 22.23.2 ← prod | 12.4.254.21-node.**56** |
| **24.21.0** | **13.6**.233.17-node.53 |

The base never leaves `12.4`. The exploit works in production on `22.23.2`, which is the newest Node 22 there is — that is the evidence the fix was not among those 56 patch rounds.

So the remediation is a runtime migration, and that needs an `isolated-vm` native built against the Node 24 ABI.

## What this does

Adds the Node 24 native **alongside** the existing Node 22 one rather than replacing it, so the `n22` runtime stays deployable and the migration can be rolled back by runtime config — the same approach the `Node22v2` variant comment describes.

- `Makefile`: `do-compile-isolated-vm-24` / `compile-isolated-vm-24`, mirroring the `-22` pair; `compile` now builds both
- `runtime/native/node24/x86_64/isolated-vm/`: the built artifact
- version bump

## Verification

- **`isolated-vm` stays pinned at 6.0.2 — no dependency change needed.** It builds and runs clean on Node 24.
- **Full suite: 857 passing / 0 failing on Node 22 *and* Node 24**, identical counts — including the thunk, the injected stubs, `executeThunk`, and the SV-138082 prototype-pollution regression test.
- Committed native is `ELF 64-bit x86-64 GNU/Linux, stripped`, isolated-vm 6.0.2, built in `public.ecr.aws/sam/build-nodejs24.x` — same shape as the node22 binary.
- Both `compile-isolated-vm-*` targets report "version matches, skipping", so CI rebuilds neither.
- No TypeScript changed, so `dist/` needs no rebuild.

## This is not sufficient on its own

Two follow-ups, neither in this PR:

1. **Monorepo companion change** — an `n24` lambda variant, the `nodejs24.x` runtime mapping, the `node24` native selection, and a bump of `@aws-sdk/client-lambda` (which has no `nodejs24x` value before **3.920.0**; the monorepo pins 3.893.0).
2. **Validation.** Nobody has yet shown the exploit *failing* on V8 13.6 — local reproduction has never worked, so the CVE attribution is still the reporter's. Before flipping the default variant, publish a pack on `n24` and re-run `NativeGetpid` against Pack 65118. Until then "fixed" is version arithmetic, not a test.

### Note for later

Node 26 (V8 **14.6**) is also an available Lambda runtime and would buy roughly another year of V8 freshness, but it needs `isolated-vm` **7.0.1** — 6.0.2 fails to compile there — and 7.0.1 changes the native layout to `prebuilds/<platform>/isolated-vm.abi<N>.<libc>.node`, which the Makefile copy steps and the monorepo's `runtime_package_helper.ts` mapping both hardcode. Worth doing as its own change, not folded into this one.


saikirangoud-grammarly:
Closing for now — not abandoned. Holding this until the Phase 2 approach is agreed with the runtime owners and the fix is validated against Pack 65118 on an n24 function, since nobody has yet observed the exploit failing on V8 13.6. The branch `sv-138435-phase2-node24-runtime` is left in place, so this can be reopened as-is. Tracking in SV-138435.

PR #3654, opened 2026-09-16
packs-sdk: release v1.17.6

peizhaomei-codaio:
### Added

- Added a `superhuman` CLI binary alias alongside the existing `packs` and `coda` aliases. `npx superhuman <cmd>` behaves identically to `npx packs <cmd>`.
- Documented installing the `create-agent-pack` skill via the `skills` CLI in the README.

### Fixed

- Fixed invalid YAML metadata in the `create-agent-pack` skill's `SKILL.md` that made it unreadable by third-party skill-installer tooling.
- `validate` now prints `Pack is valid.` on success instead of exiting silently.
- The version-bump CI check failure message now mentions the `skip-version-check` label as an escape hatch for docs-only or CI-only changes.

### Changed

- CLI commands now fail immediately in non-interactive environments instead of waiting on confirmation prompts. Use `--yes` to skip confirmations on `clone`, `link`, `register`, and `release`. `register` requires `--apiToken` (or a pasted token on a TTY). `release` accepts `--use-latest` when the manifest has no version. `--help` for each command includes examples ready to run.
- Removed the unused `bestEffortDefaultTriggers` validation option. Pack metadata validation remains strict and reports malformed triggers by array index.


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 
 ## [Unreleased]
 
+## [1.17.6] - 2026-09-16
+
```
xicheng-codaio, line 8: should we have some notes? e.g. https://github.com/coda/packs-sdk/pull/3637 seems fairly substantive?

xicheng-codaio (approved):
LGTM

PR #3658, opened 2026-09-18
Warn before overwriting an existing pack.ts on init

xicheng-codaio:
## Summary

`npx coda init` copies the example template over the current directory unconditionally, so running it a second time in a directory that already has a `pack.ts` silently overwrote the developer's code. This adds a confirmation guard before the copy, mirroring the pattern `coda clone` already uses.

## Changes
- **`cli/init.ts`** — `handleInit` now checks for an existing `pack.ts` *before any npm work* and calls `confirmOrFail`. Interactive runs get a `y/N` prompt; non-interactive runs abort (exit 1) unless `--yes` is passed.
- **`cli/index.ts`** — wired the `yes: YesArg` flag onto the `init` command (yargs runs in `.strict()` mode, so the `--yes` the prompt suggests would otherwise be rejected), added a `packs init --yes` usage example, and updated the `init` blurb since it no longer scaffolds an "empty" pack.
- **`cli/clone.ts`** — both `handleInit()` calls pass a named `overwriteAlreadyConfirmed` flag, since `clone` already confirms the overwrite itself, avoiding a double prompt.
- **`docs/reference/cli/index.md`** — regenerated CLI reference to document the new option.

## Testing
Added tests to `test/cli_headless_test.ts` (TDD — verified red before green): aborts with no side effects when `pack.ts` exists and isn't confirmed; prompts and passes `yes` through when it exists; doesn't prompt when it doesn't. Full CLI suite, lint, and type-check pass.


cli/init.ts
```diff
@@ -34,7 +35,16 @@ function escapeShellCmd(cmd: string): string {
...
-export async function handleInit() {
+export async function handleInit({yes}: {yes?: boolean} = {}) {
+  // Warn before clobbering an existing pack.ts, since init copies the template over it.
+  if (fs.existsSync(path.join(process.cwd(), 'pack.ts'))) {
+    confirmOrFail({
+      yes,
+      prompt: 'A pack.ts file already exists. Do you want to overwrite it? (y/N)?',
+      example: 'packs init --yes',
+    });
+  }
```
vivekvenkatram, line 46: This is the right place for the guard: init still copySyncs the whole example template over cwd after npm work, so a second `npx coda init` in an existing pack dir silently replaced pack.ts (Zack's GOECOS-428 dogfood). Confirming only on pack.ts matches clone and matches the ticket.

Once they say yes, the copy still replaces the rest of the template too (tsconfig, README, etc.). Same as today's init, just no longer silent on pack.ts. If we ever want this prompt to mean "overwrite only pack.ts", copySync would need to change; otherwise this is fine.

cli/clone.ts
```diff
@@ -36,6 +36,9 @@ export async function handleClone({packIdOrUrl, apiEndpoint, apiToken, yes}: Arg
     });
   }
 
+  // clone handles its own overwrite confirmation above, so init shouldn't prompt again.
+  const overwriteAlreadyConfirmed = true;
```
vivekvenkatram, line 40: Always `true` is correct here: if pack.ts existed, clone already prompted (or they passed --yes) and confirmOrFail would have exited on no. If it didn't exist, init's existsSync is also false so this flag is unused. Don't thread clone's `yes` through instead; an interactive confirm leaves `yes` false and you'd double-prompt.

test/cli_headless_test.ts
```diff
@@ -58,6 +60,60 @@ describe('Headless CLI', () => {
     });
   });
 
+  describe('init', () => {
+    // Sentinel thrown by the stubbed spawnProcess to prove execution got past the
+    // overwrite guard without doing any real npm work.
+    const SPAWN_SENTINEL = 'spawnProcess was called';
+
+    it('aborts before doing any work when pack.ts already exists and is not confirmed', async () => {
+      sinon.stub(fsExtra, 'existsSync').returns(true);
```
vivekvenkatram: This stubs every existsSync as true. Fine today because the abort happens before any other fs check, but if handleInit later existsSyncs something else this test starts lying. wdyt about `callsFake(p => String(p).endsWith('pack.ts'))` so it only models the file we care about?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This changelog keeps track of all changes to the Packs SDK. We follow convention
 ### Fixed
 
 - CLI commands now surface the server-side error message (and HTTP status) when an API request fails, instead of swallowing it behind a generic message. This affects `register`, `whoami`, `clone`, and `link`, and makes `formatResponseError` robust to non-JSON and empty error bodies.
+- `packs init` (and its `coda`/`superhuman` aliases) now prompts before overwriting an existing `pack.ts` instead of silently clobbering it. Pass `--yes` to skip the prompt (required in non-interactive environments).
```
vivekvenkatram: This is the public CLI contract now: headless `init` in a dir that already has pack.ts exits 1 unless `--yes`, so scripts that relied on the silent overwrite fail on bump. wdyt about one more clause here (keepachangelog would also allow Changed) so npm consumers see that without reading source?

cli/index.ts
```diff
@@ -55,7 +55,10 @@ const CommandExamples: Record<string, Array<[string, string]>> = {
     ['$0 execute pack.ts MySyncTable --maxRows 10', 'Sync a table and cap the number of rows.'],
   ],
   auth: [['$0 auth pack.ts', 'Start local auth setup for the Pack.']],
-  init: [['$0 init', 'Scaffold pack.ts and related starter files in the current directory.']],
+  init: [
+    ['$0 init', 'Scaffold pack.ts and related starter files in the current directory.'],
+    ['$0 init --yes', 'Overwrite an existing pack.ts without prompting.'],
```
vivekvenkatram: `Overwrite an existing pack.ts without prompting` is what `--help` shows on npm, but `--yes` still copySyncs the whole template. wdyt matching that, e.g. `Skip the overwrite prompt (replaces pack.ts and other starter files)`?

cli/index.ts
```diff
@@ -160,7 +163,10 @@ export const commands: yargs.CommandModule[] = [
   {
     command: 'init',
     describe: 'Initialize an empty Pack',
```
vivekvenkatram: `Initialize an empty Pack` is still the public `--help` / reference blurb, and it is no longer true once pack.ts exists. wdyt `Scaffold a Pack in the current directory` so we don't promise empty?

vivekvenkatram (approved):
lgtm, pls address all comments before merging

